PR #331, opened 2020-09-17
Bugfix: Reporting correct Status Code when auditing exceptions.

v-shaaal:
## Description
Fixed status code when auditing exceptions.

**Context:**
Correct status code was not getting logged in cases of exceptions. This was happening because exceptions were logged in `OnActionExecuted` method whereas HttpStatusCode were determined after this call. To get around this problem, `ExceptionHandlingHelper` is created such that it can map exceptions and get us respective status code. This status code is then set as the response status code when logging exceptions.

## Related issues
[AB#75611](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/75611).

## Testing
- Added unit tests to test the newly introduced class.
- All the existing tests are passing.

### DIFF
--- a/src/Microsoft.Health.Dicom.Api/Features/Audit/AuditLoggingFilterAttribute.cs
+++ b/src/Microsoft.Health.Dicom.Api/Features/Audit/AuditLoggingFilterAttribute.cs
@@ -8,6 +8,7 @@ using EnsureThat;
 using Microsoft.AspNetCore.Mvc.Filters;
 using Microsoft.Health.Api.Features.Audit;
 using Microsoft.Health.Core.Features.Security;
+using Microsoft.Health.Dicom.Api.Features.Exceptions;
 
 namespace Microsoft.Health.Dicom.Api.Features.Audit
 {
@@ -27,6 +28,7 @@ namespace Microsoft.Health.Dicom.Api.Features.Audit
 
             if (context.Exception != null)
             {
+                context.HttpContext.Response.StatusCode = (int)ExceptionHandlingHelper.GetStatusCode(context.Exception);
                 AuditHelper.LogExecuted(context.HttpContext, ClaimsExtractor);
             }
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.Health.Dicom.Api/Features/Audit/AuditLoggingFilterAttribute.cs
```diff
@@ -27,6 +28,7 @@ public override void OnActionExecuted(ActionExecutedContext context)
 
             if (context.Exception != null)
             {
+                context.HttpContext.Response.StatusCode = (int)ExceptionHandlingHelper.GetStatusCode(context.Exception);
```
smithago, line 31: I am curious to understand why the AuditFilter is setting the response statusCode.
Is it not going through the middleware on the way out and getting set there?

v-shaaal: That is partially correct. The way DICOM handles exception, ExceptionHandlingMiddleware is responsible for setting the correct status code. However, this middleware is called after OnActionExecuted is called in the AuditFilter which means that when OnActionExecuted is called, the status code is still set to OK (200).

Now this is different from FHIR because FHIR uses a class level annotation of OperationOutcomeExceptionFilterAttribute because of which OnResultExecuted is called in the AuditFilter. Since DICOM's layout is different and does not have this class level attribute, OnResultExecuted is not called in the cases of Exceptions, hence, we have to rely on OnActionExecuted.

smithago: Have you looked at creating a AudiLoggerMiddleware that gets called after ExceptionHandlingMiddleWare on the way out, that can extract the claim and log the the context?

Looks like for auditing we need to capture the response right before it leaves instead of here

v-shaaal:
Closing this PR. Will open a new one with AuditEgressLogger implementation that will log executed messages in AuditMiddleware as expected. Healthcare Shared Components side of PR is already out for the aforementioned change: [AB#111](https://github.com/microsoft/healthcare-shared-components/pull/111).

PR #335, opened 2020-09-18
Single Dicom File Upload 

abdullah248:
## Description
Allow for upload of single dicom file using single part (application/dicom) instead of always multipart/related

## Related issues
#349 

## Testing
Unit and integration tests added. Also uploaded the file as a post body in fiddler verification.


src/Microsoft.Health.Dicom.Client/DicomWebClient.cs
```diff
@@ -187,6 +187,26 @@ await using (var stream = GetMemoryStream())
...
+            return await PostAsync(stream, studyInstanceUid, cancellationToken);
+        }
+
+        public async Task<DicomWebResponse<DicomDataset>> StoreAsync(
+            DicomFile dicomFile,
+            string studyInstanceUid = null,
+            CancellationToken cancellationToken = default)
+        {
+            await using (var stream = GetMemoryStream())
+            {
```
smithago: using to dispose the stream created?

smithago (approved):
Please react to the comment before checkin

src/Microsoft.Health.Dicom.Client/DicomWebClient.cs
abdullah248: Updated, thanks

PR #338, opened 2020-09-19
Update retrieving study/instance/series to support more transfer syntax

pengchen0692:
## Description
Update retrieving study/instance/series to support more transfer syntax:
- 1.2.840.10008.1.2.1 (ExplicitVRLittleEndian)
- 1.2.840.10008.1.2.4.90 (JPEG2000Lossless)

## Summary
- Update code to support more transfersyntax
- Add End to end test to verify 
- Refactor and clean up RetrieveTransactionResourceTests 
   - Test matrix can be found [here](https://microsoft.sharepoint.com/teams/msh/_layouts/OneNote.aspx?id=%2Fteams%2Fmsh%2FShared%20Documents%2FProject%20Resolute%2FMedical%20Imaging%20and%20Media%2FMedical%20Imaging%20Team&wd=target%28Features.one%7C1BFE0535-89E7-49F1-80FF-410140F9C561%2FRetrieve%20Resources%20Test%20Matrix%7C492123D0-95C4-473F-8156-1E73D2CDCD15%2F%29)

## Related issues
Addresses https://microsofthealth.visualstudio.com/Health/_workitems/edit/75185


src/Microsoft.Health.Dicom.Api/Controllers/RetrieveController.cs
```diff
@@ -47,18 +46,14 @@ public RetrieveController(IMediator mediator, ILogger<RetrieveController> logger
             _logger = logger;
         }
 
-        [AcceptContentFilter(new[] { KnownContentTypes.ApplicationDicom }, allowSingle: false, allowMultiple: true)]
```
smithago, line 50: Are these filters used anywhere else? If not should we delete them?

pengchen0692: AcceptContentFilter is still used by GetxxMetadataAsync, since GetMetadata has a different code path, I didn't plan to touch them.

src/Microsoft.Health.Dicom.Tests.Common/Comparers/DicomDatasetEqualityComparer.cs
```diff
@@ -0,0 +1,50 @@
...
+        public IEnumerable<DicomTag> IgnoredTags { get; }
+
+        bool IEqualityComparer<DicomDataset>.Equals(DicomDataset x, DicomDataset y)
+        {
+            if (x == null || y == null)
+            {
+                return object.ReferenceEquals(x, y);
+            }
+
+            // Ignore PixelData here, will check later with DicomPixelDataEqualityComparer
```
snehabandla, line 31: >Ignore PixelData here, will check later with DicomPixelDataEqualityComparer [](start = 15, length = 75)

It probably is sufficient for this comment to say - "First validiate...."

test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/HttpRequestMessageBuilder.cs
```diff
@@ -0,0 +1,45 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Net.Http;
+using System.Net.Http.Headers;
+
+namespace Microsoft.Health.Dicom.Web.Tests.E2E.Rest
+{
+    public class HttpRequestMessageBuilder
```
snehabandla, line 12: >HttpRequestMessageBuilder [](start = 17, length = 25)

This now creates two paths to create requests - this class and the client. How do we know when to use which? Is there a reason we can't update the client to support some of this functionality?

pengchen0692: The client doesn't return content type (e.g: image/jp2 when retrieving frames) which should be verified by test case. I also think about updating client to support, but updating product code purely for testing is not a  good practice, besides I see existing tests ([link](https://github.com/microsoft/dicom-server/blob/882ff5f461b6946ea1d74797753245deeb8de86a/test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/RetrieveTransactionResourceTests.cs#L133)) already build httprequest by itself for validation, so I think this might be a better way.

I'm open to question, we can discuss to get the best solution.

src/Microsoft.Health.Dicom.Tests.Common/Comparers/DicomDatasetEqualityComparer.cs
pengchen0692: will update...

test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/HttpRequestMessageBuilder.cs
pengchen0692: Positive test cases have been updated to use client. Still need this for Negative tests.

snehabandla:
> > DICOM File Size Limit: there is a size limit of 2GB for a DICOM file by default.

I'm not sure if this how we want to  word it. We have plans to find ways to increase this, right?

---
Refers to: docs/resources/conformance-statement.md:48 in 505f735. [](commit_id = 505f7356b34b813df008222923446f007111b7b8, deletion_comment = False)

pengchen0692:
> > > DICOM File Size Limit: there is a size limit of 2GB for a DICOM file by default.
> 
> I'm not sure if this how we want to word it. We have plans to find ways to increase this, right?
> 
> Refers to: docs/resources/conformance-statement.md:48 in [505f735](https://github.com/microsoft/dicom-server/commit/505f7356b34b813df008222923446f007111b7b8). [](commit_id = [505f735](https://github.com/microsoft/dicom-server/commit/505f7356b34b813df008222923446f007111b7b8), deletion_comment = False)

I guess this is in another commit? I checked with Smitha, it turns out there is no customer requirement for >2GB, and also our Retrieve API still not support >2GB now, so  currently we limit the upload file size to 2GB.

smithago (approved):

PR #368, opened 2020-10-02
Support store/retrieve up to 4GB file 

pengchen0692:
## Description
This is private fix for store/retrieve up to 4GB file. (**NOT** intended for production use)

## Summary
* MemoryStream won't be able to support >2GB file, replace with FileStream
* Update client to use FileStream also to support >2GB file

## Note
* This change is not intended for production use

## Related issues
Addresses https://microsofthealth.visualstudio.com/Health/_boards/board/t/Medical%20Imaging/Stories/?workitem=76230

## Testing
* Test locally
* Tested in K8s cluster

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Common/FileStreamManager.cs
+++ b/src/Microsoft.Health.Dicom.Core/Common/FileStreamManager.cs
@@ -1,0 +1,19 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.IO;
+
+namespace Microsoft.Health.Dicom.Core.Common
+{
+    public class FileStreamManager
+    {
+#pragma warning disable CA1822 // Mark members as static
+        public Stream GetStream()
+#pragma warning restore CA1822 // Mark members as static
+        {
+            return new FileStream(Path.GetTempFileName(), FileMode.Create, FileAccess.ReadWrite, FileShare.Delete, 1024 * 1024 * 5, FileOptions.DeleteOnClose);
+        }
+    }
+}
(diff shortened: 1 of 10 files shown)


src/Microsoft.Health.Dicom.Core/Common/FileStreamManager.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Microsoft.Health.Dicom.Core.Common
+{
+    public class FileStreamManager
+    {
+#pragma warning disable CA1822 // Mark members as static
+        public Stream GetStream()
+#pragma warning restore CA1822 // Mark members as static
+        {
+            return new FileStream(Path.GetTempFileName(), FileMode.Create, FileAccess.ReadWrite, FileShare.Delete, 1024 * 1024 * 5, FileOptions.DeleteOnClose);
```
smithago, line 16: When is this stream closed?

src/Microsoft.Health.Dicom.Web/web.config
```diff
@@ -9,7 +9,7 @@
       <requestFiltering>
         <!-- When the process is hosted in IIS, change the following limit to increase maximum content length allowed. -->
         <!-- Set to 2GB by default. -->
-        <requestLimits maxAllowedContentLength="2147483648" />
+        <requestLimits maxAllowedContentLength="4294967295" />
```
smithago, line 12: We will not default support 4GB. We may reduce it even from 2GB to 1Gb, there maybe disk contention with file streams, the long term solution to support large files will be streaming

src/Microsoft.Health.Dicom.Web/Startup.cs
```diff
@@ -25,7 +25,7 @@ public virtual void ConfigureServices(IServiceCollection services)
             services.Configure<IISServerOptions>(options =>
             {
                 // When hosted on IIS, the max request body size can not over 2GB, according to Asp.net Core bug https://github.com/dotnet/aspnetcore/issues/2711
-                options.MaxRequestBodySize = int.MaxValue;
```
smithago, line 28: can we make this change out of the other changes in this PR.

PR #374, opened 2020-10-07
Dont fwd azure emulator & sql to host in compose

KnicKnic:
## Description
Do not expose azure emulator and sql ports to host. This will prevent them from colliding if sql server or azure storage emulator is installed in the host already

## Testing
Did the following simultaneously 
* Started azure storage emulator and sql server on machine 
* ran dicom tests
* started docker-compose environment
* uploaded dicom files to docker-compose environment

### DIFF
--- a/samples/docker/docker-compose.yaml
+++ b/samples/docker/docker-compose.yaml
@@ -21,9 +21,9 @@ services:
     container_name: azurite
     image: mcr.microsoft.com/azure-storage/azurite
     restart: always
-    ports:
-      - "10001:10001"
-      - "10000:10000"
+    # ports:
+    #   - "10001:10001"
+    #   - "10000:10000"
   sql:
     build: 
       context: ./../../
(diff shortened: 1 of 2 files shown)


samples/docker/docker-compose.yaml
```diff
@@ -21,18 +21,18 @@ services:
     container_name: azurite
     image: mcr.microsoft.com/azure-storage/azurite
     restart: always
-    ports:
-      - "10001:10001"
-      - "10000:10000"
+    # ports:
```
brandonpollett, line 24: nit: maybe add a comment around explaining to un-comment these if you want connect directly to these services from the host.

brandonpollett (approved):

samples/docker/docker-compose.yaml
KnicKnic: Done

PR #392, opened 2020-10-16
performance guidelines based on scale testing

moiradillon12:
## Description
Performance guidance for choosing Azure SQL and Azure App Service tiers, along with ensuring geo-redundancy

## Related issues
[AB#74896](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/74896)

## Testing
Viewed in VS Code


docs/resources/performance-guidance.md
```diff
@@ -0,0 +1,62 @@
...
+
+In addition to configuring geo-redudunant Azure Storage, we recommend configuring active geo-replication for your Azure SQL Database. This allows you to create readable secondary databases of individual databases on a server in the same or different data center region. For a tutorial on how to configure this, see [Creating and using active geo-replication - Azure SQL Database](https://docs.microsoft.com/azure/azure-sql/database/active-geo-replication-overview).
+
+## Workload Scenarios for Azure SQL Database & Azure App Service
+
+### Scenario 1: Testing out DICOM Server
+
+If you are testing out the Medical Imaging Server for DICOM and not running production workloads, we recommend using the S1 Azure App Service Tier alongside a Standard Azure SQL Database (S1, S2, S3). For a small system that does not require redundancy, with these tiers, you can spend as little as ~$70/month on Azure App Service & Azure SQL Database.
+
+If you had ~100,000 DICOM instances, you should expect a range of 2000-9000 requests/minute and a response time under 1 second for STOW, WADO & QIDO.
```
snehabandla, line 43: I don't think referring to the number of instances here makes sense! The numbers should hold decently for higher numbers as well. The key would be to monitor the percentage of SQL DB usage.

docs/resources/performance-guidance.md
```diff
@@ -0,0 +1,62 @@
...
+
+We recommend the S1 Standard Azure App Service Tier along side an Azure SQL Tier of S12. At these tiers, you can expect up a range of 1000-20,000 requests/minute with a response time under 400 ms for WADO, QIDO & STOW.
+
+### Scenario 3: Large Hospital System with high numbers of DICOM requests per day in a production workload
+
+If you need to upload a large number of DICOM instances a day (800,000+), we recommend scaling up your Azure App Service Plan and your Azure SQL Database to Premium Tiers. If you are going into production, you also need to ensure your Medical Imaging Server for DICOM supports geo-redundancy. Refer to our [geo-redundancy guidelines](##Geo-Redundancy).
+
+To support a large number of DICOM transactions per day, we recommend a P1v2 tier for Azure App Service alongside a P11 tier for Azure SQL Database. For STOW, WADO and QIDO, you should expect a range of 40,000 - 100,000 requests/minute with response times under 200 ms.
```
snehabandla, line 55: I think this is overkill for a hospital system. It's highly unlikely that they would hit that requests/minute value. We should have a chat with Steve and identify a better scenario for the best perf we have.

StevenBorg: @snehabandla is correct for this being overkill for the vast majority (maybe all) hospital systems.  That said, I like it in one way...  It's HUGE.  Much bigger than anyone would need. Customers often express fear that our system will not scale sufficiently for their needs. I often will drop a large number from our testing.  

That said, STOW is highly variable.  It's unlikely people will be able to get 40k / sec for anything STOW related.  We're stressing with a very small instance (50 kb or 500 kb, I believe).  Many people have instance an order of magnitude (or two) larger.  Plus they're on slower networks.  Maybe we should caveat a bit. rather than "you should expect a range" something along the lines of "With excellent bandwidth, relatively small images and and appropriate mix of STOW, WADO and QIDO requests, you can expect to handle between 40,000 and 100,000 requests/minutes with response times under 200 ms.  Larger files, usage patterns that lean heavily to STOW, and poor bandwidth will reduce the number of requests per minute. We recommend testing performance with your data and indented use."  Or something along those lines.

moiradillon12: updated based on this feedback!

moiradillon12:
@StevenBorg  @snehabandla thanks for the feedback today! updated based on our conversation

StevenBorg (approved):
Looks good to me.  Made one update to correct a spelling error.

PR #399, opened 2020-10-20
Bug Fix: Invalid Multipart Request Causes Internal Server Error

abdullah248:
## Description
Invalid multipart request was throwing an internal server error when missing new line. Updated to throw 400 Bad Request error.

## Related issues
https://microsofthealth.visualstudio.com/Health/_workitems/edit/73702

## Testing
Reproduced original bug and validated 400 Bad request thrown instead of 500 internal server error.

### DIFF
--- a/src/Microsoft.Health.Dicom.Api/Web/AspNetCoreMultipartReader.cs
+++ b/src/Microsoft.Health.Dicom.Api/Web/AspNetCoreMultipartReader.cs
@@ -13,6 +13,7 @@ using Microsoft.AspNetCore.WebUtilities;
 using Microsoft.Extensions.Options;
 using Microsoft.Health.Abstractions.Exceptions;
 using Microsoft.Health.Dicom.Core.Configs;
+using Microsoft.Health.Dicom.Core.Exceptions;
 using Microsoft.Health.Dicom.Core.Web;
 using Microsoft.Net.Http.Headers;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Dicom.Api/Web/AspNetCoreMultipartReader.cs
```diff
@@ -90,7 +91,15 @@ internal class AspNetCoreMultipartReader : IMultipartReader
         /// <inheritdoc />
         public async Task<MultipartBodyPart> ReadNextBodyPartAsync(CancellationToken cancellationToken)
         {
-            MultipartSection section = await _multipartReader.ReadNextSectionAsync(cancellationToken);
+            MultipartSection section;
+            try
+            {
+                section = await _multipartReader.ReadNextSectionAsync(cancellationToken);
```
smithago, line 97: Can we catch this exception inside ReadNextSectionAsync.

Generally we should only catch specific exception, otherwise we will catch system exceptions like OutOfMememory exceptions and try to continue to work instead of letting the app fail

jackliums: ReadNextSectionAsync is ASP.NET code but we can catch more specific exception here.

abdullah248: I updated it to catch the specific InvalidDataException

smithago (approved):

PR #409, opened 2020-10-29
Log exception when retrive change feed fails

pengchen0692:
## Description
We don't log exception when retrieve change feed fails, fix it

## Related issues
Addresses https://microsofthealth.visualstudio.com/Health/_workitems/edit/76891

## Testing
Verified locally when retrieve feeds fails for authentication

## Notes
FHIR doesn't have this problem.

### DIFF
--- a/converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/ChangeFeedProcessor.cs
+++ b/converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/ChangeFeedProcessor.cs
@@ -27,6 +27,11 @@ namespace Microsoft.Health.DicomCast.Core.Features.Worker
         private readonly IFhirTransactionPipeline _fhirTransactionPipeline;
         private readonly ISyncStateService _syncStateService;
         private readonly ILogger<ChangeFeedProcessor> _logger;
+        private static readonly Action<ILogger, Exception> LogFailtoRetrieveChangeFeedDelegate =
+           LoggerMessage.Define(
+               LogLevel.Warning,
+               default,
+               "Fail to retrieve change feed.");
 
         public ChangeFeedProcessor(
             IChangeFeedRetrieveService changeFeedRetrieveService,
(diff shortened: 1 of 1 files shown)


converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/ChangeFeedProcessor.cs
```diff
@@ -27,6 +27,11 @@ public class ChangeFeedProcessor : IChangeFeedProcessor
         private readonly IFhirTransactionPipeline _fhirTransactionPipeline;
         private readonly ISyncStateService _syncStateService;
         private readonly ILogger<ChangeFeedProcessor> _logger;
+        private static readonly Action<ILogger, Exception> LogFailtoRetrieveChangeFeedDelegate =
+           LoggerMessage.Define(
+               LogLevel.Warning,
+               default,
+               "Fail to retrieve change feed.");
```
snehabandla, line 34: Minor: Failed tor retrieve....

snehabandla (approved):
Looks mostly okay to me! Just would be nice to use a consistent tense for our logged statements.

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/ChangeFeedProcessor.cs
```diff
@@ -53,9 +58,18 @@ public async Task ProcessAsync(TimeSpan pollIntervalDuringCatchup, CancellationT
...
-                    state.SyncedSequence,
-                    cancellationToken);
+                IReadOnlyList<ChangeFeedEntry> changeFeedEntries;
+                try
+                {
+                    changeFeedEntries = await _changeFeedRetrieveService.RetrieveChangeFeedAsync(
+                        state.SyncedSequence,
+                        cancellationToken);
+                }
+                catch (Exception ex)
```
smithago, line 68: Generally we should not be catching all exceptions like this. What is the pattern followed in dicom-cast to catch these exceptions and logging it? We use middleware to handle and log in dicom-server

pengchen0692: I  didn't see something like middleware in dicom cast.
Exception is log and thrown like this: https://github.com/microsoft/dicom-server/blob/master/converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/FhirTransaction/LoggingFhirTransactionPipelineStep.cs

smithago: Can you bring this up in the standup tomorrow. We can catch a global exception in DicomCastWorker and log them and throw them back.

PR #410, opened 2020-10-30
Adding in a Github action to create a release.

snehabandla:
## Description
GitHub action that runs every Tuesday to trigger a powershell script that approves and generates latest release.

## Related issues
[User Story 76752](https://microsofthealth.visualstudio.com/Health/_workitems/edit/76752)

### DIFF
--- a/.github/workflows/CreateRelease.ps1
+++ b/.github/workflows/CreateRelease.ps1
@@ -46,7 +46,7 @@ while(-Not($lastRelease -eq $currentRelease)) {
 
     # Validate that only one release is pending approval - any others should be approved, rejected or queued.
     if($approval.Count -gt 1) {
-        $pendingApprovals = "Pending approval releases: "
+        $pendingApprovals = "Pending releases for approval: "
         foreach ($pendingApproval in $approval) {
             $pendingApprovals += $pendingApproval.release.name
         }
--- a/.github/workflows/CreateRelease.ps1
+++ b/.github/workflows/CreateRelease.ps1
@@ -1,0 +1,95 @@
+# Required parameter specifying the secret access token to view releases and manage approvals.
+param([Parameter(Mandatory=$true)][string]$accessToken)
+
+# Function to help log out useful information.
+function log ([parameter(mandatory)][String]$msg) { Add-Content log.txt $msg }
+
+# Generate a log file.
+New-Item ./log.txt -ItemType File -Force
+
+# Releases are only created every other week even though the action occurs every week.
+# Identify if this is the second week of a sprint.
+$currentDate = Get-Date
+$currentDate = $currentDate.ToUniversalTime().Date
+$firstRun = ([datetime]::ParseExact('10-23-2020', 'MM-dd-yyyy', $null)).Date
+$weeksSince = [math]::floor(($currentDate - $firstRun).TotalDays / 7)
+$shouldRelease = ($weeksSince % 2) -eq 0
+
+if($shouldRelease) {
+    log "We're not releasing this week, one week left!"
+    exit
+}
+
+# Set basis to call AzureDevOps' Rest API as necessary.
+[Net.ServicePointManager]::SecurityProtocol =  [Net.SecurityProtocolType]::Tls12
+$azureDevOpsAuthenicationHeader = @{Authorization = 'Basic ' + [Convert]::ToBase64String([Text.Encoding]::ASCII.GetBytes(":$($accessToken)")) }
+$apiBase = "https://vsrm.dev.azure.com/microsofthealthoss"
+
+# Hardcoded to point to Release Defenition 23 (DICOM Server Nuget and Tag Release) under the FhirServer Project
+$definitionUri = $apiBase + "/7621b231-1a7d-4364-935b-2f72b911c43d/_apis/Release/definitions/23"
+
+# Identify the latest release
+$definition = Invoke-RestMethod -Uri $definitionUri -Method get -Headers $azureDevOpsAuthenicationHeader
+$lastRelease = $definition.lastRelease.Id
+$currentRelease = $null
+
+while(-Not($lastRelease -eq $currentRelease)) {
+    # Fetch releases currently pending for approval
+    log "Fetching approvals"
+    $approvalsUri = $apiBase + "/fhirserver/_apis/release/approvals?api-version=6.0"
+    $approvals = Invoke-RestMethod -Uri $approvalsUri -Method get -Headers $azureDevOpsAuthenicationHeader
+    $approval = @()
+    foreach($instance in $approvals.value) {
+        # Isolate the one associated with dicom-server
+        if ($instance.releaseDefinition.id -eq "23") { 
+            $approval += $instance
+        }
+    }
+
+    # Validate that only one release is pending approval - any others should be approved, rejected or queued.
+    if($approval.Count -gt 1) {
+        throw "Error: More than 1 approval at a time was unexpected"
+    }
+    elseif($approval.Count -eq 0) {
+        # If there are no pending approvals, exit as there are no new changes to release.
+        log "No pending approvals, nothing to do."
+        exit
+    }
+
+    # If there is exactly one release, identify relevant properties to base further queries on.
+    $currentRelease = $approval[0].release.id
+    $currentReleaseUrl = $approval[0].release.url + "?api-version=6.0"
+
+    log "Approval found for $($($($approval[0]).release).name)"
+
+    $updateStatusObj = $null;
+
+    # All releases that are not the latest release are rejected.
+    if(-Not($currentRelease -eq $lastRelease)) {
+        $updateStatusObj = @{
+            status="rejected"
+            comment="Rejected by automation (reason: not the latest)"
+        }
+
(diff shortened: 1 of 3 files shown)


.github/workflows/CreateRelease.ps1
```diff
@@ -0,0 +1,95 @@
...
+    log "Fetching approvals"
+    $approvalsUri = $apiBase + "/fhirserver/_apis/release/approvals?api-version=6.0"
+    $approvals = Invoke-RestMethod -Uri $approvalsUri -Method get -Headers $azureDevOpsAuthenicationHeader
+    $approval = @()
+    foreach($instance in $approvals.value) {
+        # Isolate the one associated with dicom-server
+        if ($instance.releaseDefinition.id -eq "23") { 
+            $approval += $instance
+        }
+    }
```
brandonpollett, line 47: We likely need to handle if there's a continuationtoken.  https://docs.microsoft.com/en-us/rest/api/azure/devops/release/approvals/list?view=azure-devops-rest-6.1

.github/workflows/CreateRelease.ps1
```diff
@@ -0,0 +1,95 @@
...
+
+    # Validate that only one release is pending approval - any others should be approved, rejected or queued.
+    if($approval.Count -gt 1) {
+        throw "Error: More than 1 approval at a time was unexpected"
```
brandonpollett, line 51: Does this log or just show as a failed message? Should we log the pending approvals?

.github/workflows/CreateRelease.ps1
```diff
@@ -0,0 +1,95 @@
...
+
+# Identify the latest release
+$definition = Invoke-RestMethod -Uri $definitionUri -Method get -Headers $azureDevOpsAuthenicationHeader
+$lastRelease = $definition.lastRelease.Id
+$currentRelease = $null
+
+while(-Not($lastRelease -eq $currentRelease)) {
+    # Fetch releases currently pending for approval
+    log "Fetching approvals"
+    $approvalsUri = $apiBase + "/fhirserver/_apis/release/approvals?api-version=6.0"
```
brandonpollett, line 39: Even though `statusFilter` is defaulted to `pending`, we may want to go ahead and set that here.

.github/workflows/CreateRelease.ps1
```diff
@@ -0,0 +1,95 @@
...
+    }
+    elseif($approval.Count -eq 0) {
+        # If there are no pending approvals, exit as there are no new changes to release.
+        log "No pending approvals, nothing to do."
+        exit
+    }
+
+    # If there is exactly one release, identify relevant properties to base further queries on.
+    $currentRelease = $approval[0].release.id
+    $currentReleaseUrl = $approval[0].release.url + "?api-version=6.0"
```
brandonpollett, line 61: If we're using the same `api-version` everywhere, we should probably set it in a variable above.

.github/workflows/CreateRelease.ps1
```diff
@@ -0,0 +1,95 @@
...
+
+# Hardcoded to point to Release Defenition 23 (DICOM Server Nuget and Tag Release) under the FhirServer Project
+$definitionUri = $apiBase + "/7621b231-1a7d-4364-935b-2f72b911c43d/_apis/Release/definitions/23"
+
+# Identify the latest release
+$definition = Invoke-RestMethod -Uri $definitionUri -Method get -Headers $azureDevOpsAuthenicationHeader
+$lastRelease = $definition.lastRelease.Id
+$currentRelease = $null
+
+while(-Not($lastRelease -eq $currentRelease)) {
```
brandonpollett, line 36: When does this happen since if there's more than one release we throw an exception?

.github/workflows/CreateRelease.ps1
```diff
@@ -0,0 +1,95 @@
...
+            comment="Approved by automation"
+        }
+
+        log "Accepting $($($($approval[0]).release).name)"
+    }
+    
+    try {
+        # Make a patch request to the AzureDevOps' API to approve the release.
+        $patch = Invoke-WebRequest -Uri $currentReleaseUrl -Method patch -Headers $azureDevOpsAuthenicationHeader -ContentType "application/json" -Body $updateStatusObj -ErrorAction Stop
+        Start-Sleep -Milliseconds 10000
```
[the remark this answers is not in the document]
snehabandla: I think it's to wait for the patch to be applied! Otherwise the next loop will act on the same release.

.github/workflows/CreateRelease.ps1
```diff
@@ -0,0 +1,95 @@
+# Required parameter specifying the secret access token to view releases and manage approvals.
+param([Parameter(Mandatory=$true)][string]$accessToken)
+
+# Function to help log out useful information.
+function log ([parameter(mandatory)][String]$msg) { Add-Content log.txt $msg }
```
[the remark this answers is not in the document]
snehabandla: I'm not sure where the write-host gets piped/saved. Here, I'm ensuring that logged data gets uploaded and persisted in some way which helps for debugging.

.github/workflows/oss-release.yml
```diff
@@ -0,0 +1,16 @@
...
+  - name: Script to Create OSS Release
+    uses: actions/upload-artifact@v2
+    with: # Upload generated log file
+      name: log
+      path: log.txt
+  - shell: pwsh
+    env: # Use secret in repo to make necessary AzureDevOps calls.
+      AZUREDEVOPS_PAT: ${{ secrets.AZUREDEVOPS_PAT }}
+    run: | # Run powershell script to manage API calls to generate Release.
+      CreateRelease.ps1 -acessToken "$env.AZUREDEVOPS_PAT"
```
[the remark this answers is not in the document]
snehabandla: For now I've uploaded it as a secret. It has a one year expiry!

.github/workflows/CreateRelease.ps1
snehabandla: So my understanding of how this works is that only one release is ever in a pending state, all other releases are queued! So if we have 10 releases after the last released , 1 would be pending and 9 would be queued. The API called here only returns the ones that are pending.

.github/workflows/CreateRelease.ps1
snehabandla: Will see how to handle that! I also wanted to see if I can use the filter in the url. For now though, FHIR does not handle the continuation token!

---
In reply to: [516980400](https://github.com/microsoft/dicom-server/pull/410#discussion_r516980400) [](ancestors = 516980400)

.github/workflows/CreateRelease.ps1
```diff
@@ -0,0 +1,95 @@
...
+
+# Generate a log file.
+New-Item ./log.txt -ItemType File -Force
+
+# Releases are only created every other week even though the action occurs every week.
+# Identify if this is the second week of a sprint.
+$currentDate = Get-Date
+$currentDate = $currentDate.ToUniversalTime().Date
+$firstRun = ([datetime]::ParseExact('10-23-2020', 'MM-dd-yyyy', $null)).Date
+$weeksSince = [math]::floor(($currentDate - $firstRun).TotalDays / 7)
```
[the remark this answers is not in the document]
snehabandla: I think the reason to not do a %14 is because to allow the schedule to change. This way, it could be a tuesday initially and then change to a wednesday and this script wouldn't have to change, just the schedule for the github action.

---
In reply to: [516969486](https://github.com/microsoft/dicom-server/pull/410#discussion_r516969486) [](ancestors = 516969486)

.github/workflows/CreateRelease.ps1
```diff
@@ -0,0 +1,95 @@
...
+    }
+    
+    try {
+        # Make a patch request to the AzureDevOps' API to approve the release.
+        $patch = Invoke-WebRequest -Uri $currentReleaseUrl -Method patch -Headers $azureDevOpsAuthenicationHeader -ContentType "application/json" -Body $updateStatusObj -ErrorAction Stop
+        Start-Sleep -Milliseconds 10000
+    }
+    catch {
+        # Any errors are thrown and will be handled by the github action.
+        throw "Error: Patch request to update $($($($approval[0]).release).name) as $($updateStatusObj.Status) failed with: $_.Exception.patch.StatusCode.value_"
```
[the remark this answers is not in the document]
snehabandla: added a log statement

.github/workflows/CreateRelease.ps1
```diff
@@ -0,0 +1,95 @@
+# Required parameter specifying the secret access token to view releases and manage approvals.
+param([Parameter(Mandatory=$true)][string]$accessToken)
```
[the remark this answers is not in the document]
snehabandla: updated this!

.github/workflows/CreateRelease.ps1
snehabandla: added a log statement with pending approvals.

.github/workflows/CreateRelease.ps1
snehabandla: updated!

.github/workflows/CreateRelease.ps1
snehabandla: added a variable.

brandonpollett (approved):

PR #421, opened 2020-11-09
Adding BirthDate to DicomCast

snehabandla:
## Description
If PatientBirthDate is available, add when creating a patient. It will not be updated if the patient already exists in Fhir

## Related issues
[User Story 76990](https://microsofthealth.visualstudio.com/Health/_workitems/edit/76990)


converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/FhirTransaction/Patient/PatientPipelineStep.cs
```diff
@@ -71,9 +71,10 @@ public async Task PrepareRequestAsync(FhirTransactionContext context, Cancellati
             _patientSynchronizer.Synchronize(dataset, patient);
 
             if (requestMode == FhirTransactionRequestMode.None &&
-                !existingPatient.IsExactly(patient))
+                !existingPatient.IsEquivalent(patient))
```
jackliums, line 74: Can you explain what these logic supposed to do?

snehabandla: This validates that the existing patient has the same identifiers (name and gender). This is so that a patient's birthdate is not taken into consideration when comparing between Dicom and Fhir (since we were concerned about inaccuracies and decided Fhir should be the source of truth).

jackliums: I don't think this works because if the name or gender changes, it wouldn't detect it as update right?

jackliums: Is this to make sure not to overwrite existing birthday? If that's the case, why not only set the birthday in `PatientBirthDateSynchronizer` if the Patient.Birthday is null?

snehabandla: I think we don't even want to overwrite the birthdate if the patient existed before hand! And it should work since it still compares name and gender. If they are not deemed to be the same, it continues in here to set the mode to update.

jackliums: So what happens if a new field in Patient is being synced? Then the person has to known to also update this extension method to include that new property. The whole point of using the framework provided IsExactly is so that we don't have to have these logic manually written and let the system determine whether some property was updated or not.

Can you schedule some time with me to help understand what the code is trying to do?

snehabandla: Sure! Let me set up time today.

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/FhirTransaction/Patient/PatientPipelineStep.cs
```diff
@@ -68,7 +68,7 @@ public async Task PrepareRequestAsync(FhirTransactionContext context, Cancellati
                 requestMode = FhirTransactionRequestMode.Create;
             }
 
-            _patientSynchronizer.Synchronize(dataset, patient);
+            _patientSynchronizer.Synchronize(dataset, patient, requestMode);
```
jackliums, line 71: I don't think this will give you the right result you want right?

The only two possible values here are None or Create but "None" here doesn't really mean "None". I think what you want is a flag indicating this is a new resource or not so why not pass in as a boolean?

snehabandla: I'm working here on the assumption that the request mode will be 'Create' when we are creating. Do you think that is invalid?

snehabandla: updated code to address this!

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/FhirTransaction/Patient/PatientBirthDateSynchronizer.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            if (newPatient)
+            {
+                EnsureArg.IsNotNull(dataset, nameof(dataset));
+                EnsureArg.IsNotNull(patient, nameof(patient));
+
+                FhirDateTime dicomPatientBirthDate = dataset.GetDatePropertyIfNotDefaultValue(DicomTag.PatientBirthDate);
+                if (dicomPatientBirthDate != null)
+                {
+                    patient.BirthDate = dicomPatientBirthDate.ToString();
```
jackliums, line 29: One thing you might want to consider is invalid date (for example, 2001/02/29). I am not sure if creating a new FhirDateTime would fail or not. If it does indeed fail, then this step will always fail, blocking the rest of the synchronization. Ideally the data is clean and birthdate is correctly but we have all seen bad data coming in. Just something to think about.

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/FhirTransaction/Patient/PatientBirthDateSynchronizer.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        /// <inheritdoc/>
+        public void Synchronize(DicomDataset dataset, Patient patient, bool newPatient)
+        {
+            if (newPatient)
+            {
+                EnsureArg.IsNotNull(dataset, nameof(dataset));
+                EnsureArg.IsNotNull(patient, nameof(patient));
+
+                FhirDateTime dicomPatientBirthDate = dataset.GetDatePropertyIfNotDefaultValue(DicomTag.PatientBirthDate);
```
jackliums, line 26: nit: FHIR has a Date object type (`Hl7.Fhir.Model.Date`) that only stores the date and you can access throught `BirthDateElement` property. You can avoid making into an object then serialize back into string.

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/FhirTransaction/Patient/IPatientSynchronizer.cs
```diff
@@ -18,6 +18,7 @@ public interface IPatientSynchronizer
         /// </summary>
         /// <param name="dataset">The DICOM properties.</param>
         /// <param name="patient">The <see cref="Patient"/> resource.</param>
-        void Synchronize(DicomDataset dataset, Patient patient);
+        /// <param name="newPatient">Flag to determine whether or not the patient being synchronized is new.</param>
+        void Synchronize(DicomDataset dataset, Patient patient, bool newPatient);
```
jackliums, line 22: nit: for boolean variables, usually the naming starts with a question (so like `isNewPatient`).

jackliums (approved):

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/FhirTransaction/Patient/PatientBirthDateSynchronizer.cs
snehabandla: The creation of the date does not fail per my tests! It will only fail if you call to assert whether or not it is a bad value.

snehabandla: Added an assert in the extension so it returns null if the date is invalid in Fhir.

PR #422, opened 2020-11-09
Single deploy ARM template for fhir, dicom, dicomcast 

moiradillon12:
## Description
ARM template to deploy fhir, dicom & dicom cast at once 
- updated to create one app service plan for fhir & dicom 
- doesn't include auth

## Related issues
[AB#76829](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/76829)
[AB#77268](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/77268)

## Testing
Deployed via Azure Portal


samples/templates/dicomcast-azuredeploy.json
```diff
@@ -0,0 +1,385 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
```
smithago, line 2: how do we keep this in sync with the dicom, fhir and dicom-cast templates? Is there a way to compose them?

samples/templates/dicomcast-azuredeploy.json
```diff
@@ -0,0 +1,385 @@
+{
```
smithago, line 1: Can we change the title of the PR?

moiradillon12: done

samples/templates/dicomcast-azuredeploy.json
moiradillon12: @smithago  investigating how to do this , thanks!

moiradillon12: @smithago I talked to Brandon, he recommended we could have the template run as part of a scheduled pipeline to see if it is still working. While I would probably be the one who makes changes to dicom or dicom cast template, I wouldn't catch a fhir change that would break template. 

What do you think of adding it into a scheduled pipeline? Thanks

moiradillon12:
@smithago should I keep these templates in this folder or move to converter/dicomcast/samples?

StevenBorg (approved):
I didn't validate the deployment without enabling App Insights, but we may want to do so, given the recent Issue.  That said, I believe this should go to Prod to be re-tested (link in PR is to the final location, so can't do a one-click deploy until it's in place).

PR #427, opened 2020-11-11
AAD V2 and Dicom to Fhir doc updates

abdullah248:
## Description
Update docs for how to use AAD v2 and also to verify using managed identity to set up dicom-server, dicom-cast, and fhir.


StevenBorg (approved):
Looks good.  I didn't verify the links, but they all look valid.

moiradillon12 (approved):
thanks abdullah!

PR #441, opened 2020-11-20
Update usages of IOptions to IOptionsSnapshot and IOptionsMonitor

brandonpollett:
## Description
Update usages of IOptions to IOptionsSnapshot and IOptionsMonitor.

## Related issues
Addresses [issue #].

## Testing

### DIFF
--- a/src/Microsoft.Health.Dicom.Api/Features/Audit/AuditLogger.cs
+++ b/src/Microsoft.Health.Dicom.Api/Features/Audit/AuditLogger.cs
@@ -33,17 +33,17 @@ namespace Microsoft.Health.Dicom.Api.Features.Audit
             "CorrelationId: {CorrelationId}" + Environment.NewLine +
             "Claims: {Claims}";
 
-        private readonly SecurityConfiguration _securityConfiguration;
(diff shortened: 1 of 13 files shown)


src/Microsoft.Health.Dicom.Api/Features/Audit/AuditLogger.cs
```diff
@@ -33,17 +33,17 @@ public class AuditLogger : IAuditLogger
             "CorrelationId: {CorrelationId}" + Environment.NewLine +
             "Claims: {Claims}";
 
-        private readonly SecurityConfiguration _securityConfiguration;
+        private readonly IOptionsMonitor<SecurityConfiguration> _securityConfiguration;
```
smithago, line 36: SecurityConfiguration is referenced in SecurityModule when updating the serviceCollection. We will need something there that can refresh the authority for the changes I am making. Will OptionMonitor work there?

brandonpollett: I need to revisit that code and see if the ability to reload those settings beyond startup ever made it in. I once had a branch that would allow you to do that and there was some work in another area that I think wanted to do that.

PR #445, opened 2020-11-25
Fix bug for TimezoneOffset bug

abdullah248:
## Description
Update the the format specifier to match c# documentation and fixed the test cases to ensure proper testing. In previous version the format specifier was incorrect and the test case was generated non-statically using the specifier so the bug was not caught.

## Related issues
AB#77746

## Testing
Updated the unit tests.


smithago (approved):

snehabandla (approved):
:shipit:

PR #450, opened 2020-11-27
increase min name length

jeff3parker:
increase min serviceName length=6 so that default-azuredeploy.json does not error out due to too short name (min size = 11) for name+id.

## Description
When the serviceName length <= 5 results in an InvalidTemplate error in default-azuredeploy.json.  Increasing size >= 6 resolves this issue.

## Testing
Manually tested with length 5 and then length 6-8 to verify.


smithago (approved):

StevenBorg:
Thank you, @jeff3parker for both the find and the update!  We greatly appreciate it!  :-)

PR #451, opened 2020-12-01
Minor Changes to Formatting

snehabandla:
AB#75762


.github/workflows/CreateRelease.ps1
```diff
@@ -72,25 +72,25 @@ $currentRelease = $null
     # All releases that are not the latest release are rejected.
     if(-Not($currentRelease -eq $lastRelease)) {
         $updateStatusObj = @{
-            status="rejected"
-            comment="Rejected by automation (reason: not the latest)"
-        }
+            "status" = "rejected"
```
abdullah248, line 75: I think the quotes for the keys are optional. I tested it out and it worked with and without them

abdullah248 (approved):

PR #452, opened 2020-12-01
Add Accession Number

pengchen0692:
## Description
Update document for Accession Number change

The Property column is FHIR path (http://hl7.org/fhirpath/)

## Related issues
Addresses AB#77787

### DIFF
--- a/docs/concepts/dicom-cast.md
+++ b/docs/concepts/dicom-cast.md
@@ -57,6 +57,7 @@ The current implementation of DICOM Cast has the following mappings:
 | ImagingStudy.series.instance.uid | (0008,0018) | SOPInstanceUID | |
 | ImagingStudy.series.instance.sopClass | (0008,0016) | SOPClassUID | |
 | ImagingStudy.series.instance.number | (0020,0013) | InstanceNumber | |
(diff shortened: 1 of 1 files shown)


docs/concepts/dicom-cast.md
```diff
@@ -57,6 +57,7 @@ The current implementation of DICOM Cast has the following mappings:
 | ImagingStudy.series.instance.uid | (0008,0018) | SOPInstanceUID | |
 | ImagingStudy.series.instance.sopClass | (0008,0016) | SOPClassUID | |
 | ImagingStudy.series.instance.number | (0020,0013) | InstanceNumber | |
+| ImagingStudy.identifier.where(system = 'http://ginormoushospital.org/accession') | (0008,0050) | Accession Number | System name could change on different organization. |
```
abdullah248, line 60: Maybe just put a template url like <organization url>/accession so it does not link to an actual place?

pengchen0692: > Maybe just put a template url like {organization url}/accession so it does not link to an actual place?

The FHIR path has been updated, since system always be empty in our case.

abdullah248 (approved):

PR #461, opened 2020-12-14
Add CustomTag Cache

pengchen0692:
## Description
Add custom tag cache which cache custom tag list in memory , and refresh when expired

## Related issues
Addresses AB#78143

## Testing
Unit tests passed locally

### DIFF
--- a/src/Microsoft.Health.Dicom.Core.UnitTests/Features/CustomTag/CustomTagCacheTests.cs
+++ b/src/Microsoft.Health.Dicom.Core.UnitTests/Features/CustomTag/CustomTagCacheTests.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Health.Dicom.Core.UnitTests.Features.CustomTag
             List<CustomTagEntry> expectedResult = new List<CustomTagEntry>();
             expectedResult.Add(new CustomTagEntry() { Version = expectedVersion });
             ICustomTagStore customTagStore = CreateCustomTagStore(expectedResult, expectedResult);
-            CustomTagCache customTagCache = new CustomTagCache(customTagStore, DefaultExpirationInterval);
+            CustomTagListCache customTagCache = new CustomTagListCache(customTagStore, DefaultExpirationInterval);
             CustomTagList customTags = await customTagCache.GetCustomTagsAsync();
             Assert.Equal(expectedResult, customTags.CustomTags);
             Assert.NotNull(customTagCache.LastRefreshTime);
@@ -37,10 +37,12 @@ namespace Microsoft.Health.Dicom.Core.UnitTests.Features.CustomTag
             List<CustomTagEntry> expectedResult2 = new List<CustomTagEntry>() { CustomTagTestHelper.CreateCustomTagEntry(version: expectedVersion2) };
             ICustomTagStore customTagStore = CreateCustomTagStore(expectedResult1, expectedResult2);
 
-            CustomTagCache customTagCache = new CustomTagCache(customTagStore, TimeSpan.FromSeconds(1));
+            CustomTagListCache customTagCache = new CustomTagListCache(customTagStore, TimeSpan.FromMilliseconds(500));
             CustomTagList customTags1 = await customTagCache.GetCustomTagsAsync();
             DateTimeOffset lastRefreshTime = customTagCache.LastRefreshTime.Value;
-            await Task.Delay(TimeSpan.FromSeconds(2));
+
+            // Wait for 1 seconds for cache become invalid
+            await Task.Delay(TimeSpan.FromSeconds(1));
 
             CustomTagList customTags2 = await customTagCache.GetCustomTagsAsync();
             Assert.Equal(expectedResult2, customTags2.CustomTags);
@@ -55,7 +57,7 @@ namespace Microsoft.Health.Dicom.Core.UnitTests.Features.CustomTag
             List<CustomTagEntry> expectedResult2 = new List<CustomTagEntry>() { CustomTagTestHelper.CreateCustomTagEntry(expectedVersion2) };
             ICustomTagStore customTagStore = CreateCustomTagStore(expectedResult1, expectedResult2);
 
-            CustomTagCache customTagCache = new CustomTagCache(customTagStore, TimeSpan.FromSeconds(6));
+            CustomTagListCache customTagCache = new CustomTagListCache(customTagStore, DefaultExpirationInterval);
             await customTagCache.GetCustomTagsAsync();
             DateTimeOffset lastRefreshTime = customTagCache.LastRefreshTime.Value;
             await Task.Delay(TimeSpan.FromSeconds(1));
@@ -73,7 +75,7 @@ namespace Microsoft.Health.Dicom.Core.UnitTests.Features.CustomTag
             List<CustomTagEntry> expectedResult2 = new List<CustomTagEntry>() { CustomTagTestHelper.CreateCustomTagEntry(expectedVersion2) };
             ICustomTagStore customTagStore = CreateCustomTagStore(expectedResult1, expectedResult2);
 
-            CustomTagCache customTagCache = new CustomTagCache(customTagStore, DefaultExpirationInterval);
+            CustomTagListCache customTagCache = new CustomTagListCache(customTagStore, DefaultExpirationInterval);
             await customTagCache.GetCustomTagsAsync();
             DateTimeOffset lastRefreshTime = customTagCache.LastRefreshTime.Value;
             await Task.Delay(TimeSpan.FromSeconds(1));
--- a/src/Microsoft.Health.Dicom.Core.UnitTests/Features/CustomTag/CustomTagCacheTests.cs
+++ b/src/Microsoft.Health.Dicom.Core.UnitTests/Features/CustomTag/CustomTagCacheTests.cs
@@ -88,10 +88,10 @@ namespace Microsoft.Health.Dicom.Core.UnitTests.Features.CustomTag
             customTagStore.TryRefreshCustomTags(out Arg.Any<CustomTagList>(), default)
               .ReturnsForAnyArgs(true);
 
-            customTagStore.WhenForAnyArgs(x => x.TryRefreshCustomTags(out Arg.Any<CustomTagList>(), default))
-                .Do(Callback.First(x => x[0] = new CustomTagList(firstCallReturn))
-                .ThenKeepDoing(x => x[0] = new CustomTagList(secondCallReturn))
-                .AndAlways(x => { }));
+            customTagStore.WhenForAnyArgs(store => store.TryRefreshCustomTags(out Arg.Any<CustomTagList>(), default))
+                .Do(Callback.First(callInfo => callInfo[0] = new CustomTagList(firstCallReturn))
+                .ThenKeepDoing(callInfo => callInfo[0] = new CustomTagList(secondCallReturn))
+                .AndAlways(callInfo => { }));
             return customTagStore;
         }
     }
--- a/src/Microsoft.Health.Dicom.Core.UnitTests/Features/CustomTag/CustomTagCacheTests.cs
+++ b/src/Microsoft.Health.Dicom.Core.UnitTests/Features/CustomTag/CustomTagCacheTests.cs
@@ -1,0 +1,98 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.Health.Dicom.Core.Features.CustomTag;
+using NSubstitute;
+using Xunit;
+
+namespace Microsoft.Health.Dicom.Core.UnitTests.Features.CustomTag
+{
+    public class CustomTagCacheTests
+    {
+        private static readonly TimeSpan DefaultExpirationInterval = TimeSpan.FromSeconds(30);
+
+        [Fact]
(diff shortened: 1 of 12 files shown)


src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagCache.cs
```diff
@@ -0,0 +1,64 @@
...
+            EnsureArg.IsNotNull(customTagStore);
+            _customtags = null;
+            _lastRefreshTime = null;
+            _customTagStore = customTagStore;
+            _expirationInterval = expirationInterval;
+        }
+
+        public DateTimeOffset? LastRefreshTime { get => _lastRefreshTime; }
+
+        public async Task<CustomTagList> GetCustomTagsAsync(CancellationToken cancellationToken = default)
```
snehabandla, line 31: I think this should always require a cancellation token - there shouldn't be a default value.

src/Microsoft.Health.Dicom.Core.UnitTests/Features/CustomTag/CustomTagCacheTests.cs
```diff
@@ -0,0 +1,98 @@
...
+            Assert.Equal(lastRefreshTime, customTagCache.LastRefreshTime);
+        }
+
+        [Fact]
+        public async Task GivenRefreshedCustomTagCacheWithEmptyList_WhenGetCustomTagsIsCalled_ShouldNotRefreshCache()
+        {
+            long expectedVersion2 = 2;
+            List<CustomTagEntry> expectedResult1 = new List<CustomTagEntry>();
+            List<CustomTagEntry> expectedResult2 = new List<CustomTagEntry>() { CustomTagTestHelper.CreateCustomTagEntry(expectedVersion2) };
+            ICustomTagStore customTagStore = CreateCustomTagStore(expectedResult1, expectedResult2);
```
snehabandla, line 74: This second value is never used right? Is there any need to initialize it?

src/Microsoft.Health.Dicom.Core.UnitTests/Features/CustomTag/CustomTagCacheTests.cs
```diff
@@ -0,0 +1,98 @@
...
+        private ICustomTagStore CreateCustomTagStore(List<CustomTagEntry> firstCallReturn, List<CustomTagEntry> secondCallReturn)
+        {
+            ICustomTagStore customTagStore = Substitute.For<ICustomTagStore>();
+            customTagStore.TryRefreshCustomTags(out Arg.Any<CustomTagList>(), default)
+              .ReturnsForAnyArgs(true);
+
+            customTagStore.WhenForAnyArgs(store => store.TryRefreshCustomTags(out Arg.Any<CustomTagList>(), default))
+                .Do(Callback.First(callInfo => callInfo[0] = new CustomTagList(firstCallReturn))
+                .ThenKeepDoing(callInfo => callInfo[0] = new CustomTagList(secondCallReturn))
+                .AndAlways(callInfo => { }));
```
snehabandla, line 94: What if you don't need to validate multiple calls?

src/Microsoft.Health.Dicom.Core.UnitTests/Features/CustomTag/CustomTagCacheTests.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        [Fact]
+        public async Task GivenRefreshedCustomTagCache_WhenGetCustomTagsIsCalled_ShouldNotRefreshCache()
+        {
+            long expectedVersion1 = 1, expectedVersion2 = 2;
+            List<CustomTagEntry> expectedResult1 = new List<CustomTagEntry>() { CustomTagTestHelper.CreateCustomTagEntry(expectedVersion1) };
+            List<CustomTagEntry> expectedResult2 = new List<CustomTagEntry>() { CustomTagTestHelper.CreateCustomTagEntry(expectedVersion2) };
+            ICustomTagStore customTagStore = CreateCustomTagStore(expectedResult1, expectedResult2);
+
+            CustomTagCache customTagCache = new CustomTagCache(customTagStore, TimeSpan.FromSeconds(6));
```
snehabandla, line 58: >TimeSpan.FromSeconds(6) [](start = 79, length = 23)

Any reason you aren't using default here?

src/Microsoft.Health.Dicom.Core.UnitTests/Features/CustomTag/CustomTagCacheTests.cs
```diff
@@ -0,0 +1,98 @@
...
+            long expectedVersion1 = 1, expectedVersion2 = 2;
+            List<CustomTagEntry> expectedResult1 = new List<CustomTagEntry>() { CustomTagTestHelper.CreateCustomTagEntry(version: expectedVersion1) };
+            List<CustomTagEntry> expectedResult2 = new List<CustomTagEntry>() { CustomTagTestHelper.CreateCustomTagEntry(version: expectedVersion2) };
+            ICustomTagStore customTagStore = CreateCustomTagStore(expectedResult1, expectedResult2);
+
+            CustomTagCache customTagCache = new CustomTagCache(customTagStore, TimeSpan.FromSeconds(1));
+            CustomTagList customTags1 = await customTagCache.GetCustomTagsAsync();
+            DateTimeOffset lastRefreshTime = customTagCache.LastRefreshTime.Value;
+            await Task.Delay(TimeSpan.FromSeconds(2));
+
```
snehabandla, line 44: Delaying for seconds could make our tests run very long, can we do shorter delays?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        public long TagId { get; set; }
+
+        public string Path { get; set; }
+
+        public string VR { get; set; }
+
+        public CustomTagLevel Level { get; set; }
+
+        public long Version { get; set; }
```
smithago, line 18: What is the version for? Can you add comments on class and properties

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagCache.cs
```diff
@@ -0,0 +1,64 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using EnsureThat;
+
+namespace Microsoft.Health.Dicom.Core.Features.CustomTag
+{
+    public class CustomTagCache : ICustomTagCache
```
smithago, line 13: comment on the class responsibility

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagCache.cs
pengchen0692: I'm not very sure if that's true for all case, I just followed how other interface does. :) 
e.g: https://github.com/microsoft/dicom-server/blob/master/src/Microsoft.Health.Dicom.Core/Features/ChangeFeed/IChangeFeedService.cs

src/Microsoft.Health.Dicom.Core/Features/CustomTag/ICustomTagStore.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    public interface ICustomTagStore
+    {
+        /// <summary>
+        /// Refresh given CustomTagList.
+        /// </summary>
+        /// <param name="customTags">The customTags.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns>True if the custom tags has been refreshed, false otherwise</returns>
+        Task<bool> TryRefreshCustomTags(out CustomTagList customTags, CancellationToken cancellationToken = default);
```
smithago, line 19: Why do we need the API to Try? Why not just

CustomTagList GetCustomTagsAsync()?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagCache.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        private CustomTagList _customtags;
+        private ICustomTagStore _customTagStore;
+        private TimeSpan _expirationInterval;
+        private DateTimeOffset? _lastRefreshTime;
+
+        public CustomTagCache(ICustomTagStore customTagStore, TimeSpan expirationInterval)
```
smithago, line 20: Why is the expiration passed in?

src/Microsoft.Health.Dicom.Core.UnitTests/Features/CustomTag/CustomTagCacheTests.cs
pengchen0692: Ideally the method should take an array as input, but  unfortunately nsubstitute doesn't accept an array as input, so if we need to create our own version, the code will be looks like:
if(ary.length==1) {}
else if(ary.length==2){}
...

so I use a simpler way: since at most we only use 2 different returns, the method take 2 return as input, and process as the code. If only have 1 return, pass same return for both them should be good.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagCache.cs
smithago: Also do you really want to implement cache now? We should really look at using .net Memory cache or something like that, this is not thread safe. We will need a lot of advanced telemetry and ops to control the expiration and size.

src/Microsoft.Health.Dicom.Core.UnitTests/Features/CustomTag/CustomTagCacheTests.cs
pengchen0692: it can use default one. let me update

src/Microsoft.Health.Dicom.Core.UnitTests/Features/CustomTag/CustomTagCacheTests.cs
pengchen0692: would 1 second be better?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/ICustomTagStore.cs
pengchen0692: the API need to handle 2 case:
1. If given customTags is latest, won't update the customTags, and return false which means not Refreshed
2. if given customTags is not latest, will update the customTags, and return true which means the customTags is refreshed.

CustomTagList GetCustomTagsAsync() doesn't have a good way to differentiate these 2.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagCache.cs
pengchen0692: The value will be passed from config file.
I prefer an centralized place for ExpirationInterval and CustomTagPropagationTime for easy management.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
pengchen0692: updated

src/Microsoft.Health.Dicom.Core/Features/CustomTag/ICustomTagStore.cs
smithago: Store layer is just a dum data access layer, basically an interface that can be easily replaced when datastore changes. Rest of the logic should be implemented in different interfaces in the core

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
smithago: Thanks. How is it used here?

smithago: Also can you give me an example of an allowed update and the effect on the re-indexing

pengchen0692: Version match to SQL TIMESTAMP type in database. Each CustomTag row in database has a version, every time it got changed, a new version is assigned by system automatically.
TIMESTAMP is a global bigint which keep increasing and never come back. The max version in CustomTag table is the version of while CustomTag table.


I'm thinking if create another table (CustomTagVersionTable) , and put the version into that is a more clear idea. We only need version of entire CustomTag table instead of each row.  The CustomTagVersionTable will look like:

CREATE TABLE dbo.CustomTagVersion (
    Version               TIMESTAMP               NOT NULL,
   -- Audit column
   LastStatusUpdatedDate                DATETIME2(7) NOT NULL,
)

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagCache.cs
pengchen0692: * CustomTagCache is foundation of the entire optimization. QIDO and STOW depends on this.
* Yes, it's not thread safe, will figure out a way to handle with
* Could you details bit more on "need a lot of advanced telemetry and ops"?

pengchen0692: details on  "need a lot of advanced telemetry and ops":
* for Metrix, need to consider miss and hit rate
* check https://docs.microsoft.com/en-us/aspnet/core/performance/caching/memory?view=aspnetcore-5.0 to see what to consider when design cache.

pengchen0692:
Close for now, will postpone the implementation after the first simple prototype come out

PR #464, opened 2020-12-15
Add CustomTagController

pengchen0692:
## Description
Add controller for custom tag

## Related issues
Addresses AB#78129

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
+++ b/src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
@@ -1,0 +1,44 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Dicom.Core.Features.CustomTag
+{
+    /// <summary>
+    /// Represent each custom tag entry has retrieved from the store.
+    /// </summary>
+    public class CustomTagEntry
+    {
+        /// <summary>
+        /// Key of this custom tag entry.
+        /// </summary>
+        public long Key { get; set; }
+
+        /// <summary>
+        /// Path of this tag. Normally it's composed of groupid and elementid.
+        /// E.g: 00100020 is path of patient id.
+        /// </summary>
+        public string Path { get; set; }
+
+        /// <summary>
+        /// VR of this tag.
+        /// </summary>
+        public string VR { get; set; }
+
+        /// <summary>
+        /// Level of this tag. Could be Study, Series or Instance.
+        /// </summary>
+        public CustomTagLevel Level { get; set; }
+
+        /// <summary>
+        /// Status of this tag.
+        /// </summary>
+        public CustomTagStatus Status { get; set; }
+
+        public override string ToString()
+        {
+            return $"Key: {Key}, Path: {Path}, VR:{VR}, Level:{Level}, Status:{Status}";
+        }
+    }
+}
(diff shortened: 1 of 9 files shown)


src/Microsoft.Health.Dicom.Core/Messages/CustomTag/AddCustomTagResponse.cs
```diff
@@ -0,0 +1,29 @@
...
+        private readonly IEnumerable<CustomTagEntry> _customTags;
+        private readonly string _job;
+
+        public AddCustomTagResponse(IEnumerable<CustomTagEntry> customTags, string job)
+        {
+            _customTags = customTags;
+            _job = job;
+        }
+
+        public IEnumerable<CustomTagEntry> CustomTags { get => _customTags; }
```
brandonpollett, line 22: nit: you can just convert these and the others like it to auto-properties

e.g. remove the private variables and replace this line with something like
```csharp
public IEnumerable<CustomTagEntry> CustomTags { get; }
```

Then in the constructor you can change `_customTags = customTags;` to `CustomTags = customTags`

src/Microsoft.Health.Dicom.Core/Messages/CustomTag/AddCustomTagResponse.cs
```diff
@@ -0,0 +1,29 @@
...
+namespace Microsoft.Health.Dicom.Core.Messages.CustomTag
+{
+    public class AddCustomTagResponse
+    {
+        private readonly IEnumerable<CustomTagEntry> _customTags;
+        private readonly string _job;
+
+        public AddCustomTagResponse(IEnumerable<CustomTagEntry> customTags, string job)
+        {
+            _customTags = customTags;
```
brandonpollett, line 18: Should we ensure these aren't null/empty?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// <summary>
+        /// Key of this custom tag entry.
+        /// </summary>
+        public long Key { get; set; }
+
+        /// <summary>
+        /// Path of this tag. Normally it's composed of groupid and elementid.
+        /// E.g: 00100020 is path of patient id.
+        /// </summary>
+        public string Path { get; set; }
```
brandonpollett, line 22: Do we need to do any sort of validation on the `Path` or `VR` properties anywhere? I assume that that's there's some format we expect these in.

src/Microsoft.Health.Dicom.Core/Messages/CustomTag/AddCustomTagResponse.cs
pengchen0692: not sure about that. Checked several XXResponse (e.g: https://github.com/microsoft/dicom-server/blob/master/src/Microsoft.Health.Dicom.Core/Messages/Store/StoreResponse.cs), none of them validate the parameters.

pengchen0692: My best guess is that these type of classes are data model, so any input should be acceptable. The validation could be in real business logic.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
pengchen0692: yes, CustomTagEntry is used when adding custom tag, the Path and VR will be validated over there.

src/Microsoft.Health.Dicom.Core/Messages/CustomTag/AddCustomTagResponse.cs
pengchen0692: resolved.

brandonpollett (approved):

PR #471, opened 2021-01-06
enable more dependabot

KnicKnic:
## Description
1. Adds more outstanding PRs to dependabot
2. Enables dependabot on all docker files
3. enables dependabot on github actions

## Related Items
AB#78592

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,4 +5,39 @@ updates:
   - package-ecosystem: "nuget"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "daily"
+      time: "09:00"
+      timezone: "America/Los_Angeles"
+    open-pull-requests-limit: 30
+  
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "daily"
+      time: "09:00"
+      timezone: "America/Los_Angeles"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -5,4 +5,39 @@ updates:
   - package-ecosystem: "nuget"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "daily"
+      time: "09:00"
+      timezone: "America/Los_Angeles"
+    open-pull-requests-limit: 30
```
brandonpollett, line 11: I think we should leave this at weekly and probably take the limit down some at like 10 or 5.

brandonpollett (approved):

.github/dependabot.yml
KnicKnic: Weekly means you are very slow to update across the board. Dependabot will only issue 5 new changes a week. And only up to a max of 5 changes. In Github Issues you can search to filter out dependabot. We had dependabot stuck for a month because no one was unblocking it. Dependabot in general should be low maintenance. Having more outstanding PR's at a time enables us to more rapidly reset all of dependencies to fresh.

I also chose to schedule dependabot before our daily standup, to ensure we could divvy them out.

PR #490, opened 2021-01-07
possible null fxcop warnings

KnicKnic:
## Description
fix some possible null fx cop warnings

## Related issues
AB#78643

## Testing
None, relying on CI

### DIFF
--- a/src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
+++ b/src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
@@ -34,6 +34,7 @@ namespace Microsoft.Health.Dicom.Api.Features.Exceptions
 
         public async Task Invoke(HttpContext context)
         {
+            EnsureArg.IsNotNull(context, nameof(context));
             ExceptionDispatchInfo exceptionDispatchInfo = null;
             try
             {
@@ -122,6 +123,7 @@ namespace Microsoft.Health.Dicom.Api.Features.Exceptions
 
         protected internal virtual async Task ExecuteResultAsync(HttpContext context, IActionResult result)
         {
+            EnsureArg.IsNotNull(result, nameof(result));
             await result.ExecuteResultAsync(new ActionContext { HttpContext = context });
         }
     }
--- a/src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
+++ b/src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
@@ -123,6 +123,7 @@ namespace Microsoft.Health.Dicom.Api.Features.Exceptions
 
         protected internal virtual async Task ExecuteResultAsync(HttpContext context, IActionResult result)
         {
+            EnsureArg.IsNotNull(context, nameof(context));
             EnsureArg.IsNotNull(result, nameof(result));
             await result.ExecuteResultAsync(new ActionContext { HttpContext = context });
         }
--- a/src/Microsoft.Health.Dicom.Api/Features/Responses/MetadataResult.cs
+++ b/src/Microsoft.Health.Dicom.Api/Features/Responses/MetadataResult.cs
@@ -19,7 +19,6 @@ namespace Microsoft.Health.Dicom.Api.Features.Responses
         public MetadataResult(RetrieveMetadataResponse response)
             : base(response.ResponseMetadata)
         {
-            EnsureArg.IsNotNull(response, nameof(response));
             _response = response;
         }
 
--- a/src/Microsoft.Health.Dicom.Api/Features/Responses/MetadataResult.cs
+++ b/src/Microsoft.Health.Dicom.Api/Features/Responses/MetadataResult.cs
@@ -17,7 +17,7 @@ namespace Microsoft.Health.Dicom.Api.Features.Responses
         private readonly RetrieveMetadataResponse _response;
 
         public MetadataResult(RetrieveMetadataResponse response)
-            : base(response.ResponseMetadata)
+            : base(EnsureArg.IsNotNull(response, nameof(response)).ResponseMetadata)
         {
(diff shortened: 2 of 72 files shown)


src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
```diff
@@ -122,6 +123,7 @@ private IActionResult GetContentResult(HttpStatusCode statusCode, string message
 
         protected internal virtual async Task ExecuteResultAsync(HttpContext context, IActionResult result)
         {
+            EnsureArg.IsNotNull(result, nameof(result));
```
brandonpollett, line 126: Ensure context? I realize it's probably not explicitly required, but we seem to be ensuring the `HttpContext` everywhere else.

src/Microsoft.Health.Dicom.Api/Features/Responses/MetadataResult.cs
```diff
@@ -19,7 +19,6 @@ public class MetadataResult : ObjectResult
         public MetadataResult(RetrieveMetadataResponse response)
             : base(response.ResponseMetadata)
         {
-            EnsureArg.IsNotNull(response, nameof(response));
```
brandonpollett, line 22: We might want to update the call to `base(response.ResponseMetadata)` to something like `base(response?.ResponseMetadata)`. I'm not sure of the implications without digging deeper, just something to think abou.t

src/Microsoft.Health.Dicom.Client/DicomWebClientExtensions.cs
```diff
@@ -114,6 +114,7 @@ public static class DicomWebClientExtensions
             int[] frames = null,
             CancellationToken cancellationToken = default)
         {
+            EnsureArg.IsNotNull(dicomWebClient, nameof(dicomWebClient));
```
brandonpollett, line 117: Are null strings valid here for the various uids?

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Query/QueryParserTests.cs
```diff
@@ -33,6 +34,7 @@ public QueryParserTests()
         [InlineData("includefield", "StudyDate, StudyTime")]
         public void GivenIncludeField_WithValidAttributeId_CheckIncludeFields(string key, string value)
         {
+            EnsureArg.IsNotNull(value, nameof(value));
```
brandonpollett, line 37: Generally for strings we do `IsNotNullOrWhiteSpace`. Also, any reason to not do the key?

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Retrieve/RetrieveMetadataHandlerTests.cs
```diff
@@ -89,6 +94,7 @@ public async Task GivenARequestWithInvalidStudyAndSeriesInstanceIdentifier_WhenR
         [InlineData("aaaa-bbbb2", " ")]
         public async Task GivenARequestWithInvalidSeriesInstanceIdentifier_WhenRetrievingInstanceMetadata_ThenDicomInvalidIdentifierExceptionIsThrown(string seriesInstanceUid, string sopInstanceUid)
         {
+            EnsureArg.IsNotNull(seriesInstanceUid, nameof(seriesInstanceUid));
             string ifNoneMatch = null;
```
brandonpollett, line 98: Check `SopInstanceUid`?

brandonpollett (approved):
Overall, I think this makes sense. Is there any reason that in a few places we're only checking the first parameter? Also, when we're doing strings, we often do `IsNotNullOrWhitespace` depending on the context.

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Query/QueryParserTests.cs
KnicKnic: I did the ones the editor warned about. Normally this was due to accessing a member, EX: value.length . And that would only lead to checking one of the values. Now in this case I do not know why the editor warned about key and not value.

src/Microsoft.Health.Dicom.Client/DicomWebClientExtensions.cs
KnicKnic: Does not appear so

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Retrieve/RetrieveMetadataHandlerTests.cs
KnicKnic: Again same as above, only checking minimal necessary. This function actually calls .trim . Also this is tests I tend to allow as much as possible to facilitate getting results from the underlying code.

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Query/QueryParserTests.cs
KnicKnic: resolving since tests

src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
KnicKnic: addressed

src/Microsoft.Health.Dicom.Api/Features/Responses/MetadataResult.cs
KnicKnic: base(EnsureArg.IsNotNull(response, nameof(response).ResponseMetadata)

decided this would be best

src/Microsoft.Health.Dicom.Client/DicomWebClientExtensions.cs
KnicKnic: addressed

PR #492, opened 2021-01-08
[Dicom-Cast] Partial validation and storage

abdullah248:
## Description
Implement partial validation of dicom meta data. 

If partial validation is enabled then in the case of non required meta data from change feed entry being invalid, then store the error to table storage and still process the change feed entry and send to fhir. 

If partial validation is not enabled then in the case of any meta data from change feed entry being invalid then the processing stops. Currently do not store into table storage, but will in future when continue processing of other change feed entries.

## Related issues
AB#78068

## Testing
Locally tested by uploading dicom files with invalid data.


converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/FhirTransaction/ImagingStudy/ImagingStudyPropertySynchronizer.cs
```diff
@@ -30,17 +48,47 @@ public void Synchronize(FhirTransactionContext context, ImagingStudy imagingStud
...
-            AddImagingStudyEndpoint(imagingStudy, context);
-            AddModality(imagingStudy, dataset);
-            AddNote(imagingStudy, dataset);
-            AddAccessionNumber(imagingStudy, dataset);
+        private void SynchronizePropertiesAsync(ImagingStudy imagingStudy, FhirTransactionContext context, bool required, Action<ImagingStudy, FhirTransactionContext> synchronizeAction, CancellationToken cancellationToken = default)
+        {
+            try
+            {
+                synchronizeAction(imagingStudy, context);
+            }
```
KnicKnic, line 63: This function SynchronizePropertiesAsync feels repeated many times can it be abstracted

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/FhirTransaction/ImagingStudy/ImagingStudyInstancePropertySynchronizer.cs
```diff
@@ -26,14 +44,54 @@ public void Synchronize(FhirTransactionContext context, ImagingStudy.InstanceCom
                 return;
             }
 
-            // Add sopclass to instance
+            SynchronizePropertiesAsync(instance, context, true, AddSopClass, cancellationToken);
```
KnicKnic, line 47: can the relationships between instance, synchronizeAction, and required be expressed more directly as a property or table mapping. and then added to each class?

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/FhirTransaction/ImagingStudy/IImagingStudyInstancePropertySynchronizer.cs
```diff
@@ -14,6 +15,7 @@ public interface IImagingStudyInstancePropertySynchronizer
         /// </summary>
         /// <param name="context">The transaction context.</param>
         /// <param name="instance">The instance component within study.</param>
-        void Synchronize(FhirTransactionContext context, ImagingStudy.InstanceComponent instance);
+        /// <param name="cancellationToken">The cancellation token.</param>
+        void Synchronize(FhirTransactionContext context, ImagingStudy.InstanceComponent instance, CancellationToken cancellationToken);
```
pengchen0692, line 19: looks like the operations are changed to async. in that case, 
1. should return Task instead of void
2. should name changes to xxxAsync

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/FhirTransaction/ImagingStudy/ImagingStudyInstancePropertySynchronizer.cs
```diff
@@ -26,14 +44,54 @@ public void Synchronize(FhirTransactionContext context, ImagingStudy.InstanceCom
...
+            SynchronizePropertiesAsync(instance, context, false, AddInstanceNumber, cancellationToken);
```
pengchen0692, line 48: is it ok to add parameter names when calling SynchronizePropertiesAsync?
the meaning of true/false and AddInstanceNumber/AddSopClass is not very explicit when reading method invocation.

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/FhirTransaction/ImagingStudy/ImagingStudyPropertySynchronizer.cs
abdullah248: Thanks, I made it a generic and move it to a helper class

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/FhirTransaction/ImagingStudy/ImagingStudyInstancePropertySynchronizer.cs
abdullah248: This is a good idea, I made it so that it is mapped in a list and then we just loop through that instead

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/FhirTransaction/ImagingStudy/IImagingStudyInstancePropertySynchronizer.cs
abdullah248: Updated, thanks for the catch

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/FhirTransaction/ImagingStudy/ImagingStudyInstancePropertySynchronizer.cs
abdullah248: I made the argument names more clear so hopefully it shouldnt be an issue anymore

abdullah248:
Thanks for the comments on this pr, I updated as suggested and pushed. However, since these changes are also in pr #496 I just pushed it to that branch. I will close this pr and follow up on that one.

PR #494, opened 2021-01-11
Adding get custom tag to controller

snehabandla:
## Description
Adding support to get custom tags to controller.

## Related issues
[AB#78173](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/78173)
[AB#78174](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/78174)

### DIFF
--- a/src/Microsoft.Health.Dicom.Api/Controllers/CustomTagController.cs
+++ b/src/Microsoft.Health.Dicom.Api/Controllers/CustomTagController.cs
@@ -68,7 +68,7 @@ namespace Microsoft.Health.Dicom.Api.Controllers
         {
             _logger.LogInformation("DICOM Web Get Custom Tag request received for all custom tags");
 
-            GetCustomTagResponse response = await _mediator.GetCustomTagsAsync(null, HttpContext.RequestAborted);
+            GetAllCustomTagsResponse response = await _mediator.GetAllCustomTagsAsync(null, HttpContext.RequestAborted);
 
             return StatusCode(
                 (int)HttpStatusCode.OK, response);
@@ -90,9 +90,9 @@ namespace Microsoft.Health.Dicom.Api.Controllers
         [AuditEventType(AuditEventSubType.GetCustomTag)]
         public async Task<IActionResult> GetTagAsync(string tagPath)
         {
-            _logger.LogInformation("DICOM Web Get Custom Tag request received for custom tag: {0}", tagPath);
+            _logger.LogInformation("DICOM Web Get Custom Tag request received for custom tag: {tagPath}");
 
-            GetCustomTagResponse response = await _mediator.GetCustomTagsAsync(tagPath, HttpContext.RequestAborted);
+            GetCustomTagResponse response = await _mediator.GetCustomTagAsync(tagPath, HttpContext.RequestAborted);
 
             return StatusCode(
                 (int)HttpStatusCode.OK, response);
--- a/src/Microsoft.Health.Dicom.Api/Controllers/CustomTagController.cs
+++ b/src/Microsoft.Health.Dicom.Api/Controllers/CustomTagController.cs
@@ -8,6 +8,7 @@ using System.Net;
 using System.Threading.Tasks;
 using EnsureThat;
 using MediatR;
+using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Logging;
 using Microsoft.Health.Api.Features.Audit;
@@ -37,9 +38,9 @@ namespace Microsoft.Health.Dicom.Api.Controllers
             _logger = logger;
         }
 
+        [ProducesResponseType(typeof(JsonResult), (int)HttpStatusCode.Accepted)]
         [HttpPost]
         [Route(KnownRoutes.CustomTagRoute)]
-        [ProducesResponseType(typeof(JsonResult), (int)HttpStatusCode.Accepted)]
         [AuditEventType(AuditEventSubType.CustomTag)]
         public async Task<IActionResult> PostAsync([FromBody] IEnumerable<CustomTagEntry> customTags)
         {
@@ -49,6 +50,36 @@ namespace Microsoft.Health.Dicom.Api.Controllers
 
             return StatusCode(
                (int)HttpStatusCode.Accepted, response);
+        }
+
+        /// <summary>
+        /// Handles requests to get individual/all custom tags.
+        /// </summary>
+        /// <param name="tagPath">Path for requested custom tag. Value is null when retrieving all custom tags.</param>
+        /// <returns>
+        /// Returns Bad Request if given path can't be parsed. Returns Not Found if given path doesn't map to a stored
+        /// custom tag or if no custom tags are stored. Returns OK with a JSON body of requested tag(s) in other cases.
+        /// </returns>
+        [ProducesResponseType(typeof(JsonResult), (int)HttpStatusCode.OK)]
+        [ProducesResponseType((int)HttpStatusCode.BadRequest)]
+        [ProducesResponseType((int)HttpStatusCode.NotFound)]
+        [HttpGet]
+        [Route(KnownRoutes.CustomTagRoute)]
+        [AuditEventType(AuditEventSubType.CustomTag)]
+        public async Task<IActionResult> GetAsync(string tagPath)
+        {
+            if (tagPath == null)
+            {
+                _logger.LogInformation("DICOM Web Get Custom Tag request received for all custom tags");
+            }
+            else
+            {
+                _logger.LogInformation("DICOM Web Get Custom Tag request received for custom tag: {0}", tagPath);
+            }
+
+            GetCustomTagResponse response = await _mediator.GetCustomTagsAsync(tagPath, HttpContext.RequestAborted);
+
+            return StatusCode((int)HttpStatusCode.OK, null);
         }
     }
 }
--- a/src/Microsoft.Health.Dicom.Api/Controllers/CustomTagController.cs
+++ b/src/Microsoft.Health.Dicom.Api/Controllers/CustomTagController.cs
@@ -41,7 +41,7 @@ namespace Microsoft.Health.Dicom.Api.Controllers
         [ProducesResponseType(typeof(JsonResult), (int)HttpStatusCode.Accepted)]
         [HttpPost]
         [Route(KnownRoutes.CustomTagRoute)]
-        [AuditEventType(AuditEventSubType.CustomTag)]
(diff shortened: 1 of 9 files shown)


src/Microsoft.Health.Dicom.Api/Controllers/CustomTagController.cs
```diff
@@ -50,5 +51,35 @@ public async Task<IActionResult> PostAsync([FromBody] IEnumerable<CustomTagEntry
...
+        /// Returns Bad Request if given path can't be parsed. Returns Not Found if given path doesn't map to a stored
+        /// custom tag or if no custom tags are stored. Returns OK with a JSON body of requested tag(s) in other cases.
+        /// </returns>
+        [ProducesResponseType(typeof(JsonResult), (int)HttpStatusCode.OK)]
+        [ProducesResponseType((int)HttpStatusCode.BadRequest)]
+        [ProducesResponseType((int)HttpStatusCode.NotFound)]
+        [HttpGet]
+        [Route(KnownRoutes.CustomTagRoute)]
+        [AuditEventType(AuditEventSubType.CustomTag)]
+        public async Task<IActionResult> GetAsync(string tagPath)
```
pengchen0692, line 69: should the input be Id instead of tagpath?
The List api will returns all custom tag information where user can find tag id.

snehabandla: What do we do in the scenario where there are multiple tag paths with the same id? I figured tag path would return the most concrete answer for the user.

src/Microsoft.Health.Dicom.Api/Controllers/CustomTagController.cs
```diff
@@ -50,5 +51,35 @@ public async Task<IActionResult> PostAsync([FromBody] IEnumerable<CustomTagEntry
...
+        {
+            if (tagPath == null)
```
pengchen0692, line 71: would String.IsNullOrWhiteSpace be better?

src/Microsoft.Health.Dicom.Core/Messages/CustomTag/GetCustomTagResponse.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        public GetCustomTagResponse(IEnumerable<CustomTagEntry> customTagEntries)
+        {
+            CustomTags = customTagEntries;
+        }
+
+        /// <summary>
+        /// Retrieved custom tag when request specifies exact custom tag path.
+        /// </summary>
+        public CustomTagEntry CustomTag { get; }
```
pengchen0692, line 26: I'm curious why we need 2 properties? 
Could we also utilize CustomTags for single custom tag retrieving?

src/Microsoft.Health.Dicom.Api/Controllers/CustomTagController.cs
```diff
@@ -50,5 +51,35 @@ public async Task<IActionResult> PostAsync([FromBody] IEnumerable<CustomTagEntry
...
+                _logger.LogInformation("DICOM Web Get Custom Tag request received for all custom tags");
+            }
+            else
+            {
+                _logger.LogInformation("DICOM Web Get Custom Tag request received for custom tag: {0}", tagPath);
+            }
+
+            GetCustomTagResponse response = await _mediator.GetCustomTagsAsync(tagPath, HttpContext.RequestAborted);
+
+            return StatusCode((int)HttpStatusCode.OK, null);
```
pengchen0692, line 82: I guess this "null" will be fulfilled later? could you add a TODO here for future track?

src/Microsoft.Health.Dicom.Core/Messages/CustomTag/GetCustomTagResponse.cs
snehabandla: Serializing an IEnumarable returns a list of json entries which isn't accurate per what a user requested when asking for a single tag.

src/Microsoft.Health.Dicom.Api/Controllers/CustomTagController.cs
snehabandla: ah, i forgot to update this! Will change it to be the response.

src/Microsoft.Health.Dicom.Api/Controllers/CustomTagController.cs
```diff
@@ -50,5 +51,51 @@ public async Task<IActionResult> PostAsync([FromBody] IEnumerable<CustomTagEntry
...
+        /// <returns>
+        /// Returns Bad Request if given path can't be parsed. Returns Not Found if given path doesn't map to a stored
+        /// custom tag or if no custom tags are stored. Returns OK with a JSON body of all tags in other cases.
+        /// </returns>
+        [ProducesResponseType(typeof(JsonResult), (int)HttpStatusCode.OK)]
+        [ProducesResponseType((int)HttpStatusCode.NotFound)]
+        [HttpGet]
+        [Route(KnownRoutes.CustomTagRoute)]
+        [AuditEventType(AuditEventSubType.GetAllCustomTags)]
+        public async Task<IActionResult> GetAllTagsAsync()
```
pengchen0692, line 67: I'm thinking if we could do similar thing like GetFramesAsync (https://github.com/microsoft/dicom-server/blob/master/src/Microsoft.Health.Dicom.Api/Controllers/RetrieveController.cs) which can support return multiple either multiple or single frame

src/Microsoft.Health.Dicom.Api/Controllers/CustomTagController.cs
```diff
@@ -50,5 +51,51 @@ public async Task<IActionResult> PostAsync([FromBody] IEnumerable<CustomTagEntry
...
+        /// </returns>
+        [ProducesResponseType(typeof(JsonResult), (int)HttpStatusCode.OK)]
+        [ProducesResponseType((int)HttpStatusCode.BadRequest)]
+        [ProducesResponseType((int)HttpStatusCode.NotFound)]
+        [HttpGet]
+        [Route(KnownRoutes.GetCustomTagRoute)]
+        [AuditEventType(AuditEventSubType.GetCustomTag)]
+        public async Task<IActionResult> GetTagAsync(string tagPath)
+        {
+            _logger.LogInformation("DICOM Web Get Custom Tag request received for custom tag: {0}", tagPath);
```
brandonpollett, line 93: nit: instead of `{0}`, you should be able to do `{tagPath}` so that the name is carried through.

src/Microsoft.Health.Dicom.Core/Messages/CustomTag/GetCustomTagResponse.cs
brandonpollett: I would say we should probably just have different response objects for this case instead of overloading a single one for two cases.

src/Microsoft.Health.Dicom.Api/Controllers/CustomTagController.cs
snehabandla: Per our conversation during standup, I'll maintain it like this to follow REST API guidelines to not have trailing whitespaces in uris.

src/Microsoft.Health.Dicom.Api/Controllers/CustomTagController.cs
```diff
@@ -50,5 +51,51 @@ public async Task<IActionResult> PostAsync([FromBody] IEnumerable<CustomTagEntry
...
+        /// Returns Bad Request if given path can't be parsed. Returns Not Found if given path doesn't map to a stored
+        /// custom tag or if no custom tags are stored. Returns OK with a JSON body of all tags in other cases.
+        /// </returns>
+        [ProducesResponseType(typeof(JsonResult), (int)HttpStatusCode.OK)]
+        [ProducesResponseType((int)HttpStatusCode.NotFound)]
+        [HttpGet]
+        [Route(KnownRoutes.CustomTagRoute)]
+        [AuditEventType(AuditEventSubType.GetAllCustomTags)]
+        public async Task<IActionResult> GetAllTagsAsync()
+        {
```
pengchen0692, line 68: Based on https://github.com/microsoft/api-guidelines/blob/vNext/Guidelines.md#93-collection-url-patterns, looks like we should combine getCustomTag and get all custom tag them together?

brandonpollett (approved):

src/Microsoft.Health.Dicom.Core/Extensions/DicomMediatorExtensions.cs
```diff
@@ -146,5 +139,26 @@ public static class DicomMediatorExtensions
...
+
+        public static Task<AddCustomTagResponse> AddCustomTagsAsync(
+            this IMediator mediator, IEnumerable<CustomTagEntry> customTags, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNull(mediator, nameof(mediator));
+            return mediator.Send(new AddCustomTagRequest(customTags), cancellationToken);
+        }
+
+        public static Task<GetAllCustomTagsResponse> GetAllCustomTagsAsync(
+            this IMediator mediator, string customTagPath, CancellationToken cancellationToken)
```
pengchen0692, line 151: no need for customTagPath?

src/Microsoft.Health.Dicom.Api/Features/Routing/KnownRoutes.cs
```diff
@@ -40,6 +41,9 @@ public static class KnownRoutes
         public const string ChangeFeed = "changefeed";
         public const string ChangeFeedLatest = ChangeFeed + "/" + "latest";
 
+        public const string CustomTagRoute = CustomTagsRouteSegment;
```
pengchen0692, line 44: rename as GetAllCustomTagsRoute  to keep consistency?

pengchen0692 (approved):
2 minor comments

src/Microsoft.Health.Dicom.Api/Controllers/CustomTagController.cs
snehabandla: resolved offline!

src/Microsoft.Health.Dicom.Api/Features/Routing/KnownRoutes.cs
snehabandla: This is the same route that's being used to add custom tags!

src/Microsoft.Health.Dicom.Core/Extensions/DicomMediatorExtensions.cs
snehabandla: yup, put the change to remove this in the wrong branch. :(

PR #495, opened 2021-01-11
Add CustomTagEntryValidator

pengchen0692:
## Description
Add CustomTagEntryValidator

## Related issues
Addresses [AB#78138](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/78138)

## Testing
Unit tests pass

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/DicomCoreResource.Designer.cs
+++ b/src/Microsoft.Health.Dicom.Core/DicomCoreResource.Designer.cs
@@ -467,11 +467,20 @@ namespace Microsoft.Health.Dicom.Core {
         }
         
         /// <summary>
-        ///   Looks up a localized string similar to The VR code &apos;{0}&apos; is not supported for tag &apos;{1}&apos;..
+        ///   Looks up a localized string similar to The CR code &apos;{0}&apos; is not supported..
         /// </summary>
         internal static string UnsupportedVRCode {
             get {
                 return ResourceManager.GetString("UnsupportedVRCode", resourceCulture);
+            }
+        }
+        
+        /// <summary>
+        ///   Looks up a localized string similar to The VR code &apos;{0}&apos; is not supported for tag &apos;{1}&apos;..
+        /// </summary>
+        internal static string UnsupportedVRCodeOnTag {
+            get {
+                return ResourceManager.GetString("UnsupportedVRCodeOnTag", resourceCulture);
             }
         }
         
--- a/src/Microsoft.Health.Dicom.Core/DicomCoreResource.Designer.cs
+++ b/src/Microsoft.Health.Dicom.Core/DicomCoreResource.Designer.cs
@@ -88,6 +88,15 @@ namespace Microsoft.Health.Dicom.Core {
         }
         
         /// <summary>
+        ///   Looks up a localized string similar to The custom tag &apos;{0}&apos; has already been supported..
+        /// </summary>
+        internal static string CustomTagAlreadySupported {
+            get {
+                return ResourceManager.GetString("CustomTagAlreadySupported", resourceCulture);
+            }
+        }
+        
+        /// <summary>
         ///   Looks up a localized string similar to The data store operation failed..
         /// </summary>
         internal static string DataStoreOperationFailed {
@@ -102,6 +111,15 @@ namespace Microsoft.Health.Dicom.Core {
         internal static string DicomElementValidationFailed {
             get {
                 return ResourceManager.GetString("DicomElementValidationFailed", resourceCulture);
+            }
+        }
+        
+        /// <summary>
+        ///   Looks up a localized string similar to The custom tag &apos;{0}&apos; has been specified before..
+        /// </summary>
+        internal static string DuplicateCustomTag {
+            get {
+                return ResourceManager.GetString("DuplicateCustomTag", resourceCulture);
             }
         }
         
@@ -156,6 +174,15 @@ namespace Microsoft.Health.Dicom.Core {
         internal static string InstanceNotFound {
             get {
                 return ResourceManager.GetString("InstanceNotFound", resourceCulture);
+            }
+        }
+        
+        /// <summary>
+        ///   Looks up a localized string similar to The custom tag &apos;{0}&apos; is invalid..
+        /// </summary>
+        internal static string InvalidCustomTag {
+            get {
+                return ResourceManager.GetString("InvalidCustomTag", resourceCulture);
             }
         }
         
@@ -251,6 +278,15 @@ namespace Microsoft.Health.Dicom.Core {
         }
         
         /// <summary>
+        ///   Looks up a localized string similar to The VR code &apos;{0}&apos; is invalid..
+        /// </summary>
+        internal static string InvalidVRCode {
+            get {
+                return ResourceManager.GetString("InvalidVRCode", resourceCulture);
+            }
+        }
+        
+        /// <summary>
         ///   Looks up a localized string similar to The specified item cannot be found..
         /// </summary>
         internal static string ItemNotFound {
@@ -269,6 +305,15 @@ namespace Microsoft.Health.Dicom.Core {
         }
         
         /// <summary>
+        ///   Looks up a localized string similar to The custom tag(s) is missing..
+        /// </summary>
+        internal static string MissingCustomTag {
+            get {
+                return ResourceManager.GetString("MissingCustomTag", resourceCulture);
+            }
+        }
+        
+        /// <summary>
         ///   Looks up a localized string similar to The request body is missing..
         /// </summary>
         internal static string MissingRequestBody {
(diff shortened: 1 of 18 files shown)


src/Microsoft.Health.Dicom.Core/DicomCoreResource.Designer.cs
```diff
@@ -87,6 +87,15 @@ internal class DicomCoreResource {
             }
         }
         
+        /// <summary>
+        ///   Looks up a localized string similar to The custom tag &apos;{0}&apos; has already been supported..
+        /// </summary>
+        internal static string CustomTagAlreadySupported {
```
snehabandla, line 93: >Supported [](start = 47, length = 9)

This should probably be already exists to match the other exceptions!

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntryValidator.cs
```diff
@@ -0,0 +1,170 @@
...
+                    throw new CustomTagEntryValidationException(
+                         string.Format(CultureInfo.InvariantCulture, DicomCoreResource.DuplicateCustomTag, tagEntry.Path));
+                }
+
+                pathSet.Add(tagEntry.Path);
+            }
+        }
+
+        private void ValidateCustomTagEntry(CustomTagEntry tagEntry)
+        {
```
snehabandla, line 74: Can you add function summaries on these?

src/Microsoft.Health.Dicom.Core/DicomCoreResource.Designer.cs
pengchen0692: Hi Sneha, could you point out the duplicated exception?

snehabandla: Sorry, I mean the wording here for this exception should be CustomTagAlreadyExists instead of supported.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntryValidator.cs
pengchen0692: I'm not sure what methods "these" refer to.   I think as last time team discussion, only class level summary are mandatory.
I will add summary on this method

snehabandla: I think it's mostly since there are many functions here that do similar sounding things!

pengchen0692: I see. thanks Sneha. :)

src/Microsoft.Health.Dicom.Core/DicomCoreResource.Designer.cs
pengchen0692: Actually I'm not sure what's the right word to put here, maybe  you have better idea.

There are 2 case of "already exist":
1. The custom tag has been added before (currently it fail for DuplicatedCustomTag)
2. The custom tag is one of tag we supported now ( e.g: PatientName https://github.com/microsoft/dicom-server/blob/master/docs/resources/conformance-statement.md#searchable-attributes)

This error message is for #2. 
If give CustomTagAlreadyExists, would that be confused with #1?

snehabandla: hmm not sure... I'm absolutely the worst with names. I think it sounds confusing to say it is already supported since it doesn't convey what you want. I would say RequestedTagAlreadyExists or something?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/ICustomTagEntryValidator.cs
```diff
@@ -18,7 +16,6 @@ public interface ICustomTagEntryValidator
         /// Validate if given custom tag ehtries are valid.
```
smithago, line 16: nit: entries

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntryValidator.cs
```diff
@@ -0,0 +1,170 @@
...
+            DicomVRCode.PN,
+            DicomVRCode.SH,
+            DicomVRCode.SL,
+            DicomVRCode.SS,
+            DicomVRCode.TM,
+            DicomVRCode.UI,
+            DicomVRCode.UL,
+            DicomVRCode.US);
+
+        public void ValidateCustomTags(IEnumerable<CustomTagEntry> customTagEntries)
```
smithago, line 49: What is a customTagEntry record? What is stored in our DB or what is passed in from the customer. Because it has a Key and status property and I am not sure how can user pass that it.

We shouls separate user input model from stored entry

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntryValidator.cs
```diff
@@ -0,0 +1,170 @@
...
+        private static DicomTag ParseTag(string path)
+        {
+            if (string.IsNullOrWhiteSpace(path))
+            {
+                throw new CustomTagEntryValidationException(DicomCoreResource.MissingCustomTag);
+            }
+
+            try
+            {
+                return DicomTag.Parse(path);
```
smithago, line 152: This only parses just the attribute id not the path. We should support the format similar to query with a dot separator for path.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntryValidator.cs
```diff
@@ -0,0 +1,170 @@
...
+            }
+            catch (Exception ex)
```
smithago, line 154: Can we not catch all exceptions here. You will get DicomDataException if parse fails

smithago: Generally coding practice, don't catch system.exception. Some of the system.exceptions like stackoverflow should not be caught.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntryValidator.cs
```diff
@@ -0,0 +1,170 @@
...
+using Dicom;
+using EnsureThat;
+using Microsoft.Health.Dicom.Core.Exceptions;
+using Microsoft.Health.Dicom.Core.Features.Query;
+
+namespace Microsoft.Health.Dicom.Core.Features.CustomTag
+{
+    public class CustomTagEntryValidator : ICustomTagEntryValidator
+    {
+        private const string UnknownTagName = "Unknown";
```
smithago, line 20: can you use DicomDictionary.UnknownTag?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntryValidator.cs
```diff
@@ -0,0 +1,170 @@
...
+using System.Globalization;
+using System.Linq;
+using Dicom;
+using EnsureThat;
+using Microsoft.Health.Dicom.Core.Exceptions;
+using Microsoft.Health.Dicom.Core.Features.Query;
+
+namespace Microsoft.Health.Dicom.Core.Features.CustomTag
+{
+    public class CustomTagEntryValidator : ICustomTagEntryValidator
```
smithago, line 18: I am wondering if this is the right pattern to validate the input. Did you look at https://docs.microsoft.com/en-us/aspnet/core/mvc/models/validation?view=aspnetcore-5.0? There is validationAttributes and IValidationObjects you can use to validate each input attribute

pengchen0692: I see we also validate the input on Core level(https://github.com/microsoft/dicom-server/blob/dcc067d7b2628e68cca9bebfbd98fe8aa5505746/src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveTransferSyntaxHandler.cs#L57). Wondering  what's the best practice here?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntryValidator.cs
pengchen0692: I'm thinking of a common class DicomTagParser for parsing dicom tag path, since QIDO and here both need to parse dicom tag

smithago (approved):

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntryValidator.cs
pengchen0692: I tried the model validation, I think it also works. just not very sure should we implement the validation in API layer or Core layer.  I have several consideration:
If we implement the model validation, since it  bind to Asp.net Core MVC, the API layer likely to be the best place to put it
2. In the past, we moved the validation on accepted transfersyntax from API layer to Core (Update GetFrame to support transfer syntax as discussed (#228) · microsoft/dicom-server@ce73611 (github.com), should we follow same thing this time?

PR #501, opened 2021-01-14
Adding handlers and service to support getting a single custom tag/all custom tags

snehabandla:
## Description
Adding handlers and service to support getting a single custom tag/all custom tags

## Related issues
[AB#78173](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/78173) [AB#78174](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/78174)


src/Microsoft.Health.Dicom.Core.UnitTests/Features/CustomTag/AddCustomTagServiceTests.cs
```diff
@@ -18,19 +18,19 @@
 
 namespace Microsoft.Health.Dicom.Core.UnitTests.Features.ChangeFeed
 {
-    public class CustomTagServiceTests
+    public class AddCustomTagServiceTests
```
pengchen0692, line 21: wondering why split customTag APIs into 2 services?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
```diff
@@ -19,6 +21,14 @@ public CustomTagEntry(long key, string path, string vr, CustomTagLevel level, Cu
             Status = status;
         }
 
+        public CustomTagEntry(string path, string vr, CustomTagLevel level, CustomTagStatus status)
```
pengchen0692, line 24: As previous PR feedback, I created CustomTagStoreEntry standing for CustomTag entry in database, and CustomTagEntry is now only for customer input. The purpose is to isolate customer input and our internal data model. 
As per that change, Status and Key is removed.

But I think Status is also required when GetCustomTag Apis, so may be you could pull latest code and merge your changes here.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
```diff
@@ -49,5 +59,21 @@ public override string ToString()
         {
             return $"Key: {Key}, Path: {Path}, VR:{VR}, Level:{Level}, Status:{Status}";
         }
+
+        public override int GetHashCode()
+        {
+            return HashCode.Combine(Key.GetHashCode(), Path, VR, Level.GetHashCode(), Status.GetHashCode());
```
pengchen0692, line 65: VR could be null, would HashCode.Combine works with that?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/GetCustomTagHandler.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        public GetCustomTagHandler(IGetCustomTagsService getCustomTagsService)
+        {
+            EnsureArg.IsNotNull(getCustomTagsService, nameof(getCustomTagsService));
+            _getCustomTagsService = getCustomTagsService;
+        }
+
+        public async Task<GetCustomTagResponse> Handle(GetCustomTagRequest request, CancellationToken cancellationToken)
+        {
+            TagPathValidator.Validate(request.CustomTagPath);
```
pengchen0692, line 27: I see most our code follow such pattern where input parameters are checked nullable for any public methods. could you also follow this?

pengchen0692, line 27: I checked in DicomTagParser (https://github.com/microsoft/dicom-server/blob/master/src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs) in previous PR for parsing tag path.  maybe we can reuse that?

src/Microsoft.Health.Dicom.Core.UnitTests/Features/CustomTag/AddCustomTagServiceTests.cs
snehabandla: It's just to keep the services small and easy to understand! I suspect that Add especially might get long once we add support for a job framework!

---
In reply to: [558512075](https://github.com/microsoft/dicom-server/pull/501#discussion_r558512075) [](ancestors = 558512075)

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
snehabandla: I believe so! 

---
In reply to: [558518647](https://github.com/microsoft/dicom-server/pull/501#discussion_r558518647) [](ancestors = 558518647)

src/Microsoft.Health.Dicom.Core/Features/CustomTag/GetCustomTagHandler.cs
snehabandla: Sure, I'll update to validate that.

---
In reply to: [558527662](https://github.com/microsoft/dicom-server/pull/501#discussion_r558527662) [](ancestors = 558527662)

snehabandla: I'll take a look!

---
In reply to: [558528870](https://github.com/microsoft/dicom-server/pull/501#discussion_r558528870) [](ancestors = 558528870)

src/Microsoft.Health.Dicom.Core/Features/CustomTag/GetCustomTagsService.cs
```diff
@@ -0,0 +1,46 @@
...
+        private readonly ICustomTagStore _customTagStore;
+
+        public GetCustomTagsService(ICustomTagStore customTagStore)
+        {
+            EnsureArg.IsNotNull(customTagStore, nameof(customTagStore));
+
+            _customTagStore = customTagStore;
+        }
+
+        public async Task<GetCustomTagResponse> GetCustomTagAsync(string tagPath, CancellationToken cancellationToken = default)
```
pengchen0692, line 25: "CancellationToken cancellationToken = default"
Default value could be specified on interface only

src/Microsoft.Health.Dicom.Core/Features/CustomTag/GetCustomTagsService.cs
```diff
@@ -0,0 +1,46 @@
...
+        {
+            string internalTagPath = ConvertTagPathToInternalFormat(tagPath);
```
pengchen0692, line 27: use DicomTagParser  to parse Tag path instead?
BTW, we don't need to support embeded custom tags right now.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
```diff
@@ -49,5 +59,21 @@ public override string ToString()
...
+        }
+
+        public override bool Equals(object obj)
```
pengchen0692, line 68: it's good to have GetHashCode and Equals, just curious  where we use them?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/ICustomTagStore.cs
```diff
@@ -21,6 +22,21 @@ public interface ICustomTagStore
         /// <returns>The tag key.</returns>
         Task<long> AddCustomTagAsync(string path, string vr, CustomTagLevel level, CustomTagStatus status, CancellationToken cancellationToken = default);
 
+        /// <summary>
+        /// Get a requested custom tag from CustomTagStore by tagPath.
+        /// </summary>
+        /// <param name="path">Path associated with requested custom tag formatted as it is stored internally.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns>Custom tag entry with path, VR, level and status.</returns>
+        Task<CustomTagEntry> GetCustomTagAsync(string path, CancellationToken cancellationToken = default);
```
pengchen0692, line 31: I'm thinking on ICustomTagStore level, could we merge these 2 functions together?
We can do something like https://github.com/microsoft/dicom-server/blob/368a21c761c83a14288bec80dc2146ccd7c2f539/src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/1.sql#L629  to handle Null path

src/Microsoft.Health.Dicom.Core.UnitTests/Features/CustomTag/AddCustomTagServiceTests.cs
pengchen0692: There are multiple operations for CustomTag(add/remove/list/get), they might share some functions, putting in same class could make this simpler.
Another concern is if we put AddCustomTag into separated service, should we also put remove/get/list into different classes as well.

Anyway, I'm fine with current approach, if do see there is a need for change, we can do that later.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/ICustomTagStore.cs
snehabandla: I want to keep them separate till sql so that the return values are distinct!

---
In reply to: [558596602](https://github.com/microsoft/dicom-server/pull/501#discussion_r558596602) [](ancestors = 558596602)

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
snehabandla: The Equals requires GetHashCode and is used in test code to compare custom tag entries.

---
In reply to: [558595543](https://github.com/microsoft/dicom-server/pull/501#discussion_r558595543) [](ancestors = 558595543)

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
```diff
@@ -18,7 +18,7 @@ public bool TryParse(string dicomTagPath, out DicomTag[] dicomTags, bool support
             dicomTags = null;
             if (supportMultiple)
             {
-                throw new System.NotImplementedException("Multiple dicom tag in tag path will be supported later");
+                throw new System.NotImplementedException("Sequential dicom tags are currently not supported.");
```
wsugarman, line 21: >System [](start = 26, length = 6)

nit: in this case, I'd simply import `System`

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
```diff
@@ -38,6 +38,16 @@ public bool TryParse(string dicomTagPath, out DicomTag[] dicomTags, bool support
             return dicomTag != null;
         }
 
+        public string ParseFormattedTagPath(string dicomTagPath, bool supportMultiple = false)
+        {
+            if ((!supportMultiple && dicomTagPath.Contains('.', System.StringComparison.OrdinalIgnoreCase)) || supportMultiple)
```
wsugarman, line 43: >!supportMultiple [](start = 17, length = 16)

This `!supportMultiple` expression is unnecessary if you change the order of the expressions to:

```csharp
if (supportMultiple || dicomTagPath.Contains('.', System.StringComparison.OrdinalIgnoreCase))
```

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
```diff
@@ -38,6 +38,16 @@ public bool TryParse(string dicomTagPath, out DicomTag[] dicomTags, bool support
             return dicomTag != null;
         }
 
+        public string ParseFormattedTagPath(string dicomTagPath, bool supportMultiple = false)
```
wsugarman, line 41: >ParseFormattedTagPath [](start = 22, length = 21)

nit: this really isn't parsing. It's more "validating" and echoing the value

src/Microsoft.Health.Dicom.Core/Features/CustomTag/GetAllCustomTagsHandler.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            EnsureArg.IsNotNull(getCustomTagsService, nameof(getCustomTagsService));
+            _getCustomTagsService = getCustomTagsService;
+        }
+
+        public async Task<GetAllCustomTagsResponse> Handle(GetAllCustomTagsRequest request, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNull(request, nameof(request));
+
+            return await _getCustomTagsService.GetAllCustomTagsAsync(cancellationToken);
```
wsugarman, line 28: >return await [](start = 12, length = 12)

nit: This method does not need to be `async`. Instead you can simply say `return _getCustomTagsService.GetAllCustomTagsAsync(cancellationToken);` to return the `Task<T>` directly

src/Microsoft.Health.Dicom.Core/Features/CustomTag/AddCustomTagHandler.cs
```diff
@@ -13,17 +13,17 @@ namespace Microsoft.Health.Dicom.Core.Features.CustomTag
...
-            EnsureArg.IsNotNull(customTagService, nameof(customTagService));
-            _customTagService = customTagService;
+            EnsureArg.IsNotNull(addCustomTagService, nameof(addCustomTagService));
+            _addCustomTagService = addCustomTagService;
         }
 
         public async Task<AddCustomTagResponse> Handle(AddCustomTagRequest request, CancellationToken cancellationToken)
         {
-            return await _customTagService.AddCustomTagAsync(request.CustomTags, cancellationToken);
+            return await _addCustomTagService.AddCustomTagAsync(request.CustomTags, cancellationToken);
```
wsugarman, line 26: >return await _addCustomTagService.AddCustomTagAsync(request.CustomTags, cancellationToken); [](start = 12, length = 91)

nit: You can return this `Task` directly without `await`-ing in an `async` method

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagService.cs
```diff
@@ -33,7 +33,7 @@ public CustomTagService(ICustomTagStore customTagStore, IReindexJob reindexJob,
             _logger = logger;
         }
 
-        public async Task<AddCustomTagResponse> AddCustomTagAsync(IEnumerable<CustomTagEntry> customTags, CancellationToken cancellationToken = default)
```
wsugarman, line 36: > = default) [](start = 141, length = 11)

Any reason for removing the default?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
wsugarman: Yeah: [https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/HashCode.cs](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/HashCode.cs)

---
In reply to: [558589887](https://github.com/microsoft/dicom-server/pull/501#discussion_r558589887) [](ancestors = 558589887,558518647)

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Validation/TagPathValidatorTests.cs
```diff
@@ -0,0 +1,36 @@
...
+    public class TagPathValidatorTests
+    {
+        [InlineData("helloworld")]
+        [InlineData("(0101,2323)")]
+        [InlineData("0101,2323")]
+        [InlineData("(01012323)")]
+        [InlineData("01012323-01012323")]
+        [InlineData("01012323.01012323 45455656")]
+        [InlineData("0101232301012323")]
+        [Theory]
```
wsugarman, line 21: >[Theory] [](start = 8, length = 8)

nit: I have typically seen the `[Theory]` attribute first, as if to say, "This method is a theory for the following scenarios". Also, some part of me likes to see the smaller line on top of the larger lines lol. But if this is the typical order we use, then nevermind!

src/Microsoft.Health.Dicom.Core.UnitTests/Features/CustomTag/GetCustomTagsServiceTests.cs
```diff
@@ -0,0 +1,84 @@
...
+            _getCustomTagsService = new GetCustomTagsService(_customTagStore, _dicomTagParser);
+        }
+
+        [Fact]
+        public async Task GivenRequestForAllTags_WhenNoTagsAreStored_ThenEmptyListShouldBeReturned()
+        {
+            _customTagStore.GetAllCustomTagsAsync(default).Returns(new List<CustomTagEntry>());
+            GetAllCustomTagsResponse response = await _getCustomTagsService.GetAllCustomTagsAsync(default);
+
+            Assert.True(response.CustomTags.Count() == 0);
```
wsugarman, line 37: >True [](start = 19, length = 4)

nit: use `Equals` instead so the error message if it fails is easier to understand. Otherwise all we'll see is "Assert failed" lol

src/Microsoft.Health.Dicom.Core.UnitTests/Features/CustomTag/GetCustomTagsServiceTests.cs
```diff
@@ -0,0 +1,84 @@
...
+            CustomTagEntry tag2 = new CustomTagEntry("01012323", DicomVRCode.FL.ToString(), CustomTagLevel.Series, CustomTagStatus.Reindexing);
+
+            List<CustomTagEntry> storedEntries = new List<CustomTagEntry>() { tag1, tag2 };
+
+            _customTagStore.GetAllCustomTagsAsync(default).Returns(storedEntries);
+            GetAllCustomTagsResponse response = await _getCustomTagsService.GetAllCustomTagsAsync(default);
+
+            List<CustomTagEntry> result = storedEntries.Except(response.CustomTags).ToList();
+
+            Assert.True(result.Count == 0);
```
wsugarman, line 53: >True [](start = 19, length = 4)

nit: same comment as above about `Equals`

src/Microsoft.Health.Dicom.Core/Features/Validation/TagPathValidator.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Text.RegularExpressions;
+using Microsoft.Health.Dicom.Core.Exceptions;
+
+namespace Microsoft.Health.Dicom.Core.Features.Validation
+{
+    public static class TagPathValidator
+    {
+        public static void Validate(string tagPath)
+        {
+            Match match = Regex.Match(tagPath, @"\A\d{8}(\.(\d){8})*$");
```
wsugarman, line 15: >Match match = Regex.Match(tagPath, @"\A\d{8}(\.(\d){8})*$"); [](start = 12, length = 60)

If we're going to use this a lot, it is probably best to declare this regex outside of the method's scope and compile it with the other [`ctor`]([https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex.-ctor?view=net-5.0#System_Text_RegularExpressions_Regex__ctor_System_String_System_Text_RegularExpressions_RegexOptions](https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex.-ctor?view=net-5.0#System_Text_RegularExpressions_Regex__ctor_System_String_System_Text_RegularExpressions_RegexOptions)_):

```csharp
private static readonly Regex _expectedTagPath = new Regex(@"\A\d{8}(\.(\d){8})*$", RegexOptions.Compiled);
```

src/Microsoft.Health.Dicom.Core/Features/Validation/TagPathValidator.cs
```diff
@@ -0,0 +1,22 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Text.RegularExpressions;
+using Microsoft.Health.Dicom.Core.Exceptions;
+
+namespace Microsoft.Health.Dicom.Core.Features.Validation
+{
+    public static class TagPathValidator
```
wsugarman, line 11: >public [](start = 4, length = 6)

Does this class need to be `public`? This is more of a general question/comment across the PR. We should be judicious about what we decide to be `public`. `public` things become part of the assembly's contract, and can't be changed as easily. As a rule of thumb, only make what is absolutely necessary `public`/`protected`. Everything else can be `internal`/`private`.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/AddCustomTagHandler.cs
```diff
@@ -13,17 +13,17 @@ namespace Microsoft.Health.Dicom.Core.Features.CustomTag
 {
     public class AddCustomTagHandler : IRequestHandler<AddCustomTagRequest, AddCustomTagResponse>
```
wsugarman, line 14: >AddCustomTagHandler [](start = 17, length = 19)

What are the various responsibilities of the different layers here? It looks like the `IRequestHandler<TIn, TOut>` interface is for `MediatR`, but what is the semantic difference between the `IAddCustomTagSerice` and the `ICustomTagService`. It seems like if there is necessary input/output shimming, it can be done within the handler directly. However, I can imagine keeping the service and store separate if the Add action itself has specific actions that must be taken (and is worth encapsulating in its own entity)

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
```diff
@@ -3,10 +3,12 @@
...
 
+using System;
+
 namespace Microsoft.Health.Dicom.Core.Features.CustomTag
 {
     /// <summary>
-    /// Represent each custom tag entry from customer input.
+    /// External representation of a custom tag entry.
     /// </summary>
     public class CustomTagEntry
```
wsugarman, line 13: >CustomTagEntry [](start = 17, length = 14)

If the properties are "settable", I don't think constructors are really necessary. That said, can we make this entity immutable? I would probably recommend 1 of 2 ways:
1. Given the size of this entry, it may be reasonable to use a `readonly` `struct` here
2. Otherwise, we could use the new C# 9 record type, which is a fancy wrapper around `class`, to auto-generate the equality, hash-code, etc. There is a new tutorial page [here]([https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/exploration/records#:~:text=C#%209%20introduces%20records,%20a%20new%20reference%20type,classes%20in%20that%20record%20types%20use%20value-based%20equality](https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/exploration/records#:~:text=C#%209%20introduces%20records,%20a%20new%20reference%20type,classes%20in%20that%20record%20types%20use%20value-based%20equality).)

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Common/DicomTagParserTests.cs
```diff
@@ -40,6 +41,20 @@ public void GivenInvalidTag_WhenParse_ThenShouldReturnFalse(string dicomTagPath)
             Assert.False(succeed);
         }
 
+        [Fact]
+        public void GivenDicomTagPath_WhenParsingFormattedTagPath_ThenShouldRemoveUnnecessaryCharacters()
+        {
+            string tag = "00001111";
+            string parsedString = _dicomTagParser.ParseFormattedTagPath(tag);
+            Assert.True(parsedString.Equals(tag));
```
wsugarman, line 49: >True [](start = 19, length = 4)

nit: `Equals`

wsugarman (changes requested):
:clock1:

src/Microsoft.Health.Dicom.Core/Features/CustomTag/GetCustomTagsService.cs
```diff
@@ -0,0 +1,45 @@
...
+            EnsureArg.IsNotNull(customTagStore, nameof(customTagStore));
+            EnsureArg.IsNotNull(dicomTagParser, nameof(dicomTagParser));
+
+            _customTagStore = customTagStore;
+            _dicomTagParser = dicomTagParser;
+        }
+
+        public async Task<GetCustomTagResponse> GetCustomTagAsync(string tagPath, CancellationToken cancellationToken)
+        {
+            string internalTagPath = _dicomTagParser.ParseFormattedTagPath(tagPath);
```
pengchen0692, line 31: what if tagPath is Dicom Tag Keyword? like "DeviceSerialNumber"

src/Microsoft.Health.Dicom.Core/Features/Validation/TagPathValidator.cs
pengchen0692: How about using DicomTagParser.TryParse? it could filter out more invalid cases before calling into more heavier SQL?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/GetAllCustomTagsHandler.cs
snehabandla: It looks like our existing handlers and services always follow this pattern. Do you think we will increase latencies by doing this?

---
In reply to: [564813493](https://github.com/microsoft/dicom-server/pull/501#discussion_r564813493) [](ancestors = 564813493)

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagService.cs
snehabandla: I think we require the cancellation token to be provided in implementations! Based on other similar classes.

---
In reply to: [564818697](https://github.com/microsoft/dicom-server/pull/501#discussion_r564818697) [](ancestors = 564818697)

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Validation/TagPathValidatorTests.cs
snehabandla: I do agree that the smaller line below looks strange. However, this is the typical order we use!

---
In reply to: [564825555](https://github.com/microsoft/dicom-server/pull/501#discussion_r564825555) [](ancestors = 564825555)

src/Microsoft.Health.Dicom.Core/Features/Validation/TagPathValidator.cs
snehabandla: Hmm, I was following the pattern used in UidValidator. I think it's public also because of accessibility in test code?

---
In reply to: [564837026](https://github.com/microsoft/dicom-server/pull/501#discussion_r564837026) [](ancestors = 564837026)

src/Microsoft.Health.Dicom.Core/Features/CustomTag/AddCustomTagHandler.cs
snehabandla: I've split this so we have separate services for the actions on custom tag. This is primarily because I want to keep the class responsibility small and easy to read. Add I imagine will get complex, as will remove!

---
In reply to: [564841468](https://github.com/microsoft/dicom-server/pull/501#discussion_r564841468) [](ancestors = 564841468)

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
snehabandla: I think there is only one scenario where we require a settable property (Status has to be updateable).  The rest of them I've made get only.

---
In reply to: [564843380](https://github.com/microsoft/dicom-server/pull/501#discussion_r564843380) [](ancestors = 564843380)

src/Microsoft.Health.Dicom.Core/Features/CustomTag/GetCustomTagsService.cs
snehabandla: My understanding is that tagPath will always be the 8 digit tag id (multiple of 8 for lower levels). Has this expectation changed?

---
In reply to: [564890558](https://github.com/microsoft/dicom-server/pull/501#discussion_r564890558) [](ancestors = 564890558)

src/Microsoft.Health.Dicom.Core/Features/Validation/TagPathValidator.cs
snehabandla: I'm currently following the pattern for how we validate Uid for retrieve/query. This does less validation than the TryParse does - it doesn't need to validate that it is a valid tag path!

---
In reply to: [564894958](https://github.com/microsoft/dicom-server/pull/501#discussion_r564894958) [](ancestors = 564894958)

pengchen0692 (approved):
Whether we should support keyword when fetching custom tag by path is my only concern.
I will raise the question up for next standup.

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
wsugarman: Do you want to throw on `'.'` even if it doesn't support multiple?

---
In reply to: [564811735](https://github.com/microsoft/dicom-server/pull/501#discussion_r564811735) [](ancestors = 564811735)

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
```diff
@@ -44,5 +46,21 @@ public override string ToString()
         {
             return $"Path: {Path}, VR:{VR}, Level:{Level} Status:{Status}";
         }
+
+        public override int GetHashCode()
+        {
+            return HashCode.Combine(Path, VR, Level.GetHashCode(), Status.GetHashCode());
+        }
+
+        public override bool Equals(object obj)
```
wsugarman, line 55: >Equals [](start = 29, length = 6)

nit: I'd implement `IEquatable<CustomTagEntry>` if you're already going to override `Equals`

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
```diff
@@ -44,5 +46,21 @@ public override string ToString()
...
+        {
+            CustomTagEntry other = obj as CustomTagEntry;
+            if (other == null)
+            {
+                return false;
+            }
+
+            return Path == other.Path && VR == other.VR && Level == other.Level && Status == other.Status;
```
wsugarman, line 63: > Path == other.Path && VR == other.VR [](start = 18, length = 37)

Are `Path` and `VR` case-sensitive?

snehabandla: yes, nice catch! will make a case-insensitive comparison.

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
snehabandla: I think I still need to because it means it somehow contains multiple even when we explicitly don't support them.

---
In reply to: [567146103](https://github.com/microsoft/dicom-server/pull/501#discussion_r567146103) [](ancestors = 567146103,564811735)

snehabandla: Longer term, the support multiple flag won't be necessary and this entire exception can be removed


---
In reply to: [567153710](https://github.com/microsoft/dicom-server/pull/501#discussion_r567153710) [](ancestors = 567153710,567146103,564811735)

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
wsugarman: If `Path` and `VR` are `null`-able, then you can use `string.Equals` instead

---
In reply to: [567147794](https://github.com/microsoft/dicom-server/pull/501#discussion_r567147794) [](ancestors = 567147794)

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
```diff
@@ -44,5 +46,31 @@ public override string ToString()
...
+
+        public override bool Equals(object obj)
+        {
+            CustomTagEntry other = obj as CustomTagEntry;
+            if (other == null)
+            {
+                return false;
+            }
+
+            return Path.Equals(other.Path, StringComparison.OrdinalIgnoreCase) && VR.Equals(other.VR, StringComparison.OrdinalIgnoreCase) && Level == other.Level && Status == other.Status;
```
wsugarman, line 63: Let's re-use our new `Equals` method here:

```csharp
public override bool Equals(obj)
{
    return obj is CustomTagEntry other && Equals(other);
}
```

src/Microsoft.Health.Dicom.Core/Features/Validation/TagPathValidator.cs
wsugarman: You can explicitly share your `internal` components with other assemblies, typically test assemblies, using the below attribute in your *AssemblyInfo.cs*:

```csharp
[assembly: InternalsVisibleTo("Microsoft.Health.Dicom.UnitTests")]
```

---
In reply to: [566994328](https://github.com/microsoft/dicom-server/pull/501#discussion_r566994328) [](ancestors = 566994328,564837026)

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
wsugarman: Ok!

---
In reply to: [567154733](https://github.com/microsoft/dicom-server/pull/501#discussion_r567154733) [](ancestors = 567154733,567153710,567146103,564811735)

src/Microsoft.Health.Dicom.Core/Features/CustomTag/GetAllCustomTagsHandler.cs
wsugarman: No, it is just unnecessary. The IL is much more complex in this case, as the state machine needs to be constructed and the task must be fetched (essentially you must create a Task around the Task). You can leave this as-is, but we probably want to avoid this in the future just to avoid the overhead.

---
In reply to: [566992864](https://github.com/microsoft/dicom-server/pull/501#discussion_r566992864) [](ancestors = 566992864,564813493)

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagService.cs
wsugarman: Oh ok

---
In reply to: [566993274](https://github.com/microsoft/dicom-server/pull/501#discussion_r566993274) [](ancestors = 566993274,564818697)

wsugarman (approved):
:shipit:

PR #504, opened 2021-01-14
Add secondary audience with opposite trailing slash existence.

brandonpollett:
## Description
This PR updates the security module to use a list of audiences instead of just the single value from the configuration. It adds a secondary value that has the opposite existence of the trailing slash.

| Given Audience | Secondary Audience |
| --- | --- |
| https://example.com | https://example.com/ |
| https://example.com/ | https://example.com |

## Related issues
Addresses [AB#78859](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/78859)

## Testing
Manually tested hitting the endpoint with both audiences. Added unit test for generating the secondary audience.

### DIFF
--- a/src/Microsoft.Health.Dicom.Api/Modules/SecurityModule.cs
+++ b/src/Microsoft.Health.Dicom.Api/Modules/SecurityModule.cs
@@ -3,6 +3,7 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System;
 using EnsureThat;
 using Microsoft.AspNetCore.Authentication.JwtBearer;
 using Microsoft.AspNetCore.Authorization;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Dicom.Api/Modules/SecurityModule.cs
```diff
@@ -42,9 +44,16 @@ public void Load(IServiceCollection services)
...
                     options.Authority = _securityConfiguration.Authentication.Authority;
-                    options.Audience = _securityConfiguration.Authentication.Audience;
                     options.RequireHttpsMetadata = true;
                     options.Challenge = $"Bearer authorization_uri=\"{_securityConfiguration.Authentication.Authority}\", resource_id=\"{_securityConfiguration.Authentication.Audience}\", realm=\"{_securityConfiguration.Authentication.Audience}\"";
+                    options.TokenValidationParameters = new TokenValidationParameters
+                    {
+                        ValidAudiences = new[]
+                        {
+                            _securityConfiguration.Authentication.Audience,
+                            GenerateSecondaryAudience(),
```
johnstairs, line 54: The audience could be the appid, in which case it's just a GUID.

brandonpollett: What do you then think about just updating the semantics of the audience to understand a delimited list. Such that a setting of `https://example.com;https://example.com;1234` would results in the following array:

```json
[
   "https://example.com",
   "https://example.com",
   "1234"
]
```

PR #506, opened 2021-01-15
Implement ReindexJob

pengchen0692:
## Description
Implement Reindex job which reindex dicom instances in the past on custom tags

## Related issues
Addresses [AB#78130](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/78130)

## Testing
Unit test pass

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
+++ b/src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
@@ -9,6 +9,7 @@ using System.Globalization;
 using System.Linq;
 using Dicom;
 using EnsureThat;
+using Microsoft.Health.Dicom.Core.Features.CustomTag;
 using Microsoft.Health.Dicom.Core.Features.Model;
 
 namespace Microsoft.Health.Dicom.Core.Extensions
@@ -157,6 +158,38 @@ namespace Microsoft.Health.Dicom.Core.Extensions
             {
                 dicomDataset.Add(dicomTag, value);
             }
+        }
+
+        /// <summary>
+        /// Get custom tag indexes
+        /// </summary>
+        /// <param name="dicomDataset">The dicom dataset.</param>
+        /// <param name="customTagStoreEntries">The custom tag entries stored.</param>
+        /// <returns>The indexes.</returns>
+        public static Dictionary<long, DicomItem> GetCustomTagIndexes(this DicomDataset dicomDataset, IEnumerable<CustomTagStoreEntry> customTagStoreEntries)
+        {
+            // We only support index on top level custom tag for now, will support embeded tag later.
+            Dictionary<long, DicomItem> indexes = new Dictionary<long, DicomItem>();
+            Dictionary<string, CustomTagStoreEntry> tagPathDictionary = customTagStoreEntries.ToDictionary(keySelector: item => item.Path, comparer: StringComparer.OrdinalIgnoreCase);
+            if (tagPathDictionary.Count != 0)
+            {
+                foreach (var item in dicomDataset)
+                {
+                    string path = item.Tag.GetPath();
+                    if (tagPathDictionary.ContainsKey(path))
+                    {
+                        CustomTagStoreEntry entry = tagPathDictionary[path];
+
+                        // check if VR match
+                        if (string.Equals(entry.VR, item.ValueRepresentation.Code, StringComparison.OrdinalIgnoreCase))
+                        {
+                            indexes[entry.Key] = item;
+                        }
+                    }
+                }
+            }
+
+            return indexes;
         }
     }
 }
--- a/src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
+++ b/src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
@@ -166,19 +166,18 @@ namespace Microsoft.Health.Dicom.Core.Extensions
         /// <param name="dicomDataset">The dicom dataset.</param>
         /// <param name="customTagStoreEntries">The custom tag entries stored.</param>
         /// <returns>The indexes.</returns>
-        public static Dictionary<long, DicomItem> GetCustomTagIndexes(this DicomDataset dicomDataset, IEnumerable<CustomTagStoreEntry> customTagStoreEntries)
+        public static Dictionary<long, DicomItem> GetCustomTagIndexes(this DicomDataset dicomDataset, IDictionary<string, CustomTagStoreEntry> customTagStoreEntries)
         {
             // We only support index on top level custom tag for now, will support embeded tag later.
-            Dictionary<long, DicomItem> indexes = new Dictionary<long, DicomItem>();
-            Dictionary<string, CustomTagStoreEntry> tagPathDictionary = customTagStoreEntries.ToDictionary(keySelector: item => item.Path, comparer: StringComparer.OrdinalIgnoreCase);
-            if (tagPathDictionary.Count != 0)
+            var indexes = new Dictionary<long, DicomItem>();
+            if (customTagStoreEntries.Count != 0)
             {
                 foreach (var item in dicomDataset)
                 {
                     string path = item.Tag.GetPath();
-                    if (tagPathDictionary.ContainsKey(path))
+                    if (customTagStoreEntries.ContainsKey(path))
                     {
-                        CustomTagStoreEntry entry = tagPathDictionary[path];
+                        CustomTagStoreEntry entry = customTagStoreEntries[path];
 
                         // check if VR match
                         if (string.Equals(entry.VR, item.ValueRepresentation.Code, StringComparison.OrdinalIgnoreCase))
--- a/src/Microsoft.Health.Dicom.Core/Extensions/DicomTagExtensions.cs
+++ b/src/Microsoft.Health.Dicom.Core/Extensions/DicomTagExtensions.cs
@@ -24,5 +24,35 @@ namespace Microsoft.Health.Dicom.Core.Extensions
             EnsureArg.IsNotNull(dicomTag, nameof(dicomTag));
             return dicomTag.Group.ToString("X4") + dicomTag.Element.ToString("X4");
         }
+
+        /// <summary>
+        /// Get default VR for dicom tag.
+        /// </summary>
+        /// <param name="dicomTag">The dicm tag</param>
+        /// <returns>The default VR.</returns>
+        public static DicomVR GetDefaultVR(this DicomTag dicomTag)
+        {
+            EnsureArg.IsNotNull(dicomTag, nameof(dicomTag));
+            if (dicomTag.IsPrivate)
+            {
+                // Private tag don't have default VR.
+                return null;
+            }
+
+            if (dicomTag.DictionaryEntry == DicomDictionary.UnknownTag)
+            {
+                // this tag is invalid
+                return null;
+            }
+
+            DicomVR[] vrs = dicomTag.DictionaryEntry.ValueRepresentations;
+            if (vrs.Length == 0)
+            {
+                return null;
+            }
+
+            // Use the first valid VR as default VR
+            return vrs[0];
+        }
     }
 }
--- a/src/Microsoft.Health.Dicom.Core/Extensions/DicomTagExtensions.cs
+++ b/src/Microsoft.Health.Dicom.Core/Extensions/DicomTagExtensions.cs
@@ -3,6 +3,7 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System.Linq;
 using Dicom;
 using EnsureThat;
 
@@ -45,14 +46,7 @@ namespace Microsoft.Health.Dicom.Core.Extensions
                 return null;
             }
 
-            DicomVR[] vrs = dicomTag.DictionaryEntry.ValueRepresentations;
-            if (vrs.Length == 0)
-            {
-                return null;
-            }
-
-            // Use the first valid VR as default VR
-            return vrs[0];
+            return dicomTag.DictionaryEntry.ValueRepresentations.FirstOrDefault();
         }
     }
 }
--- a/src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagService.cs
+++ b/src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagService.cs
@@ -55,7 +55,7 @@ namespace Microsoft.Health.Dicom.Core.Features.CustomTag
             // Validate input
             _customTagEntryValidator.ValidateCustomTags(customTags);
 
-            Dictionary<long, CustomTagStoreEntry> addedTags = new Dictionary<long, CustomTagStoreEntry>();
+            List<CustomTagStoreEntry> addedCustomTagStoreEntries = new List<CustomTagStoreEntry>();
             foreach (var customTag in customTags)
             {
                 try
@@ -85,18 +85,17 @@ namespace Microsoft.Health.Dicom.Core.Features.CustomTag
                         }
                     }
 
-                    long key = await _customTagStore.AddCustomTagAsync(path, vr, customTag.Level, initStatus);
-                    CustomTagStoreEntry storeEntry = new CustomTagStoreEntry(key, path, vr, customTag.Level, initStatus);
-                    addedTags.Add(key, storeEntry);
+                    CustomTagStoreEntry storeEntry = await _customTagStore.AddCustomTagAsync(path, vr, customTag.Level, initStatus);
+                    addedCustomTagStoreEntries.Add(storeEntry);
                 }
                 catch (Exception ex)
                 {
-                    _logger.LogCritical(ex, "Failed to add custom tag {tag}.", customTag);
+                    _logger.LogCritical(ex, "Failed to add custom tag {customTag}.", customTag);
 
                     // clean up
-                    foreach (var tagkey in addedTags.Keys)
+                    foreach (var customTagStoreEntry in addedCustomTagStoreEntries)
                     {
-                        await _customTagStore.DeleteCustomTagAsync(tagkey);
+                        await _customTagStore.DeleteCustomTagAsync(customTagStoreEntry.Key);
                     }
 
                     throw;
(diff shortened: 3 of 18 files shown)


src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagService.cs
```diff
@@ -48,18 +56,31 @@ public async Task<AddCustomTagResponse> AddCustomTagAsync(IEnumerable<CustomTagE
...
-            foreach (var tag in customTags)
+            foreach (var customTag in customTags)
             {
                 try
                 {
                     CustomTagStatus initStatus = CustomTagStatus.Reindexing;
-                    long key = await _customTagStore.AddCustomTagAsync(tag.Path, tag.VR, tag.Level, initStatus);
-                    CustomTagStoreEntry storeEntry = new CustomTagStoreEntry(key, tag.Path, tag.VR, tag.Level, initStatus);
+
+                    // we only support path that has 1 attrubteId for now, will support path which is composed of multiple attributeId later (e.g: 00100010.00100020)
```
pengchen0692, line 65: we only support top level dicomtag right now, will support embeded dicomtag soon

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagService.cs
```diff
@@ -48,18 +56,31 @@ public async Task<AddCustomTagResponse> AddCustomTagAsync(IEnumerable<CustomTagE
...
+                    if (!_dicomTagParser.TryParse(customTag.Path, out tags, supportMultiple: false))
+                    {
+                        // not a valid dicom tag path
+                        throw new CustomTagEntryValidationException(
+                            string.Format(CultureInfo.InvariantCulture, DicomCoreResource.InvalidCustomTag, customTag));
+                    }
+
+                    DicomTag tag = tags[0];
+                    string path = tag.GetPath();
+                    string vr = tag.GetDefaultVR().Code;
```
pengchen0692, line 76: should use given vr if specified

src/Microsoft.Health.Dicom.Core/Features/CustomTag/IInstanceIndexer.cs
```diff
@@ -0,0 +1,24 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Health.Dicom.Core.Features.Model;
+
+namespace Microsoft.Health.Dicom.Core.Features.CustomTag
+{
+    public interface IInstanceIndexer
```
pengchen0692, line 13: summary for the interface

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -158,5 +159,37 @@ public static void AddValueIfNotNull<T>(this DicomDataset dicomDataset, DicomTag
...
+        /// <summary>
+        /// Get custom tag indexes
+        /// </summary>
+        /// <param name="dicomDataset">The dicom dataset.</param>
+        /// <param name="customTagStoreEntries">The custom tag entries stored.</param>
+        /// <returns>The indexes.</returns>
+        public static Dictionary<long, DicomItem> GetCustomTagIndexes(this DicomDataset dicomDataset, IEnumerable<CustomTagStoreEntry> customTagStoreEntries)
+        {
+            // We only support index on top level custom tag for now, will support embeded tag later.
+            Dictionary<long, DicomItem> indexes = new Dictionary<long, DicomItem>();
```
brandonpollett, line 172: nit: can use `var` here since the type is explicitly stated.

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -158,5 +159,37 @@ public static void AddValueIfNotNull<T>(this DicomDataset dicomDataset, DicomTag
...
+            Dictionary<string, CustomTagStoreEntry> tagPathDictionary = customTagStoreEntries.ToDictionary(keySelector: item => item.Path, comparer: StringComparer.OrdinalIgnoreCase);
```
brandonpollett, line 173: Would it make sense to do this conversion a few layers up and just pass the dictionary here? My reading of the code seems like we'll be doing this work many times and will always get the same results.

src/Microsoft.Health.Dicom.Core/Extensions/DicomTagExtensions.cs
```diff
@@ -24,5 +24,35 @@ public static string GetPath(this DicomTag dicomTag)
...
+            }
+
+            DicomVR[] vrs = dicomTag.DictionaryEntry.ValueRepresentations;
+            if (vrs.Length == 0)
+            {
+                return null;
+            }
+
+            // Use the first valid VR as default VR
+            return vrs[0];
```
brandonpollett, line 55: Would it be possible to just use `return dicomTag.DictionaryEntry.ValueRepresentations.FirstOrdDefault()` here?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagService.cs
```diff
@@ -48,18 +56,42 @@ public async Task<AddCustomTagResponse> AddCustomTagAsync(IEnumerable<CustomTagE
...
+                    // when VR is not specified for standard tag,
+                    if (!tag.IsPrivate && tag.DictionaryEntry != DicomDictionary.UnknownTag)
+                    {
+                        if (string.IsNullOrWhiteSpace(vr))
+                        {
+                            vr = tag.GetDefaultVR()?.Code;
+                        }
+                    }
+
+                    long key = await _customTagStore.AddCustomTagAsync(path, vr, customTag.Level, initStatus);
```
brandonpollett, line 88: Would it make more sense for the `AddCustomTagAsync` to return a `CustomTagStoreEntry` with the `key` populated?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagService.cs
```diff
@@ -48,18 +56,42 @@ public async Task<AddCustomTagResponse> AddCustomTagAsync(IEnumerable<CustomTagE
...
+                    CustomTagStoreEntry storeEntry = new CustomTagStoreEntry(key, path, vr, customTag.Level, initStatus);
                     addedTags.Add(key, storeEntry);
                 }
                 catch (Exception ex)
                 {
-                    _logger.LogCritical(ex, "Failed to add custom tag {tag}.", tag);
+                    _logger.LogCritical(ex, "Failed to add custom tag {tag}.", customTag);
```
brandonpollett, line 94: nit: Update the placeholder too

src/Microsoft.Health.Dicom.Core/Features/CustomTag/ReindexJob.cs
```diff
@@ -0,0 +1,61 @@
...
+            // 1. The reindex status (endWatermark) should be updated as reindexing
+            // 2. job framework should allow resume, so that if reindex fail in the middle, user is able to resume.
+            if (customTagStoreEntries.Count() == 0)
+            {
+                return;
+            }
+
+            while (true)
+            {
+                IEnumerable<VersionedInstanceIdentifier> instances = await _customTagStore.GetInstancesInThePastAsync(endWatermark, top: Top, indexStatus: IndexStatus.Created, cancellationToken);
```
brandonpollett, line 44: There doesn't appear to be an implementation and the documentation isn't clear - is `endWatermark` inclusive or exclusive?

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
brandonpollett: In looking further up the stack, it appears that we'll be doing this for every single instance.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/ReindexJob.cs
```diff
@@ -0,0 +1,61 @@
...
+                if (instances.Count() == 0)
+                {
+                    break;
+                }
+
+                instances = instances.OrderByDescending(item => item.Version);
+                foreach (var instance in instances)
+                {
+                    await _instanceIndexer.IndexInstanceAsync(customTagStoreEntries, instance, cancellationToken);
```
brandonpollett, line 53: With this code, we're going to await for each instance to complete. You likely want to actually add all of these to an `IEnumerable<Task>` and use something like `await Task.WhenAll(tasks)`

src/Microsoft.Health.Dicom.Core/Features/CustomTag/ReindexJob.cs
pengchen0692: I rename it to maxWatermark, would it make more sense? Or better idea?

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -158,5 +159,36 @@ public static void AddValueIfNotNull<T>(this DicomDataset dicomDataset, DicomTag
...
             }
         }
+
+        /// <summary>
+        /// Get custom tag indexes
+        /// </summary>
+        /// <param name="dicomDataset">The dicom dataset.</param>
+        /// <param name="customTagStoreEntries">The custom tag entries stored.</param>
+        /// <returns>The indexes.</returns>
+        public static Dictionary<long, DicomItem> GetCustomTagIndexes(this DicomDataset dicomDataset, IDictionary<string, CustomTagStoreEntry> customTagStoreEntries)
```
smithago, line 169: Why would we not convert the path to DicomTag and use GetSingleValueOrDefault instead?

src/Microsoft.Health.Dicom.Core/Extensions/DicomTagExtensions.cs
```diff
@@ -24,5 +25,28 @@ public static string GetPath(this DicomTag dicomTag)
             EnsureArg.IsNotNull(dicomTag, nameof(dicomTag));
             return dicomTag.Group.ToString("X4") + dicomTag.Element.ToString("X4");
         }
+
+        /// <summary>
+        /// Get default VR for dicom tag.
+        /// </summary>
+        /// <param name="dicomTag">The dicm tag</param>
+        /// <returns>The default VR.</returns>
+        public static DicomVR GetDefaultVR(this DicomTag dicomTag)
```
smithago, line 34: Should'nt all tags have VR? So this function should be able to get it based on it right. Also do you think it should return null versus throwing exception?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagService.cs
```diff
@@ -47,24 +55,47 @@ public async Task<AddCustomTagResponse> AddCustomTagAsync(IEnumerable<CustomTagE
...
+                        if (string.IsNullOrWhiteSpace(vr))
+                        {
+                            vr = tag.GetDefaultVR()?.Code;
+                        }
+                    }
+
+                    CustomTagStoreEntry storeEntry = await _customTagStore.AddCustomTagAsync(path, vr, customTag.Level, initStatus);
+                    addedCustomTagStoreEntries.Add(storeEntry);
                 }
                 catch (Exception ex)
```
smithago, line 91: Why are we catching and deleting here? Add should be last thing to happen.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagService.cs
```diff
@@ -79,14 +110,14 @@ public async Task<AddCustomTagResponse> AddCustomTagAsync(IEnumerable<CustomTagE
             {
                 // Reindex from latest one to earliest
                 // Current solution won't be able to handle when reindex fail in the middle, when moving to job framework, should solve it.
-                await _reindexJob.ReindexAsync(addedTags.Values, lastWatermark.Value);
+                await _reindexJob.ReindexAsync(addedCustomTagStoreEntries, lastWatermark.Value);
```
smithago, line 113: What happens if Reindexing fails?

smithago: Job framework will not handle it. It just executes the task you give to it. If it fails it is app failure like dicom tag value parsing fails, it will return failed status to reindex job

src/Microsoft.Health.Dicom.Core/Features/CustomTag/ReindexJob.cs
```diff
@@ -0,0 +1,79 @@
...
+            {
+                IEnumerable<VersionedInstanceIdentifier> instances = await _customTagStore.GetInstancesInThePastAsync(endWatermark, top: Top, indexStatus: IndexStatus.Created, cancellationToken);
+                if (instances.Count() == 0)
+                {
+                    break;
+                }
+
+                instances = instances.OrderByDescending(item => item.Version);
+
+                // Please note that, if reindexing any instances fails in IndexInstances, the excution throws exception and stop.
```
smithago, line 57: nit: typo execution

src/Microsoft.Health.Dicom.Core/Features/CustomTag/ReindexJob.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            EnsureArg.IsNotNull(customTagStore, nameof(customTagStore));
+            EnsureArg.IsNotNull(instanceIndexer, nameof(instanceIndexer));
+            _customTagStore = customTagStore;
+            _instanceIndexer = instanceIndexer;
+        }
+
+        public async Task ReindexAsync(IEnumerable<CustomTagStoreEntry> customTagStoreEntries, long endWatermark, CancellationToken cancellationToken)
+        {
+            // TODO: this is transient solution for main scenario to go through. There are several things need to be done when move onto job framework.
```
smithago, line 35: Lets have a conversation with Brandon on what is realisticaly covered by Job framework. Lets not assume things.
There are couple of things missing in this solution
1. The work needs to happen in batches, this won't scale from get go
2. Log everything, long running jobs are bound to fail, this helps to debug issues in prod easily
3. No matter where the code get executed on persisting the intermediate state, code needs to be return CustomTag.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/InstanceIndexer.cs
```diff
@@ -0,0 +1,47 @@
...
+            }
+
+            DicomDataset dataset = await _metadataStore.GetInstanceMetadataAsync(instance);
+            Dictionary<long, DicomItem> indexes = dataset.GetCustomTagIndexes(customTagStoreEntries);
+            if (indexes.Count == 0)
+            {
+                return;
+            }
+
+            // TODO:  AddCustomTagIndexes could fail, once move to job framework, resumeing the job should be able to solve this problem. Double check after moving.
```
smithago, line 43: nit: type resuming

src/Microsoft.Health.Dicom.Core/Features/CustomTag/ICustomTagStore.cs
```diff
@@ -44,5 +47,15 @@ public interface ICustomTagStore
...
+
+        /// <summary>
+        /// Get instances with watermark upto maxWatermark.
+        /// </summary>
+        /// <param name="maxWatermark">The max watermark</param>
+        /// <param name="top">Top X instanances.</param>
+        /// <param name="indexStatus">The index status.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns>The instances.</returns>
+        Task<IEnumerable<VersionedInstanceIdentifier>> GetInstancesInThePastAsync(long maxWatermark, int top, IndexStatus indexStatus = IndexStatus.Created, CancellationToken cancellationToken = default);
```
smithago, line 59: This does not fit in CustomtagStore, it is accessing the IndexStore, you should probably move it there

smithago: Also, any reason to make IndexStatus a param, do we want to pull anything other than created?

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
pengchen0692: hi Smitha, I'm not getting this.

When we support embed custom tag, path could be a list of DicomTag.

src/Microsoft.Health.Dicom.Core/Extensions/DicomTagExtensions.cs
pengchen0692: DicomElement(https://github.com/fo-dicom/fo-dicom/blob/development/DICOM/DicomElement.cs) has VR specified, while DicomTag is not.
Based on definition of DicomTag(https://github.com/fo-dicom/fo-dicom/blob/development/DICOM/DicomTag.cs), it has a bunch of candidate VRs -- defined in DictionaryEntry.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagService.cs
pengchen0692: I suppose the job  should have status like Thread, so that when job fails, it should allow resume based on the status.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagService.cs
pengchen0692: Actually I'm not very sure about the right solution here before know how the job framework works. That's why add several lines of TODO above.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/ICustomTagStore.cs
pengchen0692: > This does not fit in CustomtagStore, it is accessing the IndexStore, you should probably move it there

I'm thinking if also move GetLatestIntanceAsync into IndexStore




> Also, any reason to make IndexStatus a param, do we want to pull anything other than created?

I don't think we need IndexStatus rather than Created for now, adding it as param is just for extensibility.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagService.cs
pengchen0692: Since we don't have general job framework, these logic will be changed as our specific job framework comes out. Current plan is that adding custom tags and starting reindex job are done in one SQL transaction, the reindex job will reindex and update custom tag status after completion. 
The new logic should be able to solve all our concerns.

Since another PR is pending on this, and the logic here will be rewritten after our job framework is finished, it is OK to get this PR approved for now, so that dependent  code (for customTagStore) could be checked-in.

pengchen0692:
This code will be revisited once have job framework

Issue #505, opened 2021-01-15
Support multiple audiences


brandonpollett:
## Problem
Today, the security configuration allows a single audience for verification of JWT tokens. This audience is expected to be available in the `aud` claim of the JWT token. The `aud` claim per the [RFC](https://tools.ietf.org/html/rfc7519#section-4.1.3) is of the type `StringOrUri`. 

Using AAD v1 endpoints, the audience is passed as a `resource` parameter and returned in the `aud` claim. Generally speaking, these often are in the form of a URI that contain a trailing `/` but can also be a `guid`. 

Using AAD v2 endpoints, the audience is passed as part of a `scopes` parameter. A default scope is available such that you can pass a scope of `https://example.com/.default` and the resulting `aud` claim will be `https://example.com`. The behavior of AAD is to return everything before the final `/` as the `aud`. 

## User story
As an administrator, I want to support multiple `Security.Authentication.Audience` so that I can use AAD v1 and v2 easily.

## Proposals
Ideally, this change would not result in a breaking change to the service. I currently see three different ways this can be accomplished. 

- Update the `Security.Authentication.Audience` to allow a delimited list of audiences that will be allowed.
- Add a property `Security.Authentication.Audiences` that will be an array and will take precedence over the `Audience` property. 
- Add a property `Security.Authentication.AdditionalAudiences` that when present will add additional audiences in addition tot he `Audience` specified.

## Acceptance criteria
1. When I make a request with the `aud` claim of `http://example.com`, then the request is allowed.
1. When I make a request with the `aud` claim of `http://example.com/`, then the request is allowed.
1. When I make a request with the `aud` claim of `12345`, then the request is allowed.
1. A user can specific 1 or more audiences that will be acceptable to the service.

PR #510, opened 2021-01-20
Update to support multiple audiences

brandonpollett:
## Description
This PR updates the code to support having multiple audiences for authentication. It adds a secondary property called `Audiences` that takes precedence over the `Audience` property.

## Related issues
Addresses #505, [AB#78859](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/78859)

## Testing
Added unit tests and manually tested options.

### DIFF
--- a/src/Microsoft.Health.Dicom.Api/Modules/SecurityModule.cs
+++ b/src/Microsoft.Health.Dicom.Api/Modules/SecurityModule.cs
@@ -5,6 +5,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using EnsureThat;
 using Microsoft.AspNetCore.Authentication.JwtBearer;
 using Microsoft.AspNetCore.Authorization;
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.Dicom.Api/Modules/SecurityModule.cs
```diff
@@ -64,5 +70,13 @@ public void Load(IServiceCollection services)
 
             services.AddSingleton<IClaimsExtractor, PrincipalClaimsExtractor>();
         }
+
+        internal IEnumerable<string> GetAuthenticationAudiences()
+        {
+            return _securityConfiguration.Authentication.Audiences ?? new[]
+            {
+                _securityConfiguration.Authentication.Audience,
```
johnstairs, line 78: Do you get a reasonable error message when one of the entries is empty, like `Audience` being null or ""?

johnstairs (approved):

src/Microsoft.Health.Dicom.Api/Modules/SecurityModule.cs
brandonpollett: In the current implementation, it actually changes the behavior slightly if you don't have either being set. 

I'm going to update the code to maintain the current behavior if neither is specified.

brandonpollett: Also, you get a message like `Microsoft.IdentityModel.Tokens.SecurityTokenInvalidAudienceException: IDX10214: Audience validation failed. Audiences: 'https://example.com'. Did not match: validationParameters.ValidAudience: 'null' or validationParameters.ValidAudiences: 'empty'.`

smithago (approved):

PR #511, opened 2021-01-25
Add Basic RBAC Support

brandonpollett:
## Description
This PR introduces the basic implementation of RBAC. It follows the same model as the FHIR Server for Azure. Currently there are 3 data actions defined (`read`, `write`, `delete`) and two roles (`globalReader`, `globalAdmin`). 

The implementation is based on the idea that there's an `IDicomAuthorizationService` that is called in each handler to check if the appropriate `DataAction` is allowed for that user. 

## Related issues
Addresses [AB#79125](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/79125).

## Testing
Automated test for user with only read. Existing tests continue to pass with updated `GlobalAdmin` role.


src/Microsoft.Health.Dicom.Api/Features/Context/DicomRequestContextAfterAuthenticationMiddleware.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            EnsureArg.IsNotNull(next, nameof(next));
+
+            _next = next;
+        }
+
+        public async Task Invoke(HttpContext context, IDicomRequestContextAccessor dicomRequestContextAccessor)
+        {
+            // Now the authentication is completed successfully, sets the user.
+            if (context.User != null)
```
smithago, line 30: Should we check security enabled and fail if user is null here?

src/Microsoft.Health.Dicom.Core/Exceptions/UnauthorizedDicomActionException.cs
```diff
@@ -0,0 +1,15 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Dicom.Core.Exceptions
+{
+    public class UnauthorizedDicomActionException : DicomServerException
+    {
+        public UnauthorizedDicomActionException()
```
smithago, line 10: Should we add more details on what was expected and what was passed, so it is easier to troubleshoot for our customers

src/Microsoft.Health.Dicom.Core/Features/Security/roles.schema.json
```diff
@@ -0,0 +1,68 @@
+﻿{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "http://microsoft.com/fhir/roles.schema.json",
+    "definitions": {
+        "dataActions": {
+            "enum": [
+                "*",
+                "read",
```
smithago, line 8: copied from fhir?

brandonpollett: yep, good catch.

src/Microsoft.Health.Dicom.Web/DevelopmentIdentityProviderRegistrationExtensions.cs
```diff
@@ -0,0 +1,238 @@
...
+using Microsoft.AspNetCore.Builder;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Configuration.Json;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using Microsoft.Health.Dicom.Core.Configs;
+
+namespace Microsoft.Health.Dicom.Web
+{
+    public static class DevelopmentIdentityProviderRegistrationExtensions
```
smithago, line 24: I am assuming this all goes to the share health.development.identity

src/Microsoft.Health.Dicom.Web/roles.json
```diff
@@ -0,0 +1,25 @@
+﻿{
+    "$schema": "../Microsoft.Health.Dicom.Core/Features/Security/roles.schema.json",
+    "roles": [
+        {
+            "name": "globalReader",
+            "dataActions": [
+                "read"
+            ],
+            "notDataActions": [],
```
smithago, line 9: Is this the expected schema? I am not sure if we will ever use NotDataActions

src/Microsoft.Health.Dicom.Core/Features/ChangeFeed/ChangeFeedHandler.cs
```diff
@@ -8,24 +8,33 @@
...
-
-            _changeFeedService = changeFeedService;
+            _changeFeedService = EnsureArg.IsNotNull(changeFeedService, nameof(changeFeedService));
         }
 
         public async Task<ChangeFeedResponse> Handle(ChangeFeedRequest request, CancellationToken cancellationToken)
         {
             EnsureArg.IsNotNull(request, nameof(request));
+
+            if (await AuthorizationService.CheckAccess(DataActions.Read, cancellationToken) != DataActions.Read)
```
smithago, line 33: This signature is interesting, I was expecting
bool CheckAccess(ExpectedActions), instead we pass and return Actions and then compare here, why?

src/Microsoft.Health.Dicom.Web/roles.json
brandonpollett: Yes, this is the expected schema once defined for Azure RBAC - https://docs.microsoft.com/en-us/azure/role-based-access-control/role-definitions#notdataactions

src/Microsoft.Health.Dicom.Api/Features/Context/DicomRequestContextAfterAuthenticationMiddleware.cs
brandonpollett: I don't believe that would get us anything useful. At this point we're just seeing if we can decorate the `DicomRequestContext` with a user after the authentication has run. We still want the code to flow though because we have non-authenticated endpoints such as `/health/check`

src/Microsoft.Health.Dicom.Web/DevelopmentIdentityProviderRegistrationExtensions.cs
brandonpollett: Yes, a large portion of this can be abstracted to shared components.

src/Microsoft.Health.Dicom.Core/Features/ChangeFeed/ChangeFeedHandler.cs
brandonpollett: We're following the FHIR implementation for this currently. The reasoning for FHIR was to allow the following scenarios (maybe more in the future):
- Allow for the possibility to have a downgraded experience based on the permissions
- Allow a single call to get all possible permissions when in the context of a batch/transaction

src/Microsoft.Health.Dicom.Web/DevelopmentIdentityProviderRegistrationExtensions.cs
```diff
@@ -0,0 +1,238 @@
...
+                jsonConfigurationSource.ResolveFileProvider();
+                return new Provider(jsonConfigurationSource, _existingConfiguration);
+            }
+
+            private class Provider : JsonConfigurationProvider
+            {
+                private const string AuthorityKey = "FhirServer:Security:Authentication:Authority";
+                private const string AudienceKey = "FhirServer:Security:Authentication:Audience";
+                private const string DevelopmentIdpEnabledKey = "DevelopmentIdentityProvider:Enabled";
+                private const string PrincipalClaimsKey = "FhirServer:Security:PrincipalClaims";
```
brandonpollett, line 175: Needs updated.

src/Microsoft.Health.Dicom.Core/Exceptions/UnauthorizedDicomActionException.cs
smithago: @brandonpollett, will this be addressed?

brandonpollett: I think it would be a good idea to log the exceptions, but I don't think we should make the error message to the end user any more verbose.

src/Microsoft.Health.Dicom.Web/DevelopmentIdentityProviderRegistrationExtensions.cs
smithago: I already have initial authentication. Will this move in the next PR?

smithago (approved):

PR #518, opened 2021-01-26
Remove reindex from AddCustomTagService

pengchen0692:
## Description
Remove reindex from AddCustomTagService

## Related issues
Addresses [AB#78352](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/78352)


src/Microsoft.Health.Dicom.Core/Extensions/DicomTagExtensions.cs
```diff
@@ -22,7 +23,30 @@ public static class DicomTagExtensions
         public static string GetPath(this DicomTag dicomTag)
         {
             EnsureArg.IsNotNull(dicomTag, nameof(dicomTag));
-            return dicomTag.Group.ToString("X4") + dicomTag.Element.ToString("X4");
+            return dicomTag.Group.ToString("X4").ToUpperInvariant() + dicomTag.Element.ToString("X4").ToUpperInvariant();
```
wsugarman, line 26: >Group [](start = 28, length = 5)

Are `Group` and `Element` always non-`null`?

wsugarman, line 26: >X [](start = 44, length = 1)

Uppercase `"X"` should be sufficient to make the characters uppercase, as per this doc: [https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings#the-hexadecimal-x-format-specifier](https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings#the-hexadecimal-x-format-specifier)

src/Microsoft.Health.Dicom.Core/Extensions/DicomTagExtensions.cs
```diff
@@ -22,7 +23,30 @@ public static class DicomTagExtensions
...
+        }
+
+        /// <summary>
+        /// Get default VR for dicom tag.
+        /// </summary>
+        /// <param name="dicomTag">The dicm tag</param>
+        /// <returns>The default VR.</returns>
```
wsugarman, line 33: nit: there should probably be some mention that the method returns `null` under certain circumstances, probably in the `<returns>`, but a `<remarks>` may also be appropriate.

src/Microsoft.Health.Dicom.Core/Extensions/DicomTagExtensions.cs
```diff
@@ -22,7 +23,30 @@ public static class DicomTagExtensions
...
+                return null;
+            }
+
+            if (dicomTag.DictionaryEntry == DicomDictionary.UnknownTag)
+            {
+                // this tag is invalid
+                return null;
+            }
+
+            return dicomTag.DictionaryEntry.ValueRepresentations.FirstOrDefault();
```
wsugarman, line 49: >FirstOrDefault [](start = 65, length = 14)

nit: while convenient, checking and using array access is faster and doesn't involve the extra allocation from the `IEnumerable<T>`:
```csharp
return dicomTag.DictionaryEntry.ValueRepresentations.Length > 0 ? dicomTag.DictionaryEntry.ValueRepresentation[0] : null;
 ```

File: src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs [hunk 4]
```diff
@@ -10,11 +10,16 @@ namespace Microsoft.Health.Dicom.Core.Features.CustomTag
     /// </summary>
     public class CustomTagEntry
```
wsugarman, line 11: >CustomTagEntry [](start = 17, length = 14)

I think I wrote this comment in Sneha's PR, but this class is currently mutable (as the properties have `public` setters). If this is intended, then we don't really need the ctors. If we want to make it immutable (which is generally preferable), then we can either make it a `struct` or perhaps even a new `record` type.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntryFormalizer.cs
```diff
@@ -0,0 +1,53 @@
...
+            DicomTag[] tags;
+            if (!_dicomTagParser.TryParse(customTagEntry.Path, out tags, supportMultiple: false))
+            {
+                // not a valid dicom tag path
+                throw new CustomTagEntryValidationException(
+                    string.Format(CultureInfo.InvariantCulture, DicomCoreResource.InvalidCustomTag, customTagEntry));
+            }
+
+            DicomTag tag = tags[0];
+            string path = tag.GetPath().ToUpperInvariant();
```
wsugarman, line 36: >ToUpperInvariant [](start = 40, length = 16)

This is already uppercase, right?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntryFormalizer.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Globalization;
+using Dicom;
+using EnsureThat;
+using Microsoft.Health.Dicom.Core.Exceptions;
+using Microsoft.Health.Dicom.Core.Extensions;
+using Microsoft.Health.Dicom.Core.Features.Common;
+
+namespace Microsoft.Health.Dicom.Core.Features.CustomTag
+{
+    public class CustomTagEntryFormalizer : ICustomTagEntryFormalizer
```
wsugarman, line 15: >public [](start = 4, length = 6)

Does this need to be `public`? This goes for all classes in the PR

src/Microsoft.Health.Dicom.Core.UnitTests/Extensions/DicomTagExtensionsTests.cs
```diff
@@ -0,0 +1,35 @@
...
+    public class DicomTagExtensionsTests
+    {
+        [Fact]
+        public void GivenValidDicomTag_WhenGetPath_ThenShouldReturnCorrectValue()
+        {
+            Assert.Equal("0014408B", DicomTag.UserSelectedGainY.GetPath());
+        }
+
+        [MemberData(nameof(MemberDataForTestingGetDefaultVR))]
+        [Theory]
```
wsugarman, line 22: >[Theory] [](start = 8, length = 8)

nit: I have typically seen the `[Theory]` and `[Fact]` attributes on top, but this may be the style of the repo

src/Microsoft.Health.Dicom.Core.UnitTests/Features/CustomTag/CustomTagEntryFormalizerTests.cs
```diff
@@ -0,0 +1,92 @@
...
+
+namespace Microsoft.Health.Dicom.Core.UnitTests.Features.ChangeFeed
+{
+    public class CustomTagEntryFormalizerTests
+    {
+        private ICustomTagEntryFormalizer _customTagEntryFormalizer;
+
+        public CustomTagEntryFormalizerTests()
+        {
+            _customTagEntryFormalizer = new CustomTagEntryFormalizer(new DicomTagParser());
```
wsugarman, line 22: >_customTagEntryFormalizer = new CustomTagEntryFormalizer(new DicomTagParser()); [](start = 12, length = 79)

This isn't stateful, right? So we can probably just initialize this as a `static` `readonly` `private` member

src/Microsoft.Health.Dicom.Core.UnitTests/Features/CustomTag/CustomTagEntryFormalizerTests.cs
```diff
@@ -0,0 +1,92 @@
...
+            string expectedPath = tag.GetPath();
+            CustomTagEntry formalized = _customTagEntryFormalizer.Formalize(entry);
+            Assert.Equal(formalized.Path, expectedPath);
+        }
+
+        public static IEnumerable<object[]> GetValidCustomTagEntries()
+        {
+            yield return new object[] { DicomTag.DeviceSerialNumber.BuildCustomTagEntry() }; // standard custom tag with VR
+            yield return new object[] { new CustomTagEntry("12051003", DicomVRCode.OB, CustomTagLevel.Instance, CustomTagStatus.Added) }; // private tag with VR
+        }
```
wsugarman, line 90: I don't think you have to make a function for `MemberData`. Alternatively, and perhaps more simply, you can have a property or even field that is enumerable:

```csharp
public static IEnumerable<object[]> GetValidCustomTagEntries =>
    new object[][]
    {
        new object[] { DicomTag.DeviceSerialNumber.BuildCustomTagEntry() }, // standard custom tag with VR
        new object[] { new CustomTagEntry("12051003", DicomVRCode.OB, CustomTagLevel.Instance, CustomTagStatus.Added) }, // private tag with VR
    };
```

src/Microsoft.Health.Dicom.Core.UnitTests/Extensions/DicomTagExtensionsTests.cs
pengchen0692: it's perfect if there is some tools for this. :)
Updating...

src/Microsoft.Health.Dicom.Core/Features/CustomTag/ICustomTagEntryFormalizer.cs
```diff
@@ -0,0 +1,20 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Dicom.Core.Features.CustomTag
+{
+    /// <summary>
+    /// Formalize custom tag entry.
+    /// </summary>
+    public interface ICustomTagEntryFormalizer
```
wsugarman, line 11: >ICustomTagEntryFormalizer [](start = 21, length = 25)

Is this something that requires an abstraction like an interface, or is this better off being simply part of the `CustomTagService`?

wsugarman (changes requested):
:clock1:

src/Microsoft.Health.Dicom.Core.UnitTests/Features/CustomTag/CustomTagEntryFormalizerTests.cs
pengchen0692: This is really good idea, never think of this.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
pengchen0692: As Sneha and I share same file, I leave this to Sneha, since her code has more functions.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntryFormalizer.cs
pengchen0692: **internal** also works here.  what's the benefit with that?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntryFormalizer.cs
pengchen0692: yeap

src/Microsoft.Health.Dicom.Core/Features/CustomTag/ICustomTagEntryFormalizer.cs
pengchen0692: Extract out mostly from unit test perspective. 
It's fine to put into a private method in CustomTagService, but would make it harder to write unit tests.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntryFormalizer.cs
pengchen0692: Moved it to internal static method as we discussed.

src/Microsoft.Health.Dicom.Core/Extensions/CustomTagEntryExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// Extension methods for <see cref="CustomTagEntry"/>.
+    /// </summary>
+    internal static class CustomTagEntryExtensions
+    {
+        /// <summary>
+        /// Normalize custom tag entry before saving to CustomTagStore.
+        /// </summary>
+        /// <param name="customTagEntry">The custom tag entry.</param>
+        /// <returns>Formalized custom tag entry.</returns>
+        internal static CustomTagEntry Normalize(this CustomTagEntry customTagEntry)
```
wsugarman, line 24: >internal [](start = 8, length = 8)

Typically I write the class's access modifiers as if the class were `public`, and then adjust the class's modifier as appropriate. So in this case, I think it's ok to make these `public`

src/Microsoft.Health.Dicom.Core/Extensions/CustomTagEntryExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// <summary>
+    /// Extension methods for <see cref="CustomTagEntry"/>.
+    /// </summary>
+    internal static class CustomTagEntryExtensions
+    {
+        /// <summary>
+        /// Normalize custom tag entry before saving to CustomTagStore.
+        /// </summary>
+        /// <param name="customTagEntry">The custom tag entry.</param>
+        /// <returns>Formalized custom tag entry.</returns>
```
wsugarman, line 23: >F [](start = 21, length = 1)

nit: `Normalized tag entry`

src/Microsoft.Health.Dicom.Core/Extensions/DicomTagExtensions.cs
wsugarman: I would use `<see langword="null">` instead of the word "null"

---
In reply to: [564902240](https://github.com/microsoft/dicom-server/pull/518#discussion_r564902240) [](ancestors = 564902240)

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagService.cs
```diff
@@ -3,91 +3,46 @@
...
-
-            // if lastWatermark doesn't exist, means no instance in database
-            if (lastWatermark.HasValue)
+            IEnumerable<CustomTagEntry> result = customTags.Select(item =>
             {
-                // Reindex from latest one to earliest
-                // Current solution won't be able to handle when reindex fail in the middle, when moving to job framework, should solve it.
-                await _reindexJob.ReindexAsync(addedTags.Values, lastWatermark.Value);
-            }
+                CustomTagEntry formalized = item.Normalize();
```
wsugarman, line 40: >formalized [](start = 31, length = 10)

nit: `normalized`

wsugarman (approved):
:shipit:

PR #519, opened 2021-01-27
[Dicom-Cast] ARM template for table storage and partial validation

abdullah248:
## Description
Update ARM template to configure partial validation and storage to table storage.

## Related issues
[AB#78974](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/78974)

## Testing
Created deployment with the template.


converter/dicom-cast/samples/templates/default-azuredeploy.json
```diff
@@ -94,6 +94,20 @@
             "metadata": {
                 "description": "The endpoint of the FHIR server."
             }
+        },
+        "enableTableStorage": {
```
smithago, line 98: Do we want to expose these has params? Or should we just have the default behavior of, enableTableStorage: True and set the IgnoreSyncOfInvalidTags: true in the container.

smithago: Basically reduce the settings here to get the default dicom-cast deployment.

abdullah248: Updated to set tablestorage to true and leave IgnoreSyncOfInvalidTags as configurable

converter/dicom-cast/samples/templates/default-azuredeploy.json
```diff
@@ -94,6 +94,13 @@
             "metadata": {
                 "description": "The endpoint of the FHIR server."
             }
+        },
+        "IgnoreSyncOfInvalidTagValue": {
+            "defaultValue": false,
```
smithago, line 99: Sorry Abdullah, should we make this true by default?

smithago (approved):

converter/dicom-cast/samples/templates/default-azuredeploy.json
abdullah248: Updated

PR #523, opened 2021-01-27
Adding in tables for CustomTag

snehabandla:
## Description
Adding table for custom tag, custom tag data and indexes.

## Related issues
[AB#79129](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/79129)


src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -364,6 +364,274 @@ CREATE NONCLUSTERED INDEX IX_ChangeFeed_StudyInstanceUid_SeriesInstanceUid_SopIn
...
+    Stores added custom tags
+    TagPath is represented without any delimiters and each level takes 8 bytes
+    TagLevel can be 0, 1 or 2 to represent Instance, Series or Study level
+**************************************************************/
+CREATE TABLE dbo.CustomTag (
+    TagKey                  BIGINT               NOT NULL, --PK
+    TagPath                 VARCHAR(64)          NOT NULL,
+    TagVR                   VARCHAR(2)           NOT NULL,
+    TagLevel                TINYINT              NOT NULL
+)
```
wsugarman, line 378: If we're creating clustered indices on the primary key, should we just create a primary key constraint instead?

```sql
    TagLevel TINYINT NOT NULL,
    CONSTRAINT PK_CustomTag_TagKey PRIMARY KEY CLUSTERED (TagKey)
```

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -364,6 +364,274 @@ CREATE NONCLUSTERED INDEX IX_ChangeFeed_StudyInstanceUid_SeriesInstanceUid_SopIn
...
+
+/*************************************************************
+    Custom Tag Data Table for VR Types mapping to Int
+**************************************************************/
+CREATE TABLE dbo.CustomTagInt (
+    TagKey                  BIGINT               NOT NULL, --PK
+    TagValue                BIGINT               NOT NULL,
+    StudyKey                BIGINT               NOT NULL, --FK
+    SeriesKey               BIGINT               NULL,     --FK
+    InstanceKey             BIGINT               NULL,     --FK
```
wsugarman, line 446: >FK [](start = 61, length = 2)

Do we avoid declaring foreign key constraints to avoid performance penalties when inserting?

wsugarman (approved):
:shipit:

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -364,6 +364,274 @@ CREATE NONCLUSTERED INDEX IX_ChangeFeed_StudyInstanceUid_SeriesInstanceUid_SopIn
...
+(
+    TagPath
+)
+
+/*************************************************************
+    Custom Tag Data Table for VR Types mapping to String
+**************************************************************/
+CREATE TABLE dbo.CustomTagString (
+    TagKey                  BIGINT               NOT NULL, --PK
+    TagValue                VARCHAR(64)          NOT NULL,
```
pengchen0692, line 395: should tagValue be NVARCHAR(64) to better support unicode? https://stackoverflow.com/questions/144283/what-is-the-difference-between-varchar-and-nvarchar

wsugarman: Oh that's a good catch!

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -364,6 +364,274 @@ CREATE NONCLUSTERED INDEX IX_ChangeFeed_StudyInstanceUid_SeriesInstanceUid_SopIn
...
+    TagKey,
+    TagValue
+)
+
+/*************************************************************
+    Custom Tag Data Table for VR Types mapping to PersonName
+**************************************************************/
+CREATE TABLE dbo.CustomTagPersonName (
+    TagKey                  BIGINT               NOT NULL, --PK
+    TagValue                NVARCHAR(200)        COLLATE SQL_Latin1_General_CP1_CI_AI NOT NULL,
```
pengchen0692, line 587: how about NVARCHAR(324)?
As DICOM  standard, the possible max name length should be 324, I'll find the link later.

pengchen0692: http://dicom.nema.org/dicom/2013/output/chtml/part05/sect_6.2.html

There are totally 5 component group , and max length of each is 64, plus 4 delimiters, add up to 324

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -364,6 +364,274 @@ CREATE NONCLUSTERED INDEX IX_ChangeFeed_StudyInstanceUid_SeriesInstanceUid_SopIn
     SopInstanceUid
 )
 
+/*************************************************************
```
pengchen0692, line 367: I'm thinking if we should put all custom tag changes together, and put on the tail, mark as **version2**, so that can easily differentiate 2 vs 1.sql

smithago: THis shoud be fine. 2.diff.sql will capture that once you add it.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -364,6 +364,274 @@ CREATE NONCLUSTERED INDEX IX_ChangeFeed_StudyInstanceUid_SeriesInstanceUid_SopIn
...
+    Custom Tag Table
+    Stores added custom tags
+    TagPath is represented without any delimiters and each level takes 8 bytes
```
smithago, line 370: Why is the models not generated and included in this PR?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -364,6 +364,274 @@ CREATE NONCLUSTERED INDEX IX_ChangeFeed_StudyInstanceUid_SeriesInstanceUid_SopIn
...
+    StudyKey,
+    SeriesKey,
+    InstanceKey
+)
+INCLUDE
+(
+    WaterMark
+)
+
+CREATE NONCLUSTERED INDEX IXC_CustomTagString_TagKey_StudyKey ON dbo.CustomTagString
```
smithago, line 414: This is not needed, the unique cluster already covers this

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -364,6 +364,274 @@ CREATE NONCLUSTERED INDEX IX_ChangeFeed_StudyInstanceUid_SeriesInstanceUid_SopIn
...
+(
+    TagKey,
+    StudyKey,
+)
+
+CREATE NONCLUSTERED INDEX IXC_CustomTagString_TagKey_SeriesKey ON dbo.CustomTagString
```
smithago, line 420: When will we use the below 3 indexed? We should always have StudyKey

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -364,6 +364,274 @@ CREATE NONCLUSTERED INDEX IX_ChangeFeed_StudyInstanceUid_SeriesInstanceUid_SopIn
...
+    StudyKey,
+    SeriesKey,
+    InstanceKey
+)
+INCLUDE
+(
+    WaterMark
+)
+
+CREATE NONCLUSTERED INDEX IXC_CustomTagInt_TagKey_StudyKey ON dbo.CustomTagInt
```
smithago, line 462: same has above

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
pengchen0692: Hi Smitha, would 2.diff.sql be generated from any tool?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -400,6 +668,13 @@ CREATE SEQUENCE dbo.InstanceKeySequence
     NO CYCLE
     CACHE 1000000
 
+CREATE SEQUENCE dbo.TagKeySequence
+    AS BIGINT
+    START WITH 1
+    INCREMENT BY 1
+    MINVALUE 1
+    NO CYCLE
+    CACHE 1000000
```
pengchen0692, line 677: Customer not likely to have large number of custom tags, would 1000000 too big? how about 10000?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
snehabandla: It seems like our current table for PersonName only uses 200. Let me check and see why that works!

---
In reply to: [565726566](https://github.com/microsoft/dicom-server/pull/523#discussion_r565726566) [](ancestors = 565726566)

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
snehabandla: Hmm, that's a good point. We don't currently seem to follow that pattern, I'll find out why!

---
In reply to: [565686889](https://github.com/microsoft/dicom-server/pull/523#discussion_r565686889) [](ancestors = 565686889)

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Model/V2.Generated.cs
```diff
@@ -0,0 +1,439 @@
...
+        }
+
+        internal class CustomTagDoubleTable : Table
+        {
+            internal CustomTagDoubleTable() : base("dbo.CustomTagDouble")
+            {
+            }
+
+            internal readonly BigIntColumn TagKey = new BigIntColumn("TagKey");
+            internal readonly FloatColumn TagValue = new FloatColumn("TagValue", 53);
```
snehabandla, line 92: >FloatColumn [](start = 30, length = 11)

This is currently not supported. Will need to add to shared components first!

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -364,6 +364,184 @@ CREATE NONCLUSTERED INDEX IX_ChangeFeed_StudyInstanceUid_SeriesInstanceUid_SopIn
...
+    Stores added custom tags
+    TagPath is represented without any delimiters and each level takes 8 bytes
+    TagLevel can be 0, 1 or 2 to represent Instance, Series or Study level
+**************************************************************/
+CREATE TABLE dbo.CustomTag (
+    TagKey                  BIGINT               NOT NULL, --PK
+    TagPath                 VARCHAR(64)          NOT NULL,
+    TagVR                   VARCHAR(2)           NOT NULL,
+    TagLevel                TINYINT              NOT NULL
+)
```
pengchen0692, line 378: is TagStatus missed here?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
pengchen0692: I brought up this question before in standup, and created user story https://microsofthealth.visualstudio.com/Health/_workitems/edit/79257. But we could double check Smitha who wrote this SQL before.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
smithago: No, needs to be manually created

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -364,6 +364,184 @@ CREATE NONCLUSTERED INDEX IX_ChangeFeed_StudyInstanceUid_SeriesInstanceUid_SopIn
...
+CREATE NONCLUSTERED INDEX IXC_CustomTagString_TagKey_TagValue ON dbo.CustomTagString
+(
+    TagKey,
+    TagValue
+)
+
+/*************************************************************
+    Custom Tag Data Table for VR Types mapping to Int
+**************************************************************/
+CREATE TABLE dbo.CustomTagInt (
```
pengchen0692, line 423: CustomTagLong or CustomTagBigInt? It actually maps to BigInt.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -364,6 +364,184 @@ CREATE NONCLUSTERED INDEX IX_ChangeFeed_StudyInstanceUid_SeriesInstanceUid_SopIn
...
+) WITH (DATA_COMPRESSION = PAGE)
+
+CREATE UNIQUE CLUSTERED INDEX IXC_CustomTagString ON dbo.CustomTagString
+(
+    TagKey,
+    StudyKey,
+    SeriesKey,
+    InstanceKey
+)
+INCLUDE
```
pengchen0692, line 409: cannot include for Clustered index

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -429,19 +427,15 @@ CREATE TABLE dbo.CustomTagInt (
     Watermark               BIGINT               NOT NULL
 ) WITH (DATA_COMPRESSION = PAGE)
 
-CREATE UNIQUE CLUSTERED INDEX IXC_CustomTagInt ON dbo.CustomTagInt
+CREATE UNIQUE CLUSTERED INDEX IXC_CustomTagBigInt ON dbo.CustomTagInt
```
pengchen0692, line 430: ON dbo.CustomTagInt => dbo.CustomTagBigInt

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -364,6 +364,166 @@ CREATE NONCLUSTERED INDEX IX_ChangeFeed_StudyInstanceUid_SeriesInstanceUid_SopIn
...
+)
+
+CREATE NONCLUSTERED INDEX IXC_CustomTagString_TagKey_TagValue ON dbo.CustomTagString
+(
+    TagKey,
+    TagValue
+)
+
+/*************************************************************
+    Custom Tag Data Table for VR Types mapping to Int
```
pengchen0692, line 419: nit: mapping to BigInt?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -364,6 +364,166 @@ CREATE NONCLUSTERED INDEX IX_ChangeFeed_StudyInstanceUid_SeriesInstanceUid_SopIn
...
+    InstanceKey
+)
+
+CREATE NONCLUSTERED INDEX IXC_CustomTagPersonName_TagKey_TagValue ON dbo.CustomTagPersonName
+(
+    TagKey,
+    TagValue
+)
+
+CREATE FULLTEXT INDEX ON CustomTagPersonName(TagValueWords LANGUAGE 1033)
```
pengchen0692, line 523: I got failure when executing:
'IXC_CustomTagPersonName' is not a valid index to enforce a full-text search key. A full-text search key must be a unique, non-nullable, single-column index which is not offline, is not defined on a non-deterministic or imprecise nonpersisted computed column, does not have a filter, and has maximum size of 900 bytes. Choose another index for the full-text key.

I'm thinking we could exclude personname from the min scope.

PR #527, opened 2021-01-29
Add Semver

abdullah248:
## Description
Implement semver based on how it was implemented in FHIR and healthcare-shared-components repos.

Below are instructions copied from Brendans pr into healthcare shared components:
Adds semver with GitVersioning

This package allows controlling versioning through Git, version bumps can be utilized through +semver: major|minor|feature|breaking syntax in commit messages.

For all incremental bug fixes, nothing needs to be changed, GitVersion will increment the patch version automatically based on commits in main.

For breaking changes you can use the following syntax in the commit message:

```
+semver: breaking
or
+semver: major
```
Features can use:

```
+semver: feature
or
+semver: minor
```

## Addresses
[AB#77545](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/77545)

## Testing
Created and ran the first few stages in a test pipeline.

### DIFF
--- a/GitVersion.yml
+++ b/GitVersion.yml
@@ -1,0 +1,40 @@
+mode: Mainline
+# assembly-version: If this number changes, other assemblies have to update references to the assembly
+assembly-versioning-scheme: Major
+assembly-file-versioning-scheme: MajorMinorPatch
+ignore:
+  sha: []
+branches:
+  master:
+    regex: ^master$|^main$
+    tag: ''
+    increment: Patch
+    prevent-increment-of-merged-branch-version: true
+    track-merge-target: false
+    tracks-release-branches: false
+    is-release-branch: true
+  pull-request:
+    regex: ^(pull|pull\-requests|pr)[/-]
+    tag: pr
+    increment: Inherit
+    prevent-increment-of-merged-branch-version: false
+    tag-number-pattern: '[/-](?<number>\d+)[-/]'
+    track-merge-target: false
+    tracks-release-branches: false
+    is-release-branch: false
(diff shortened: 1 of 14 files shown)


GitVersion.yml
```diff
@@ -0,0 +1,40 @@
...
+    tag: pr
+    increment: Inherit
+    prevent-increment-of-merged-branch-version: false
+    tag-number-pattern: '[/-](?<number>\d+)[-/]'
+    track-merge-target: false
+    tracks-release-branches: false
+    is-release-branch: false
+  hotfix:
+    regex: ^hotfix(es)?[/-]
+    tag: beta
```
brandonpollett, line 27: It looks like in the fhir-server repo, they're using `useBranchName` here for a hotfix -- is this something we want to follow?

brandonpollett (approved):
Not sure if we have any pre-release nugets, but it may be an issue if we try to publish a non pre-release version with a reference to one and we don't add the following to `Directory.Build.props`

```
    <!-- allow pre-release dependencies -->
    <NoWarn>$(NoWarn);NU5104</NoWarn>
```

snehabandla (approved):
Still would like to find a way to normalize this across our repos without duplicating code. It seems too easy to miss a change and diverge how we handle versioning.

wsugarman (approved):

GitVersion.yml
abdullah248: Updated and also updated the Directory.Build.Props

PR #531, opened 2021-02-01
Update NuGet Config

wsugarman:
## Description
Remove obsolete elements from NuGet config

## Related issues
N/A

## Testing
Build pipelines should still succeed

### DIFF
--- a/nuget.config
+++ b/nuget.config
@@ -1,18 +1,12 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
-  <config>
-    <add key="repositoryPath" value="packages" />
-  </config>
   <packageSources>
     <!-- When <clear /> is present, previously defined sources are ignored -->
     <clear />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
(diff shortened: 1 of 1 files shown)


nuget.config
```diff
@@ -1,18 +1,12 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
-  <config>
```
wsugarman, line 3: This key only applies to when using a *package.config* according to the docs: https://docs.microsoft.com/en-us/nuget/reference/nuget-config-file#config-section

nuget.config
```diff
@@ -1,18 +1,12 @@
...
-    <add key="repositoryPath" value="packages" />
-  </config>
   <packageSources>
     <!-- When <clear /> is present, previously defined sources are ignored -->
     <clear />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
     <add key="Microsoft Health OSS" value="https://microsofthealthoss.pkgs.visualstudio.com/FhirServer/_packaging/Public/nuget/v3/index.json" />
   </packageSources>
-  <activePackageSource>
```
wsugarman, line 12: `activePackageSource` only applies up to NuGet 2.x: https://docs.microsoft.com/en-us/nuget/reference/nuget-config-file#activepackagesource

brandonpollett (approved):

PR #532, opened 2021-02-02
[Dicom-Cast] Move Sync State to Table Storage and Remove Blob Storage

abdullah248:
## Description
Moves storage of sync state from blob storage to azure table storage. Also removes the flag for table storage and makes it enabled at all times by default.

## Related issues
[AB#78065](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/78065)

## Testing
Current unit tests pass and manual end to end testing.


converter/dicom-cast/src/Microsoft.Health.DicomCast.TableStorage/Features/Storage/TableSyncStateStore.cs
```diff
@@ -0,0 +1,70 @@
...
+        public async Task<SyncState> ReadAsync(CancellationToken cancellationToken = default)
+        {
+            CloudTable table = _client.GetTableReference(Constants.SyncStateTableName);
+            TableOperation retrieveOperation = TableOperation.Retrieve<SyncStateEntity>(Constants.SyncStatePartitionKey, Constants.SyncStateRowKey);
+
+            TableResult result;
+
+            try
+            {
+                result = await table.ExecuteAsync(retrieveOperation);
```
wsugarman, line 37: `ExecuteAsync` appears to have an overload that accepts `CancellationToken`, so we should also use that to propogate the cancellation if desired. If we do that, it also means re-throwing any `OperationCanceledException`

converter/dicom-cast/src/Microsoft.Health.DicomCast.TableStorage/Features/Storage/TableSyncStateStore.cs
```diff
@@ -0,0 +1,70 @@
...
+            }
+
+            SyncStateEntity syncState = result.Result as SyncStateEntity;
+
+            if (syncState != null)
+            {
+                return new SyncState(syncState.SyncedSequence, syncState.Timestamp);
+            }
+
+            return SyncState.CreateInitialSyncState();
```
wsugarman, line 51: I think we can simplify this to:

```csharp
return result.Result is SyncStateEntity sse // do we need to check for null here too?
    ? new SyncState(sse.SyncedSequence, sse.Timestamp)
    : SyncState.CreateInitialSyncState();
```

converter/dicom-cast/src/Microsoft.Health.DicomCast.TableStorage/Features/Storage/TableSyncStateStore.cs
```diff
@@ -0,0 +1,70 @@
...
+
+            TableOperation operation = TableOperation.InsertOrReplace(entity);
+            try
+            {
+                await table.ExecuteAsync(operation, cancellationToken);
+            }
+            catch (Exception ex)
+            {
+                throw new DataStoreException(ex);
+            }
```
wsugarman, line 67: You should not wrap the possible `OperationCanceledException`

wsugarman (approved):

converter/dicom-cast/src/Microsoft.Health.DicomCast.TableStorage/Features/Storage/TableSyncStateStore.cs
wsugarman: Although, this is also perhaps my personal taste lol

abdullah248: Thanks, I looked into it and the "is" operator will actually check for null as well and will return false if it is null. I updated to match your suggestion.

converter/dicom-cast/src/Microsoft.Health.DicomCast.TableStorage/Features/Storage/TableSyncStateStore.cs
abdullah248: Removed the try catch, thanks

abdullah248: Removed try catch

converter/dicom-cast/src/Microsoft.Health.DicomCast.TableStorage/Features/Storage/TableSyncStateStore.cs
abdullah248: Thanks for the catch

KnicKnic (approved):

PR #528, opened 2021-01-30
Update Shared Packages and Metadata

wsugarman:
## Description
- Bump version of the shared healthcare packages
- Update NuGet package metadata and dependencies
- Fix malformed XML doc comments
- Update .NET Core runtime
- Avoid rebuilding assemblies in pipelines to ensure consistent build arguments

## Related issues
N/A

## Testing
Run unit tests and validate pipeline output

PR #534, opened 2021-02-03
Upgrading healthcare-shared-components packages

snehabandla:
Including change to support float as a SQL column type.


wsugarman:
I can bump to 1.1.3 in my PR https://github.com/microsoft/dicom-server/pull/528 if you'd like

PR #543, opened 2021-02-04
[Dicom-Cast] Add HTTPRequestException to Retryable

abdullah248:
## Description
Add HTTPRequestException as a retryable exception and add some unit tests.

## Related issues
[AB#78067](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/78067)

## Testing
Updated unit tests and manually tests


converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/FhirTransaction/RetryableFhirTransactionPipeline.cs
```diff
@@ -66,16 +68,20 @@ public Task ProcessAsync(ChangeFeedEntry changeFeedEntry, CancellationToken canc
...
                     try
                     {
-                       return _fhirTransactionPipeline.ProcessAsync(changeFeedEntry, cancellationToken);
+                       await _fhirTransactionPipeline.ProcessAsync(changeFeedEntry, cancellationToken);
                     }
                     catch (TaskCanceledException ex) when (!cancellationToken.IsCancellationRequested)
                     {
                         throw new RetryableException(ex);
                     }
+                    catch (HttpRequestException ex)
```
wsugarman, line 81: I think this makes sense to add to your policy instead, as the `TaskCanceledException` needed special-casing to differeniate cancellation scenarios

wsugarman (approved):

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/FhirTransaction/RetryableFhirTransactionPipeline.cs
abdullah248: Hmm so I added it here because in ChangeFeedProcessor I will otherwise need to catch HttpRequestExceptioin there as well. I thought it would be easier if I just wrap all exceptions that are retryable before so I can just catch retryable exception instead of each individual one.

KnicKnic (approved):

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/FhirTransaction/RetryableFhirTransactionPipeline.cs
wsugarman: I see.

Yeah, you can rethrow it like that. I haven't thought through it completely it, but it may be more accurate to push this sort of logic directly into the retrying pipeline since some of this logic is based specifically on it. But that can be done another time

abdullah248: Okay, Thanks!

PR #552, opened 2021-02-05
Add Controller/Handler/Service for deleting custom tag

pengchen0692:
## Description
Add Controller/Handler/Service for deleting custom tag

## Related issues
Addresses [AB#78139](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/78139)

## Testing
Unit tests pass


src/Microsoft.Health.Dicom.Core/Modules/ServiceModule.cs
```diff
@@ -144,7 +144,7 @@ public void Load(IServiceCollection services)
                     .AsSelf()
                     .AsImplementedInterfaces();
 
-                services.Add<CustomTagService>()
+                services.Add<AddCustomTagService>()
```
pengchen0692, line 147: add deletecustomtagservice as well.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/DeleteCustomTagService.cs
```diff
@@ -0,0 +1,78 @@
...
+    public class DeleteCustomTagService : IDeleteCustomTagService
+    {
+        private readonly ICustomTagStore _customTagStore;
+        private readonly IDicomTagParser _dicomTagParser;
+        private readonly ILogger<DeleteCustomTagService> _logger;
+
+        /// <summary>
+        /// Mapping from CustomTagVR to IndexType
+        /// </summary>
+        private static readonly IReadOnlyDictionary<string, CustomTagIndexType> CustomTagVRAndIndexTypeMapping = new Dictionary<string, CustomTagIndexType>()
```
smithago, line 28: Is'nt this mapping needed in Add? Would'nt this be a common customTag thing?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagIndexType.cs
```diff
@@ -0,0 +1,38 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Dicom.Core.Features.CustomTag
+{
+    /// <summary>
+    /// Index type of custom tag.
+    /// </summary>
+    public enum CustomTagIndexType
```
smithago, line 11: nit: DataType?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/DeleteCustomTagService.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            _customTagStore = customTagStore;
+            _dicomTagParser = dicomTagParser;
+            _logger = logger;
+        }
+
+        public async Task DeleteCustomTagAsync(string tagPath, CancellationToken cancellationToken)
+        {
+            DicomTag[] tags;
+            if (!_dicomTagParser.TryParse(tagPath, out tags, supportMultiple: false))
```
smithago, line 65: Would'nt it be better for the Parse to throw the more specific error, instead of catching and throw a generic error here?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/DeleteCustomTagService.cs
```diff
@@ -0,0 +1,78 @@
...
+            {
+                throw new InvalidCustomTagPathException(
+                    string.Format(CultureInfo.InvariantCulture, DicomCoreResource.InvalidCustomTag, tagPath ?? string.Empty));
+            }
+
+            string normalizedPath = tags[0].GetPath();
+
+            CustomTagEntry customTagEntry = await _customTagStore.GetCustomTagAsync(normalizedPath, cancellationToken);
```
smithago, line 73: Does this throw if the tag is not available? If not we should handle it here.

src/Microsoft.Health.Dicom.Core/Messages/CustomTag/DeleteCustomTagResponse.cs
```diff
@@ -0,0 +1,11 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Dicom.Core.Messages.CustomTag
```
smithago, line 6: Why do we have a empty class?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/DeleteCustomTagService.cs
pengchen0692: the GetCustomTagAsync should throw exception if not exist.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/DeleteCustomTagService.cs
pengchen0692: This is how QIDO currently handle with input, I'm thinking if we could leave it right now, and evolving as per custom tag feedback?

src/Microsoft.Health.Dicom.Core/Messages/CustomTag/DeleteCustomTagResponse.cs
pengchen0692: Following https://github.com/microsoft/dicom-server/blob/main/src/Microsoft.Health.Dicom.Core/Messages/Delete/DeleteResourcesResponse.cs.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/AddCustomTagService.cs
```diff
@@ -3,6 +3,7 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System;
```
wsugarman, line 6: Is this necessary?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/ICustomTagStore.cs
```diff
@@ -39,11 +39,12 @@ public interface ICustomTagStore
...
         /// </summary>
-        /// <param name="key">The tag key.</param>
-        /// <param name="cancellationToken">The cancellation token.</param>
-        /// <returns>Result.</returns>
-        Task DeleteCustomTagAsync(long key, CancellationToken cancellationToken = default);
+        /// <param name="tagPath">The tag path.</param>
+        /// <param name="customTagDataType">The custom tag data type.</param>
+        /// <param name="cancellationToken">The cancellation token</param>
+        /// <returns>The task.</returns>
+        Task DeleteCustomTagAsync(string tagPath, CustomTagDataType customTagDataType, CancellationToken cancellationToken = default);
```
wsugarman, line 48: I think the `CustomTagDataType` enum sort of implies that there should be separate handling per VR, but that seems more like an implementation detail. I would say to either:
1. Pass `CustomTagEntry` to the method instead of both `tagpath` and `customTagDataType` OR
2. Replace `customTagDataType` with the VR directly

Nevertheless, I think the concrete implementation should perform some logic equivalent to the current `CustomTagLimit` class by analyzing the VR.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/DeleteCustomTagService.cs
wsugarman: I think it makes sense too where `Parse` acts like an assertion that throws the most specific exception. However, we should be consistent across APIs with our error handling and it can probably wait if the change would affect too many endpoints.

wsugarman (changes requested):

src/Microsoft.Health.Dicom.Core/Features/CustomTag/AddCustomTagService.cs
pengchen0692: I think you find a bug in Visual Studio -- it is indicated as used, but deleting that line doesn't fail any thing.
image

src/Microsoft.Health.Dicom.Core/Features/CustomTag/AddCustomTagService.cs
```diff
@@ -2,7 +2,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
-
```
wsugarman, line 5: nit: newline between copyright and namespaces

wsugarman (approved):

PR #555, opened 2021-02-08
Update c# tutorial with nuget information

abdullah248:
## Description
Add link to public nuget feed for Dicom Web Client.

## Related issues
#553

### DIFF
--- a/docs/tutorials/use-dicom-web-standard-apis-with-c#.md
+++ b/docs/tutorials/use-dicom-web-standard-apis-with-c#.md
@@ -22,9 +22,14 @@ Once you have deployed an instance of the Medical Imaging Server for DICOM, retr
 1. Search for **App Services** and select your Medical Imaging Server for DICOM App Service.
 1. Copy the **URL** of your App Service.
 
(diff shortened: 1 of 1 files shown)


docs/tutorials/use-dicom-web-standard-apis-with-c#.md
```diff
@@ -22,9 +22,14 @@ Once you have deployed an instance of the Medical Imaging Server for DICOM, retr
 1. Search for **App Services** and select your Medical Imaging Server for DICOM App Service.
 1. Copy the **URL** of your App Service.
 
+In your application install the following nuget packages:
```
wsugarman, line 25: Would it be worth it instead to have a *NuGet.config* and *.csproj* movie inline?

wsugarman (approved):

docs/tutorials/use-dicom-web-standard-apis-with-c#.md
abdullah248: Updated to have links to nugets

PR #560, opened 2021-02-09
Implement SQLStore for DeleteCustomTag

pengchen0692:
## Description
Implement SQLStore for DeleteCustomTag

## Related issues
Addresses [AB#78132](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/78132)

### DIFF
--- a/src/Microsoft.Health.Dicom.SqlServer/DicomSqlServerResource.resx
+++ b/src/Microsoft.Health.Dicom.SqlServer/DicomSqlServerResource.resx
@@ -117,8 +117,8 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
-  <data name="InvalidCustomTagStatus" xml:space="preserve">
-    <value>The custom tag '{0}' is not in valid status.</value>
+  <data name="CustomTagIsBusy" xml:space="preserve">
+    <value>The custom tag '{0}' is being processed.</value>
     <comment>{0} is tag path</comment>
   </data>
   <data name="SchemaVersionNeedsToBeUpgraded" xml:space="preserve">
--- a/src/Microsoft.Health.Dicom.SqlServer/DicomSqlServerResource.resx
+++ b/src/Microsoft.Health.Dicom.SqlServer/DicomSqlServerResource.resx
@@ -117,6 +117,10 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="InvalidCustomTagStatus" xml:space="preserve">
+    <value>The custom tag '{0}' is not in invalid status.</value>
+    <comment>{0} is tag path</comment>
+  </data>
   <data name="SchemaVersionNeedsToBeUpgraded" xml:space="preserve">
     <value>Cannot carry out the SQL datastore operation because the SQL schema needs to be upgraded.</value>
   </data>
--- a/src/Microsoft.Health.Dicom.SqlServer/DicomSqlServerResource.resx
+++ b/src/Microsoft.Health.Dicom.SqlServer/DicomSqlServerResource.resx
@@ -118,7 +118,7 @@
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
   <data name="InvalidCustomTagStatus" xml:space="preserve">
-    <value>The custom tag '{0}' is not in invalid status.</value>
+    <value>The custom tag '{0}' is not in valid status.</value>
     <comment>{0} is tag path</comment>
   </data>
   <data name="SchemaVersionNeedsToBeUpgraded" xml:space="preserve">
--- a/src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/CustomTagDataType.cs
+++ b/src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/CustomTagDataType.cs
@@ -8,7 +8,7 @@ namespace Microsoft.Health.Dicom.SqlServer.Features.CustomTag
     /// <summary>
     /// Date type of custom tag.
     /// </summary>
-    public enum CustomTagDataType
+    internal enum CustomTagDataType
     {
         /// <summary>
         /// The custom tag is treated as String.
--- a/src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/CustomTagDataType.cs
+++ b/src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/CustomTagDataType.cs
@@ -1,0 +1,38 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Dicom.SqlServer.Features.CustomTag
+{
+    /// <summary>
+    /// Date type of custom tag.
+    /// </summary>
+    public enum CustomTagDataType
+    {
+        /// <summary>
+        /// The custom tag is treated as String.
+        /// </summary>
+        StringData = 0,
+
+        /// <summary>
+        /// The custom tag is treated as Long.
+        /// </summary>
+        LongData = 1,
+
+        /// <summary>
+        /// The custom tag is treated as Double.
+        /// </summary>
+        DoubleData = 2,
+
+        /// <summary>
+        /// The custom tag is treated as DateTime.
+        /// </summary>
+        DateTimeData = 3,
+
+        /// <summary>
+        /// The custom tag is treated as PersonName.
+        /// </summary>
+        PersonNameData = 4,
+    }
+}
--- a/src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/CustomTagLimit.cs
+++ b/src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/CustomTagLimit.cs
@@ -1,0 +1,41 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+using System.Collections.Generic;
+using Dicom;
+
+namespace Microsoft.Health.Dicom.SqlServer.Features.CustomTag
+{
+    /// <summary>
+    /// Limits on CustomTag feature.
+    /// </summary>
+    public static class CustomTagLimit
+    {
+        /// <summary>
+        /// Mapping from CustomTagVR to IndexType
+        /// </summary>
+        public static readonly IReadOnlyDictionary<string, CustomTagDataType> CustomTagVRAndIndexTypeMapping = new Dictionary<string, CustomTagDataType>()
+        {
+            { DicomVR.AE.Code, CustomTagDataType.StringData },
+            { DicomVR.AS.Code, CustomTagDataType.StringData },
+            { DicomVR.AT.Code, CustomTagDataType.LongData },
+            { DicomVR.CS.Code, CustomTagDataType.StringData },
+            { DicomVR.DA.Code, CustomTagDataType.DateTimeData },
+            { DicomVR.DS.Code, CustomTagDataType.StringData },
+            { DicomVR.DT.Code, CustomTagDataType.DateTimeData },
+            { DicomVR.FL.Code, CustomTagDataType.DoubleData },
+            { DicomVR.FD.Code, CustomTagDataType.DoubleData },
+            { DicomVR.IS.Code, CustomTagDataType.StringData },
+            { DicomVR.LO.Code, CustomTagDataType.StringData },
+            { DicomVR.PN.Code, CustomTagDataType.StringData },
+            { DicomVR.SH.Code, CustomTagDataType.StringData },
+            { DicomVR.SL.Code, CustomTagDataType.LongData },
+            { DicomVR.SS.Code, CustomTagDataType.LongData },
+            { DicomVR.TM.Code, CustomTagDataType.DateTimeData },
+            { DicomVR.UI.Code, CustomTagDataType.StringData },
+            { DicomVR.UL.Code, CustomTagDataType.LongData },
+            { DicomVR.US.Code, CustomTagDataType.LongData },
+        };
+    }
+}
--- a/src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/CustomTagLimit.cs
+++ b/src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/CustomTagLimit.cs
@@ -10,7 +10,7 @@ namespace Microsoft.Health.Dicom.SqlServer.Features.CustomTag
     /// <summary>
     /// Limits on CustomTag feature.
     /// </summary>
-    public static class CustomTagLimit
+    internal static class CustomTagLimit
     {
         /// <summary>
         /// Mapping from CustomTagVR to DateType
--- a/src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/CustomTagLimit.cs
+++ b/src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/CustomTagLimit.cs
@@ -13,9 +13,9 @@ namespace Microsoft.Health.Dicom.SqlServer.Features.CustomTag
     public static class CustomTagLimit
     {
         /// <summary>
-        /// Mapping from CustomTagVR to IndexType
+        /// Mapping from CustomTagVR to DateType
         /// </summary>
-        public static readonly IReadOnlyDictionary<string, CustomTagDataType> CustomTagVRAndIndexTypeMapping = new Dictionary<string, CustomTagDataType>()
+        public static readonly IReadOnlyDictionary<string, CustomTagDataType> CustomTagVRAndDataTypeMapping = new Dictionary<string, CustomTagDataType>()
         {
             { DicomVR.AE.Code, CustomTagDataType.StringData },
             { DicomVR.AS.Code, CustomTagDataType.StringData },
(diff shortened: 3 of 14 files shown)


src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/CustomTagLimit.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    /// <summary>
+    /// Limits on CustomTag feature.
+    /// </summary>
+    public static class CustomTagLimit
+    {
+        /// <summary>
+        /// Mapping from CustomTagVR to IndexType
+        /// </summary>
+        public static readonly IReadOnlyDictionary<string, CustomTagDataType> CustomTagVRAndIndexTypeMapping = new Dictionary<string, CustomTagDataType>()
```
pengchen0692, line 18: =>CustomTagVRAndDataTypeMapping

src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/CustomTagDataType.cs
```diff
@@ -0,0 +1,38 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Dicom.SqlServer.Features.CustomTag
+{
+    /// <summary>
+    /// Date type of custom tag.
+    /// </summary>
+    public enum CustomTagDataType
```
wsugarman, line 11: `internal`?

src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/CustomTagLimit.cs
```diff
@@ -0,0 +1,41 @@
...
+// -------------------------------------------------------------------------------------------------
+using System.Collections.Generic;
+using Dicom;
+
+namespace Microsoft.Health.Dicom.SqlServer.Features.CustomTag
+{
+    /// <summary>
+    /// Limits on CustomTag feature.
+    /// </summary>
+    public static class CustomTagLimit
```
wsugarman, line 13: Should this exist outside of the `SqlCustomTagStore`? I don't think any other class/component is going to use this mapping

src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/SqlCustomTagStore.cs
```diff
@@ -87,9 +87,34 @@ public Task<IEnumerable<CustomTagEntry>> GetAllCustomTagsAsync(CancellationToken
...
-            throw new System.NotImplementedException();
+            if (_schemaInformation.Current < SchemaVersionConstants.SupportCustomTagSchemaVersion)
+            {
+                throw new BadRequestException(DicomSqlServerResource.SchemaVersionNeedsToBeUpgraded);
+            }
+
+            using (SqlConnectionWrapper sqlConnectionWrapper = await _sqlConnectionWrapperFactory.ObtainSqlConnectionWrapperAsync(cancellationToken))
+            using (SqlCommandWrapper sqlCommandWrapper = sqlConnectionWrapper.CreateSqlCommand())
+            {
+                VLatest.DeleteCustomTag.PopulateCommand(sqlCommandWrapper, tagPath, (byte)CustomTagLimit.CustomTagVRAndDataTypeMapping[vr]);
```
wsugarman, line 100: What if the VR value is not found?

src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/CustomTagDataType.cs
wsugarman: And where is this now used?

wsugarman (changes requested):

src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/CustomTagDataType.cs
pengchen0692: should be internal, suppose only within this project.

src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/CustomTagLimit.cs
pengchen0692: shoudl be internal

src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/SqlCustomTagStore.cs
pengchen0692: technically should not happen, invalid VR should not be saved into database at first place.
If not found, should be a bug, and request fails for internal server failure.
But I'm not sure if this is how we normally do ? :)

src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/SqlCustomTagStore.cs
```diff
@@ -87,9 +87,34 @@ public Task<IEnumerable<CustomTagEntry>> GetAllCustomTagsAsync(CancellationToken
             throw new System.NotImplementedException();
         }
 
-        public Task DeleteCustomTagAsync(string tagPath, string vr, CancellationToken cancellationToken = default)
+        public async Task DeleteCustomTagAsync(string tagPath, string vr, CancellationToken cancellationToken = default)
         {
-            throw new System.NotImplementedException();
+            if (_schemaInformation.Current < SchemaVersionConstants.SupportCustomTagSchemaVersion)
+            {
+                throw new BadRequestException(DicomSqlServerResource.SchemaVersionNeedsToBeUpgraded);
```
smithago, line 94: Feature not supported right? notsupportedexception?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
```diff
@@ -95,3 +94,64 @@ AS
...
+        DECLARE @tagKey TINYINT
+        SELECT @tagKey = TagKey, @tagStatus = TagStatus
+        FROM dbo.CustomTag WITH(HOLDLOCK) 
+        WHERE dbo.CustomTag.TagPath = @tagPath
+
+        IF @@ROWCOUNT = 0
+            THROW 50404, 'custom tag not found', 1 
+
+        -- check if status is Added
+        IF @tagStatus <> 1
```
smithago, line 132: THis is not handled in the store layer

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
```diff
@@ -95,3 +94,64 @@ AS
...
+        IF @@ROWCOUNT = 0
+            THROW 50404, 'custom tag not found', 1 
+
+        -- check if status is Added
+        IF @tagStatus <> 1 
+            THROW 50412, 'custom tag is not in status Added', 1
+
+        -- Update status to deindexing
+        UPDATE dbo.CustomTag
+        SET TagStatus = 2
```
smithago, line 137: What are we getting by setting this state?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
```diff
@@ -95,3 +94,64 @@ AS
...
+        WHERE TagKey = @tagKey
+
+        -- Delete index data
+        IF @dataType = 0
+            DELETE FROM dbo.CustomTagString WHERE TagKey = @tagKey
```
smithago, line 142: These tables are not available yet.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
```diff
@@ -95,3 +94,64 @@ AS
...
+        ELSE IF @dataType = 1
+            DELETE FROM dbo.CustomTagBigInt WHERE TagKey = @tagKey
```
smithago, line 144: Can you check how long it will take to delete say 2 million rows? You may have to do it in batches.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
pengchen0692: this pends on Sneha's code.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
pengchen0692: about 15 seconds

src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/SqlCustomTagStore.cs
wsugarman: Sounds reasonable

src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/CustomTagLimit.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    /// <summary>
+    /// Limits on CustomTag feature.
+    /// </summary>
+    internal static class CustomTagLimit
+    {
+        /// <summary>
+        /// Mapping from CustomTagVR to DateType
+        /// </summary>
+        public static readonly IReadOnlyDictionary<string, CustomTagDataType> CustomTagVRAndDataTypeMapping = new Dictionary<string, CustomTagDataType>()
```
wsugarman, line 18: I think this should just be inside of `SqlCustomTagStore`, as there don't seem to be any other consumers of this dictionary.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
```diff
@@ -95,3 +94,64 @@ AS
...
+        ELSE IF @dataType = 2
+            DELETE FROM dbo.CustomTagDouble WHERE TagKey = @tagKey
+        ELSE IF @dataType = 3
+            DELETE FROM dbo.CustomTagDateTime WHERE TagKey = @tagKey
+        ELSE
```
wsugarman, line 149: Perhaps we should explicitly handle unknown data types by throwing an exception?

src/Microsoft.Health.Dicom.SqlServer/Registration/DicomServerBuilderSqlServerRegistrationExtensions.cs
```diff
@@ -10,6 +10,7 @@
 using Microsoft.Health.Dicom.Core.Features.Store;
 using Microsoft.Health.Dicom.Core.Registration;
 using Microsoft.Health.Dicom.SqlServer.Features.ChangeFeed;
+using Microsoft.Health.Dicom.SqlServer.Features.CustomTag;
```
wsugarman, line 13: Is this namespace used?

wsugarman (changes requested):

src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/CustomTagLimit.cs
pengchen0692: it will be used by STOW and QIDO

src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/SqlCustomTagStore.cs
pengchen0692: Following FHIR

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
pengchen0692: I'm not sure whether we should handle it.
Adding exception could handle some unexpected situation, like we added a new datatype not the SQL is not updated yet. 
While in another side, this should not happen unless a bug. this is storage procedure is only used internally and we have the control on input to avoid such thing happen.

src/Microsoft.Health.Dicom.SqlServer/Registration/DicomServerBuilderSqlServerRegistrationExtensions.cs
pengchen0692: yes. previous code has a bug, to fix that, one class in this file is put in this namesapce.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
```diff
@@ -29,22 +23,143 @@ CREATE UNIQUE NONCLUSTERED INDEX IX_CustomTag_TagPath ON dbo.CustomTag
...
+	Note: The primary key is designed on the assumption that tags only occur once in an instance.
+**************************************************************/
+CREATE TABLE dbo.CustomTagPersonName (
+    TagKey                  BIGINT               NOT NULL, --FK
+    TagValue                NVARCHAR(200)        COLLATE SQL_Latin1_General_CP1_CI_AI NOT NULL,
+    StudyKey                BIGINT               NOT NULL, --FK
+    SeriesKey               BIGINT               NULL,     --FK
+    InstanceKey             BIGINT               NULL,     --FK
+    Watermark               BIGINT               NOT NULL,
+	WatermarkAndTagKey      AS CONCAT(TagKey, '.', Watermark), --PK
```
wsugarman, line 141: nit: this formatting is off slightly

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
wsugarman: In SQL perhaps this is fine to leave. If it is truly an impossible situation, I would add a `Debug.Assert` in C# to convey the precondition to other developers

src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/SqlCustomTagStore.cs
```diff
@@ -87,9 +88,36 @@ public Task<IEnumerable<CustomTagEntry>> GetAllCustomTagsAsync(CancellationToken
...
+                }
+                catch (SqlException ex)
+                {
+                    switch (ex.Number)
+                    {
+                        case SqlErrorCodes.NotFound:
+                            throw new CustomTagNotFoundException();
+                        case SqlErrorCodes.PreconditionFailed:
+                            throw new PreconditionFailedException(
+                                string.Format(CultureInfo.InvariantCulture, DicomSqlServerResource.InvalidCustomTagStatus, tagPath));
```
wsugarman, line 115: Given that the deletions are not jobs yet, the status is always `1`, right? Although it's good to handle this case before the implementation is in place!

However, I just figured out that we're sort of piggy-backing off of HTTP status codes, right? It's like the mandatory 50,000 range + error code. So, I would probably not make an association to the precondition failed error code, only because that is pretty specific to the HTTP headers, as per [Mozilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/412). This error is really indicating that you can't delete this custom tag because it is being processed (presumably it is mid-index or mid-delete right?). 

So I would probably say either come up with a tag-specific error like `CustomTagBusyException` or something more generic that's built-in like `InvalidOperationException` to convey this

src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/CustomTagLimit.cs
```diff
@@ -0,0 +1,41 @@
...
+// -------------------------------------------------------------------------------------------------
+using System.Collections.Generic;
+using Dicom;
+
+namespace Microsoft.Health.Dicom.SqlServer.Features.CustomTag
+{
+    /// <summary>
+    /// Limits on CustomTag feature.
+    /// </summary>
+    internal static class CustomTagLimit
```
wsugarman, line 13: I don't really know about this class name. Perhaps something like `CustomTagCategories` with a method like `GetDataType`?

src/Microsoft.Health.Dicom.SqlServer/DicomSqlServerResource.resx
```diff
@@ -117,6 +117,10 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="InvalidCustomTagStatus" xml:space="preserve">
+    <value>The custom tag '{0}' is not in valid status.</value>
```
wsugarman, line 121: Perhaps something a bit more specific. As a user I'd say, "The heck does that mean?" lol

wsugarman (approved):

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
pengchen0692: I guess it's problem with github, it looks good on VS
image

src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/CustomTagLimit.cs
pengchen0692: I follow this class https://github.com/microsoft/dicom-server/blob/main/src/Microsoft.Health.Dicom.Core/Features/Query/QueryLimit.cs :)

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
wsugarman: You used a tab there instead of spaces. You can show white space in VS to help see it

src/Microsoft.Health.Dicom.SqlServer/DicomSqlServerResource.resx
pengchen0692: how about "The custom tag '{0}' is being processed, please retry later."

src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/SqlCustomTagStore.cs
pengchen0692: will change to CustomTagBusyException

src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/CustomTagLimit.cs
pengchen0692: will leave as it is now.

src/Microsoft.Health.Dicom.SqlServer/DicomSqlServerResource.resx
pengchen0692: will change to "The custom tag '{0}' is being processed"

PR #567, opened 2021-02-10
Handle custom tags in a QIDO request

snehabandla:
[AB#78177](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/78177)


src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
```diff
@@ -44,7 +54,16 @@ public class QueryService : IQueryService
 
             ValidateRequestIdentifiers(message);
 
-            QueryExpression queryExpression = _queryParser.Parse(message);
+            IEnumerable<CustomTagEntry> customTags = await _customTagStore.GetCustomTagsAsync(null, cancellationToken);
+
+            Dictionary<QueryResource, HashSet<DicomTag>> queryResourceToCustomTagMapping = null;
```
pengchen0692, line 59: nit: not setting queryResourceToCustomTagMapping  to null, instead, returns what from GenerateQueryResourceToCustomTagMapping which is empty dictionary. it could simplify the code little bit.

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
```diff
@@ -63,6 +82,77 @@ public class QueryService : IQueryService
...
+            {
+                DicomTag[] result;
+                DicomTag dicomTag;
+                if (_dicomTagPathParser.TryParse(customTag.Path, out result))
+                {
+                    dicomTag = result[0];
+                    switch (customTag.Level)
+                    {
+                        case CustomTagLevel.Instance:
+                            if (!ret.ContainsKey(QueryResource.AllInstances))
```
pengchen0692, line 99: nit: ret.TryAdd(QueryResource.AllInstances, new HashSet<DicomTag>())?

src/Microsoft.Health.Dicom.SqlServer/Features/Query/SqlQueryGenerator.cs
```diff
@@ -234,6 +234,30 @@ public override void Visit(StringSingleValueMatchCondition stringSingleValueMatc
                 .AppendLine();
         }
 
+        public override void Visit(DoubleSingleValueMatchCondition doubleSingleValueMatchCondition)
+        {
+            var dicomTagSqlEntry = DicomTagSqlEntry.GetDicomTagSqlEntry(doubleSingleValueMatchCondition.DicomTag);
```
pengchen0692, line 239: missing changes on GetDicomTagSqlEntry and GetTableAlias

src/Microsoft.Health.Dicom.SqlServer/Features/Query/SqlQueryGenerator.cs
```diff
@@ -234,6 +234,30 @@ public override void Visit(StringSingleValueMatchCondition stringSingleValueMatc
...
+            var tableAlias = GetTableAlias(dicomTagSqlEntry);
+            _stringBuilder
```
pengchen0692, line 241: does this pattern also work for custom tag? should tagKey be included in query as well?

src/Microsoft.Health.Dicom.Core/Features/Query/IQueryParser.cs
```diff
@@ -3,13 +3,15 @@
...
+using Dicom;
 using Microsoft.Health.Dicom.Core.Features.Query.Model;
 using Microsoft.Health.Dicom.Core.Messages.Query;
 
 namespace Microsoft.Health.Dicom.Core.Features.Query
 {
     public interface IQueryParser
     {
-        QueryExpression Parse(QueryResourceRequest request);
+        QueryExpression Parse(QueryResourceRequest request, IDictionary<QueryResource, HashSet<DicomTag>> customTagMapping = null);
```
pengchen0692, line 15: should customtag status also  be passed through? Or just filter out Reindexing and Deindexing custom tags before calling this?

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.cs
```diff
@@ -180,10 +201,15 @@ private bool TryParseDicomAttributeId(string attributeId, out DicomTag dicomTag)
...
         }
 
-        private static void ValidateIfTagSupported(DicomTag dicomTag, string attributeId, QueryResource resourceType)
+        private static void ValidateIfTagSupported(DicomTag dicomTag, string attributeId, QueryResource resourceType, IDictionary<QueryResource, HashSet<DicomTag>> customTagMapping = null)
         {
             HashSet<DicomTag> supportedQueryTags = QueryLimit.QueryResourceTypeToTagsMapping[resourceType];
 
+            if (customTagMapping != null && customTagMapping.ContainsKey(resourceType))
+            {
+                supportedQueryTags.UnionWith(customTagMapping[resourceType]);
```
pengchen0692, line 210: nit: another way to write this is:
if (customTagMapping != null)
 {
                supportedQueryTags.UnionWith(customTagMapping.GetValueOrDefault(resourceType,new HashSet());
}

just personal preference to save some word. :)

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -364,6 +364,173 @@ CREATE NONCLUSTERED INDEX IX_ChangeFeed_StudyInstanceUid_SeriesInstanceUid_SopIn
...
+    TagLevel                TINYINT              NOT NULL,
+    TagStatus               TINYINT              NOT NULL
+)
+
+CREATE UNIQUE CLUSTERED INDEX IXC_CustomTag ON dbo.CustomTag
+(
+    TagKey
+)
+
+CREATE NONCLUSTERED INDEX IX_CustomTag_TagPath ON dbo.CustomTag
```
smithago, line 387: unique

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -364,6 +364,173 @@ CREATE NONCLUSTERED INDEX IX_ChangeFeed_StudyInstanceUid_SeriesInstanceUid_SopIn
...
+/*************************************************************
+    Custom Tag Data Table for VR Types mapping to String
+**************************************************************/
+CREATE TABLE dbo.CustomTagString (
+    TagKey                  BIGINT               NOT NULL, --PK
+    TagValue                NVARCHAR(64)         NOT NULL,
+    StudyKey                BIGINT               NOT NULL, --FK
+    SeriesKey               BIGINT               NULL,     --FK
+    InstanceKey             BIGINT               NULL,     --FK
+    Watermark               BIGINT               NOT NULL
```
smithago, line 401: nit: can we add comment on how watermark will be used

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -364,6 +364,173 @@ CREATE NONCLUSTERED INDEX IX_ChangeFeed_StudyInstanceUid_SeriesInstanceUid_SopIn
...
+
+CREATE UNIQUE CLUSTERED INDEX IXC_CustomTagString ON dbo.CustomTagString
+(
+    TagKey,
+    StudyKey,
+    SeriesKey,
+    InstanceKey
+)
+
+CREATE NONCLUSTERED INDEX IXC_CustomTagString_TagKey_TagValue ON dbo.CustomTagString
```
smithago, line 412: can you share examples on how this index will be used?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -364,6 +364,173 @@ CREATE NONCLUSTERED INDEX IX_ChangeFeed_StudyInstanceUid_SeriesInstanceUid_SopIn
...
+    InstanceKey
+)
+
+CREATE NONCLUSTERED INDEX IXC_CustomTagBigInt_TagKey_TagValue ON dbo.CustomTagBigInt
+(
+    TagKey,
+    TagValue
+)
+
+/*************************************************************
```
smithago, line 444: 2.diff.sql should have all the incremental changes since 1.sql

src/Microsoft.Health.Dicom.Core/Exceptions/MultipleCustomTagsFoundException.cs
```diff
@@ -0,0 +1,18 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Dicom.Core.Exceptions
+{
+    /// <summary>
+    /// Exception thrown when two custom tags are found for a given tag path.
+    /// </summary>
+    public class MultipleCustomTagsFoundException : DicomServerException
```
smithago, line 11: When can this happen? Does'nt this mean we allowed bad data?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
```diff
@@ -16,17 +16,17 @@ public class CustomTagEntry : IEquatable<CustomTagEntry>
         /// Path of this tag. Normally it's composed of groupid and elementid.
         /// E.g: 00100020 is path of patient id.
         /// </summary>
-        public string Path { get; }
+        public string Path { get; set; }
```
smithago, line 19: Should the set be on constructor? so that the model is immutable?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/GetCustomTagsService.cs
```diff
@@ -30,14 +32,28 @@ public async Task<GetCustomTagResponse> GetCustomTagAsync(string tagPath, Cancel
         {
             string internalTagPath = _dicomTagParser.ParseFormattedTagPath(tagPath);
```
smithago, line 33: Is this checking validation of tagPath including null/empty and throwing validation exception?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/GetCustomTagsService.cs
```diff
@@ -30,14 +32,28 @@ public async Task<GetCustomTagResponse> GetCustomTagAsync(string tagPath, Cancel
...
 
-            CustomTagEntry customTag = await _customTagStore.GetCustomTagAsync(internalTagPath, cancellationToken);
+            IEnumerable<CustomTagEntry> customTags = await _customTagStore.GetCustomTagsAsync(internalTagPath, cancellationToken);
 
-            return new GetCustomTagResponse(customTag);
+            if (!customTags.Any())
+            {
+                throw new CustomTagNotFoundException(string.Format(DicomCoreResource.CustomTagNotFound, tagPath));
+            }
+            else if (customTags.Count() > 1)
```
smithago, line 41: just check if this is even possible

src/Microsoft.Health.Dicom.Core/Features/CustomTag/GetCustomTagsService.cs
```diff
@@ -30,14 +32,28 @@ public async Task<GetCustomTagResponse> GetCustomTagAsync(string tagPath, Cancel
...
+
+            return new GetCustomTagResponse(customTags.First());
         }
 
         public async Task<GetAllCustomTagsResponse> GetAllCustomTagsAsync(CancellationToken cancellationToken)
         {
-            IEnumerable<CustomTagEntry> customTags = await _customTagStore.GetAllCustomTagsAsync(cancellationToken);
+            IEnumerable<CustomTagEntry> customTags = await _customTagStore.GetCustomTagsAsync(null, cancellationToken);
+
+            if (!customTags.Any())
```
smithago, line 53: should we just return empty array instead of exception?

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Query/DicomQueryServiceTests.cs
```diff
@@ -24,7 +25,9 @@ public DicomQueryServiceTests()
             _queryService = new QueryService(
                 Substitute.For<IQueryParser>(),
                 Substitute.For<IQueryStore>(),
-                Substitute.For<IMetadataStore>());
+                Substitute.For<IMetadataStore>(),
+                Substitute.For<ICustomTagStore>(),
```
smithago, line 29: We will need more UTs in general for all the new things that are added for QueryParser and how the dynamic SQL will be generated.

So try to separate it out with FF. So that we can turn off the feature if we want to

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -364,6 +364,173 @@ CREATE NONCLUSTERED INDEX IX_ChangeFeed_StudyInstanceUid_SeriesInstanceUid_SopIn
...
+    Custom Tag Data Table for VR Types mapping to DateTime
+**************************************************************/
+CREATE TABLE dbo.CustomTagDateTime (
+    TagKey                  BIGINT               NOT NULL, --PK
+    TagValue                DATETIME2(7)         NOT NULL,
```
smithago, line 475: can you check if you can use the same value for both DA and DT and Time VRs we are planning to support on this PR. Particualary around Range query on DA. 

If it is easy to support everything lets do that, otherwise we should just start with DA

src/Microsoft.Health.Dicom.SqlServer/Registration/DicomServerBuilderSqlServerRegistrationExtensions.cs
```diff
@@ -10,6 +10,7 @@
 using Microsoft.Health.Dicom.Core.Features.Store;
 using Microsoft.Health.Dicom.Core.Registration;
 using Microsoft.Health.Dicom.SqlServer.Features.ChangeFeed;
+using Microsoft.Health.Dicom.SqlServer.Features.CustomTag;
```
wsugarman, line 13: Is this necessary?

src/Microsoft.Health.Dicom.Core/Exceptions/MultipleCustomTagsFoundException.cs
snehabandla: This is part of a bad merge.

---
In reply to: [574701326](https://github.com/microsoft/dicom-server/pull/567#discussion_r574701326) [](ancestors = 574701326)

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
snehabandla: I'll remove support!

---
In reply to: [574773278](https://github.com/microsoft/dicom-server/pull/567#discussion_r574773278) [](ancestors = 574773278)

src/Microsoft.Health.Dicom.Core/Features/Query/QueryLimit.cs
```diff
@@ -57,12 +58,12 @@ public static class QueryLimit
 
         public static bool IsValidRangeQueryTag(DicomTag tag)
         {
-            return tag == DicomTag.StudyDate;
+            return tag.GetDefaultVR().Equals(DicomVR.DA);
```
pengchen0692, line 61: Do we consider private tag here?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
```diff
@@ -12,27 +12,43 @@ namespace Microsoft.Health.Dicom.Core.Features.CustomTag
...
+            Level = storedEntry.Level;
+            Status = storedEntry.Status;
+        }
+
         /// <summary>
         /// Path of this tag. Normally it's composed of groupid and elementid.
         /// E.g: 00100020 is path of patient id.
         /// </summary>
-        public string Path { get; set; }
+        public string Path { get; }
```
pengchen0692, line 35: I got failure when calling add custom tag API without "set;". The reason is that CustomTagEntry is used by Asp.net to deserialize and store input.

src/Microsoft.Health.Dicom.Core/Features/Query/IQueryParser.cs
```diff
@@ -3,13 +3,14 @@
...
+using System.Collections.Generic;
 using Microsoft.Health.Dicom.Core.Features.Query.Model;
 using Microsoft.Health.Dicom.Core.Messages.Query;
 
 namespace Microsoft.Health.Dicom.Core.Features.Query
 {
     public interface IQueryParser
     {
-        QueryExpression Parse(QueryResourceRequest request);
+        QueryExpression Parse(QueryResourceRequest request, IDictionary<QueryResource, HashSet<CustomTagFilterDetails>> customTagMapping = null);
```
pengchen0692, line 14: nit: IReadOnlyDictionary if no plan to modify.

src/Microsoft.Health.Dicom.Core/Features/Query/Model/FilterConditions/CustomTagFilterDetails.cs
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        public override int GetHashCode()
+        {
+            return HashCode.Combine(Tag.GetHashCode());
+        }
+
+        public override bool Equals(object obj)
+        {
+            CustomTagFilterDetails other = obj as CustomTagFilterDetails;
```
pengchen0692, line 47: nit: call   public bool Equals(CustomTagFilterDetails other) instead?

`return Equals(obj as CustomTagFilterDetais)`

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.cs
```diff
@@ -164,6 +195,8 @@ private bool ParseFilterCondition(KeyValuePair<string, StringValues> queryParame
                 condition = valueParser(dicomTag, trimmedValue);
```
pengchen0692, line 195: should 'var tagTypeCode = dicomTag.DictionaryEntry.ValueRepresentations.FirstOrDefault()?.Code;' be updated for cusomtag? Since it should be retrieved from database instead.

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
```diff
@@ -63,6 +76,53 @@ public class QueryService : IQueryService
...
+                        ret.TryAdd(QueryResource.StudySeriesInstances, new HashSet<CustomTagFilterDetails>());
+
+                        ret[QueryResource.StudySeriesInstances].Add(new CustomTagFilterDetails(customTag.Key, customTag.Level, dicomTag));
+                    }
+
+                    if (customTag.Level.Equals(CustomTagLevel.Instance) || customTag.Level.Equals(CustomTagLevel.Series))
+                    {
+                        ret.TryAdd(QueryResource.AllSeries, new HashSet<CustomTagFilterDetails>());
+
+                        ret[QueryResource.AllSeries].Add(new CustomTagFilterDetails(customTag.Key, customTag.Level, dicomTag));
```
pengchen0692, line 110: AllSeries should not contains customtag on instance level. 
https://github.com/microsoft/dicom-server/blob/fca9af145d1b183e18543d5e795635cb1036a38c/src/Microsoft.Health.Dicom.Core/Features/Query/QueryLimit.cs#L42
As defined above, AllSeriesTags is composed of AllStudiesTags and StudySeriesTags.

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
```diff
@@ -63,6 +76,53 @@ public class QueryService : IQueryService
...
+
+                        ret.TryAdd(QueryResource.StudySeries, new HashSet<CustomTagFilterDetails>());
+
+                        ret[QueryResource.StudySeries].Add(new CustomTagFilterDetails(customTag.Key, customTag.Level, dicomTag));
+                    }
+
+                    ret.TryAdd(QueryResource.AllStudies, new HashSet<CustomTagFilterDetails>());
+
+                    ret[QueryResource.AllStudies].Add(new CustomTagFilterDetails(customTag.Key, customTag.Level, dicomTag));
```
pengchen0692, line 119: Study level should also be added to AllInstancesTags and AllSeriesTags.

src/Microsoft.Health.Dicom.SqlServer/Features/Query/SqlQueryGenerator.cs
```diff
@@ -121,6 +132,85 @@ private void AppendFilterTable(string filterAlias)
             _stringBuilder.AppendLine($") {filterAlias}");
         }
 
+        private void AppendCustomTagTables()
+        {
+            Dictionary<CustomTagDataType, CustomTagLevel> queriedCustomTagDataTypesPerLevel = new Dictionary<CustomTagDataType, CustomTagLevel>();
+            foreach (CustomTagFilterDetails filterDetails in _queryExpression.QueriedCustomTagFilterDetails)
+            {
+                CustomTagDataType dataType;
+                CustomTagLimit.CustomTagVRAndDataTypeMapping.TryGetValue(filterDetails.Tag.GetDefaultVR().Code, out dataType);
```
pengchen0692, line 141: GetDefaultVR cannot be used on private tag.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -1123,7 +1123,8 @@ BEGIN
     SET NOCOUNT     ON
     SET XACT_ABORT  ON
 
-    SELECT  TagPath,
+    SELECT  TagKey,
```
smithago, line 1126: 2.diff.sql is not updated?

src/Microsoft.Health.Dicom.SqlServer/Features/Query/SqlQueryGenerator.cs
```diff
@@ -121,6 +132,85 @@ private void AppendFilterTable(string filterAlias)
             _stringBuilder.AppendLine($") {filterAlias}");
         }
 
+        private void AppendCustomTagTables()
+        {
+            Dictionary<CustomTagDataType, CustomTagLevel> queriedCustomTagDataTypesPerLevel = new Dictionary<CustomTagDataType, CustomTagLevel>();
```
smithago, line 137: Where is this data used? I don't see it being used in the SQL generation

src/Microsoft.Health.Dicom.SqlServer/Features/Query/SqlQueryGenerator.cs
```diff
@@ -121,6 +132,85 @@ private void AppendFilterTable(string filterAlias)
...
+
+            foreach (CustomTagFilterDetails filterDetails in _queryExpression.QueriedCustomTagFilterDetails)
+            {
+                CustomTagDataType dataType;
+                CustomTagLimit.CustomTagVRAndDataTypeMapping.TryGetValue(filterDetails.Tag.GetDefaultVR().Code, out dataType);
+                string customTagTableAlias = null;
+                _stringBuilder.Append("INNER JOIN ");
+                switch (dataType)
+                {
+                    case CustomTagDataType.StringData:
```
smithago, line 156: Does this work for all level queries.

src/Microsoft.Health.Dicom.SqlServer/Features/Query/SqlQueryGenerator.cs
```diff
@@ -121,6 +132,85 @@ private void AppendFilterTable(string filterAlias)
...
+                        customTagTableAlias = CustomTagStringTableAlias + filterDetails.Key;
+                        _stringBuilder.AppendLine($"{VLatest.CustomTagString.TableName} {customTagTableAlias}");
```
smithago, line 158: How will this work for multiple tags on the same table? Will you not reuse the same alias which will fail the SQL syntax

src/Microsoft.Health.Dicom.SqlServer/Features/Query/SqlQueryGenerator.cs
```diff
@@ -239,11 +371,17 @@ public override void Visit(DateRangeValueMatchCondition rangeValueMatchCondition
             var dicomTagSqlEntry = DicomTagSqlEntry.GetDicomTagSqlEntry(rangeValueMatchCondition.DicomTag);
             var tableAlias = GetTableAlias(dicomTagSqlEntry);
             _stringBuilder
-                .Append("AND ")
+                .Append("AND ");
+
+            tableAlias = AppendCustomTagKeyFilter(dicomTagSqlEntry, tableAlias, rangeValueMatchCondition) ?? tableAlias;
```
smithago, line 376: This code is very confusing to me. You already have the right tableAlias from GetTableAlias()

Why would you not just call AppendCustomTagKeyFilter() without expecting any return

src/Microsoft.Health.Dicom.SqlServer/Features/Query/SqlQueryGenerator.cs
```diff
@@ -239,11 +371,17 @@ public override void Visit(DateRangeValueMatchCondition rangeValueMatchCondition
...
+
+            _stringBuilder
                 .Append(dicomTagSqlEntry.SqlColumn, tableAlias).Append(" BETWEEN ")
                 .Append(_parameters.AddParameter(dicomTagSqlEntry.SqlColumn, rangeValueMatchCondition.Minimum.ToString(SqlDateFormat)))
                 .Append(" AND ")
-                .Append(_parameters.AddParameter(dicomTagSqlEntry.SqlColumn, rangeValueMatchCondition.Maximum.ToString(SqlDateFormat)))
+                .Append(_parameters.AddParameter(dicomTagSqlEntry.SqlColumn, rangeValueMatchCondition.Maximum.ToString(SqlDateFormat)));
+
+            _stringBuilder
```
smithago, line 384: nit: add it to previous _stringBuilder.

src/Microsoft.Health.Dicom.SqlServer/Features/Query/SqlQueryGenerator.cs
```diff
@@ -110,6 +119,8 @@ private void AppendFilterTable(string filterAlias)
                 AppendStatusClause(InstanceTableAlias);
             }
 
+            AppendCustomTagTables();
```
smithago, line 122: will you be adding unit tests to these new changes

src/Microsoft.Health.Dicom.SqlServer/Features/Query/DicomTagSqlEntry.cs
```diff
@@ -26,12 +28,37 @@ internal class DicomTagSqlEntry
                 { DicomTag.SOPInstanceUID, new DicomTagSqlEntry(DicomTag.SOPInstanceUID, SqlTableType.InstanceTable, VLatest.Instance.SopInstanceUid) },
         };
 
-        private DicomTagSqlEntry(DicomTag dicomTag, SqlTableType sqlTableType, Column sqlColumn, string fullTextIndexColumnName = null)
+        private static Dictionary<DicomVR, DicomTagSqlEntry> _customTagToSqlMappingCache = new Dictionary<DicomVR, DicomTagSqlEntry>()
+        {
+                { DicomVR.DA, new DicomTagSqlEntry(null, SqlTableType.CustomTagDateTimeTable, VLatest.CustomTagDateTime.TagValue, null, VLatest.CustomTagDateTime.TagKey) },
```
smithago, line 33: I see you are passing null to DicomTag. Do you want to change the constructor to not include dicomTag and remove that property. It is not being used. in the queryGenerator

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Query/QueryParserTests.cs
```diff
@@ -126,7 +127,71 @@ public void GivenFilterCondition_WithKnownTagButNotSupportedAtLevel_Throws(strin
...
+        {
+            var queryString = "FloatingPointValue=1.1";
+            var filterDetails = new CustomTagFilterDetails(1, CustomTagLevel.Series, DicomTag.FloatingPointValue);
+            EnsureArg.IsNotNull(queryString, nameof(queryString));
+            Dictionary<QueryResource, HashSet<CustomTagFilterDetails>> filterDetailsByResource = new Dictionary<QueryResource, HashSet<CustomTagFilterDetails>>();
+            filterDetailsByResource.Add(QueryResource.AllSeries, new HashSet<CustomTagFilterDetails>() { filterDetails });
+            QueryExpression queryExpression = _queryParser.Parse(CreateRequest(GetQueryCollection(queryString), QueryResource.AllSeries), filterDetailsByResource);
+            Assert.Contains(filterDetails, queryExpression.QueriedCustomTagFilterDetails);
+            Assert.Equal(filterDetails, queryExpression.FilterConditions.First().CustomTagFilterDetails);
         }
```
smithago, line 195: what about tests around invalid cases? Wrong level. wrong tag, wrong value

src/Microsoft.Health.Dicom.Core/Features/Query/Model/FilterConditions/CustomTagFilterDetails.cs
```diff
@@ -0,0 +1,66 @@
...
+using System;
+using Dicom;
+using Microsoft.Health.Dicom.Core.Features.CustomTag;
+
+namespace Microsoft.Health.Dicom.Core.Features.Query
+{
+    /// <summary>
+    /// Details required to create a custom tag filter condition
+    /// </summary>
+    public class CustomTagFilterDetails : IEquatable<CustomTagFilterDetails>
```
smithago, line 15: Can we reuse CustomTagStoreEntry here? I see the only new property here is DicomTag which can be derived

src/Microsoft.Health.Dicom.Core/Features/Query/Model/FilterConditions/QueryFilterCondition.cs
```diff
@@ -15,6 +15,8 @@ public QueryFilterCondition(DicomTag tag)
 
         public DicomTag DicomTag { get; }
 
+        public CustomTagFilterDetails CustomTagFilterDetails { get; set; }
```
smithago, line 18: nullable?

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.cs
```diff
@@ -55,12 +55,31 @@ public QueryParser(IDicomTagParser dicomTagPathParser, ILogger<QueryParser> logg
...
+            _valueParsers.Add(DicomVRCode.FL, ParseDoubleTagValue);
+            _valueParsers.Add(DicomVRCode.FD, ParseDoubleTagValue);
         }
 
-        public QueryExpression Parse(QueryResourceRequest request)
+        public QueryExpression Parse(QueryResourceRequest request, IDictionary<QueryResource, HashSet<CustomTagFilterDetails>> customTagMapping)
         {
             EnsureArg.IsNotNull(request, nameof(request));
 
+            HashSet<CustomTagFilterDetails> queriedCustomTags = new HashSet<CustomTagFilterDetails>();
```
smithago, line 81: nit: var

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.cs
```diff
@@ -180,12 +213,18 @@ private bool TryParseDicomAttributeId(string attributeId, out DicomTag dicomTag)
             return succeed;
         }
 
-        private static void ValidateIfTagSupported(DicomTag dicomTag, string attributeId, QueryResource resourceType)
+        private static void ValidateIfTagSupported(DicomTag dicomTag, string attributeId, QueryResource resourceType, out CustomTagFilterDetails customTagFilterDetails, IDictionary<QueryResource, HashSet<CustomTagFilterDetails>> customTagMapping = null)
```
smithago, line 216: out should be the last param for readability

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
```diff
@@ -63,6 +76,53 @@ public class QueryService : IQueryService
             return new QueryResourceResponse(responseMetadata);
         }
 
+        private Dictionary<QueryResource, HashSet<CustomTagFilterDetails>> GenerateQueryResourceToCustomTagMapping(IEnumerable<CustomTagStoreEntry> customTags)
```
smithago, line 79: Can we move this to CustomTagService. This is very specific to custom tag, it does not belong here.

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
```diff
@@ -44,7 +53,11 @@ public class QueryService : IQueryService
 
             ValidateRequestIdentifiers(message);
 
-            QueryExpression queryExpression = _queryParser.Parse(message);
+            IEnumerable<CustomTagStoreEntry> customTags = await _customTagStore.GetCustomTagsAsync(null, cancellationToken);
```
smithago, line 56: Can you log how long this takes. We may won't to cache it based on that.

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.cs
```diff
@@ -180,12 +213,18 @@ private bool TryParseDicomAttributeId(string attributeId, out DicomTag dicomTag)
...
         {
+            customTagFilterDetails = null;
             HashSet<DicomTag> supportedQueryTags = QueryLimit.QueryResourceTypeToTagsMapping[resourceType];
 
             if (!supportedQueryTags.Contains(dicomTag))
             {
+                if (customTagMapping != null && customTagMapping.ContainsKey(resourceType) && customTagMapping[resourceType].TryGetValue(new CustomTagFilterDetails(dicomTag), out customTagFilterDetails))
```
smithago, line 223: can you simply this? Maybe service only gets tags at a sepcific resourceType. Because a query will be at only one level.

Also, HashSet<DicomTag> supportedQueryTags = QueryLimit.QueryResourceTypeToTagsMapping[resourceType];
can we replaced by a function to get the list of both static and dynamic custom tahs

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
smithago: Also need unit tests to make sure we are testing these calculations

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
```diff
@@ -44,7 +53,11 @@ public class QueryService : IQueryService
...
+
+            Dictionary<QueryResource, HashSet<CustomTagFilterDetails>> queryResourceToCustomTagMapping = GenerateQueryResourceToCustomTagMapping(customTags);
+
+            QueryExpression queryExpression = _queryParser.Parse(message, queryResourceToCustomTagMapping);
```
smithago, line 61: Are the custom tags in the "include" or response tested? Can you make sure it works and add unit tests if not already.

src/Microsoft.Health.Dicom.Core/Features/Query/Model/FilterConditions/CustomTagFilterDetails.cs
snehabandla: VR and Status are also not used!

---
In reply to: [582547138](https://github.com/microsoft/dicom-server/pull/567#discussion_r582547138) [](ancestors = 582547138)

src/Microsoft.Health.Dicom.Core/Features/Query/Model/FilterConditions/QueryFilterCondition.cs
snehabandla: This implicitly allows it to be null I think.

---
In reply to: [582547364](https://github.com/microsoft/dicom-server/pull/567#discussion_r582547364) [](ancestors = 582547364)

src/Microsoft.Health.Dicom.SqlServer/Features/Query/SqlQueryGenerator.cs
snehabandla: Yes! I've added in tests to validate it as well.

---
In reply to: [582212608](https://github.com/microsoft/dicom-server/pull/567#discussion_r582212608) [](ancestors = 582212608)

src/Microsoft.Health.Dicom.SqlServer/Features/Query/SqlQueryGenerator.cs
snehabandla: The custom tag key is appended to the table alias so that we can join against the same custom tag table multiple times!

---
In reply to: [582517527](https://github.com/microsoft/dicom-server/pull/567#discussion_r582517527) [](ancestors = 582517527)

src/Microsoft.Health.Dicom.SqlServer/Features/Query/SqlQueryGenerator.cs
snehabandla: This is old code, that I removed.

---
In reply to: [582212411](https://github.com/microsoft/dicom-server/pull/567#discussion_r582212411) [](ancestors = 582212411)

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
snehabandla: It's awkward to move since it describes things by QueryResource...There is also currently no singular CustomTagService! It's broken out by action. I can create a separate extension class if you still fell like it's too specific!

---
In reply to: [582550008](https://github.com/microsoft/dicom-server/pull/567#discussion_r582550008) [](ancestors = 582550008)

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
snehabandla: It's tested and there are already unit tests covering it!

---
In reply to: [583002754](https://github.com/microsoft/dicom-server/pull/567#discussion_r583002754) [](ancestors = 583002754)

src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/SqlCustomTagStore.cs
```diff
@@ -73,36 +75,41 @@ public async Task AddCustomTagsAsync(IEnumerable<CustomTagEntry> customTagEntrie
...
+                        results.Add(new CustomTagStoreEntry(tagKey, tagPath, tagVR, (CustomTagLevel)tagLevel, (CustomTagStatus)tagStatus));
                     }
+
+                    executionTimeWatch.Stop();
+                    _logger.LogInformation(executionTimeWatch.ElapsedMilliseconds.ToString());
                 }
             }
 
-            return results;
+            return results.AsReadOnly();
```
wsugarman, line 112: This isn't necessary unless you want to make doubly-sure that the caller isn't going to cast the result to a `List<T>` and mutate it. Instead, you can simply return `results`

src/Microsoft.Health.Dicom.SqlServer/Features/Query/SqlQueryGenerator.cs
```diff
@@ -121,6 +131,74 @@ private void AppendFilterTable(string filterAlias)
             _stringBuilder.AppendLine($") {filterAlias}");
         }
 
+        private void AppendCustomTagTables()
+        {
+            foreach (CustomTagFilterDetails filterDetails in _queryExpression.QueriedCustomTagFilterDetails)
+            {
+                CustomTagDataType dataType;
+                CustomTagLimit.CustomTagVRAndDataTypeMapping.TryGetValue(filterDetails.Tag.GetDefaultVR().Code, out dataType);
```
wsugarman, line 139: If you know this will never return `false`, use the indexer instead

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.cs
snehabandla: The optional parameters have to be added at the end of the parameter list!

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
```diff
@@ -12,6 +12,22 @@ namespace Microsoft.Health.Dicom.Core.Features.CustomTag
...
     {
+        public CustomTagEntry(string path, string vR, CustomTagLevel level, CustomTagStatus status)
+        {
+            Path = path;
+            VR = vR;
+            Level = level;
+            Status = status;
+        }
+
+        public CustomTagEntry(CustomTagStoreEntry storedEntry)
```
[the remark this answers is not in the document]
snehabandla: fixed this!

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
```diff
@@ -12,6 +12,22 @@ namespace Microsoft.Health.Dicom.Core.Features.CustomTag
...
+        {
+            Path = storedEntry.Path;
+            VR = storedEntry.VR;
+            Level = storedEntry.Level;
+            Status = storedEntry.Status;
+        }
+
         /// <summary>
         /// Path of this tag. Normally it's composed of groupid and elementid.
         /// E.g: 00100020 is path of patient id.
```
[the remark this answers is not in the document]
snehabandla: The setters are for serialization I believe!

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.cs
```diff
@@ -120,7 +146,7 @@ private static void PostProcessFilterConditions(QueryExpressionImp parsedQuery)
             if (parsedQuery.FuzzyMatch == true)
             {
                 IEnumerable<QueryFilterCondition> potentialFuzzyConds = parsedQuery.FilterConditions
-                                                                            .Where(c => QueryLimit.IsValidFuzzyMatchingQueryTag(c.DicomTag)).ToList();
+                                                                            .Where(c => QueryLimit.IsValidFuzzyMatchingQueryTag(c.DicomTag, c.CustomTagFilterDetails?.VR)).ToList();
```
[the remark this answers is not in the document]
snehabandla: Unfortunately the foreach modifies values so it needs to be a list!

---
In reply to: [586590074](https://github.com/microsoft/dicom-server/pull/567#discussion_r586590074) [](ancestors = 586590074)

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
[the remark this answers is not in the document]
wsugarman: Ah, if we're using some sort of auto-generated serialization where we are deserializing these objects, then we can make the setters `private` at least

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.cs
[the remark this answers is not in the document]
wsugarman: Ah, I see in the loop. I think we can refactor this then a bit to perform the replacements by enumerating over indices. Luckily we're only replacing elements instead of changing the count:

```csharp
IEnumerable<(StringSingleValueMatchCondition, int)> potentialFuzzyConds = parsedQuery.FilterConditions
    .Where(c => QueryLimit.IsValidFuzzyMatchingQueryTag(c.DicomTag, c.CustomTagFilterDetails?.VR))
    .Cast<StringSingleValueMatchCondition>()
    .Select((x, i) => (x, i));

foreach (var (cond, i) in potentialFuzzyConds)
{
    parsedQuery.FilterConditions[i] = new PersonNameFuzzyMatchCondition(cond.DicomTag, cond.Value);
}
```

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.cs
```diff
@@ -144,7 +175,8 @@ private bool ParseFilterCondition(KeyValuePair<string, StringValues> queryParame
                 return false;
             }
 
-            ValidateIfTagSupported(dicomTag, attributeId, resourceType);
+            CustomTagFilterDetails customTagFilterDetails;
+            ValidateIfTagSupported(dicomTag, attributeId, resourceType, out customTagFilterDetails, supportedCustomTags);
 
             // parse tag value
             if (!queryParameter.Value.Any() || queryParameter.Value.Count > 1)
```
wsugarman, line 182: nit: I think you can just say `queryParameter.Value.Count != 1`

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.cs
```diff
@@ -75,7 +94,7 @@ public QueryExpression Parse(QueryResourceRequest request)
                 }
 
                 // filter conditions with attributeId as key
-                if (ParseFilterCondition(queryParam, request.QueryResourceType, out QueryFilterCondition condition))
+                if (ParseFilterCondition(queryParam, request.QueryResourceType, supportedCustomTags?.ToHashSet(), out QueryFilterCondition condition))
```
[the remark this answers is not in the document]
snehabandla: It's actually the input ISet<T>!

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.cs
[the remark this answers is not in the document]
snehabandla: Unfortunately, that won't work either since parsedQuery.FilterConditions is still being modified!

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.cs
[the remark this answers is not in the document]
wsugarman: Can we change it to `ISet<T>` the whole way down?

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.cs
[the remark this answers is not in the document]
wsugarman: Ah! My mistake -- I didn't realize the internal [version was incremented on set too](https://github.com/dotnet/runtime/blob/f7de865465532de372f65b55e00a8ec2be1e5274/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/List.cs#L162). My second one should still work since we avoid the enumerator

src/Microsoft.Health.Dicom.Core/Features/Query/Model/FilterConditions/CustomTagFilterDetails.cs
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        public CustomTagFilterDetails(DicomTag tag)
+        {
+            Tag = tag;
+        }
+
+        public long Key { get; }
+
+        public string VR { get; }
```
smithago, line 32: Can this be typed DicomVR? fo-dicom has a type for it

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
```diff
@@ -63,6 +76,32 @@ public class QueryService : IQueryService
             return new QueryResourceResponse(responseMetadata);
         }
 
+        private HashSet<CustomTagFilterDetails> RetrieveSuupportedCustomTagsForQueryResourceType(IReadOnlyList<CustomTagStoreEntry> customTags, QueryResource queryResource)
```
smithago, line 79: nit: spelling, RetrieveSupported..

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
```diff
@@ -63,6 +76,32 @@ public class QueryService : IQueryService
...
+        {
+            HashSet<CustomTagFilterDetails> ret = new HashSet<CustomTagFilterDetails>();
```
smithago, line 81: nit: var

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
```diff
@@ -63,6 +76,32 @@ public class QueryService : IQueryService
...
+
+            foreach (CustomTagStoreEntry customTag in customTags)
+            {
+                DicomTag[] result;
+                DicomTag dicomTag;
+                if (customTag.Status.Equals(CustomTagStatus.Added) && _dicomTagPathParser.TryParse(customTag.Path, out result))
```
smithago, line 87: We need to move this method CustomTagService. It needs to expose GetActiveTagsAtLevel(QueryResourceLevel)

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
```diff
@@ -44,7 +53,11 @@ public class QueryService : IQueryService
 
             ValidateRequestIdentifiers(message);
 
-            QueryExpression queryExpression = _queryParser.Parse(message);
+            IReadOnlyList<CustomTagStoreEntry> customTags = await _customTagStore.GetCustomTagsAsync(null, cancellationToken);
+
+            HashSet<CustomTagFilterDetails> supportedCustomTags = RetrieveSuupportedCustomTagsForQueryResourceType(customTags, message.QueryResourceType);
```
smithago, line 58: Also instead of getting this and passing it to QueryParser, see if you can get the QueriableTags

smithago: I think the same think we talked with Peng. We already have CustomTagService. Can we create IndexedTagService and make its problem to combine and get the full list and use that service everywhere?

src/Microsoft.Health.Dicom.SqlServer/Features/Query/SqlQueryGenerator.cs
smithago: Okay can you add that logic to GetTableAlias?

src/Microsoft.Health.Dicom.Core/Features/Query/QueryResource.cs
```diff
@@ -7,11 +7,11 @@ namespace Microsoft.Health.Dicom.Core.Features.Query
 {
     public enum QueryResource
```
wsugarman, line 8: nit: as we discussed, I think it's ok to leave the numbers out (as they are the same values if left unspecified), but let's leave a comment about how the order works

wsugarman (approved):

PR #572, opened 2021-02-16
Add Integration Tests for Add and Remove Custom Tag

pengchen0692:
## Description
Add Integration Tests for Add and Remove Custom Tag

## Related issues
Addresses [AB#79467](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/79467)


test/Microsoft.Health.Dicom.Tests.Integration/Features/CustomTagServiceTests.cs
```diff
@@ -0,0 +1,105 @@
...
+using Microsoft.Health.Dicom.Core.Messages.CustomTag;
+using Microsoft.Health.Dicom.Tests.Common.Extensions;
+using Microsoft.Health.Dicom.Tests.Integration.Persistence;
+using Xunit;
+
+namespace Microsoft.Health.Dicom.Tests.Integration.Features
+{
+    public class CustomTagServiceTests : IClassFixture<CustomTagServiceTestsFixture>
+    {
+        private static readonly CancellationToken DefaultCancellationToken = new CancellationTokenSource().Token;
```
wsugarman, line 22: Use the `default` keyword or `CancellationToken.None` throughout the tests since you're not using a `CancellationTokenSource`

test/Microsoft.Health.Dicom.Tests.Integration/Features/CustomTagServiceTests.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        private readonly CustomTagServiceTestsFixture _customTagServiceTestsFixture;
+
+        public CustomTagServiceTests(CustomTagServiceTestsFixture customTagServiceTestsFixture)
+        {
+            EnsureArg.IsNotNull(customTagServiceTestsFixture, nameof(customTagServiceTestsFixture));
+            _customTagServiceTestsFixture = customTagServiceTestsFixture;
```
wsugarman, line 29: I don't think we do this, but `EnsureArg` methods typically return the value I think, so they can be used like this:

```csharp
_customTagServiceTestsFixture = EnsureArg.IsNotNull(customTagServiceTestsFixture, nameof(customTagServiceTestsFixture));
```

Without this package, people typically just do:

```csharp
_customTagServiceTestsFixture = customTagServiceTestsFixture ?? throw new ArgumentNullException(nameof(customTagServiceTestsFixture));
```

test/Microsoft.Health.Dicom.Tests.Integration/Features/CustomTagServiceTests.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        [Fact(Skip = "Feature not ready")]
+        public async Task GivenValidCustomTags_WhenAddCustomTag_ThenTagShouldBeAdded()
+        {
+            DicomTag tag1 = DicomTag.DeviceSerialNumber;
```
wsugarman, line 35: A couple questions:

1. Do we need to handle multiple instances of our integration tests running? Would that cause problems?
2. We are using `finally` blocks to clean up our tests should they fail, but do we need to worry about if the process stops/fails in them middle of a test such that the DBs are not clean?

test/Microsoft.Health.Dicom.Tests.Integration/Persistence/CustomTagServiceTestsFixture.cs
```diff
@@ -0,0 +1,44 @@
...
+            GetCustomTagsService = new GetCustomTagsService(CustomTagStore, new DicomTagParser());
+            DeleteCustomTagService = new DeleteCustomTagService(CustomTagStore, new DicomTagParser(), NullLogger<DeleteCustomTagService>.Instance);
+        }
+
+        public IAddCustomTagService AddCustomTagService { get; private set; }
+
+        public IGetCustomTagsService GetCustomTagsService { get; private set; }
+
+        public IDeleteCustomTagService DeleteCustomTagService { get; private set; }
+
```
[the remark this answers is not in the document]
pengchen0692: The file is removed

test/Microsoft.Health.Dicom.Tests.Integration/Features/CustomTagServiceTests.cs
pengchen0692: 1. are you talking about each fixture having one database instance? 
2. when test fail, the db stays for diagnosis. the db will be clean up when the PR is closed by Github Action https://github.com/microsoft/dicom-server/actions?query=workflow%3ACleanupAzureResources

wsugarman (approved):

PR #574, opened 2021-02-19
Migrate to .NET Analyzers

wsugarman:
## Description
With .NET 5, .NET Analyzers are included as part of the SDK and replace the older FxCop Analyzers. This PR adds these analyzers and addresses the suggestions/warnings/errors brought up by both the analyzers and our own *.editorconfig* file. See [here](https://docs.microsoft.com/en-us/visualstudio/code-quality/migrate-from-fxcop-analyzers-to-net-analyzers?view=vs-2019) for details.

Below is a list of the addressed issues and a brief description:

- CA1002 - Do not use generic list arguments
- CA1008 - Enums should have a default member
- CA1012 - Abstract types should not have public ctors
- CA1027 - Mark enums with flags attributes
- CA1031 - Catch exceptions more specific than `Exception` (ignored)
- CA1062 - Check non-`null` args on `public` method before use
- CA1725 - Match parameter names between interfaces and implementations
- CA1805 - Avoid initializing members to their defaults
- CA1813 - Avoid unsealed attributes
- CA1827 - Use `Any()` instead of `Count()`
- CA1309 - Use ordinal comparison
- CA1508 - Avoid dead conditional code
- CA1822 - Method should be static as it doesn't access instance data
- CA1829 - Use `Array.Length` over `Enumerable.Count()`
- CA1834 - Use `StringBuilder.Append(char)` when possible
- CA2000 - Call dispose or leverage using statement
- CA2012 - `ValueTask<T>` should be immediately used
- CA2016 - Forward cancellation token
- CA5394 - Do not use insecure randomness for security-related purposes
- IDE0007 - Use `var` instead of explicit type
- IDE0017 - Object initialization can be simplified
- IDE0018 - Inline variable declaration
- IDE0017 - Collection initialization can be simplified
- IDE0052 - Private member is never used

## Related issues
N/A

## Testing
Run unit tests

## Versioning
+semver: major


src/Microsoft.Health.Dicom.SqlServer/Features/Schema/SchemaVersion.cs
```diff
@@ -10,6 +10,7 @@ namespace Microsoft.Health.Dicom.SqlServer.Features.Schema
     /// </summary>
     public enum SchemaVersion
     {
+        Unknown = 0,
```
brandonpollett, line 13: What's the rule that wanted this added?

src/Microsoft.Health.Dicom.Api/Features/Filters/DicomRequestContextRouteDataPopulatingFilterAttribute.cs
```diff
@@ -15,7 +15,7 @@
 namespace Microsoft.Health.Dicom.Api.Features.Filters
 {
     [AttributeUsage(AttributeTargets.Class)]
-    public class DicomRequestContextRouteDataPopulatingFilterAttribute : ActionFilterAttribute
+    public sealed class DicomRequestContextRouteDataPopulatingFilterAttribute : ActionFilterAttribute
```
KnicKnic, line 18: breaking change?

wsugarman: Yup, this is breaking along with some other access modifier changes

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.cs
```diff
@@ -169,15 +165,16 @@ private bool ParseFilterCondition(KeyValuePair<string, StringValues> queryParame
...
-            dicomTag = null;
-            DicomTag[] result;
-            bool succeed = _dicomTagPathParser.TryParse(attributeId, out result, supportMultiple: false);
-            if (succeed)
+            if (_dicomTagPathParser.TryParse(attributeId, out DicomTag[] result, supportMultiple: false))
             {
                 dicomTag = result[0];
+                return true;
+            }
+            else
```
brandonpollett, line 173: else isn't needed here

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
```diff
@@ -41,6 +42,8 @@ public bool TryParse(string dicomTagPath, out DicomTag[] dicomTags, bool support
 
         public string ParseFormattedTagPath(string dicomTagPath, bool supportMultiple = false)
         {
+            EnsureArg.IsNotNull(dicomTagPath, nameof(dicomTagPath));
```
brandonpollett, line 45: I know a lot of places we do `IsNotNullOrWhitespace` not sure if that's appropriate here or not.

KnicKnic (approved):
:shipit:

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/SchemaVersion.cs
wsugarman: [CA1008](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1008)

The rule describes a desire to have `default(SchemaVersion)` or `default` to be a defined value of the enum where appropriate.

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
wsugarman: For now, given the volume of changes, I'll do the `null` check alone as if there was a `null` passed to this method today it would already throw an exception (albeit `NullReferenceException`).

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/SchemaVersion.cs
brandonpollett: I would err on the side of not adding the default here unless we know how this interacts with the schema migration/schema API changes.

wsugarman: Reading the code there doesn't appear any usage that would interfere. I think we're ok to add this

PR #585, opened 2021-02-24
Fix bug which block service to start

pengchen0692:
## Description
Fix bug which block service to start

## Related issues
Addresses [AB#78174](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/78174)

## Testing
Locally tested with Add/Remove/Get/List custom tag

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Modules/ServiceModule.cs
+++ b/src/Microsoft.Health.Dicom.Core/Modules/ServiceModule.cs
@@ -144,6 +144,11 @@ namespace Microsoft.Health.Dicom.Core.Modules
                     .AsSelf()
                     .AsImplementedInterfaces();
 
+                services.Add<GetCustomTagsService>()
+                   .Scoped()
+                   .AsSelf()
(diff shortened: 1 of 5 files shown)


src/Microsoft.Health.Dicom.Core/Modules/ServiceModule.cs
```diff
@@ -144,6 +144,11 @@ public void Load(IServiceCollection services)
                     .AsSelf()
                     .AsImplementedInterfaces();
 
+                services.Add<GetCustomTagsService>()
```
wsugarman, line 147: Was it that this service was missing in the DI?

wsugarman (approved):

src/Microsoft.Health.Dicom.Core/Modules/ServiceModule.cs
pengchen0692: yes

PR #588, opened 2021-02-24
TagKey should be BIGINT

pengchen0692:
## Description
@TagKey is a variable to save tagKey, so that it should be as same type as TagKey. this is a typo.

### DIFF
--- a/src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
+++ b/src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
@@ -249,7 +249,7 @@ AS
         
         -- Check if tag exsit
         DECLARE @tagStatus TINYINT
-        DECLARE @tagKey TINYINT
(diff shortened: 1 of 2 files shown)


snehabandla (approved):

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
```diff
@@ -249,7 +249,7 @@ AS
         
         -- Check if tag exsit
         DECLARE @tagStatus TINYINT
-        DECLARE @tagKey TINYINT
+        DECLARE @tagKey BIGINT
```
smithago, line 252: Can you add a little bit more reasoning on why the change in the description. Will we ever support 8bytes, 2power63 different custom tags.

pengchen0692: @tagKey is a variable to save tagKey, so that it should be as same type as TagKey. this is a typo.

PR #590, opened 2021-02-25
Add Security Tasks to PR and CI Pipelines

wsugarman:
## Description
- Add *analyze.yml* which contains a few 1ES security tasks to ensure secure code quality
- Incorporate *analyze.yml* into both PR and CI

## Related issues
N/A

## Testing
PR pipeline should be sufficient as it runs the same tasks as CI


build/package.yml
```diff
@@ -26,11 +26,11 @@ steps:
     inputs:
       command: pack
       configuration: '$(buildConfiguration)'
+      packagesToPack: '**/*.csproj;!test/**/*.csproj;!**/*.UnitTests.csproj'
```
wsugarman, line 29: I remove the packing of test projects, and have removed the web tests and integration tests from the `nuget` artifact. The pipeline instead uses these projects from the `IntegrationTests` artifact, as they are built in a previous task.

build/build.yml
```diff
@@ -17,14 +18,17 @@ steps:
     inputs:
       command: 'build'
       projects: '**/*.csproj'
-      arguments: '--configuration $(buildConfiguration) -p:AssemblyVersion="$(assemblySemVer)" -p:FileVersion="$(assemblySemFileVer)" -p:InformationalVersion="$(informationalVersion)" -p:ContinuousIntegrationBuild=true'
+      arguments: '-c $(buildConfiguration) -p:AssemblyVersion="$(assemblySemVer)" -p:FileVersion="$(assemblySemFileVer)" -p:InformationalVersion="$(informationalVersion)" -p:ContinuousIntegrationBuild=true'
```
brandonpollett, line 21: nit: Generally, I think we should keep the full argument name in scripts like this for discoverability.

brandonpollett (approved):

build/build.yml
wsugarman: That makes sense

PR #597, opened 2021-03-01
Fix bugs in Get/List Custom Tags

pengchen0692:
## Description
Fix multiple bugs in Get/List custom tags

## Related issues
Addresses [AB#80037](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/80037) and [AB#80038](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/80038)

## Testing
unit tests, locally manually tests.

### DIFF
--- a/src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
+++ b/src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
@@ -5,6 +5,7 @@
 
 using System;
 using System.Reflection;
+using System.Text.Json.Serialization;
 using Dicom;
 using Dicom.Serialization;
 using EnsureThat;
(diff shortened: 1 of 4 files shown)


src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
```diff
@@ -71,6 +72,9 @@ public static class DicomServerServiceCollectionExtensions
                 options.EnableEndpointRouting = false;
                 options.RespectBrowserAcceptHeader = true;
                 options.OutputFormatters.Insert(0, new DicomJsonOutputFormatter());
+            }).AddJsonOptions(jsonOptions =>
+            {
+                jsonOptions.JsonSerializerOptions.Converters.Add(new JsonStringEnumConverter());
             });
```
wsugarman, line 78: nit: Given that we're not chaining together calls elsewhere, let's just use:

```csharp
services.AddJsonOptions(jsonOptions => ... )
```

wsugarman (approved):

src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
pengchen0692: Hill, it's chained... AddJsonOptions is a method of IMvcBuilder which returns from previous call.

PR #600, opened 2021-03-02
Update MinimumValidator to support custom tags

pengchen0692:
## Description
Update MinimumValidator to support custom tags

## Related issues
Addresses #AB80094

## Testing
Unit tests

### DIFF
--- a/src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/DicomDatasetValidatorTests.cs
+++ b/src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/DicomDatasetValidatorTests.cs
@@ -9,6 +9,7 @@ using Dicom;
 using Microsoft.Extensions.Options;
 using Microsoft.Health.Dicom.Core.Configs;
 using Microsoft.Health.Dicom.Core.Exceptions;
+using Microsoft.Health.Dicom.Core.Features.CustomTag;
 using Microsoft.Health.Dicom.Core.Features.Store;
 using Microsoft.Health.Dicom.Core.Features.Validation;
 using Microsoft.Health.Dicom.Tests.Common;
@@ -43,7 +44,7 @@ namespace Microsoft.Health.Dicom.Core.UnitTests.Features.Store
         [Fact]
         public void GivenAValidDicomDataset_WhenValidated_ThenItShouldSucceed()
         {
-            _dicomDatasetValidator.Validate(_dicomDataset, requiredStudyInstanceUid: null);
+            _dicomDatasetValidator.Validate(_dicomDataset, new List<CustomTagEntry>(), requiredStudyInstanceUid: null);
         }
 
         [Fact]
@@ -55,6 +56,7 @@ namespace Microsoft.Health.Dicom.Core.UnitTests.Features.Store
 
             _dicomDatasetValidator.Validate(
                 _dicomDataset,
+                new List<CustomTagEntry>(),
                 studyInstanceUid);
         }
 
@@ -166,14 +168,14 @@ namespace Microsoft.Health.Dicom.Core.UnitTests.Features.Store
         public void GivenDatasetWithEmptyIndexedTagValue_WhenValidating_ThenValidationPasses()
         {
             _dicomDataset.AddOrUpdate(DicomTag.ReferringPhysicianName, string.Empty);
-            _dicomDatasetValidator.Validate(_dicomDataset, null);
+            _dicomDatasetValidator.Validate(_dicomDataset, new List<CustomTagEntry>(), null);
         }
 
         private void ExecuteAndValidateException<T>(ushort failureCode, string requiredStudyInstanceUid = null)
             where T : Exception
         {
             var exception = Assert.Throws<T>(
-                () => _dicomDatasetValidator.Validate(_dicomDataset, requiredStudyInstanceUid));
+                () => _dicomDatasetValidator.Validate(_dicomDataset, new List<CustomTagEntry>(), requiredStudyInstanceUid));
 
             if (exception is DatasetValidationException)
             {
--- a/src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/DicomDatasetValidatorTests.cs
+++ b/src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/DicomDatasetValidatorTests.cs
@@ -13,6 +13,7 @@ using Microsoft.Health.Dicom.Core.Features.CustomTag;
 using Microsoft.Health.Dicom.Core.Features.Store;
 using Microsoft.Health.Dicom.Core.Features.Validation;
 using Microsoft.Health.Dicom.Tests.Common;
+using Microsoft.Health.Dicom.Tests.Common.Extensions;
 using NSubstitute;
 using Xunit;
 
@@ -171,11 +172,37 @@ namespace Microsoft.Health.Dicom.Core.UnitTests.Features.Store
             _dicomDatasetValidator.Validate(_dicomDataset, new List<CustomTagEntry>(), null);
         }
 
-        private void ExecuteAndValidateException<T>(ushort failureCode, string requiredStudyInstanceUid = null)
+        [Fact]
+        public void GivenAValidDicomDatasetWithCustomTag_WhenValidated_ThenItShouldSucceed()
+        {
+            DicomTag tag = DicomTag.PatientAge;
+            CustomTagEntry entry = tag.BuildCustomTagEntry();
+            _dicomDataset.Add(tag, "001W");
+            _dicomDatasetValidator.Validate(_dicomDataset, new List<CustomTagEntry>() { entry }, requiredStudyInstanceUid: null);
+        }
+
+        [Fact]
+        public void GivenInvalidDicomDatasetWithCustomTag_WhenValidated_ThenValidationExceptionShouldBeThrown()
+        {
+            DicomTag tag = DicomTag.PatientAge;
+            CustomTagEntry entry = tag.BuildCustomTagEntry();
+#pragma warning disable CS0618 // Type or member is obsolete
+            DicomValidation.AutoValidation = false;
+#pragma warning restore CS0618 // Type or member is obsolete
+
+            _dicomDataset.Add(tag, "001w");
+
+#pragma warning disable CS0618 // Type or member is obsolete
+            DicomValidation.AutoValidation = true;
+#pragma warning restore CS0618 // Type or member is obsolete
+            ExecuteAndValidateException<DicomElementValidationException>(TestConstants.ValidationFailureReasonCode, customTagEntries: new List<CustomTagEntry>() { entry });
+        }
+
+        private void ExecuteAndValidateException<T>(ushort failureCode, string requiredStudyInstanceUid = null, IReadOnlyList<CustomTagEntry> customTagEntries = null)
             where T : Exception
         {
             var exception = Assert.Throws<T>(
-                () => _dicomDatasetValidator.Validate(_dicomDataset, new List<CustomTagEntry>(), requiredStudyInstanceUid));
+                () => _dicomDatasetValidator.Validate(_dicomDataset, customTagEntries ?? new List<CustomTagEntry>(), requiredStudyInstanceUid));
 
             if (exception is DatasetValidationException)
             {
--- a/src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/DicomDatasetValidatorTests.cs
+++ b/src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/DicomDatasetValidatorTests.cs
@@ -45,7 +45,7 @@ namespace Microsoft.Health.Dicom.Core.UnitTests.Features.Store
         [Fact]
         public void GivenAValidDicomDataset_WhenValidated_ThenItShouldSucceed()
         {
-            _dicomDatasetValidator.Validate(_dicomDataset, new List<CustomTagEntry>(), requiredStudyInstanceUid: null);
+            _dicomDatasetValidator.Validate(_dicomDataset, requiredStudyInstanceUid: null);
         }
 
         [Fact]
@@ -57,7 +57,6 @@ namespace Microsoft.Health.Dicom.Core.UnitTests.Features.Store
 
             _dicomDatasetValidator.Validate(
                 _dicomDataset,
-                new List<CustomTagEntry>(),
                 studyInstanceUid);
         }
 
@@ -169,7 +168,7 @@ namespace Microsoft.Health.Dicom.Core.UnitTests.Features.Store
         public void GivenDatasetWithEmptyIndexedTagValue_WhenValidating_ThenValidationPasses()
         {
             _dicomDataset.AddOrUpdate(DicomTag.ReferringPhysicianName, string.Empty);
-            _dicomDatasetValidator.Validate(_dicomDataset, new List<CustomTagEntry>(), null);
+            _dicomDatasetValidator.Validate(_dicomDataset, null);
         }
 
         [Fact]
@@ -198,7 +197,7 @@ namespace Microsoft.Health.Dicom.Core.UnitTests.Features.Store
             ExecuteAndValidateException<DicomElementValidationException>(TestConstants.ValidationFailureReasonCode, customTagEntries: new List<CustomTagEntry>() { entry });
         }
 
-        private void ExecuteAndValidateException<T>(ushort failureCode, string requiredStudyInstanceUid = null, IReadOnlyList<CustomTagEntry> customTagEntries = null)
+        private void ExecuteAndValidateException<T>(ushort failureCode, string requiredStudyInstanceUid = null, IList<CustomTagEntry> customTagEntries = null)
             where T : Exception
         {
             var exception = Assert.Throws<T>(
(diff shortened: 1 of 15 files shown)


src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/DicomStoreServiceTests.cs
```diff
@@ -50,7 +53,10 @@ public DicomStoreServiceTests()
                 _storeResponseBuilder,
                 _dicomDatasetValidator,
                 _storeOrchestrator,
+                _customTagStore,
                 NullLogger<StoreService>.Instance);
+            _customTagStore.GetCustomTagsAsync(Arg.Any<string>(), Arg.Any<CancellationToken>())
```
pengchen0692, line 58: formatting

src/Microsoft.Health.Dicom.Core/Features/Store/DicomDatasetValidator.cs
```diff
@@ -90,22 +94,44 @@ string EnsureRequiredTagIsPresent(DicomTag dicomTag)
...
             {
-                ValidateIndexedItems(dicomDataset);
+                ValidateIndexedItems(dicomDataset, customTagEntries);
             }
         }
 
-        private void ValidateIndexedItems(DicomDataset dicomDataset)
+        private void ValidateIndexedItems(DicomDataset dicomDataset, IReadOnlyList<CustomTagEntry> customTags)
         {
+            // TODO: imporve performance
```
pengchen0692, line 103: refine the code

src/Microsoft.Health.Dicom.Core/Features/Store/DicomDatasetValidator.cs
```diff
@@ -90,22 +94,44 @@ string EnsureRequiredTagIsPresent(DicomTag dicomTag)
...
+            // Process Standard Tag
             HashSet<DicomTag> indexableTags = QueryLimit.AllInstancesTags;
 
+            indexableTags.UnionWith(customTags.Select(x => DicomTag.Parse(x.Path)).Where(y => !y.IsPrivate));
```
pengchen0692, line 107: add unit tests

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMinimumValidation.cs
```diff
@@ -17,8 +19,11 @@ public static class DicomElementMinimumValidation
         private static readonly Regex ValidIdentifierCharactersFormat = new Regex("^[0-9\\.]*$", RegexOptions.Compiled);
         private const string DateFormat = "yyyyMMdd";
 
-        public static void ValidateCS(string value, string name)
+        public static void ValidateCS(DicomElement element)
         {
+            EnsureArg.IsNotNull(element, nameof(element));
+            string name = GetName(element.Tag);
```
pengchen0692, line 25: GetName move to throw exeption

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/DicomDatasetValidatorTests.cs
```diff
@@ -166,14 +169,40 @@ public void GivenDatasetWithInvalidIndexedTagValue_WhenValidating_ThenValidation
         public void GivenDatasetWithEmptyIndexedTagValue_WhenValidating_ThenValidationPasses()
         {
             _dicomDataset.AddOrUpdate(DicomTag.ReferringPhysicianName, string.Empty);
-            _dicomDatasetValidator.Validate(_dicomDataset, null);
+            _dicomDatasetValidator.Validate(_dicomDataset, new List<CustomTagEntry>(), null);
         }
 
-        private void ExecuteAndValidateException<T>(ushort failureCode, string requiredStudyInstanceUid = null)
+        [Fact]
+        public void GivenAValidDicomDatasetWithCustomTag_WhenValidated_ThenItShouldSucceed()
```
pengchen0692, line 176: nit: whenValidating_thenVlaidationPasses (just keep concistency(

src/Microsoft.Health.Dicom.Core/Features/Store/DicomDatasetValidator.cs
```diff
@@ -90,22 +93,56 @@ string EnsureRequiredTagIsPresent(DicomTag dicomTag)
...
                 if (dicomElement != null)
                 {
-                    string value = dicomDataset.GetSingleValueOrDefault<string>(indexableTag, default);
-                    _minimumValidator.Validate(indexableTag, value);
+                    _minimumValidator.Validate(dicomElement);
+                }
+            }
+
+            // Process Private Tag
+            // dicomDataset.GetDicomItem<DicomElement>() is not able to get the private tag, we need to compare through path
```
pengchen0692, line 132: nit: cannot get value for private tag, we need to loop and compare with path

src/Microsoft.Health.Dicom.Core/Features/Store/IDicomDatasetValidator.cs
```diff
@@ -16,10 +18,11 @@ public interface IDicomDatasetValidator
         /// Validates the <paramref name="dicomDataset"/>.
         /// </summary>
         /// <param name="dicomDataset">The DICOM dataset to validate.</param>
+        /// <param name="customTagEntries">The DICOM dataset to validatesd.</param>
```
pengchen0692, line 21: update summary

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMinimumValidation.cs
```diff
@@ -131,6 +148,42 @@ public static void ValidateUI(string value, string name)
...
 
+        public static void ValidateUI(DicomElement element)
+        {
+            EnsureArg.IsNotNull(element, nameof(element));
+            string name = GetName(element.Tag);
+            string value = element.Get<string>();
+            ValidateUI(value, name);
+        }
+
+        public static void DefaultValidate(DicomElement element)
```
pengchen0692, line 159: add summary

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMinimumValidator.cs
```diff
@@ -5,45 +5,38 @@
...
         }
 
         // only works for single value dicom element
-        public void Validate(DicomTag dicomTag, string value)
+        public void Validate(DicomElement element)
         {
-            EnsureArg.IsNotNull(dicomTag, nameof(dicomTag));
-            DicomVR dicomVR = dicomTag.DictionaryEntry.ValueRepresentations.FirstOrDefault();
-
-            if (dicomVR == null)
```
pengchen0692, line 35: should not remove check on dicomVR

src/Microsoft.Health.Dicom.Core/Features/Store/DicomDatasetValidator.cs
```diff
@@ -90,22 +93,56 @@ string EnsureRequiredTagIsPresent(DicomTag dicomTag)
...
+                    privateTags.Add(customTagEntry.Path, customTagEntry);
+                }
+                else
+                {
+                    standardTags.Add(dicomTag);
+                }
+            }
+
+            // Process Standard Tag
             HashSet<DicomTag> indexableTags = QueryLimit.AllInstancesTags;
```
wsugarman, line 118: `QueryLimit.AllInstanceTags` is mutable, as it is a plain ol' `HashSet<T>` (the `readonly` keyword only denotes that the reference itself is immutable). So the next `UnionWith` method is modifying the `static` `QueryLimit.AllInstancesTags` value!!

This is probably not what you want, right? Instead, we should make some changes:
1. Can `QueryLimit` be `internal`?
2. All of the `HashSet<T>` fields in `QueryLimit` should be:
    1. Either `IReadOnlyCollectiont<T>` or `ImmutableHashSet<T>` depending on your need
    2. Readonly properties (only `{ get; }`) instead of fields to make future changes easier

src/Microsoft.Health.Dicom.Core/Features/Store/StoreService.cs
```diff
@@ -81,12 +87,13 @@ public class StoreService : IStoreService
             {
                 _dicomInstanceEntries = instanceEntries;
                 _requiredStudyInstanceUid = requiredStudyInstanceUid;
+                List<CustomTagEntry> customTagEntries = (await _customTagStore.GetCustomTagsAsync(path: null, cancellationToken: cancellationToken)).ToList();
```
wsugarman, line 90: Instead of making this a list, can we simply return some sort of `List<T>`-like type from `GetCustomTagsAsync`?

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMinimumValidation.cs
wsugarman, line 25: If you're calling this to throw an exception for a particular case, and you're using quite a lot in the class, I'd change this method (or create a new one) for that purpose explicitly. For example, often `IDisposable` classes have a method called `ThrowIfDisposed()` whose entire purpose is to throw `IObjectDisposedException` if the class has already been disposed.

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/DicomDatasetValidatorTests.cs
```diff
@@ -43,7 +45,7 @@ public DicomDatasetValidatorTests()
         [Fact]
         public void GivenAValidDicomDataset_WhenValidated_ThenItShouldSucceed()
         {
-            _dicomDatasetValidator.Validate(_dicomDataset, requiredStudyInstanceUid: null);
+            _dicomDatasetValidator.Validate(_dicomDataset, new List<CustomTagEntry>(), requiredStudyInstanceUid: null);
```
wsugarman, line 48: nit: `T[]` implements `IList<T>` and `IReadOnlyList<T>`, so you can pass `Array.Empty<CustomTagEntry>()` here and below where you need some empty list of stuff (just don't try adding to it).

I recommend this especially outside of test code as `Empty` doesn't allocate a new array.

wsugarman (changes requested):

src/Microsoft.Health.Dicom.Core/Features/Store/IDicomDatasetValidator.cs
```diff
@@ -16,10 +18,11 @@ public interface IDicomDatasetValidator
         /// Validates the <paramref name="dicomDataset"/>.
         /// </summary>
         /// <param name="dicomDataset">The DICOM dataset to validate.</param>
+        /// <param name="customTagEntries">The custom tag entries.</param>
         /// <param name="requiredStudyInstanceUid">
         /// If supplied, the StudyInstanceUID in the <paramref name="dicomDataset"/> must match to be considered valid.
         /// </param>
         /// <exception cref="DatasetValidationException">Thrown when the validation fails.</exception>
-        void Validate(DicomDataset dicomDataset, string requiredStudyInstanceUid);
+        void Validate(DicomDataset dicomDataset, IReadOnlyList<CustomTagEntry> customTagEntries, string requiredStudyInstanceUid);
```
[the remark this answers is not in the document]
pengchen0692: I'm thinking of just using IList, since you wont' be able to convert a IReadonlyList to IList back.

src/Microsoft.Health.Dicom.Core/Features/Store/DicomDatasetValidator.cs
pengchen0692: I think the _public_ vs _internal_ is very common.

src/Microsoft.Health.Dicom.Core/Features/Store/IDicomDatasetValidator.cs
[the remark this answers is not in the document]
wsugarman: Yeah, unfortunately, `IList<T>` does not implement `IReadOnlyList<T>` (what a pain!). But it does implement `IReadOnlyCollection<T>`. At least for a validator, I wouldn't expect it to require the ability to mutate my list

wsugarman (approved):

src/Microsoft.Health.Dicom.Core/Features/Store/DicomDatasetValidator.cs
```diff
@@ -32,9 +34,10 @@ public DicomDatasetValidator(IOptions<FeatureConfiguration> featureConfiguration
             _minimumValidator = minimumValidator;
         }
 
-        public void Validate(DicomDataset dicomDataset, string requiredStudyInstanceUid)
+        public void Validate(DicomDataset dicomDataset, IList<CustomTagEntry> customTagEntries, string requiredStudyInstanceUid)
```
smithago, line 37: I am little worried how this is passed in has a new param.

We need a model that can handle this correctly without passing custom tags like this.

src/Microsoft.Health.Dicom.Core/Features/Store/DicomDatasetValidator.cs
```diff
@@ -90,22 +93,66 @@ string EnsureRequiredTagIsPresent(DicomTag dicomTag)
...
 
-        private void ValidateIndexedItems(DicomDataset dicomDataset)
+        private void ValidateIndexedItems(DicomDataset dicomDataset, IList<CustomTagEntry> customTagEntries)
         {
-            HashSet<DicomTag> indexableTags = QueryLimit.AllInstancesTags;
+            HashSet<DicomTag> standardTags = new HashSet<DicomTag>();
+            Dictionary<string, CustomTagEntry> privateTags = new Dictionary<string, CustomTagEntry>();
+            foreach (var customTagEntry in customTagEntries)
+            {
+                DicomTag dicomTag = DicomTag.Parse(customTagEntry.Path);
```
smithago, line 106: Validate should not be doing this. Have knowledge of what is private or not. Its job is just to validate the value against the VR type. The models passed should have enough information for validators to do this

smithago (changes requested):
I have left some comments to reduce the knowledge of custom tags in the validator code

src/Microsoft.Health.Dicom.Core/Features/Store/StoreService.cs
```diff
@@ -81,12 +87,14 @@ public class StoreService : IStoreService
             {
                 _dicomInstanceEntries = instanceEntries;
                 _requiredStudyInstanceUid = requiredStudyInstanceUid;
+                IList<CustomTagEntry> customTagEntries = await _customTagStore.GetCustomTagsAsync(path: null, cancellationToken: cancellationToken);
```
smithago, line 90: You should consider using IReadonlyCollection here. These lists need to be immutable.

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMinimumValidator.cs
```diff
@@ -5,45 +5,38 @@
...
+            if (_minValidators.TryGetValue(element.ValueRepresentation, out Action<DicomElement> validator))
             {
-                validator(value, dicomTag.DictionaryEntry.Keyword);
+                validator(element);
             }
             else
             {
-                Debug.Fail($"Missing validation action for for VR :{dicomVR.Code}, add a new validation and register in the constructor.");
+                // Use default validator provided by Fo-dicom, if we see problems in the feature, could create custom ones.
+                DicomElementMinimumValidation.DefaultValidate(element);
```
smithago, line 39: We should not do this. We should build our own. The fo-dicom validator may do more than we need.

smithago: That is why it is called min validator

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMinimumValidation.cs
smithago: We should not mix fo-dicom with our validation

src/Microsoft.Health.Dicom.Core/Features/Store/DicomDatasetValidator.cs
pengchen0692: how about:  IDictionary<DicomTag,DicomVR> validatingTags..

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMinimumValidator.cs
pengchen0692: maybe for string like VR, just verify length?

pengchen0692: btw, what's the purpose of minvalidation? is it for successfully saving data to database?

src/Microsoft.Health.Dicom.Core/Features/Store/DicomDatasetValidator.cs
pengchen0692: The reason for seperating standard and private tag is that  we need retrieve associated DicomElement in different way.
Standard custom tag can be retrieved as: DicomDataset.GetDicomItem, while Private custom tag need to loop entire dataset and locate with tagPath, since we don't have privateCreator.

pengchen0692:
Moved to https://github.com/microsoft/dicom-server/pull/605

PR #612, opened 2021-03-07
Implement SQL for STOW

pengchen0692:
## Description
Implement SQL for STOW

## Related issues
Addresses [AB#80088](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/80088)

## Testing


src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -158,5 +149,71 @@ public static void AddValueIfNotNull<T>(this DicomDataset dicomDataset, DicomTag
                 dicomDataset.Add(dicomTag, value);
             }
         }
+
+        public static IDictionary<IndexTag, DicomTag> GetDicomTags(this DicomDataset dicomDataset, IEnumerable<IndexTag> indexTags)
```
pengchen0692, line 153: add summary, add unit tests

pengchen0692: unit tests:
1.  when contains both standtag and private tags, both are processed correctly

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -158,5 +149,71 @@ public static void AddValueIfNotNull<T>(this DicomDataset dicomDataset, DicomTag
...
+                            }
+                        }
+                    }
+                }
+            }
+
+            return result;
+        }
+
+        public static IReadOnlyDictionary<IndexTag, object> GetIndexTagValues(this DicomDataset dicomDataset, IEnumerable<IndexTag> indexTags)
```
pengchen0692, line 194: summary and unit tests
1. when contains both standard and private tag, able to get  data

src/Microsoft.Health.Dicom.Core/Extensions/DicomElementExtensions.cs
```diff
@@ -0,0 +1,146 @@
...
+            { DicomVR.SH, GetStringValue },
+            { DicomVR.SL, GetIntValue },
+            { DicomVR.SS, GetShortValue },
+            { DicomVR.TM, GetTMValue },
+            { DicomVR.UI, GetStringValue },
+            { DicomVR.UL, GetUIntValue },
+            { DicomVR.US, GetUShortValue },
+        };
+
+        public static object GetSingleValue(this DicomElement element)
```
pengchen0692, line 43: 1. summary
2. unit tests
 Verify all supported DicomElement are able to get data

src/Microsoft.Health.Dicom.Core/Extensions/DicomElementExtensions.cs
```diff
@@ -0,0 +1,146 @@
...
+
+        private static object SafeGetValue<T>(DicomElement element)
+        {
+            try
+            {
+                return element.Get<T>();
+            }
+            catch
+            {
+                return null;
```
pengchen0692, line 81: add log

src/Microsoft.Health.Dicom.Core/Features/Store/StoreOrchestrator.cs
```diff
@@ -53,8 +58,8 @@ public class StoreOrchestrator : IStoreOrchestrator
             EnsureArg.IsNotNull(dicomInstanceEntry, nameof(dicomInstanceEntry));
 
             DicomDataset dicomDataset = await dicomInstanceEntry.GetDicomDatasetAsync(cancellationToken);
-
-            long version = await _indexDataStore.CreateInstanceIndexAsync(dicomDataset, cancellationToken);
+            var dicomTags = await _indexTagService.GetIndexTagsAsync(cancellationToken);
```
pengchen0692, line 61: nit:indextags

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
```diff
@@ -171,6 +171,61 @@ CREATE TYPE dbo.AddCustomTagsInputTableType_1 AS TABLE
...
+)
+GO
+
+/*************************************************************
+    Table valued parameter to insert into Custom table for data type Big Int
+*************************************************************/
+CREATE TYPE dbo.InsertBigIntCustomTagTableType_1 AS TABLE
+(
+    TagKey                     BIGINT,
+    TagValue                BIGINT,
```
pengchen0692, line 202: nit formatting

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
```diff
@@ -305,3 +360,326 @@ AS
...
+        AND StudyKey = @studyKey
+    END
+
+    -- Insert Custom Tags
+    DECLARE @rowCount AS BIGINT
+    DECLARE @temp AS INT
+
+    -- String tags
+    SET @rowCount = (SELECT COUNT(TagKey) FROM @stringCustomTags)
+	IF @rowCount <> 0
```
pengchen0692, line 507: nit: formatting

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Model/ImmutableTypes.Generated.cs
```diff
@@ -44,4 +44,174 @@ internal AddCustomTagsInputTableTypeV1Row(System.String TagPath, System.String T
...
+    internal struct InsertStringCustomTagTableTypeV1Row
+    {
+        internal InsertStringCustomTagTableTypeV1Row(System.Int64 TagKey, System.String TagValue, System.Byte TagLevel)
+        {
+            this.TagKey = TagKey;
+            this.TagValue = TagValue;
+            this.TagLevel = TagLevel;
+        }
+
+        internal System.Int64 TagKey { get; }
```
pengchen0692, line 213: why int64?

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStore.cs
```diff
@@ -61,11 +73,13 @@ public async Task<long> CreateInstanceIndexAsync(DicomDataset instance, Cancella
...
                     instance.GetSingleValueOrDefault<string>(DicomTag.Modality),
                     instance.GetStringDateAsDateTime(DicomTag.PerformedProcedureStepStartDate),
-                    (byte)IndexStatus.Creating);
+                    (byte)IndexStatus.Creating,
+                    parameters);
 
                 try
                 {
-                    return (long)(await sqlCommandWrapper.ExecuteScalarAsync(cancellationToken));
+                    object result = await sqlCommandWrapper.ExecuteScalarAsync(cancellationToken);
```
pengchen0692, line 81: remove object

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStore.cs
```diff
@@ -87,6 +105,45 @@ public async Task<long> CreateInstanceIndexAsync(DicomDataset instance, Cancella
             }
         }
 
+        private static VLatest.AddInstanceTableValuedParameters BuildParameters(IReadOnlyDictionary<IndexTag, object> customTagValues)
```
pengchen0692, line 108: BuildAddInstnaceTableValuedParameters

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStore.cs
```diff
@@ -87,6 +105,45 @@ public async Task<long> CreateInstanceIndexAsync(DicomDataset instance, Cancella
...
+        {
+            List<InsertStringCustomTagTableTypeV1Row> stringRows = new List<InsertStringCustomTagTableTypeV1Row>();
+            List<InsertDoubleCustomTagTableTypeV1Row> doubleRows = new List<InsertDoubleCustomTagTableTypeV1Row>();
+            List<InsertBigIntCustomTagTableTypeV1Row> bigIntRows = new List<InsertBigIntCustomTagTableTypeV1Row>();
+            List<InsertDateTimeCustomTagTableTypeV1Row> dateTimeRows = new List<InsertDateTimeCustomTagTableTypeV1Row>();
+            List<InsertPersonNameCustomTagTableTypeV1Row> personNameRows = new List<InsertPersonNameCustomTagTableTypeV1Row>();
+            foreach (var item in customTagValues.Keys)
```
pengchen0692, line 115: item=>indexTag

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStore.cs
```diff
@@ -87,6 +105,45 @@ public async Task<long> CreateInstanceIndexAsync(DicomDataset instance, Cancella
...
+                    case CustomTagDataType.PersonNameData:
+                        personNameRows.Add(new InsertPersonNameCustomTagTableTypeV1Row(entry.Key, (string)value, (byte)entry.Level));
+                        break;
+                    default:
+                        Debug.Fail($"{dataType} cannot be processed");
+                        break;
+                }
+            }
+
+            VLatest.AddInstanceTableValuedParameters parameters = new VLatest.AddInstanceTableValuedParameters(stringRows, bigIntRows, doubleRows, dateTimeRows, personNameRows);
```
pengchen0692, line 143: remove parameters

src/Microsoft.Health.Dicom.Core/Extensions/DicomElementExtensions.cs
pengchen0692: how to add log for static methods?

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
pengchen0692: Unit test:
2. If standard tag not exist in dataset, should not return
3. If private tag not exist in dataset, should not return (VR not match, tagpath not exist)

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -158,5 +149,85 @@ public static void AddValueIfNotNull<T>(this DicomDataset dicomDataset, DicomTag
...
+            Dictionary<string, IndexTag> privateTags = new Dictionary<string, IndexTag>();
+            foreach (IndexTag indexTag in indexTags)
+            {
+                if (!indexTag.Tag.IsPrivate && dicomDataset.Contains(indexTag.Tag))
+                {
+                    result.Add(indexTag, indexTag.Tag);
+                }
+                else
+                {
+                    privateTags.Add(indexTag.Tag.GetPath(), indexTag);
```
pengchen0692, line 174: the logic is not right, fix it.

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -158,5 +149,88 @@ public static void AddValueIfNotNull<T>(this DicomDataset dicomDataset, DicomTag
...
+            EnsureArg.IsNotNull(indexTags, nameof(indexTags));
+            var dicomTags = dicomDataset.GetDicomTags(indexTags);
+            Dictionary<IndexTag, object> result = new Dictionary<IndexTag, object>();
+
+            foreach (var pair in dicomTags)
+            {
+                DicomElement element = dicomDataset.GetDicomItem<DicomElement>(pair.Value);
+
+                // we only support single value
+                if (element != null && element.Count == 1)
```
pengchen0692: simplify this

src/Microsoft.Health.Dicom.Core/Extensions/DicomElementExtensions.cs
```diff
@@ -0,0 +1,164 @@
...
+using Dicom;
+using EnsureThat;
+using Microsoft.Health.Dicom.Core.Features.Validation;
+
+namespace Microsoft.Health.Dicom.Core.Extensions
+{
+    /// <summary>
+    /// Extension methods for <see cref="DicomElement"/>.
+    /// </summary>
+    public static class DicomElementExtensions
```
pengchen0692, line 18: Or DicomElementReader?

src/Microsoft.Health.Dicom.Core/Extensions/DicomElementExtensions.cs
pengchen0692: make it instance class?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -708,6 +775,181 @@ AS
...
+    IF @rowCount <> 0
+    BEGIN
+        -- Validate and lock custom tags
+        SELECT @temp = dbo.CustomTag.TagKey
+        FROM dbo.CustomTag WITH (UPDLOCK)
+        INNER JOIN @stringCustomTags input
+        ON input.TagKey = dbo.CustomTag.TagKey 
+        AND ( dbo.CustomTag.TagStatus = 0 OR  dbo.CustomTag.TagStatus = 1)
+    
+        IF @@ROWCOUNT<>@rowCOUNT
```
pengchen0692, line 794: nit: formatting

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -708,6 +775,181 @@ AS
...
+        AND ISNULL(T.InstanceKey, @instanceKey) = @instanceKey
+        WHEN MATCHED THEN 
+            UPDATE SET T.Watermark = @newWatermark, T.TagValue = S.TagValue
+        WHEN NOT MATCHED THEN 
+            INSERT (TagKey, TagValue, StudyKey, SeriesKey, InstanceKey, Watermark)
+            VALUES(
+            S.TagKey,
+            S.TagValue,
+            @studyKey,
+            (CASE WHEN S.TagLevel<>2 THEN @seriesKey ELSE NULL END),
```
pengchen0692, line 812: nit: formatting

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -708,6 +775,181 @@ AS
...
+            (CASE WHEN S.TagLevel=0 THEN @instanceKey ELSE NULL END),
```
pengchen0692, line 813: nit: formatting

src/Microsoft.Health.Dicom.Core/Extensions/DicomElementExtensions.cs
wsugarman: I think it makes sense as-is

wsugarman, line 18: Does this need to be `public` though?

src/Microsoft.Health.Dicom.Core/Extensions/DicomElementExtensions.cs
```diff
@@ -0,0 +1,164 @@
...
+    {
+        private static readonly IReadOnlyDictionary<DicomVR, Func<DicomElement, object>> Readers = new Dictionary<DicomVR, Func<DicomElement, object>>()
```
wsugarman, line 20: nit: `()` in ctor are not necessary

src/Microsoft.Health.Dicom.Core/Extensions/DicomElementExtensions.cs
```diff
@@ -0,0 +1,164 @@
...
+        /// <returns>The value.</returns>
+        public static object GetSingleValue(this DicomElement element)
+        {
+            EnsureArg.IsNotNull(element, nameof(element));
+            if (element.Count != 1)
+            {
+                return null;
+            }
+
+            if (!Readers.ContainsKey(element.ValueRepresentation))
```
wsugarman, line 57: nit: use `TryGetValue` so you don't need to look up the func twice

src/Microsoft.Health.Dicom.Core/Extensions/DicomElementExtensions.cs
```diff
@@ -0,0 +1,164 @@
...
+            }
+
+            return Readers[element.ValueRepresentation].Invoke(element);
+        }
+
+        private static object GetDAValue(DicomElement element)
+        {
+            string stringDate = (string)SafeGetValue<string>(element);
+
+            if (stringDate == null ||
```
wsugarman, line 70: nit: `TryParse` can handle `null`

wsugarman: You can even return a single expression:
```csharp
return DicomElementMinimumValidation.TryParseDA(stringDate, out DateTime result) ? result : null;
```

src/Microsoft.Health.Dicom.Core/Extensions/DicomElementExtensions.cs
wsugarman: You could add the option of passing a logger

wsugarman: Or you could have a statically-scoped `ILoggerFactory<T>` that was configured in start-up that this class can use. Given this is a library, it is a bit more difficult to do that though.

wsugarman: However, should we really be swallowing type-mismatch exceptions like this?

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -55,15 +53,8 @@ public static T GetSingleValueOrDefault<T>(this DicomDataset dicomDataset, Dicom
...
-            string stringDate = dicomDataset.GetSingleValueOrDefault<string>(dicomTag, default);
-
-            if (stringDate == null ||
-                !DateTime.TryParseExact(stringDate, DateFormat, CultureInfo.InvariantCulture, DateTimeStyles.AdjustToUniversal, out DateTime result))
-            {
-                return null;
-            }
-
-            return result;
+            DicomElement element = dicomDataset.GetDicomItem<DicomElement>(dicomTag);
```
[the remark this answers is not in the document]
wsugarman: Also as an aside, at some point we'll probably want to fix `GetSingleValueOrDefault<T>` to avoid unnecessary boxing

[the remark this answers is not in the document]
pengchen0692: this is good idea, I'm thinking of getting rid of DicomElementExtension here, since most time it's equal with GetSingleValueOrDefault

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
```diff
@@ -13,6 +13,14 @@ namespace Microsoft.Health.Dicom.Core.Features.CustomTag
     /// </summary>
     public class CustomTagEntry : IEquatable<CustomTagEntry>
     {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CustomTagEntry"/> class.
+        /// As used as customer input, constructor without parameters is required.
```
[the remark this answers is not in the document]
pengchen0692: will fix it

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
```diff
@@ -35,23 +43,23 @@ public CustomTagEntry(CustomTagStoreEntry storedEntry)
         /// Path of this tag. Normally it's composed of groupid and elementid.
         /// E.g: 00100020 is path of patient id.
         /// </summary>
-        public string Path { get; private set; }
```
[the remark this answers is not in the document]
pengchen0692: sorry, not get you. with "private set", Asp.net core is not able to set the value, hence fail out.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/IndexTag.cs
```diff
@@ -42,6 +42,11 @@ public IndexTag(DicomTag tag, DicomVR vr, CustomTagLevel level, bool isCustomTag
         /// <summary>
         /// Gets whether this is custom tag or not.
         /// </summary>
-        public bool IsCustomTag { get; }
+        public bool IsCustomTag { get => CustomTagStoreEntry != null; }
```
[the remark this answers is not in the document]
pengchen0692: CustomTag is used everywhere, Sneha is creating  a user story to get all them renamed

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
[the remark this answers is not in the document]
wsugarman: I see that you deleted the setter, and without it no code would set it via reflection (like presumably some automatically deserializer, like those in ASP.NET Core or JSON.NET). Are you keeping these values, or is the serializer using the ctor instead?

src/Microsoft.Health.Dicom.Core/Extensions/DicomElementExtensions.cs
pengchen0692: This class is removed

src/Microsoft.Health.Dicom.Core/Extensions/DicomElementExtensions.cs
pengchen0692: This class is removed

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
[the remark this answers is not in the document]
pengchen0692: eh...my intention is to keep both getter and setter, so that serializer  works.

Not  sure if there is a way for serializer to use the constructor with parameters.

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMinimumValidation.cs
```diff
@@ -123,25 +123,40 @@ public static void ValidateDA(string value, string name)
...
-            if (!DateTime.TryParseExact(value, DateFormatDA, CultureInfo.InvariantCulture, DateTimeStyles.AdjustToUniversal, out _))
+            if (!TryParseDA(value, out _))
             {
                 throw new DicomElementValidationException(name, value, DicomVR.DA, DicomCoreResource.ValueIsInvalidDate);
             }
         }
 
+        internal static bool TryParseDA(string value, out DateTime dateTime)
+        {
+            return DateTime.TryParseExact(value, DateFormatDA, CultureInfo.InvariantCulture, DateTimeStyles.AdjustToUniversal, out dateTime);
```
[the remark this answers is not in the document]
pengchen0692: This is original logic, I just copied from https://github.com/microsoft/dicom-server/blob/bd33cf21595a1ab1d21d2e5bcea8e83f3e4710ac/src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs#L60. 

Maybe we need more investigation.

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMinimumValidation.cs
```diff
@@ -123,25 +123,40 @@ public static void ValidateDA(string value, string name)
...
+
+            if (!TryParseDT(value, out _))
             {
                 throw new DicomElementValidationException(name, value, DicomVR.DT, DicomCoreResource.ValueIsInvalidDate);
             }
         }
 
+        internal static bool TryParseDT(string value, out DateTime dateTime)
+        {
+            return DateTime.TryParseExact(value, DateFormatDT, CultureInfo.InvariantCulture, DateTimeStyles.NoCurrentDateDefault, out dateTime);
```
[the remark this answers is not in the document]
pengchen0692: This code is copied from fo-dicom https://github.com/fo-dicom/fo-dicom/blob/fc6347e30586feacf6152f4a38109906c93839f1/DICOM/DicomElement.cs#L253. May need more investigation.

src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/IndexTagValueReader.cs
```diff
@@ -0,0 +1,144 @@
...
+using System.Diagnostics;
+using System.Linq;
+using Dicom;
+using Microsoft.Health.Dicom.Core.Extensions;
+using Microsoft.Health.Dicom.Core.Features.CustomTag;
+
+namespace Microsoft.Health.Dicom.SqlServer.Features.CustomTag
+{
+    /// <summary>
+    /// Date type of custom tag.
```
pengchen0692, line 17: update summary

src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/IndexTagValueReader.cs
```diff
@@ -0,0 +1,144 @@
...
+    /// </summary>
+    internal static class IndexTagValueReader
+    {
+        public static void Read(
```
pengchen0692, line 21: Add summary

src/Microsoft.Health.Dicom.Core/Extensions/DicomElementExtensions.cs
pengchen0692: This class is removed.

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMinimumValidation.cs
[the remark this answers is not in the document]
pengchen0692: I think we don't consider timezone attribute for simplicity at the present.
I tried both AdjustToUniversal and None, they return Unspecified.
`    DateTime dt = DateTime.ParseExact("20210512", "yyyyMMdd", CultureInfo.InvariantCulture, DateTimeStyles.AdjustToUniversal);`

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -158,5 +188,57 @@ public static void AddValueIfNotNull<T>(this DicomDataset dicomDataset, DicomTag
...
+
+        /// <summary>
+        /// Get DicomTags for index Tag from Dicom Dataset.
+        /// </summary>
+        /// <remarks>If indextag not exist in dataset, should not return.</remarks>
+        /// <param name="dicomDataset">The dicom dataset.</param>
+        /// <param name="indexTags">The index Tags.</param>
+        /// <returns>Mapping between IndexTag and DicomTag.</returns>
+        public static IDictionary<IndexTag, DicomTag> GetDicomTags(this DicomDataset dicomDataset, IEnumerable<IndexTag> indexTags)
+        {
```
smithago, line 200: I am not sure about this whole method. Should'nt DicomTag be contained in IndexTag? Why are transforming it again based on the DicomDataset?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/IndexTag.cs
```diff
@@ -13,15 +13,15 @@ namespace Microsoft.Health.Dicom.Core.Features.CustomTag
...
-        public IndexTag(DicomTag tag, DicomVR vr, CustomTagLevel level, bool isCustomTag)
+        public IndexTag(DicomTag tag, DicomVR vr, CustomTagLevel level, CustomTagStoreEntry customTagStoreEntry)
         {
             EnsureArg.IsNotNull(tag, nameof(tag));
             EnsureArg.IsNotNull(vr, nameof(vr));
 
             Tag = tag;
             VR = vr;
-            IsCustomTag = isCustomTag;
             Level = level;
```
smithago, line 23: You probably need 2 different constructors, one for standard and one for custom. Not sure you still need to take CustomTagLevel

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMinimumValidation.cs
```diff
@@ -235,12 +250,17 @@ internal static void ValidateSS(IByteBuffer value, string name)
...
         internal static void ValidateTM(string value, string name)
         {
-            if (!DateTime.TryParseExact(value, DataFromatTM, CultureInfo.InvariantCulture, DateTimeStyles.NoCurrentDateDefault, out _))
+            if (!TryParseTM(value, out _))
             {
                 throw new DicomElementValidationException(name, value, DicomVR.DT, DicomCoreResource.ValueIsInvalidDate);
             }
         }
 
+        internal static bool TryParseTM(string value, out DateTime dateTime)
```
smithago, line 259: Can we not support TM? We need to support only minimal things needed. I am not sure why would anyone index on TM

smithago: We can add if customer needs

src/Microsoft.Health.Dicom.Core/DicomCoreResource.resx
```diff
@@ -334,5 +334,8 @@ The first part date {1} should be lesser than or equal to the second part date {
   <data name="ValueLengthIsNotRequiredLength" xml:space="preserve">
     <value>Value length is not {0}.</value>
     <comment>{0} required length.</comment>
-  </data>  
+  </data>
+  <data name="CustomTagsAlreadyChanged" xml:space="preserve">
+    <value>One or more custom tags have been changed.</value>
```
smithago, line 339: Exception messages should be understandable by customers and actionable. I am not able to understand what you are trying to convey here.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -708,6 +775,181 @@ AS
         AND StudyKey = @studyKey
     END
 
+    -- Insert Custom Tags
+    DECLARE @rowCount AS BIGINT
+    -- This variable is used for not returning value from SELECT.
+    DECLARE @temp AS INT
+
+    -- String Key tags
+    SET @rowCount = (SELECT COUNT(TagKey) FROM @stringCustomTags)
```
smithago, line 784: IF EXISTS (SELECT * FROM @stringCustomTags)
BEGIN
...
END

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -708,6 +775,181 @@ AS
...
+    IF @rowCount <> 0
+    BEGIN
+        -- Validate and lock custom tags
+        SELECT @temp = dbo.CustomTag.TagKey
```
smithago, line 788: @temp won't work of the join results in more than one row

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -708,6 +775,181 @@ AS
...
+        FROM dbo.CustomTag WITH (UPDLOCK)
```
smithago, line 789: You don't need this select.

Instead change the MERGE condition to add only those customTag whose status is 2.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -708,6 +775,181 @@ AS
...
+        INNER JOIN @stringCustomTags input
+        ON input.TagKey = dbo.CustomTag.TagKey 
+        AND ( dbo.CustomTag.TagStatus = 0 OR  dbo.CustomTag.TagStatus = 1)
+    
+        IF @@ROWCOUNT<>@rowCOUNT
+            THROW 50409, 'Custom Tags have been changed', 10
```
smithago, line 795: Can remove this and the exception

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -708,6 +775,181 @@ AS
         AND StudyKey = @studyKey
     END
 
+    -- Insert Custom Tags
```
smithago, line 778: can you move all of this to after inserting into the instnce table

test/Microsoft.Health.Dicom.Tests.Integration/Persistence/CustomTagStoreTests.STOW.cs
```diff
@@ -0,0 +1,203 @@
...
+            DicomTag tag = DicomTag.ConversionType;
+            string value = "SYN";
+            dataset.Add(tag, value);
+
+            IndexTag indexTag = await AddCustomTag(tag.BuildCustomTagEntry(level: level));
+            try
+            {
+                long watermark = await _indexDataStore.CreateInstanceIndexAsync(dataset, new IndexTag[] { indexTag });
+                Instance instance = await _testHelper.GetInstanceAsync(studyInstanceUid, seriesInstanceUid, sopInstanceUid, watermark);
+                long? seriesKey = level != CustomTagLevel.Study ? instance.SeriesKey : null;
```
smithago, line 130: Are you validating the sql code that is adding SeriesKey and InstanceKey to the custom table based on tagLevel?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
smithago: hint it with repeatableread

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStore.cs
```diff
@@ -56,12 +90,13 @@ public async Task<long> CreateInstanceIndexAsync(DicomDataset instance, Cancella
...
-                    instance.GetStringDateAsDateTime(DicomTag.StudyDate),
+                    instance.GetStringDateAsDate(DicomTag.StudyDate),
                     instance.GetSingleValueOrDefault<string>(DicomTag.StudyDescription),
                     instance.GetSingleValueOrDefault<string>(DicomTag.AccessionNumber),
                     instance.GetSingleValueOrDefault<string>(DicomTag.Modality),
-                    instance.GetStringDateAsDateTime(DicomTag.PerformedProcedureStepStartDate),
-                    (byte)IndexStatus.Creating);
+                    instance.GetStringDateAsDate(DicomTag.PerformedProcedureStepStartDate),
+                    (byte)IndexStatus.Creating,
+                    parameters);
```
smithago, line 99: How will the compat work? Basically I have V1 DB and V2 Code. You will have to check the schema version and call different SQL procedures.

Make sure same is achieved for other existing v1 SQL procedures who's signature is changing

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -708,6 +775,181 @@ AS
...
+        AND ( dbo.CustomTag.TagStatus = 0 OR  dbo.CustomTag.TagStatus = 1)
+    
+        IF @@ROWCOUNT<>@rowCOUNT
+            THROW 50409, 'Custom Tags have been changed', 10
+    
+        -- Merge into dbo.CustomTagString table        
+        MERGE INTO dbo.CustomTagString AS T
+        USING @stringCustomTags AS S
+        ON T.TagKey = S.TagKey
+        AND T.StudyKey = @studyKey
```
smithago, line 801: indent AND

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
pengchen0692: Rename it to GetMatchingDicomTags which get matching dicom tag from dataset.
It's mostly for private tag -- we don't save private creator, so DicomDataset.GetSingleValueOrDefault(DicomTag) doesn't work for private tag, need to loop dataset and locate it by path.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/IndexTag.cs
pengchen0692: Still keep existing constructor as simple as now, created another 2 static method for converting.

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMinimumValidation.cs
pengchen0692: removed TM

test/Microsoft.Health.Dicom.Tests.Integration/Persistence/CustomTagStoreTests.STOW.cs
pengchen0692: these integration tests verify :
* adding new custom tag index tag  
* updating existing data ( on series/study custom tag)

test/Microsoft.Health.Dicom.Tests.Integration/Persistence/SqlIndexDataStoreTestHelper.cs
```diff
@@ -218,5 +229,92 @@ public async Task<IReadOnlyList<ChangeFeedRow>> GetChangeFeedRowsAsync(string st
                 return result;
             }
         }
+
+        internal async Task<IReadOnlyList<CustomTagDataRow>> GetCustomTagDataAsync(
```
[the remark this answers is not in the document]
pengchen0692: one is to get all rows with given tagKey,  the other one can also accept studyKey and seriesKey,  the Select statement is different,  may be we could merge them together.

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMinimumValidation.cs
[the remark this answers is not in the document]
pengchen0692: I will bring this up in standup  parking lot

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMinimumValidation.cs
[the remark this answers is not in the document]
pengchen0692: I will bring this up in standup  parking lot

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -45,25 +44,43 @@ public static T GetSingleValueOrDefault<T>(this DicomDataset dicomDataset, Dicom
             return dicomDataset.GetSingleValueOrDefault<T>(dicomTag, default);
         }
 
+        public static long? GetDicomTagValueAsLong(this DicomDataset dicomDataset, DicomTag dicomTag)
```
snehabandla, line 47: >GetDicomTagValueAsLong [](start = 28, length = 22)

Summary?

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -45,25 +44,43 @@ public static T GetSingleValueOrDefault<T>(this DicomDataset dicomDataset, Dicom
...
+        {
+            EnsureArg.IsNotNull(dicomDataset, nameof(dicomDataset));
+
+            DicomTag tag = dicomDataset.GetSingleValueOrDefault<DicomTag>(dicomTag, default);
+            if (tag == null)
```
snehabandla, line 52: Why are you getting a tag here? Isn't the goal to get the long value for the tag?

pengchen0692: maybe I should rename the "DicomTag tag" to "DicomTag value" to avoid this confusion.
Attribute AT is supposed to store a DicomTag value, we need to convert it to long  to store to databse.

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
pengchen0692: will add

test/Microsoft.Health.Dicom.Tests.Integration/Persistence/CustomTagStoreTests.STOW.cs
```diff
@@ -0,0 +1,180 @@
...
+using Dicom;
+using Microsoft.Health.Dicom.Core.Features.CustomTag;
+using Microsoft.Health.Dicom.SqlServer.Features.CustomTag;
+using Microsoft.Health.Dicom.Tests.Common;
+using Microsoft.Health.Dicom.Tests.Common.Extensions;
+using Microsoft.Health.Dicom.Tests.Integration.Persistence.Models;
+using Xunit;
+
+namespace Microsoft.Health.Dicom.Tests.Integration.Persistence
+{
```
snehabandla, line 19: Can we name this file differently? It isn't exactly testing the CustomTagStore directly, it more tests the index data store. So perhaps IndexDataStoreTests.CustomTag.cs?

test/Microsoft.Health.Dicom.Tests.Integration/Persistence/SqlIndexDataStoreTestHelper.cs
```diff
@@ -16,6 +18,15 @@ public class SqlIndexDataStoreTestHelper : IIndexDataStoreTestHelper
     {
         private readonly string _connectionString;
 
+        private static readonly IReadOnlyDictionary<CustomTagDataType, string> TableNames = new Dictionary<CustomTagDataType, string>()
+            {
```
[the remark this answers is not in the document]
snehabandla: Sorry, I meant that the original mapping of this is relevant. Can it be a static mapping that is used in both test code and production code?

---
In reply to: [590904575](https://github.com/microsoft/dicom-server/pull/612#discussion_r590904575) [](ancestors = 590904575)

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
snehabandla: Can you add to the summary that this needs to exist for private tag?

---
In reply to: [590593679](https://github.com/microsoft/dicom-server/pull/612#discussion_r590593679) [](ancestors = 590593679)

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
snehabandla: Value still maps to the value for a given tag... potentially you should be more specific in tying this whole function to AT. It doesn't clearly tie to that for me. I would say the tag variable should be something like valueForATTag and the function name as well should specify AT.

---
In reply to: [591944229](https://github.com/microsoft/dicom-server/pull/612#discussion_r591944229) [](ancestors = 591944229)

src/Microsoft.Health.Dicom.Core/Features/CustomTag/IndexTag.cs
```diff
@@ -13,15 +14,15 @@ namespace Microsoft.Health.Dicom.Core.Features.CustomTag
     /// </summary>
     public class IndexTag
     {
-        public IndexTag(DicomTag tag, DicomVR vr, CustomTagLevel level, bool isCustomTag)
+        public IndexTag(DicomTag tag, DicomVR vr, CustomTagLevel level, CustomTagStoreEntry customTagStoreEntry)
```
snehabandla, line 17: >public [](start = 8, length = 6)

Why does this need to public? We have public static constructors right?

pengchen0692: probably could make it private, let me check.

test/Microsoft.Health.Dicom.Tests.Integration/Persistence/CustomTagStoreTests.STOW.cs
pengchen0692: sure

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
pengchen0692: may be GetAttributeTagValueAsLong?

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
pengchen0692: will put it in remark section

test/Microsoft.Health.Dicom.Tests.Integration/Persistence/SqlIndexDataStoreTestHelper.cs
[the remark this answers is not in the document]
pengchen0692: Hi Sneha, I didn't find where production code use it, could you point me out?
I will make it static

src/Microsoft.Health.Dicom.Core/Features/CustomTag/IndexTag.cs
pengchen0692: changed to private

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -45,25 +44,49 @@ public static T GetSingleValueOrDefault<T>(this DicomDataset dicomDataset, Dicom
             return dicomDataset.GetSingleValueOrDefault<T>(dicomTag, default);
         }
 
+        /// <summary>
+        /// Get value of Dicom AttributeTag to long?.
```
snehabandla, line 48: >AttributeTag [](start = 31, length = 12)

Can you specify that this is the VR?

pengchen0692: will do

test/Microsoft.Health.Dicom.Tests.Integration/Persistence/SqlIndexDataStoreTestHelper.cs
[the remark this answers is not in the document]
snehabandla: AddInstanceTableValuedParametersBuilder uses this in a hardcoded manner since it maps StringData -> InsertStringCustomTagTableTypeV1Row which is then mapped in the sql code to the CustomTagString table. It might be useful to encode it a more clear configuration based mapping.

---
In reply to: [591963573](https://github.com/microsoft/dicom-server/pull/612#discussion_r591963573) [](ancestors = 591963573)

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -714,6 +781,152 @@ AS
...
+    IF EXISTS (SELECT 1 FROM @stringCustomTags)
+    BEGIN      
+        MERGE INTO dbo.CustomTagString AS T
+        USING 
+        (
+            SELECT input.TagKey, input.TagValue, input.TagLevel 
+            FROM @stringCustomTags input
+            INNER JOIN dbo.CustomTag WITH (REPEATABLEREAD) 
+            ON dbo.CustomTag.TagKey = input.TagKey
+            AND dbo.CustomTag.TagStatus <> 2
```
snehabandla, line 795: ><> [](start = 40, length = 2)

Do you want custom tags that are being reindexed?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -714,6 +781,152 @@ AS
...
+            AND ISNULL(T.InstanceKey, @instanceKey) = @instanceKey
+        WHEN MATCHED THEN 
+            UPDATE SET T.Watermark = @newWatermark, T.TagValue = S.TagValue
+        WHEN NOT MATCHED THEN 
+            INSERT (TagKey, TagValue, StudyKey, SeriesKey, InstanceKey, Watermark)
+            VALUES(
+            S.TagKey,
+            S.TagValue,
+            @studyKey,
+            (CASE WHEN S.TagLevel <> 2 THEN @seriesKey ELSE NULL END),
```
snehabandla, line 809: ><> [](start = 34, length = 2)

Personally, I'd prefer a more consistent comparison - perhaps if tagLevel < 2 and then if tagLevel < 1?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -714,6 +781,152 @@ AS
...
+            (CASE WHEN S.TagLevel = 0 THEN @instanceKey ELSE NULL END),
+            @newWatermark);        
+    END
+
+    -- BigInt Key tags
+    IF EXISTS (SELECT 1 FROM @bigintCustomTags)
```
snehabandla, line 815: >bigintCustomTags [](start = 30, length = 16)

bigIntCustomTags

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -714,6 +781,152 @@ AS
     VALUES
         (@studyKey, @seriesKey, @instanceKey, @studyInstanceUid, @seriesInstanceUid, @sopInstanceUid, @newWatermark, @initialStatus, @currentDate, @currentDate)
 
+    -- Insert Custom Tags
+    -- String Key tags
+    IF EXISTS (SELECT 1 FROM @stringCustomTags)
```
snehabandla, line 786: >IF EXISTS (SELECT 1 FROM @stringCustomTags) [](start = 4, length = 43)

It would be useful in the future if a few comments on what this merge is achieving are added. For example, why filter on tagStatus? How is tagLevel being used to decide to store series/instance keys? etc.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
pengchen0692: yes, the reindex job would take care of instances in the past,  AddInstance would take new instance.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
pengchen0692: I think both should be fine. There are slightly different on the meaning:
* <>2 => TagLevel is not Study
* <2 => TagLevel is any level that below than Study.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
pengchen0692: will fix

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
pengchen0692: will add

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
smithago: I still don't get this. Are we missing a peace of information we need to add to PrivateTags then as part of AddTag?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/IndexTag.cs
smithago: It still has the same problem. 
CustomTagLevel level. Why do callers need to pass this? When you can get this from the customTagStoreEntry

Adding 2 constructors helps to check for nulls also

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMinimumValidation.cs
```diff
@@ -123,25 +83,40 @@ public static void ValidateDA(string value, string name)
...
+        {
+            return DateTime.TryParseExact(value, DateFormatDA, CultureInfo.InvariantCulture, DateTimeStyles.AdjustToUniversal, out dateTime);
+        }
+
         internal static void ValidateDS(string value, string name)
         {
             ValidateLength(value.Length, 0, 16, DicomVR.DS, name, value);
         }
 
         internal static void ValidateDT(string value, string name)
```
smithago, line 102: why are some internal? and some methods public here?

This class should be purely used to expose validate only

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
pengchen0692: Yes. We don't have privateCreator for private tags.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/IndexTag.cs
pengchen0692: will update.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -714,6 +781,158 @@ AS
...
+    BEGIN      
+        MERGE INTO dbo.CustomTagString AS T
+        USING 
+        (
+            SELECT input.TagKey, input.TagValue, input.TagLevel 
+            FROM @stringCustomTags input
+            INNER JOIN dbo.CustomTag WITH (REPEATABLEREAD) 
+            ON dbo.CustomTag.TagKey = input.TagKey
+            -- Not merge on custom tag which is being deleted.
+            AND dbo.CustomTag.TagStatus <> 2
```
smithago, line 797: Can you remind me all the TagStatus. I was thinking this would be only for Successfully provisioned/re-indexed tags. What does 0 and 1 mean?

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMinimumValidation.cs
pengchen0692: I found these method don't need to be public, and haven't' change old code. I'll try to change all to internal

pengchen0692: I can make the tryParseXX method private, but I have a question:
If the purpose of MinimumValidation is to make sure store works, why not use same set of Parse logic?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
pengchen0692: * 0: the tag is being reindexing
* 1: the tag has been added (reindex has been completed)
* 2: the tag is being deleted.
https://github.com/microsoft/dicom-server/blob/feature/customtags/src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagStatus.cs

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStore.cs
```diff
@@ -87,6 +83,50 @@ public async Task<long> CreateInstanceIndexAsync(DicomDataset instance, Cancella
             }
         }
 
+        private void PopulateAddInstanceCommand(SqlCommandWrapper wrapper, DicomDataset instance, IEnumerable<IndexTag> indexTags)
+        {
+            if (_schemaInformation.Current >= SchemaVersionConstants.SupportCustomTagSchemaVersion)
```
smithago, line 88: Is there a better pattern than this? Seems non-managable long term.

Think of it like this...

Each n.sql needs different store service.

SqlIndexDataStore : IIndexDataStore
virtual CreateInstanceIndexAsync()
static SqlIndexDataStore GetStoreForSchemaVersion(IEnumberable<IIndexDataStore>)
{
    switch(Schemaversion)
    ...
}

SqlIndexDataStore1 : SqlIndexDataStore
override CreateInstanceIndexAsync()
{ version 1 method here }

SqlIndexDataStore2 : SqlIndexDataStore1
override CreateInstanceIndexAsync()
{ version 2 method here}

Others methods are inherited and work as before.

Now in ServiceClass, you take Ienumerable of SqlIndexStore 
and call SqlIndexDataStore.GetStoreForSchemaVersion()
and use that to call the store APIs.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
smithago: Thanks Peng. Condition makes sense

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
smithago: Will it help if we get this instead

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMinimumValidation.cs
smithago: Parse could be different depending on the data store right? Like for SQL we convert it into a specific preformat, while validation is to just check if the string is a valid date

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStore.cs
pengchen0692: I have a different way to implement this, please take a look.

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
pengchen0692: > Will it help if we get this instead
Hi Smitha,
Sorry, not get you.
We need DicomTag with privateCreator to call DicomDataSet.GetSingleOrDefaultValue(DicomTag) to retrieve value. Since we don't require customer to input privateCreator when create customtag, we have to loop through dataset to locate by tag path (groupid + elementid).

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStore.cs
```diff
@@ -5,284 +5,99 @@
...
 
-        public async Task<int> IncrementDeletedInstanceRetryAsync(VersionedInstanceIdentifier versionedInstanceIdentifier, DateTimeOffset cleanupAfter, CancellationToken cancellationToken = default)
+        private async Task InitInnerDataStoreAsync()
         {
             await _sqlServerIndexSchema.EnsureInitialized();
 
-            using (SqlConnectionWrapper sqlConnectionWrapper = await _sqlConnectionFactoryWrapper.ObtainSqlConnectionWrapperAsync(cancellationToken, true))
-            using (SqlCommandWrapper sqlCommandWrapper = sqlConnectionWrapper.CreateSqlCommand())
+            if (_schemaInformation.Current == 1)
             {
```
snehabandla, line 95: Instead of if else conditions, can this be a static mapping?

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStore.cs
```diff
@@ -5,284 +5,99 @@
...
         private readonly SqlIndexSchema _sqlServerIndexSchema;
         private readonly SqlConnectionWrapperFactory _sqlConnectionFactoryWrapper;
+        private readonly SchemaInformation _schemaInformation;
+        private IIndexDataStore _innerDataStore;
 
         public SqlIndexDataStore(
             SqlIndexSchema indexSchema,
-            SqlConnectionWrapperFactory sqlConnectionWrapperFactory)
+            SqlConnectionWrapperFactory sqlConnectionWrapperFactory,
+            SchemaInformation schemaInformation)
```
snehabandla, line 32: The only concern I have is here is how little value this class actually provides- it's mostly a pass-through and seems to be more of an unnecessary initialization. The method Smitha had mentioned minimizes that overhead!

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStore.cs
pengchen0692: sure, will change

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStore.cs
pengchen0692: I didn't find a good way to fit the static method with direct injection.

Theoretically, the static method should be called inside SqlServer layer instead of Core layer, since Schema version is something Core should not know.

The static method is like this:
`  
public static SqlIndexDataStore GetIndexDataStore(SchemaVersion version, SqlConnectionWrapperFactory sqlConnectionWrapperFactory)
        {
            switch (version)
            {
                case SchemaVersion.V1:
                    return new SqlIndexDataStoreV1(sqlConnectionWrapperFactory);
                case SchemaVersion.V2:
                case SchemaVersion.Unknown:
                default:
                    return new SqlIndexDataStoreV2(sqlConnectionWrapperFactory);
            }
        }
`

This method should be called in DicomServerBuilderSqlServerRegistrationExtensions.AddSqlServer, but how to pass SchemaVersion  and SqlConnectionWrapperFactory is something I didn't figure out.

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStore.cs
```diff
@@ -5,284 +5,99 @@
...
-                try
-                {
-                    await sqlCommandWrapper.ExecuteScalarAsync(cancellationToken);
-                }
-                catch (SqlException ex)
-                {
-                    throw new DataStoreException(ex);
-                }
-            }
+            await InitInnerDataStoreAsync();
```
smithago, line 86: can this be done in a constructor instead of each method

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStore.cs
smithago: +1, you can keep a dictionary of version and class type mapping and use that

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStore.cs
```diff
@@ -5,284 +5,99 @@
...
     {
         private readonly SqlIndexSchema _sqlServerIndexSchema;
         private readonly SqlConnectionWrapperFactory _sqlConnectionFactoryWrapper;
+        private readonly SchemaInformation _schemaInformation;
+        private IIndexDataStore _innerDataStore;
 
         public SqlIndexDataStore(
             SqlIndexSchema indexSchema,
-            SqlConnectionWrapperFactory sqlConnectionWrapperFactory)
+            SqlConnectionWrapperFactory sqlConnectionWrapperFactory,
```
smithago, line 31: What is the lifetime of the SchemaInformation object? Just wanted to understand how a online schema upgrade would work from v1 to v2 without breaking POST requests

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStore.cs
pengchen0692: Would this to be too much heavy to put in constructor?

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStore.cs
pengchen0692: it's singleton as initalized here
https://github.com/microsoft/dicom-server/blob/85939f0a647eff28e71dfe151e5c7ceeac71faae/src/Microsoft.Health.Dicom.SqlServer/Registration/DicomServerBuilderSqlServerRegistrationExtensions.cs#L50

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStoreFactory.cs
```diff
@@ -0,0 +1,29 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Health.Dicom.Core.Features.Store;
+using Microsoft.Health.SqlServer.Features.Schema;
+
+namespace Microsoft.Health.Dicom.SqlServer.Features.Store
+{
+    internal class SqlIndexDataStoreFactory : IIndexDataStoreFactory
```
pengchen0692, line 13: add sumary

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStoreV1.cs
```diff
@@ -13,55 +13,50 @@
...
 using Microsoft.Health.Dicom.Core.Models;
 using Microsoft.Health.Dicom.SqlServer.Features.Schema.Model;
-using Microsoft.Health.Dicom.SqlServer.Features.Storage;
 using Microsoft.Health.SqlServer.Features.Client;
 using Microsoft.Health.SqlServer.Features.Storage;
 
 namespace Microsoft.Health.Dicom.SqlServer.Features.Store
 {
-    internal class SqlIndexDataStore : IIndexDataStore
+    internal class SqlIndexDataStoreV1 : ISqlIndexDataStore
```
pengchen0692, line 25: add summary

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStoreV1.cs
```diff
@@ -13,55 +13,50 @@
...
         {
-            EnsureArg.IsNotNull(indexSchema, nameof(indexSchema));
             EnsureArg.IsNotNull(sqlConnectionWrapperFactory, nameof(sqlConnectionWrapperFactory));
-
-            _sqlServerIndexSchema = indexSchema;
             _sqlConnectionFactoryWrapper = sqlConnectionWrapperFactory;
         }
 
-        public async Task<long> CreateInstanceIndexAsync(DicomDataset instance, CancellationToken cancellationToken)
+        public virtual int Version => 1;
```
pengchen0692, line 36: constant for 1

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStoreV2.cs
```diff
@@ -0,0 +1,92 @@
...
+using Microsoft.Health.Dicom.Core.Features.CustomTag;
+using Microsoft.Health.Dicom.Core.Models;
+using Microsoft.Health.Dicom.SqlServer.Features.CustomTag;
+using Microsoft.Health.Dicom.SqlServer.Features.Schema.Model;
+using Microsoft.Health.SqlServer.Features.Client;
+using Microsoft.Health.SqlServer.Features.Storage;
+
+namespace Microsoft.Health.Dicom.SqlServer.Features.Store
+{
+    internal class SqlIndexDataStoreV2 : SqlIndexDataStoreV1
```
pengchen0692, line 25: summary

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStoreV2.cs
```diff
@@ -0,0 +1,92 @@
...
+
+        public SqlIndexDataStoreV2(
+            SqlConnectionWrapperFactory sqlConnectionWrapperFactory)
+            : base(sqlConnectionWrapperFactory)
+        {
+            EnsureArg.IsNotNull(sqlConnectionWrapperFactory, nameof(sqlConnectionWrapperFactory));
+            _sqlConnectionFactoryWrapper = sqlConnectionWrapperFactory;
+        }
+
+        public override int Version => 2;
```
pengchen0692, line 37: constant for 2

src/Microsoft.Health.Dicom.SqlServer/Registration/DicomServerBuilderSqlServerRegistrationExtensions.cs
```diff
@@ -52,11 +51,11 @@ public static class DicomServerBuilderSqlServerRegistrationExtensions
...
                 .AsSelf();
 
-            services.Add<SqlIndexSchema>()
-                .Singleton()
-                .AsSelf();
-
-            services.Add<SqlIndexDataStore>()
+            services.AddScoped<ISqlIndexDataStore, SqlIndexDataStoreV1>();
+            services.AddScoped<ISqlIndexDataStore, SqlIndexDataStoreV2>();
+            services.AddScoped<IIndexDataStore>(
```
pengchen0692, line 56: nit: format it better

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
pengchen0692: These code will be cleaned up by https://microsofthealth.visualstudio.com/Health/_workitems/edit/80419

smithago (approved):

PR #631, opened 2021-03-09
Update to use Shared AuthZ code

brandonpollett:
## Description
This PR updates the dicom server to use the shared authz code from the `healthcare-shared-components` repo.

## Related issues
Addresses [AB#79713](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/79713).

## Testing


nuget.config
```diff
@@ -9,4 +9,4 @@
   <disabledPackageSources>
     <clear />
   </disabledPackageSources>
-</configuration>
+</configuration>
```
wsugarman, line 12: nit: I have been typically adding newlines to each file I touch, and I think either VS or VS Code is also of the same persuasion. I think this is an old POSIX behavior which didn't process text that didn't end with a newline lol

src/Microsoft.Health.Dicom.Api/Modules/SecurityModule.cs
```diff
@@ -71,24 +72,25 @@ public void Load(IServiceCollection services)
...
                 else
                 {
-                    services.AddSingleton<IDicomAuthorizationService, DisabledDicomAuthorizationService>();
+                    services.AddSingleton<IAuthorizationService<DataActions>, DisabledAuthorizationService<DataActions>>();
                 }
             }
             else
             {
-                services.AddSingleton<IDicomAuthorizationService, DisabledDicomAuthorizationService>();
+                services.AddSingleton<IAuthorizationService<DataActions>, DisabledAuthorizationService<DataActions>>();
```
wsugarman, line 87: Ah, the existing versions of the shared packages have these types?

src/Microsoft.Health.Dicom.Core/Features/Security/DataActions.cs
```diff
@@ -13,10 +13,16 @@ namespace Microsoft.Health.Dicom.Core.Features.Security
     public enum DataActions : ulong
 #pragma warning restore CA1028 // Enum Storage should be Int32
     {
+        [EnumMember(Value = "none")]
```
wsugarman, line 16: These values are the defaults, aren't they? Although I suppose casing is property of some serializers settings, as most things in .NET will be PascalCase

wsugarman: Oh, I see where this comes into play in *DicomRoleLoader.cs*. We could also assume the name of the action is the name of the value unless `EnumMember` is explicitly annotated, right?

src/Microsoft.Health.Dicom.Core/Features/Security/DicomRoleLoader.cs
```diff
@@ -0,0 +1,53 @@
...
+                var enumMemberAttribute = ((EnumMemberAttribute[])enumType.GetField(name).GetCustomAttributes(typeof(EnumMemberAttribute), true)).Single();
+                _dataActionsMap.TryAdd(enumMemberAttribute.Value, Enum.Parse<DataActions>(name));
+            }
+        }
+
+        protected override Role<DataActions> RoleContractToRole(RoleContract roleContract)
+        {
+            EnsureArg.IsNotNull(roleContract, nameof(roleContract));
+
+            DataActions dataActions = roleContract.DataActions.Aggregate(default(DataActions), (acc, a) => acc | ToEnum(a));
```
wsugarman, line 37: nit: `DataActions.None` may be clearer as the initial aggregate state instead of `default(DataActions)`

wsugarman (approved):

src/Microsoft.Health.Dicom.Api/Modules/SecurityModule.cs
brandonpollett: Yep, these were abstracted away.

src/Microsoft.Health.Dicom.Core/Features/Security/DataActions.cs
brandonpollett: The casing matters, but generally yes they're basically the same. The actual value that's allowed in the json is from the `roles.schema.json`.

PR #632, opened 2021-03-10
Fix Build Warnings for Newer .NET SDK

### LINKED ISSUE #614: Build fails with the update as of Mar 8th commit in main
**Describe the bug**
local build of the OSS repo is failing with the following errors  - 
image

Severity	Code	Description	Project	File	Line	Suppression State	Detail Description
Error	CA1802	Field 'DicomServerBlobConfigurationSectionName' is declared as 'readonly' but is initialized with a constant value. Mark this field as 'const' instead.	Microsoft.Health.Dicom.Blob	{local path|\dicom-server\src\Microsoft.Health.Dicom.Blob\Registration\DicomServerBuilderBlobRegistrationExtensions.cs	23	Active	A field is declared static and read-only (Shared and ReadOnly in Visual Basic), and is initialized by using a value that is computable at compile time. Because the value that is assigned to the targeted field is computable at compile time, change the declaration to a const (Const in Visual Basic) field so that the value is computed at compile time instead of at run?time.

image

Your help will be appreciated. Thank you.

**To Reproduce**
1. Update the repo the commit point - bd33cf2 - Refactor Project Files (#606)
2. Build the solution using VS 16.9.0
3. Observer all the errors as outlined in the first screenshot

**Expected behavior**
Solution Build resulting in no errors 

**Actual behavior**
See the attached screenshot above.

wsugarman:
## Description
Address build warnings from later .NET presumably due to upcoming changes for constant interpolated strings in C# 10

## Related issues
Fixes #614 

## Testing
PR pipeline and local build


wsugarman:
I purposefully avoided the use of C# 10's preview features, as we currently use the target framework's corresponding language version

brandonpollett (approved):

PR #646, opened 2021-03-17
Enable End to end tests for ExtendQueryTag (Locally)

pengchen0692:
## Description
Enable End to end tests for ExtendQueryTag

## Related issues
Addresses [AB#80131](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/80131)

## Testing
locally passed.

## Notes
Since the feature is not enabled yet, we still need to skip end to end test.  The best we can do now is to make sure it passes locally.


src/Microsoft.Health.Dicom.Client/Models/ExtendedQueryTagLevel.cs
```diff
@@ -6,22 +6,22 @@
 namespace Microsoft.Health.Dicom.Client.Models
 {
     /// <summary>
-    /// Level of a custom tag.
+    /// Level of a queryable tag.
     /// </summary>
-    public enum CustomTagLevel
+    public enum ExtendedQueryTagLevel
```
wsugarman, line 11: Is this `enum` really specific to extended query tags? Does such an `enum` already exist for DICOM itself? Otherwise, perhaps this is just `TagLevel`?

src/Microsoft.Health.Dicom.Client/Models/ExtendedQueryTagStatus.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        /// <summary>
+        /// The query tag has been added to system.
+        /// </summary>
+        Added = 1,
+
+        /// <summary>
+        /// The query tag is being deindexed.
+        /// </summary>
+        Deindexing = 2,
```
wsugarman, line 26: On 2nd though, is this exposing too much of our internals. Should this just be like `Adding`, `Ready`, and Removing`?

test/Microsoft.Health.Dicom.Web.Tests.E2E/Comparers/ExtendedQueryTagEqualityComparer.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            if (x == null || y == null)
+            {
+                return x == y;
+            }
+
+            return x.Path.Equals(y.Path, StringComparison.OrdinalIgnoreCase)
+                && string.Equals(x.VR, y.VR, StringComparison.OrdinalIgnoreCase)
+                && x.Level == y.Level
+                && (IgnoreStatus ? true : x.Status == y.Status);
```
wsugarman, line 32: nit: `(IgnoreStatus || x.Status == y.Status)`

src/Microsoft.Health.Dicom.Client/DicomWebClient.cs
```diff
@@ -29,7 +29,7 @@ public class DicomWebClient : IDicomWebClient
     {
         private const string TransferSyntaxHeaderName = "transfer-syntax";
 
-        private static readonly Uri BaseCustomTagUri = new Uri("/tags", UriKind.Relative);
+        private const string BaseExtendedQueryTagUri = "/tags";
```
wsugarman, line 32: Let's keep this as a `Uri` unless there's a reason we need it as a `string`. There are `Uri` ctors for combining paths

wsugarman:
As a general note, I know we're referring to the feature as "Extended Query Tags," but in the context of methods to add/remove or even the data model, should we simply use `QueryTag`? Or is it better to say `ExtendedQueryTag` because there is a set of query tags the user cannot modify?

wsugarman (approved):

src/Microsoft.Health.Dicom.Client/Models/ExtendedQueryTagLevel.cs
pengchen0692: This enum should be applied to any DicomTag, so TagLevel could be a better name. 
Unfortunately, there is no existing enum (AFAIK)
will fix it.

test/Microsoft.Health.Dicom.Web.Tests.E2E/Comparers/ExtendedQueryTagEqualityComparer.cs
pengchen0692: will fix it.

src/Microsoft.Health.Dicom.Client/Models/ExtendedQueryTagStatus.cs
pengchen0692: will fix it.

src/Microsoft.Health.Dicom.Client/DicomWebClient.cs
pengchen0692: The original idea with Uri instead of string is that we can build other Uris by new Uri(Uri, relativePath), but I find it requires the Uri to be absolute while we always relative.

e.g: 
for delete tag 00100121, uri is "tags/00100121", i, new Uri(BaseExtendedQueryTagUri, "00100121") throws exception.

pengchen0692:
> As a general note, I know we're referring to the feature as "Extended Query Tags," but in the context of methods to add/remove or even the data model, should we simply use `QueryTag`? Or is it better to say `ExtendedQueryTag` because there is a set of query tags the user cannot modify?

My understanding is there are Core Query tag(https://github.com/microsoft/dicom-server/blob/main/docs/resources/conformance-statement.md#searchable-attributes) and Extended query Tag, user cannot modify the Core ones. From this perspective, ExtendedQueryTag is more accurate, but I feel it's a  really long name.

src/Microsoft.Health.Dicom.Client/Models/ExtendedQueryTagLevel.cs
pengchen0692: may be QueryTagLevel is more accurate? Since it  only applies to QIDO

src/Microsoft.Health.Dicom.Client/Models/ExtendedQueryTagStatus.cs
pengchen0692: may be deleting instead of Removing to keep consistency with API name DeleteExtendedQueryTag?

wsugarman (approved):

PR #647, opened 2021-03-17
Update AddCustomTag API to support PrivateCreator

pengchen0692:
## Description
Update AddCustomTag API to support PrivateCreator

## Related issues
Addresses [AB#80425](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/80425)


src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
```diff
@@ -21,20 +20,17 @@ public CustomTagEntry()
             // used as customer input, constructor without parameters is required.
         }
 
-        public CustomTagEntry(string path, string vR, CustomTagLevel level, CustomTagStatus status)
```
pengchen0692, line 24: Not sure if it's good idea to remove this constructor. My though is that since we already have setter on  all properties, do we still this?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
```diff
@@ -50,6 +46,11 @@ public CustomTagEntry(CustomTagStoreEntry storedEntry)
         /// </summary>
         public string VR { get; set; }
 
+        /// <summary>
+        /// Private Creator of this Tag.
```
pengchen0692, line 50: update comments

src/Microsoft.Health.Dicom.Tests.Common/Comparers/CustomTagEntryEqualityComparer.cs
```diff
@@ -0,0 +1,33 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.Health.Dicom.Core.Features.CustomTag;
+
+namespace Microsoft.Health.Dicom.Tests.Common.Comparers
+{
+    public class CustomTagEntryEqualityComparer : IEqualityComparer<CustomTagEntry>
+    {
+        public bool Equals(CustomTagEntry x, CustomTagEntry y)
```
pengchen0692, line 15: should also compare PrivateCreator

src/Microsoft.Health.Dicom.Core.UnitTests/Features/CustomTag/GetCustomTagsServiceTests.cs
```diff
@@ -50,9 +50,9 @@ public async Task GivenRequestForAllTags_WhenMultipleTagsAreStored_ThenCustomTag
             _customTagStore.GetCustomTagsAsync(default).Returns(storedEntries);
             GetAllCustomTagsResponse response = await _getCustomTagsService.GetAllCustomTagsAsync();
```
pengchen0692, line 52: change tag2 to private tag

src/Microsoft.Health.Dicom.Core/DicomCoreResource.resx
```diff
@@ -335,4 +335,12 @@ The first part date {1} should be lesser than or equal to the second part date {
     <value>Value length is not {0}.</value>
     <comment>{0} required length.</comment>
   </data>
+  <data name="InvalidPrivateCreator" xml:space="preserve">
```
smithago, line 338: Can we be more specific here?
PrivateCreatorNotNull
Private creator is a string set to represent the creator of a private tag. Since the tag being submitted is a standard tag private creator is expected to be empty.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntryValidator.cs
```diff
@@ -90,23 +88,35 @@ private void ValidateCustomTagEntry(CustomTagEntry tagEntry)
             if (QueryLimit.AllInstancesTags.Contains(tag))
             {
                 throw new CustomTagEntryValidationException(
-                   string.Format(CultureInfo.InvariantCulture, DicomCoreResource.InvalidCustomTag, tag.DictionaryEntry.Name));
+                   string.Format(CultureInfo.InvariantCulture, DicomCoreResource.InvalidCustomTag, tagEntry.Path));
             }
 
             if (tag.IsPrivate)
             {
+                if (string.IsNullOrWhiteSpace(tagEntry.PrivateCreator))
```
smithago, line 96: can you move this logic to a function ValidatePrivateCreator(Tag) . In there you may have to do length check along with current null or not null check

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
```diff
@@ -169,6 +169,7 @@ CREATE TYPE dbo.AddCustomTagsInputTableType_1 AS TABLE
 (
     TagPath                    VARCHAR(64),  -- Custom Tag Path. Each custom tag take 8 bytes, support upto 8 levels, no delimeter between each level.
```
smithago, line 170: can we make the columns NOT NULL, tagPath, TagVR, TagLevel

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
```diff
@@ -169,6 +169,7 @@ CREATE TYPE dbo.AddCustomTagsInputTableType_1 AS TABLE
...
     TagVR                      VARCHAR(2),  -- Custom Tag VR.
+    TagPrivateCreator          NVARCHAR(64),  -- Custom Tag Private Creator, only valid for private tag.
```
smithago, line 172: you should check the length during validation or do all the checks done by VR.LO. Because Private data element is of type LO. Check for control character and all

src/Microsoft.Health.Dicom.Core/Extensions/CustomTagStoreEntryExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// Extension methods for <see cref="CustomTagStoreEntry"/>.
+    /// </summary>
+    public static class CustomTagStoreEntryExtensions
+    {
+        /// <summary>
+        /// Create  <see cref="CustomTagEntry"/> from  <see cref="CustomTagStoreEntry"/>.
+        /// </summary>
+        /// <param name="storeEntry">The custom tag store entry.</param>
+        /// <returns>The custom tag entry.</returns>
+        public static CustomTagEntry ToCustomTagEntry(this CustomTagStoreEntry storeEntry)
```
smithago, line 21: Why not add this method to CustomTagStoreEntry class?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntryValidator.cs
smithago: All LO VR checks

src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/SqlCustomTagStore.cs
```diff
@@ -115,7 +115,7 @@ public async Task<IReadOnlyList<CustomTagStoreEntry>> GetCustomTagsAsync(string
 
         private static AddCustomTagsInputTableTypeV1Row ToAddCustomTagsInputTableTypeV1Row(CustomTagEntry entry)
         {
-            return new AddCustomTagsInputTableTypeV1Row(entry.Path, entry.VR, (byte)entry.Level);
+            return new AddCustomTagsInputTableTypeV1Row(entry.Path, entry.VR, entry.PrivateCreator, (byte)entry.Level);
```
smithago, line 118: We started this whole conversation because of the mapping you where doing from IndexTag to DicomTag, I did not see that method removed.

src/Microsoft.Health.Dicom.Core/DicomCoreResource.resx
pengchen0692: Will fix it.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntryValidator.cs
pengchen0692: will fix it.

src/Microsoft.Health.Dicom.Core/Extensions/CustomTagStoreEntryExtensions.cs
pengchen0692: will fix it.

src/Microsoft.Health.Dicom.SqlServer/Features/CustomTag/SqlCustomTagStore.cs
pengchen0692: I split the whole user story to 3  tasks, that method will be removed as I update STOW
image

src/Microsoft.Health.Dicom.Core/DicomCoreResource.resx
pengchen0692: how about PrivateCreatorNotEmpty?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
```diff
@@ -3,42 +3,13 @@
...
-using System;
-using EnsureThat;
-
 namespace Microsoft.Health.Dicom.Core.Features.CustomTag
 {
     /// <summary>
     /// External representation of a custom tag entry.
     /// </summary>
-    public class CustomTagEntry : IEquatable<CustomTagEntry>
+    public class CustomTagEntry
```
wsugarman, line 11: This is all now `ExtendedQueryTag` in the other PR, right? So the net change is really just the addition of `PrivateCreator`?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
wsugarman: There are now three options in C# now for setting data:
- A constructor/ctor
    - Can ensure multiple values are set in the resulting object
    - Can perform complex multi-input validation
    - Can be inherited to enforce/help creation patterns
    - Can enforce readonly semantics
- `public` setters (with optional validation)
    - Useful for "record" types that simply holds data: any combination of the fields/properties can have non-null data
    - Unfortunately the object becomes mutable as the setters must be `public`
- New "`init`-only" setters in C# 9
    - Very much the same as setters, but now the object can be immutable!

In this scenario, if we want to ensure all of the fields (or some subset) must be specified, then we should use a ctor

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntryValidator.cs
```diff
@@ -134,6 +134,32 @@ private void ValidateCustomTagEntry(CustomTagEntry tagEntry)
             }
         }
 
+        private static void ValidatePrivateCreator(DicomTag tag, string privateCreator, string originalTagPath)
+        {
+            if (tag.IsPrivate)
```
wsugarman, line 139: Are these checks we could have performed earlier in the class ctor?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagStoreEntry.cs
```diff
@@ -52,9 +52,18 @@ public CustomTagStoreEntry(int key, string path, string vr, string privateCreato
...
+        /// <returns>The custom tag entry.</returns>
+        public CustomTagEntry ToCustomTagEntry()
+        {
+            return new CustomTagEntry() { Path = Path, VR = VR, PrivateCreator = PrivateCreator, Level = Level, Status = Status };
+        }
+
         public override string ToString()
         {
-            return $"Key: {Key}, Path: {Path}, VR:{VR}, Level:{Level}, Status:{Status}";
+            return $"Key: {Key}, Path: {Path}, VR:{VR}, PrivateCreator:{PrivateCreator}, Level:{Level}, Status:{Status}";
```
wsugarman, line 66: Where is this `ToString()` used? Is it simply for debugging?

src/Microsoft.Health.Dicom.Tests.Common/Comparers/CustomTagEntryEqualityComparer.cs
```diff
@@ -0,0 +1,34 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using Microsoft.Health.Dicom.Core.Features.CustomTag;
+
+namespace Microsoft.Health.Dicom.Tests.Common.Comparers
+{
+    public class CustomTagEntryEqualityComparer : IEqualityComparer<CustomTagEntry>
```
wsugarman, line 13: A typical pattern for comparers is to provide a static instance for people to use instead of each class keeping their own cached instance. Eg.

```csharp
// Could also be called "Instance?"
public static readonly CustomTagEntryEqualityComparer Default { get; } = new CustomTagEntryEqualityComparer();
```

src/Microsoft.Health.Dicom.Tests.Common/Comparers/CustomTagEntryEqualityComparer.cs
```diff
@@ -0,0 +1,34 @@
...
+            }
+
+            return x.Path.Equals(y.Path, StringComparison.OrdinalIgnoreCase)
+                && string.Equals(x.VR, y.VR, StringComparison.OrdinalIgnoreCase)
+                && x.PrivateCreator == y.PrivateCreator
+                && x.Level == y.Level
+                && x.Status == y.Status;
+        }
+
+        public int GetHashCode([DisallowNull] CustomTagEntry obj)
```
wsugarman, line 29: Unless the `nullable` context is enable, this attribute doesn't provide any analysis. Furthermore, it doesn't prevent `null` values from being passed; it is simply to convey what _should_ happen. So you'll need a `null` check still!

src/Microsoft.Health.Dicom.Tests.Common/Extensions/DicomTagExtensions.cs
```diff
@@ -11,9 +11,9 @@ namespace Microsoft.Health.Dicom.Tests.Common.Extensions
 {
     public static class DicomTagExtensions
     {
-        public static CustomTagEntry BuildCustomTagEntry(this DicomTag tag, string vr = null, CustomTagLevel level = CustomTagLevel.Series, CustomTagStatus status = CustomTagStatus.Ready)
+        public static CustomTagEntry BuildCustomTagEntry(this DicomTag tag, string vr = null, string privateCreator = null, CustomTagLevel level = CustomTagLevel.Series, CustomTagStatus status = CustomTagStatus.Ready)
         {
-            return new CustomTagEntry(tag.GetPath(), vr ?? tag.GetDefaultVR()?.Code, level, status);
+            return new CustomTagEntry() { Path = tag.GetPath(), VR = vr ?? tag.GetDefaultVR()?.Code, PrivateCreator = privateCreator, Level = level, Status = status };
```
wsugarman, line 16: nit: remove `()`

src/Microsoft.Health.Dicom.Tests.Common/Comparers/CustomTagEntryEqualityComparer.cs
pengchen0692: will fix it.

src/Microsoft.Health.Dicom.Core/Extensions/CustomTagEntryExtensions.cs
```diff
@@ -48,7 +48,7 @@ public static CustomTagEntry Normalize(this CustomTagEntry customTagEntry, Custo
 
             vr = vr.ToUpperInvariant();
 
-            return new CustomTagEntry(path, vr, customTagEntry.Level, status);
+            return new CustomTagEntry() { Path = path, VR = vr, PrivateCreator = customTagEntry.PrivateCreator, Level = customTagEntry.Level, Status = status };
```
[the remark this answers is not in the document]
pengchen0692: Yeah, you mentioned this before, will fix it.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
pengchen0692: * There are 2 `CustomTagEntry`, one is server side (https://github.com/microsoft/dicom-server/blob/feature/customtags/src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs), the other is client side (https://github.com/microsoft/dicom-server/blob/feature/customtags/src/Microsoft.Health.Dicom.Client/Models/ExtendedQueryTag.cs).  There is user story https://microsofthealth.visualstudio.com/Health/_workitems/edit/80286 which Sneha is working on for renaming everything, so I didn't rename here.
*  Yes, the change is just addtional of `PrivateCreator`

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntry.cs
pengchen0692: CustomTagEntry is used for API input/output.
* As input, no property is mandatory, and we do validation in https://github.com/microsoft/dicom-server/blob/feature/customtags/src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntryValidator.cs later.
* As output, it's simply mapping to database record, no validation is supposed to be there.
Based on these facts, I guess we can remove the constructor, am I right?

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagEntryValidator.cs
pengchen0692: CustomTagEntry is used in this way:

HttpRequest Body => Deserialized by Asp.net Core to CustomTagEntry list =>Passed to AddCustomTags controller which does further validation.

So only the non-parameter constructor is used when deserialization.

src/Microsoft.Health.Dicom.Core/Features/CustomTag/CustomTagStoreEntry.cs
pengchen0692: It's used here https://github.com/microsoft/dicom-server/blob/1addc3c0297b7defdde291a1dbbdd8986101038e/src/Microsoft.Health.Dicom.Api/Controllers/CustomTagController.cs#L47 for easy diagnosis from log.

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMinimumValidation.cs
```diff
@@ -81,21 +81,38 @@ internal static void ValidateIS(string value, string name)
         }
 
         internal static void ValidateLO(string value, string name)
+        {
+            ValidationError error = ValidateLO(value);
```
wsugarman, line 85: Why don't we throw directly in the helper `ValidateLO` method?

src/Microsoft.Health.Dicom.Core/Features/Validation/ValidationError.cs
```diff
@@ -0,0 +1,28 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Dicom.Core.Features.Validation
+{
+    /// <summary>
+    /// Different type of  Validation Errors.
+    /// </summary>
+    public enum ValidationError
```
wsugarman, line 11: Does this need to be `public`?

src/Microsoft.Health.Dicom.Tests.Common/Comparers/CustomTagEntryEqualityComparer.cs
wsugarman: `public` fields are not recommended, even `static` ones. One of the reasons is that it's harder to make changes to the API going forward, as properties are glorified methods whose implementation can change.

For example, `Comparer<T>.Default` uses this pattern: https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.comparer-1.default?view=net-5.0#System_Collections_Generic_Comparer_1_Default

wsugarman (approved):

src/Microsoft.Health.Dicom.Core/Features/Validation/ValidationError.cs
pengchen0692: will change to internal.

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMinimumValidation.cs
pengchen0692: ValidateLO is used on 2 places:
1. STOW: when customer store dicom instance, it validate  DicomElement of VR LO.
2. AddCustomTag: when custom add custom tag, it validate privateCreator (whose VR is LO) 

Those logic is same except error message, I didn't figure out a simpler way to do this.

smithago (approved):

PR #656, opened 2021-03-19
Add web app to test STOW API with random instances

snehabandla:

### DIFF
--- a/tools/scale-testing/StowWebApp/Program.cs
+++ b/tools/scale-testing/StowWebApp/Program.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.Net.Http;
+using System.Threading.Tasks;
 using Common;
 using Dicom;
 using Microsoft.Health.Dicom.Client;
@@ -16,7 +17,7 @@ namespace StowWebApp
         private static Random rand;
         private static IDicomWebClient client;
 
-        public static async Task Main()
+        public static void Main()
         {
             using var httpClient = new HttpClient
             {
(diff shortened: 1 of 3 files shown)


tools/scale-testing/StowWebApp/Program.cs
```diff
@@ -0,0 +1,111 @@
...
+                }
+            }
+        }
+
+        private static void SetupDicomWebClient(HttpClient httpClient)
+        {
+            client = new DicomWebClient(httpClient);
+        }
+
+        private static List<(string, string, string)> InstanceGenerator()
```
pengchen0692, line 88: InstanceGenerator looks like a class name, how about GenerateInstanceIdentifier, or BuildInstanceIdentifier? Since it's generating identifier instead of actual instance.

tools/scale-testing/StowWebApp/Program.cs
```diff
@@ -0,0 +1,111 @@
...
+                                columns: 100,
+                                frames: 50);
+                    try
+                    {
+                        DicomWebResponse<DicomDataset> response = await client.StoreAsync(new List<DicomFile>() { dicomFile });
+
+                        int statusCode = (int)response.StatusCode;
+                        if (statusCode != 409 && statusCode < 200 && statusCode > 299)
+                        {
+                            throw new HttpRequestException("Stow operation failed", null, response.StatusCode);
```
pengchen0692, line 70: StoreAsync throws exception when fail, so I suppose it never goes to here?
https://github.com/microsoft/dicom-server/blob/109c3004ad643ed8df98065deb881f050960071d/src/Microsoft.Health.Dicom.Client/DicomWebClient.cs#L266

PR #679, opened 2021-03-30
ExtendedQueryTag services respect config setting

pengchen0692:
## Description
Update extended query tag services (add/remove/get) to respect feature  settings: EnableExtendedQueryTags

## Related issues
Addresses [AB#80934](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/80934)

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Modules/ServiceModule.cs
+++ b/src/Microsoft.Health.Dicom.Core/Modules/ServiceModule.cs
@@ -5,7 +5,6 @@
 
 using EnsureThat;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Health.Dicom.Core.Configs;
 using Microsoft.Health.Dicom.Core.Features.ChangeFeed;
 using Microsoft.Health.Dicom.Core.Features.Common;
 using Microsoft.Health.Dicom.Core.Features.Delete;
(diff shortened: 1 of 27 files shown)


src/Microsoft.Health.Dicom.Core/Modules/ServiceModule.cs
```diff
@@ -142,28 +133,25 @@ public void Load(IServiceCollection services)
                    .AsSelf()
                    .AsImplementedInterfaces();
 
-            if (_featureConfiguration.EnableExtendedQueryTags)
-            {
-                services.Add<ExtendedQueryTagEntryValidator>()
-                    .Singleton()
-                    .AsSelf()
-                    .AsImplementedInterfaces();
+            services.Add<ExtendedQueryTagEntryValidator>()
```
wsugarman, line 136: Why are we always adding these services if they're functionality is dependent on the config? Is it to provide better errors to the user if they try to use a feature that's not enabled?

wsugarman (approved):

src/Microsoft.Health.Dicom.Core/Modules/ServiceModule.cs
pengchen0692: if services are not created, customer would get internal server failure.

pengchen0692:
As discussed with team, should have gating in controller level

PR #701, opened 2021-04-06
Fix bug: Invalid private  TagPath "Unknown" is accepted

pengchen0692:
## Description
Fix bug: Invalid private  TagPath "Unknown" is accepted

## Related issues
Addresses [AB#80766](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/80766)

## Testing
All automation tests pass

## Notes
The reason why EnsureDefaultDictionariesLoaded(false) doesnt' work:
* that method is widely used. One popular is DicomTag.Parse. So it's very hard to make sure it's always initialized without private tags.

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
+++ b/src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
@@ -27,6 +27,9 @@ namespace Microsoft.Health.Dicom.Core.Features.Common
                 return false;
             }
 
+            // DicomDictionary.Default contains private tags by default which we don't need       
+            DicomDictionary.EnsureDefaultDictionariesLoaded(false);
+
             // Try Keyword match, returns null if not found
             DicomTag dicomTag = DicomDictionary.Default[dicomTagPath];
 
--- a/src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
+++ b/src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
@@ -27,11 +27,8 @@ namespace Microsoft.Health.Dicom.Core.Features.Common
                 return false;
             }
 
-            // DicomDictionary.Default contains private tags by default which we don't need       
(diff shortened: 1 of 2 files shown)


wsugarman (approved):

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
```diff
@@ -39,6 +39,20 @@ public bool TryParse(string dicomTagPath, out DicomTag[] dicomTags, bool support
             return dicomTag != null;
         }
 
+        private static DicomTag ParseDicomTagKeyword(string keyword)
+        {
+            // Try Keyword match, returns null if not found
+            DicomTag dicomTag = DicomDictionary.Default[keyword];
+
+            if (dicomTag == null)
```
wsugarman, line 47: nit: `return dicomTag == null || dicomTag.IsPrivate ? null : dicomTag`

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
```diff
@@ -39,6 +39,20 @@ public bool TryParse(string dicomTagPath, out DicomTag[] dicomTags, bool support
             return dicomTag != null;
         }
 
+        private static DicomTag ParseDicomTagKeyword(string keyword)
```
wsugarman, line 42: Should the method indicate somehow that only public tags are allowed? Like `ParsePublicDicomTagKeyword`?

wsugarman (approved):

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
pengchen0692: sure, might be ParseStandardDicomTagKeyword as normally called Standard instead of Public http://dicom.nema.org/dicom/2013/output/chtml/part05/sect_7.8.html

pengchen0692:
The change has been merged with another PR

PR #708, opened 2021-04-07
When VR doesn't match, skip the dicom element

pengchen0692:
## Description
When element has different VR than expected, skip both validation and store.

## Related issues
Addresses [AB#81009](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/81009)

## Testing
* All automated tests pass.
* Manually verified bug not repro.

### DIFF
--- a/src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/AddInstanceTableValuedParametersBuilder.cs
+++ b/src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/AddInstanceTableValuedParametersBuilder.cs
@@ -19,7 +19,7 @@ namespace Microsoft.Health.Dicom.SqlServer.Features.ExtendedQueryTag
     /// </summary>
     internal static class AddInstanceTableValuedParametersBuilder
     {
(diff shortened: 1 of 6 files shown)


src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/AddInstanceTableValuedParametersBuilder.cs
```diff
@@ -83,7 +83,7 @@ internal static class AddInstanceTableValuedParametersBuilder
 
         private static void AddPersonNameRow(DicomDataset instance, List<InsertPersonNameExtendedQueryTagTableTypeV1Row> personNamRows, QueryTag queryTag)
         {
-            string personNameVal = instance.GetSingleValueOrDefault<string>(queryTag.Tag);
+            string personNameVal = instance.GetSingleValueOrDefault<string>(queryTag.Tag, expectedVR: queryTag.VR);
```
wsugarman, line 86: How could the VR be an unexpected value? Don't we only call these functions based on VR in `Build`? So isn't it already asserted that the VR type for the tag here is `PersonNameData` because we checked it beforehand?

wsugarman: Discussed offline

wsugarman (approved):

PR #709, opened 2021-04-07
Fix bug 81015 - 500 error when adding a new instance

pengchen0692:
## Description
Fix bug 81015 - 500 error when adding a new instance

## Related issues
Addresses [AB#81015](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/81015)

## Rootcause
We should do null check when call vr.ToUppper()

### DIFF
--- a/src/Microsoft.Health.Dicom.Core.UnitTests/Extensions/ExtendedQueryTagEntryExtensionsTests.cs
+++ b/src/Microsoft.Health.Dicom.Core.UnitTests/Extensions/ExtendedQueryTagEntryExtensionsTests.cs
@@ -36,6 +36,15 @@ namespace Microsoft.Health.Dicom.Core.UnitTests.Extensions
             Assert.Null(normalized.PrivateCreator);
         }
 
+        [Fact]
+
+        public void GivenPrivateIdentificationCodeTagWithoutVR_WhenNormalizing_ThenVRShouldBeFilled()
+        {
+            ExtendedQueryTagEntry entry = CreateExtendedQueryTagEntry("22010010", null, null, QueryTagLevel.Instance, ExtendedQueryTagStatus.Ready);
+            ExtendedQueryTagEntry normalized = entry.Normalize(ExtendedQueryTagStatus.Adding);
+            Assert.Equal(DicomVRCode.LO, normalized.VR);
+        }
+
         [Theory]
         [InlineData(null)]
         [InlineData("")]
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Dicom.Core.UnitTests/Extensions/ExtendedQueryTagEntryExtensionsTests.cs
```diff
@@ -87,10 +96,19 @@ public void GivenStandardTagAsKeyword_WhenNormalizing_ThenVRShouldBeFilled()
             Assert.Equal(normalized.Path, expectedPath);
         }
 
+        [Fact]
+
+        public void GivenInvalidTagWithoutVR_WhenNormalizing_ThenShouldNotThrowException()
+        {
+            // track bug  https://microsofthealth.visualstudio.com/Health/_workitems/edit/81015
+            ExtendedQueryTagEntry entry = CreateExtendedQueryTagEntry(path: "00111011", null, null, QueryTagLevel.Series, ExtendedQueryTagStatus.Ready);
+            entry.Normalize(ExtendedQueryTagStatus.Adding);
```
wsugarman, line 105: nit: there should be an assertion the VR is unchanged as `null`

src/Microsoft.Health.Dicom.Core.UnitTests/Extensions/ExtendedQueryTagEntryExtensionsTests.cs
```diff
@@ -87,10 +96,19 @@ public void GivenStandardTagAsKeyword_WhenNormalizing_ThenVRShouldBeFilled()
             Assert.Equal(normalized.Path, expectedPath);
         }
 
+        [Fact]
+
+        public void GivenInvalidTagWithoutVR_WhenNormalizing_ThenShouldNotThrowException()
+        {
+            // track bug  https://microsofthealth.visualstudio.com/Health/_workitems/edit/81015
```
wsugarman, line 103: Won't this comment become moot as soon as you check this in because you're fixing it? Probably best not to leave the URL in here

src/Microsoft.Health.Dicom.Core.UnitTests/Extensions/ExtendedQueryTagEntryExtensionsTests.cs
```diff
@@ -36,6 +36,15 @@ public void GivenPrivateTagWithNonEmptyPrivateCreator_WhenNormalizing_ThenPrivat
             Assert.Null(normalized.PrivateCreator);
         }
 
+        [Fact]
+
+        public void GivenPrivateIdentificationCodeTagWithoutVR_WhenNormalizing_ThenVRShouldBeFilled()
+        {
+            ExtendedQueryTagEntry entry = CreateExtendedQueryTagEntry("22010010", null, null, QueryTagLevel.Instance, ExtendedQueryTagStatus.Ready);
+            ExtendedQueryTagEntry normalized = entry.Normalize(ExtendedQueryTagStatus.Adding);
+            Assert.Equal(DicomVRCode.LO, normalized.VR);
```
wsugarman, line 45: Should there be a test case for when the VR is lowercase?

wsugarman (approved):

src/Microsoft.Health.Dicom.Core.UnitTests/Extensions/ExtendedQueryTagEntryExtensionsTests.cs
pengchen0692: yes, https://github.com/microsoft/dicom-server/blob/db1e430e47a3084b4a9b8d1fd4701caf914ed2d5/src/Microsoft.Health.Dicom.Core.UnitTests/Extensions/ExtendedQueryTagEntryExtensionsTests.cs#L71 covers that

src/Microsoft.Health.Dicom.Core.UnitTests/Extensions/ExtendedQueryTagEntryExtensionsTests.cs
pengchen0692: The purpose of this unit test is for regresson: when tag is invalid and vr is null, NullPointerException is thrown, so I don't think we need to verify the VR?

PR #718, opened 2021-04-09
AddDicomBackgroundWorkers to extended on IDicomServerBuilder 

pengchen0692:
## Description
AddDicomBackgroundWorkers to extended on IDicomServerBuilder  instead of IServiceCollection

## Related issues
Addresses [AB#80433](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/80433)

## Testing
All automated test passed

## Notes
The extension method AddDicomBackgroundWorkers(this IServiceCollection services) should probably extend IDicomServerBuilder instead of IServiceCollection. This could make startUp code more clear.

### DIFF
--- a/src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
+++ b/src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
@@ -42,11 +42,14 @@ namespace Microsoft.AspNetCore.Builder
         /// <summary>
         /// Add services for DICOM background workers.
         /// </summary>
-        /// <param name="services">The services collection.</param>
-        public static void AddDicomBackgroundWorkers(this IServiceCollection services)
+        /// <param name="serverBuilder">The DICOM server builder instance.</param>
+        /// <returns>The DICOM server builder instance.</returns>
+        public static IDicomServerBuilder AddBackgroundWorkers(this IDicomServerBuilder serverBuilder)
         {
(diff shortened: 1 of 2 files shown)


brandonpollett (approved):
This makes sense to me. This is likely considered a breaking change so we'll need to semver it as such. Also, in the commit message you may want to mention what needs to change.

src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
```diff
@@ -42,11 +42,14 @@ public static class DicomServerServiceCollectionExtensions
         /// <summary>
         /// Add services for DICOM background workers.
         /// </summary>
-        /// <param name="services">The services collection.</param>
-        public static void AddDicomBackgroundWorkers(this IServiceCollection services)
+        /// <param name="serverBuilder">The DICOM server builder instance.</param>
+        /// <returns>The DICOM server builder instance.</returns>
+        public static IDicomServerBuilder AddBackgroundWorkers(this IDicomServerBuilder serverBuilder)
         {
-            services.AddScoped<DeletedInstanceCleanupWorker>();
```
wsugarman, line 48: What is the purpose of this refactor? I can't quite tell from the PR

wsugarman: Discussed offline

wsugarman (approved):

PR #723, opened 2021-04-12
Limit for SinglePartReader

abdullah248:
## Description
Adds in byte limit for when reading a single dicom file.

## Related issues
[AB#79839](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/79839)

## Testing
Tested locally by uploading files above and below the limits and making sure the expected behavior was observed. Additionally added some unit tests were possible.


src/Microsoft.Health.Dicom.Api/Web/LimitStream.cs
```diff
@@ -0,0 +1,114 @@
...
+            int amountRead = await _stream.ReadAsync(buffer, cancellationToken);
+
+            _bytesLeft -= amountRead;
+            makeCheck();
+
+            return amountRead;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private void makeCheck()
```
wsugarman, line 91: The method name should be PascalCase and be something like `ThrowIf...`, like `ThrowIfOverflow` or `ThrowIfLimitReached`

src/Microsoft.Health.Dicom.Api/Web/LimitStream.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            _bytesLeft -= amountRead;
+            makeCheck();
+
+            return amountRead;
+        }
+
+        public override async ValueTask<int> ReadAsync(Memory<byte> buffer, CancellationToken cancellationToken = default)
+        {
+            if (_bytesLeft <= 0)
```
wsugarman, line 77: This must be strictly `<` as 0 is fine. Furthermore, you could re-use the helper method

src/Microsoft.Health.Dicom.Api/Web/LimitStream.cs
```diff
@@ -0,0 +1,114 @@
...
+using System;
+using System.IO;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System.Threading.Tasks;
+using EnsureThat;
+
+namespace Microsoft.Health.Dicom.Api.Web
+{
+    class LimitStream : Stream
```
wsugarman, line 15: nit: visibility of classes should be explicit. This class is probably `internal`

src/Microsoft.Health.Dicom.Api/Web/LimitStream.cs
```diff
@@ -0,0 +1,114 @@
...
+    {
+        private Stream _stream;
```
wsugarman, line 17: nit: `readonly`

src/Microsoft.Health.Dicom.Api/Web/LimitStream.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        public override bool CanRead => _stream.CanRead;
+
+        public override bool CanSeek => _stream.CanSeek;
+
+        public override bool CanWrite => _stream.CanWrite;
+
+        public override long Length => _stream.Length;
+
+        public override long Position { get => _stream.Position; set => throw new NotImplementedException(); }
```
wsugarman, line 29: nit: should be `NotSupportedException`

src/Microsoft.Health.Dicom.Api/Web/LimitStream.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        public LimitStream(Stream stream, long limit)
+        {
+            EnsureArg.IsNotNull(stream, nameof(stream));
```
wsugarman, line 33: nit: `limit` cannot be negative either, right?

src/Microsoft.Health.Dicom.Api/Web/LimitStream.cs
```diff
@@ -0,0 +1,114 @@
...
+        public override void Flush()
+        {
+            _stream.Flush();
+        }
+
+        public override int Read(byte[] buffer, int offset, int count)
+        {
+            if (_bytesLeft <= 0)
+            {
+                throw new DicomFileLengthLimitExceededException(_bytesLeft);
```
wsugarman, line 48: This is the number of bytes remaining, right? Which I don't think is the proper value to pass to this exception. If you require it for the error, then we'll need to preserve the initial limit

src/Microsoft.Health.Dicom.Api/Web/LimitStream.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            int amountRead = _stream.Read(buffer, offset, count);
+
+            _bytesLeft -= amountRead;
+            makeCheck();
+
+            return amountRead;
+        }
+
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Performance", "CA1835:Prefer the 'Memory'-based overloads for 'ReadAsync' and 'WriteAsync'", Justification = "Buffer is pass through")]
```
wsugarman, line 59: nit: import the namespace

src/Microsoft.Health.Dicom.Api/Web/LimitStream.cs
```diff
@@ -0,0 +1,114 @@
...
+        {
+            if (_bytesLeft < 0)
+            {
+                throw new DicomFileLengthLimitExceededException(_bytesLeft);
+            }
+        }
+
+        public override long Seek(long offset, SeekOrigin origin)
+        {
+            return _stream.Seek(offset, origin);
```
wsugarman, line 101: We should throw `NotSupportedException` here too

src/Microsoft.Health.Dicom.Api/Web/LimitStream.cs
wsugarman: Alternatively, if you want to support `Position` and `Seek`, then we should nail down the semantics to do some math. Does this `Stream` limit the number of `bytes` read? Or does it limit the `Length` of the `Stream`?

wsugarman (changes requested):
Generally, where is this class used? Some light reading led me to `StreamOriginatedDicomInstanceEntry` which requires the `Stream` to be "seekable" so that it can return the `Stream` of data even when it might have already been read. If that's the case, we should definitely delay the "seekable" stuff until this class.

src/Microsoft.Health.Dicom.Api/Web/LimitStream.cs
abdullah248: Added a check for that

src/Microsoft.Health.Dicom.Api/Web/LimitStream.cs
abdullah248: Thanks for the catch, missed it when we were refactoring it yesterday!

src/Microsoft.Health.Dicom.Api/Web/LimitStream.cs
abdullah248: Hmm so actually for this, for canSeek we are just passing if the underlying stream can seek or not. Wont we run into issues if the underlying stream is able to seek but we thrown an error?

wsugarman: The problem isn't whether our implementation is out-of-step with the underlying `Stream`, but how does that impact our class. If someone attempts to `Seek`, will that screw up our validation of the file size?

abdullah248: Updated

src/Microsoft.Health.Dicom.Api/Web/LimitStream.cs
abdullah248: Implemented as discussed offline

src/Microsoft.Health.Dicom.Core/Features/Store/Entries/DicomInstanceEntryReaderForSinglePartRequest.cs
```diff
@@ -5,14 +5,20 @@
...
+using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 using EnsureThat;
+using Microsoft.Extensions.Options;
 using Microsoft.Health.Abstractions.Exceptions;
+using Microsoft.Health.Dicom.Core.Configs;
+using Microsoft.Health.Dicom.Core.Exceptions;
 using Microsoft.Health.Dicom.Core.Web;
+using NotSupportedException = System.NotSupportedException;
```
wsugarman, line 21: Is there a name conflict?

src/Microsoft.Health.Dicom.Core/Features/Store/Entries/DicomInstanceEntryReaderForSinglePartRequest.cs
```diff
@@ -22,12 +28,16 @@ namespace Microsoft.Health.Dicom.Core.Features.Store.Entries
     public class DicomInstanceEntryReaderForSinglePartRequest : IDicomInstanceEntryReader
     {
         private readonly ISeekableStreamConverter _seekableStreamConverter;
+        private readonly IOptions<StoreConfiguration> _storeConfiguration;
 
-        public DicomInstanceEntryReaderForSinglePartRequest(ISeekableStreamConverter seekableStreamConverter)
+
```
wsugarman, line 33: nit: extra line

src/Microsoft.Health.Dicom.Core/Features/Store/Entries/DicomInstanceEntryReaderForSinglePartRequest.cs
```diff
@@ -22,12 +28,16 @@ namespace Microsoft.Health.Dicom.Core.Features.Store.Entries
...
+        public DicomInstanceEntryReaderForSinglePartRequest(ISeekableStreamConverter seekableStreamConverter, IOptions<StoreConfiguration> storeConfiguration)
         {
             EnsureArg.IsNotNull(seekableStreamConverter, nameof(seekableStreamConverter));
+            EnsureArg.IsNotNull(storeConfiguration, nameof(storeConfiguration));
 
             _seekableStreamConverter = seekableStreamConverter;
+            _storeConfiguration = storeConfiguration;
```
wsugarman, line 40: Store the configuration and not the options object. Calling the `Value` property resolve the underlying configuration binding and we can eagerly perform that in the ctor

src/Microsoft.Health.Dicom.Core/Features/Store/Entries/DicomInstanceEntryReaderForSinglePartRequest.cs
```diff
@@ -52,10 +62,110 @@ public async Task<IReadOnlyList<IDicomInstanceEntry>> ReadAsync(string contentTy
                     string.Format(CultureInfo.InvariantCulture, DicomCoreResource.UnsupportedContentType, contentType));
             }
 
-            Stream seekableStream = await _seekableStreamConverter.ConvertAsync(stream, cancellationToken);
+            Stream seekableStream;
```
wsugarman, line 65: nit: there should be a comment as to why you can dispose of the underlying stream so early. It's not very obvious

src/Microsoft.Health.Dicom.Core/Features/Store/Entries/DicomInstanceEntryReaderForSinglePartRequest.cs
```diff
@@ -52,10 +62,110 @@ public async Task<IReadOnlyList<IDicomInstanceEntry>> ReadAsync(string contentTy
...
+                _bytesLeft = limit;
+                _limit = limit;
+            }
+
+            public override void Flush()
+            {
+                _stream.Flush();
+            }
+
+            public override int Read(byte[] buffer, int offset, int count)
```
wsugarman, line 109: I forgot. I think we also need to overwrite `ReadByte`!

src/Microsoft.Health.Dicom.Core/Features/Store/Entries/DicomInstanceEntryReaderForSinglePartRequest.cs
```diff
@@ -52,10 +62,110 @@ public async Task<IReadOnlyList<IDicomInstanceEntry>> ReadAsync(string contentTy
...
+
+            private long _bytesLeft;
+
+            private readonly long _limit;
+
+            public override bool CanRead => _stream.CanRead;
+
+            public override bool CanSeek => false;
+
+            public override bool CanWrite => _stream.CanWrite;
```
wsugarman, line 88: It may be best just to remove write capability like we did with `Seek`. Then we could call this stream something like `ReadOnlyLimitStream`

src/Microsoft.Health.Dicom.Core/Exceptions/DicomFileLengthLimitExceededException.cs
```diff
@@ -3,14 +3,12 @@
...
-using Microsoft.Health.Dicom.Core.Exceptions;
-
-namespace Microsoft.Health.Dicom.Api.Web
+namespace Microsoft.Health.Dicom.Core.Exceptions
 {
     public class DicomFileLengthLimitExceededException : ValidationException
     {
         public DicomFileLengthLimitExceededException(long maxAllowedLength)
-           : base(string.Format(DicomApiResource.DicomFileLengthLimitExceeded, maxAllowedLength))
+           : base(string.Format(DicomCoreResource.DicomFileLengthLimitExceeded, maxAllowedLength))
```
wsugarman, line 11: Does this not complain about a culture?

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/Entries/DicomInstanceEntryReaderForSinglePartRequestTests.cs
```diff
@@ -80,5 +87,56 @@ public async Task GivenBodyPartWithValidContentType_WhenReading_ThenCorrectResul
...
+
+        private class TestSeekableStreamConverter : ISeekableStreamConverter
+        {
+            public async Task<Stream> ConvertAsync(Stream stream, CancellationToken cancellationToken = default)
+            {
+                Stream seekableStream = null;
+
+                if (!stream.CanSeek)
+                {
+                    seekableStream = new FileBufferingReadStream(stream, 1024 * 30000, null, Environment.GetEnvironmentVariable("ASPNETCORE_TEMP") ?? Path.GetTempPath());
```
wsugarman, line 128: You don't really need to use the `FileBufferingReadStream` and in turn the new nuget package. You could also just use `MemoryStream` and copy everything into there for the test

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/Entries/DicomInstanceEntryReaderForSinglePartRequestTests.cs
```diff
@@ -80,5 +87,56 @@ public async Task GivenBodyPartWithValidContentType_WhenReading_ThenCorrectResul
                     Assert.Same(_stream, await item.GetStreamAsync(source.Token));
                 });
         }
+
+        [Fact]
+        public async Task GivenBodyPartWithValidContentTypeExceedLimit_ThrowsError()
```
wsugarman, line 92: nit: `ThrowError`*

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/Entries/DicomInstanceEntryReaderForSinglePartRequestTests.cs
```diff
@@ -80,5 +87,56 @@ public async Task GivenBodyPartWithValidContentType_WhenReading_ThenCorrectResul
...
+        {
+            DicomInstanceEntryReaderForSinglePartRequest dicomInstanceEntryReaderLowLimit = new DicomInstanceEntryReaderForSinglePartRequest(_seekableStreamConverter, CreateStoreConfiguration(1));
```
wsugarman, line 94: nit: `var` in this particular case

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/Entries/DicomInstanceEntryReaderForSinglePartRequestTests.cs
```diff
@@ -80,5 +87,56 @@ public async Task GivenBodyPartWithValidContentType_WhenReading_ThenCorrectResul
...
+
+            using var source = new CancellationTokenSource();
+
+            Stream stream = new MemoryStream();
+            stream.Write(Encoding.UTF8.GetBytes("someteststring"));
+            stream.Seek(0, SeekOrigin.Begin);
+
+            await Assert.ThrowsAsync<DicomFileLengthLimitExceededException>(
+                () => dicomInstanceEntryReaderLowLimit.ReadAsync(
+                DefaultContentType,
```
wsugarman, line 104: nit: indent the args to `ReadAsync` one more level

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/Entries/DicomInstanceEntryReaderForSinglePartRequestTests.cs
```diff
@@ -65,12 +71,13 @@ public async Task GivenUnSupportedContentType_WhenReading_ThenShouldThrowUnsuppo
         public async Task GivenBodyPartWithValidContentType_WhenReading_ThenCorrectResultsShouldBeReturned()
         {
             using var source = new CancellationTokenSource();
-            _seekableStreamConverter.ConvertAsync(_stream, source.Token).Returns(_stream);
             IReadOnlyCollection<IDicomInstanceEntry> results = await _dicomInstanceEntryReader.ReadAsync(
                 DefaultContentType,
                 _stream,
                 source.Token);
 
+            _stream.Write(Encoding.UTF8.GetBytes("someteststring"));
```
wsugarman, line 79: Doesn't the read need to happen after you've written bytes?

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/Entries/DicomInstanceEntryReaderForSinglePartRequestTests.cs
wsugarman, line 92: Can you add one more test where the data is exactly the max length?

src/Microsoft.Health.Dicom.Api.UnitTests/Web/SeekableStreamConverterTests.cs
```diff
@@ -72,5 +72,6 @@ private Stream SetupNonSeekableStreamException<TException>()
 
             return nonseekableStream;
         }
+
```
wsugarman, line 75: nit: extra new line

src/Microsoft.Health.Dicom.Api.UnitTests/Web/AspNetCoreMultipartReaderTests.cs
```diff
@@ -253,6 +299,7 @@ private async Task ValidateMultipartBodyPartAsync(string expectedContentType, st
             }
         }
 
+
```
wsugarman, line 302: nit: newline

src/Microsoft.Health.Dicom.Api.UnitTests/Web/AspNetCoreMultipartReaderTests.cs
```diff
@@ -178,6 +180,50 @@ public void GivenStartParameter_WhenReading_ThenDicomNotSupportedExceptionShould
             Assert.Throws<NotSupportedException>(() => Create(requestContentType));
         }
 
+        [Fact]
+        public async Task GivenAnIOExceptionReadingStream_WhenConverted_ShouldReturnNull()
+        {
+            _seekableStreamConverter = Substitute.For<ISeekableStreamConverter>();
```
wsugarman, line 186: Use a local variable here instead of override the class's state. Same for below

src/Microsoft.Health.Dicom.Core/Features/Store/Entries/DicomInstanceEntryReaderForSinglePartRequest.cs
abdullah248: Yes, there is one that we created for some dicom things that are note supported.

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/Entries/DicomInstanceEntryReaderForSinglePartRequestTests.cs
abdullah248: Still need the package for drain async, but did change to memory stream

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/Entries/DicomInstanceEntryReaderForSinglePartRequestTests.cs
abdullah248: Added

src/Microsoft.Health.Dicom.Core/Features/Store/Entries/DicomInstanceEntryReaderForSinglePartRequest.cs
```diff
@@ -22,12 +28,15 @@ namespace Microsoft.Health.Dicom.Core.Features.Store.Entries
     public class DicomInstanceEntryReaderForSinglePartRequest : IDicomInstanceEntryReader
     {
         private readonly ISeekableStreamConverter _seekableStreamConverter;
+        private readonly StoreConfiguration _storeConfiguration;
 
-        public DicomInstanceEntryReaderForSinglePartRequest(ISeekableStreamConverter seekableStreamConverter)
+        public DicomInstanceEntryReaderForSinglePartRequest(ISeekableStreamConverter seekableStreamConverter, IOptions<StoreConfiguration> storeConfiguration)
         {
             EnsureArg.IsNotNull(seekableStreamConverter, nameof(seekableStreamConverter));
+            EnsureArg.IsNotNull(storeConfiguration, nameof(storeConfiguration));
```
wsugarman, line 36: nit: I think we should check value too, like: `EnsureArg.IsNotNull(storeConfiguration?.Value, nameof(storeConfiguration));`

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/Entries/DicomInstanceEntryReaderForSinglePartRequestTests.cs
```diff
@@ -65,7 +70,55 @@ public async Task GivenUnSupportedContentType_WhenReading_ThenShouldThrowUnsuppo
...
+            Stream stream = new MemoryStream();
+            stream.Write(Encoding.UTF8.GetBytes("someteststring"));
+            stream.Seek(0, SeekOrigin.Begin);
+
+            await Assert.ThrowsAsync<DicomFileLengthLimitExceededException>(
+                () => dicomInstanceEntryReaderLowLimit.ReadAsync(
+                    DefaultContentType,
+                    stream,
+                    source.Token));
+
```
wsugarman, line 108: nit: some extra newlines here

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/Entries/DicomInstanceEntryReaderForSinglePartRequestTests.cs
```diff
@@ -65,7 +70,55 @@ public async Task GivenUnSupportedContentType_WhenReading_ThenShouldThrowUnsuppo
...
             using var source = new CancellationTokenSource();
-            _seekableStreamConverter.ConvertAsync(_stream, source.Token).Returns(_stream);
+
+            _stream.Write(Encoding.UTF8.GetBytes("someteststring"));
+
+            IReadOnlyCollection<IDicomInstanceEntry> results = await _dicomInstanceEntryReader.ReadAsync(
+                DefaultContentType,
+                _stream,
+                source.Token);
+
```
wsugarman, line 80: nit: extra newline

src/Microsoft.Health.Dicom.Api.UnitTests/Web/AspNetCoreMultipartReaderTests.cs
```diff
@@ -25,11 +27,11 @@ public class AspNetCoreMultipartReaderTests
         private const string DefaultBodyPartSeparator = "--+b+";
         private const string DefaultBodyPartFinalSeparator = "--+b+--";
 
-        private readonly ISeekableStreamConverter _seekableStreamConverter;
+        private ISeekableStreamConverter _seekableStreamConverter;
```
wsugarman, line 30: nit: we can make this `readonly` again, right?

wsugarman (approved):

PR #730, opened 2021-04-19
Throttling when adding extended query tag.

pengchen0692:
## Description
Allow up to 128 tags with one call.

## Related issues
Addresses [AB#79598](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/79598)

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Configs/ExtendedQueryTagConfiguration.cs
+++ b/src/Microsoft.Health.Dicom.Core/Configs/ExtendedQueryTagConfiguration.cs
@@ -5,12 +5,14 @@
 
 namespace Microsoft.Health.Dicom.Core.Configs
 {
-    public class ServicesConfiguration
+    /// <summary>
+    /// Configuration for extended query tag feature.
+    /// </summary>
+    public class ExtendedQueryTagConfiguration
     {
-        public DeletedInstanceCleanupConfiguration DeletedInstanceCleanup { get; } = new DeletedInstanceCleanupConfiguration();
-
-        public StoreConfiguration StoreServiceSettings { get; } = new StoreConfiguration();
-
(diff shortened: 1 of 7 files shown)


src/Microsoft.Health.Dicom.Core/Configs/ExtendedQueryTagConfiguration.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    /// <summary>
+    /// Configuration for extended query tag feature.
+    /// </summary>
+    public class ExtendedQueryTagConfiguration
+    {
+        /// <summary>
+        /// Maximum allowed number of tags when adding.
+        /// </summary>
+        public int MaxBulkAddSize { get; set; } = 128;
```
pengchen0692, line 16: Not find a good name, please suggest better name if found. :)

src/Microsoft.Health.Dicom.Core/Features/ExtendedQueryTag/ExtendedQueryTagEntryValidator.cs
```diff
@@ -60,6 +65,12 @@ public void ValidateExtendedQueryTags(IEnumerable<ExtendedQueryTagEntry> extende
                 throw new ExtendedQueryTagEntryValidationException(DicomCoreResource.MissingExtendedQueryTag);
             }
 
+            if (extendedQueryTagEntries.Count() > _maxBulkAddSize)
```
smithago, line 68: I thought this was to limit max allowed per service, instead of each request. Why this limit?

pengchen0692: * will update the user story
* This need updates database storage procedure, thinking to combine the change with other coming updates like reindexing.

Issue #444, opened 2020-11-24
Wrong media type for Retrieve Instance transaction


hackermd:
**Describe the bug**
The [Retrieve Instance transaction](https://github.com/microsoft/dicom-server/blob/master/docs/resources/conformance-statement.md#retrieve-an-instance) of the DICOMweb service uses the wrong media type. It should be `multipart/related; type="application/dicom"` instead of `application/dicom`. See [DICOM Correction Proposal 2040](https://www.dicomstandard.org/news-dir/current/docs/cpack108/cp2040.pdf) for clarification.

StevenBorg:
Thank you, @hackermd !  That certainly simplifies the retrieve transactions and makes all of them consistent with each other. We're looking into this now, and it certainly makes sense.  Thanks!  We'll keep you posted here.

hackermd:
Great! Thanks for looking into it @StevenBorg

StevenBorg:
@hackermd After some discussion, our take is that we should adopt the newer framework, even prior to acceptance under the standard, while still maintaining backward compatibility (in this case) for APIs which might be using the current standard. Do you see any concerns with this approach?

hackermd:
No concerns. I agree that this would be the best approach.

The "newer" framework has actually always been the standard - at least it was supposed to. The goal of Supplement 183 was to refactor the documentation of Part 18 without actually changing the behavior. Unfortunately, a few things got broken and we are currently trying to correct those. However, since Supplement 183 was officially accepted, we should probably consider `application/dicom` to be standard compliant or at least support it in a backwards compatible manner. We are going to do the same client-side (see https://github.com/MGHComputationalPathology/dicomweb-client/pull/43).

smithago:
Issue is fixed. Closing

PR #735, opened 2021-04-21
Allow multipart/related media type for retrieved instance transaction

bcarthic:
## Description
Support `multipart/related; type="application/dicom";` media type for retrieve instance transaction

## Related issues
Addresses [issue #444, AB#80402].

## Testing
- Added new unit test and e2e test to test both media types

### DIFF
--- a/src/Microsoft.Health.Dicom.Client/DicomWebClient.cs
+++ b/src/Microsoft.Health.Dicom.Client/DicomWebClient.cs
@@ -66,7 +66,7 @@ namespace Microsoft.Health.Dicom.Client
 
             request.Headers.TryAddWithoutValidation(
                 "Accept",
-                CreateAcceptHeader(DicomWebConstants.MediaTypeApplicationDicom, dicomTransferSyntax));
+                CreateAcceptHeader(CreateMultipartMediaTypeHeader(DicomWebConstants.ApplicationDicomMediaType), dicomTransferSyntax));
 
             HttpResponseMessage response = await HttpClient.SendAsync(request, cancellationToken)
                 .ConfigureAwait(false);
--- a/test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/RetrieveTransactionResourceTests.Instance.cs
+++ b/test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/RetrieveTransactionResourceTests.Instance.cs
@@ -6,6 +6,7 @@
 using System;
 using System.Net;
 using System.Net.Http;
+using System.Net.Http.Headers;
 using System.Threading.Tasks;
 using Dicom;
 using Microsoft.Health.Dicom.Client;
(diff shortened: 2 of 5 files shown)


src/Microsoft.Health.Dicom.Client/DicomWebClient.cs
```diff
@@ -66,7 +66,7 @@ public DicomWebClient(HttpClient httpClient)
 
             request.Headers.TryAddWithoutValidation(
                 "Accept",
-                CreateAcceptHeader(DicomWebConstants.MediaTypeApplicationDicom, dicomTransferSyntax));
+                CreateAcceptHeader(CreateMultipartMediaTypeHeader(DicomWebConstants.ApplicationDicomMediaType), dicomTransferSyntax));
```
smithago, line 69: Since we support both, we can leave the default as application/dicom in Client dll, what do you think?

smithago: We also need to update the documentation on the new support. The conformance doc needs to be updated

test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/RetrieveTransactionResourceTests.Instance.cs
```diff
@@ -37,7 +38,14 @@ public async Task GivenSupportedAcceptHeaders_WhenRetrieveInstance_ThenServerSho
 
             using DicomWebResponse<DicomFile> response = await _client.RetrieveInstanceAsync(instanceId.StudyInstanceUid, instanceId.SeriesInstanceUid, instanceId.SopInstanceUid, transferSyntax);
 
-            Assert.Equal(DicomWebConstants.ApplicationDicomMediaType, response.ContentHeaders.ContentType.MediaType);
+            Assert.Equal(DicomWebConstants.MultipartRelatedMediaType, response.ContentHeaders.ContentType.MediaType);
```
smithago, line 41: This will be not needed if the client dll by default uses application/dicom.

Should we add another E2E test to test E2E multipart accept works?

src/Microsoft.Health.Dicom.Client/DicomWebClient.cs
bcarthic: What understood from the issue conversation, we need to be backward compatible but going forward we need to support the new standard. Let me know if I understood that correctly. Good catch on the documentation. I will update it.

test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/RetrieveTransactionResourceTests.Instance.cs
bcarthic: Yes you are right, this is not needed if the client uses older format. Sure, if the client uses single part by default then we need to have E2E for multipart

src/Microsoft.Health.Dicom.Client/DicomWebClient.cs
bcarthic: Reverted to leave the client to accept application/dicom as default

smithago (approved):

PR #738, opened 2021-04-23
Addressing minor leftover changes to cleanup input for ExtendedQueryTags

snehabandla:
[AB#81504](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/81504)
[AB#81024](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/81024)
[AB#81532](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/81532)


src/Microsoft.Health.Dicom.Api/Features/Filters/BodyModelStateValidatorAttribute.cs
```diff
@@ -17,7 +17,7 @@ public override void OnActionExecuting(ActionExecutingContext context)
             EnsureArg.IsNotNull(context, nameof(context));
             if (!context.ModelState.IsValid)
             {
-                throw new InvalidRequestBodyException(string.Join(",", context.ModelState.SelectMany(x => x.Value.Errors).Select(x => x.ErrorMessage)));
+                throw new InvalidRequestBodyException(string.Join("/n", context.ModelState.SelectMany(x => x.Value.Errors).Select(x => x.ErrorMessage)));
```
pengchen0692, line 20: * Environment.NewLine instead of "/n"?  btw, seems "/n" doesn't work 
* Add a new line just after _Details:_, so that all error message are in individual lines

image

docs/how-to-guides/extended-query-tags.md
```diff
@@ -74,17 +74,17 @@ Add extended query tags to supported set with route:  /extendedquerytags
 		"Path":"04011001",
 		"VR":"SS",
 		"PrivateCreator":"MicrosoftPC",
-		"Level":"Instance"
+		"QueryTagLevel":"Instance"
```
pengchen0692, line 77: if we use _QueryTagLevel_ here, should we also append _QueryTag_ to other properties?

src/Microsoft.Health.Dicom.Client/Models/AddExtendedQueryTagEntry.cs
```diff
@@ -0,0 +1,34 @@
+﻿// -------------------------------------------------------------------------------------------------
```
pengchen0692, line 1: The file name and class name doesn't match.

src/Microsoft.Health.Dicom.Client/Models/AddExtendedQueryTagEntry.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        /// <summary>
+        /// VR of this tag.
+        /// </summary>
+        public string VR { get; set; }
+
+        /// <summary>
+        /// Level of this tag. Could be Study, Series or Instance.
+        /// </summary>
+        public string QueryTagLevel { get; set; }
```
pengchen0692, line 27: * Should we use same name on Level for both AddExtendedQueryTag and  GetExtendedQueryTag?
* Curious why QueryTagLevel instead of Level? Since if we use QueryTagLevel here, we may want to append QueryTag to all other properties -- QueryTagPath, QueryTagVR etc.
* Wondering if we could use enum QueryTagLevel instead of string, and use https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_Converters_StringEnumConverter.htm for to serialize to string instead of integer. The converter could be specified on class Property or Enum or JsonSerializerSettings, Personal preference is JsonSerializerSettings, can be setup here: https://github.com/microsoft/dicom-server/blob/243437502e472f0a7e65d4b96081f474e1c90b28/src/Microsoft.Health.Dicom.Client/DicomWebClient.cs#L43
* Some thought: are AddExtendedQueryTag and GetExtendedQueryTag good names? They are more like some actions instead of data model, but I don't have good candidates, maybe _AddExtendedQueryTagEntry_ and _ExtendedQueryTagEntry_ ? I'm not sure.

docs/how-to-guides/extended-query-tags.md
pengchen0692: QueryTagPath, QueryTagVR etc.

src/Microsoft.Health.Dicom.Client/Models/AddExtendedQueryTagEntry.cs
snehabandla: Reverted back to just 'Level' everywhere.

I added the enum in the client but I prefer a serializer specific to the class to ensure other enums are serialized in the way they need to be.

I think these names work for now...I'll keep it in mind and see if longer term I have a better name for the two entries!

pengchen0692 (approved):

PR #749, opened 2021-04-29
Adding in new sql schema for additional core tags

snehabandla:
[AB#81722](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/81722)


src/Microsoft.Health.Dicom.SqlServer/Microsoft.Health.Dicom.SqlServer.csproj
```diff
@@ -5,6 +5,10 @@
     <TargetFramework>net5.0</TargetFramework>
   </PropertyGroup>
 
+  <ItemGroup>
```
pengchen0692, line 8: nit: I know this is added by VS automatically, it's better if could be removed to keep consistency (we don't have it for 1.sql and 2.sql)

src/Microsoft.Health.Dicom.SqlServer/Microsoft.Health.Dicom.SqlServer.csproj
```diff
@@ -54,6 +58,11 @@
       <InputToMutableSqlGenerator>true</InputToMutableSqlGenerator>
       <MutableClassVersion>Latest</MutableClassVersion>
     </EmbeddedResource>
+    <EmbeddedResource Include="Features\Schema\Migrations\3.sql">
+      <InputToImmutableSqlGenerator>true</InputToImmutableSqlGenerator>
+      <InputToMutableSqlGenerator>true</InputToMutableSqlGenerator>
+      <MutableClassVersion>3</MutableClassVersion>
```
pengchen0692, line 64: Should _MutableClassVersion_ be _Latest_  for 3.sql , while the 2.sql become _2_?

snehabandla: I think I will switch it to that in a bit. For now I'll add the schema separately and rename when I integrate code to call it!

pengchen0692 (approved):

PR #750, opened 2021-04-29
Replace blob.DownloadToAsync with blob.OpenReadAsync

brandonpollett:
## Description
This PR replaces our calls on getting file from `blob.DownloadToAsync` to `blob.OpenReadAsync`. The behavior difference is that the `OpenReadAsync` returns a lazy `Stream` which only starts to download when it is actually being read. Previously, when using `DownloadToAsync` we realize the entire stream in memory. 

The result of this is that our single file read allocations go down significantly, but we we will face more garbage collection as we're not using the `RecyclableMemoryStreamManager` any longer.

In the long term I believe we can strike a balance between which method to use based on file size if we record the dicom file size during the ingestion. 

## Related issues
Addresses [AB#81741](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/81741).

## Testing

### DIFF
--- a/src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
+++ b/src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
@@ -18,7 +18,6 @@ using Microsoft.Health.Blob.Configs;
 using Microsoft.Health.Dicom.Core.Exceptions;
 using Microsoft.Health.Dicom.Core.Features.Common;
 using Microsoft.Health.Dicom.Core.Features.Model;
-using Microsoft.IO;
 
 namespace Microsoft.Health.Dicom.Blob.Features.Storage
 {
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -107,17 +101,13 @@ public class BlobFileStore : IFileStore
...
-            {
-                MaximumConcurrency = _blobDataStoreConfiguration.RequestOptions.DownloadMaximumConcurrency,
-            };
+            var blobOpenReadOptions = new BlobOpenReadOptions(allowModifications: false);
 
             await ExecuteAsync(async () =>
             {
-                stream = _recyclableMemoryStreamManager.GetStream(GetFileStreamTagName);
-                await blob.DownloadToAsync(stream, conditions: null, storageTransferOptions, cancellationToken);
+                stream = await blob.OpenReadAsync(blobOpenReadOptions, cancellationToken);
```
wsugarman, line 108: I took a look at the underlying `LazyLoadingReadOnlyStream` that is returned here, and the buffer it uses to hold chunks of the blob are rented from the shared `ArrayPool<byte>`. So I think this will be an overall win!

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -107,17 +101,13 @@ public class BlobFileStore : IFileStore
             BlockBlobClient blob = GetInstanceBlockBlob(versionedInstanceIdentifier);
 
             Stream stream = null;
-            StorageTransferOptions storageTransferOptions = new StorageTransferOptions
-            {
-                MaximumConcurrency = _blobDataStoreConfiguration.RequestOptions.DownloadMaximumConcurrency,
-            };
+            var blobOpenReadOptions = new BlobOpenReadOptions(allowModifications: false);
```
wsugarman, line 104: Should we make this a `private` `static` member?

wsugarman (approved):

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
brandonpollett: I shied away from that because one of the properties you can set is `Position`. That lead me to believe that we would likely vary it in the future to make interesting use of this method.

PR #754, opened 2021-05-03
Remove AspNetCore Dependency from Core Package

wsugarman:
## Description
- Remove inadvertent AspNetCore dependency from core DICOM packages
- Disable `BackgroundServiceHealthCheckCache` for now

## Related issues
[AB#81936](https://microsofthealth.visualstudio.com/Health/_workitems/edit/81936)
[AB#81963](https://microsofthealth.visualstudio.com/Health/_workitems/edit/81963)

## Testing
Unit tests and local web server


src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
```diff
@@ -105,6 +105,9 @@ public static IDicomServerBuilder AddBackgroundWorkers(this IDicomServerBuilder
 
             services.TryAddSingleton<RecyclableMemoryStreamManager>();
 
+            // Used by the BackgroundServiceHealthCheck
+            services.AddApplicationInsightsTelemetry();
```
brandonpollett, line 109: We do this already in the web project https://github.com/microsoft/dicom-server/blob/805968201c83be0a9cb841efdedb5a0bb272bf9d/src/Microsoft.Health.Dicom.Web/Startup.cs#L63

It feels like this is an implementation detail about what instrumentation is chosen to be used for the given deployment scenario.

src/Microsoft.Health.Dicom.Core/Features/HealthCheck/BackgroundServiceHealthCheckCache.cs
```diff
@@ -13,13 +13,17 @@ namespace Microsoft.Health.Dicom.Core.Features.HealthCheck
...
     {
-        private IMemoryCache _cache;
+        private readonly MemoryCache _cache;
 
         private const string OldestDeleteInstanceCacheKey = "_oldestDeleted";
-
         private const string NumDeleteMaxRetryCacheKey = "_numMaxRetries";
 
-        public BackgroundServiceHealthCheckCache(IMemoryCache memoryCache)
+        public BackgroundServiceHealthCheckCache()
```
brandonpollett, line 21: Why this change to the actual implementation?

src/Microsoft.Health.Dicom.Core/Modules/ServiceModule.cs
```diff
@@ -143,8 +143,6 @@ public void Load(IServiceCollection services)
                    .AsSelf()
                    .AsImplementedInterfaces();
 
-            services.AddApplicationInsightsTelemetry();
-
             services.AddSingleton<BackgroundServiceHealthCheckCache>();
 
             services.AddHealthChecks().AddCheck<BackgroundServiceHealthCheck>(name: "BackgroundServiceHealthCheck");
```
brandonpollett, line 148: I realize this isn't part of this PR, but I would imagine that `services.AddSingleton<BackgroundServiceHealthCheckCache>();` and  `.AddCheck<BackgroundServiceHealthCheck>(name: "BackgroundServiceHealthCheck");` would be part of `AddBackgroundWorkers`. This seems like a cross of concerns.

brandonpollett: Also, it may be worth seeing if the background stuff fits into our `Module` paradigm.

wsugarman: That looks like the appropriate place for this. This background check appears to be only a proxy for the service's health.

src/Microsoft.Health.Dicom.Core/Features/HealthCheck/BackgroundServiceHealthCheckCache.cs
wsugarman: Before it was unintentionally (I think) using a shared memory cache. Instead, I suspect we'll want our own cache for this scenario

src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
wsugarman: The health check leverages app insights directly via its `ITelemetryClient`, so we're sort of prescribing the instrumentation, right? But at least if I add the `services.AddApplicationInsightsTelemetry` in the same method as the health check, ensure all to the dependencies are added with the health check. If the user wants to override the instrumentation key, they can still call this method later to successfully update it.

brandonpollett (approved):

PR #776, opened 2021-05-12
Make Sql schema migration idempotent

rbans96:
## Description
This PR updates latest schema migration script(2.diff.sql) for schema migration idempotency.
Currently schema-manager tool runs schema initialization scripts(x.sql) in transaction but runs the migration scripts(x.diff.sql) without wrapping in transaction so this PR ensures that our migration scripts(x.diff.sql) is idempotent.

## Related issues
Addresses [#AB81400](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Resolute/Stories/?workitem=81400)

## Testing
Integration test is added

### DIFF
--- a/src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
+++ b/src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
@@ -1,3 +1,12 @@
+
+/*************************************************************
+Wrapping up in a transaction except CREATE FULLTEXT INDEX which is non-transactional script. Since there are no slow scripts(all the statements i.e. CREATE TABLE/INDEX/STORED PROC and ALTER STORED PROC are faster) so keeping all in one transaction.
+Guidelines to create migration scripts - https://github.com/microsoft/healthcare-shared-components/tree/master/src/Microsoft.Health.SqlServer/SqlSchemaScriptsGuidelines.md
+**************************************************************/
+SET XACT_ABORT ON
+
+BEGIN TRANSACTION
+
 /*************************************************************
     Extended Query Tag Table
     Stores added extended query tags
@@ -255,17 +264,6 @@ BEGIN
         WatermarkAndTagKey
     )
 END
-
-IF NOT EXISTS (
-    SELECT * 
-	FROM sys.fulltext_indexes 
-	where object_id = object_id('dbo.ExtendedQueryTagPersonName'))
-BEGIN
-    CREATE FULLTEXT INDEX ON ExtendedQueryTagPersonName(TagValueWords LANGUAGE 1033)
-    KEY INDEX IXC_ExtendedQueryTagPersonName_WatermarkAndTagKey
-    WITH STOPLIST = OFF;
-END
-GO
 
 /*************************************************************
     The user defined type for AddExtendedQueryTagsInput
@@ -991,4 +989,18 @@ AS
     SELECT @newWatermark
 
     COMMIT TRANSACTION
+GO
+
+COMMIT TRANSACTION
+GO
+
+IF NOT EXISTS (
+    SELECT * 
+	FROM sys.fulltext_indexes 
+	where object_id = object_id('dbo.ExtendedQueryTagPersonName'))
+BEGIN
+    CREATE FULLTEXT INDEX ON ExtendedQueryTagPersonName(TagValueWords LANGUAGE 1033)
+    KEY INDEX IXC_ExtendedQueryTagPersonName_WatermarkAndTagKey
+    WITH STOPLIST = OFF;
+END
 GO
--- a/src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
+++ b/src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
@@ -6,24 +6,31 @@
     TagLevel can be 0, 1 or 2 to represent Instance, Series or Study level
     TagStatus can be 0, 1 or 2 to represent Adding, Ready or Deleting
 **************************************************************/
-CREATE TABLE dbo.ExtendedQueryTag (
-    TagKey                  INT                  NOT NULL, --PK
-    TagPath                 VARCHAR(64)          NOT NULL,
-    TagVR                   VARCHAR(2)           NOT NULL,
-    TagPrivateCreator       NVARCHAR(64)         NULL, 
-    TagLevel                TINYINT              NOT NULL,
-    TagStatus               TINYINT              NOT NULL
-)
+IF NOT EXISTS (
+    SELECT * 
+    FROM sys.tables
+    WHERE name = 'ExtendedQueryTag')
+BEGIN
+    CREATE TABLE dbo.ExtendedQueryTag (
+        TagKey                  INT                  NOT NULL, --PK
+        TagPath                 VARCHAR(64)          NOT NULL,
+        TagVR                   VARCHAR(2)           NOT NULL,
+        TagPrivateCreator       NVARCHAR(64)         NULL, 
+        TagLevel                TINYINT              NOT NULL,
+        TagStatus               TINYINT              NOT NULL
+    )
 
-CREATE UNIQUE CLUSTERED INDEX IXC_ExtendedQueryTag ON dbo.ExtendedQueryTag
-(
-    TagKey
-)
+    CREATE UNIQUE CLUSTERED INDEX IXC_ExtendedQueryTag ON dbo.ExtendedQueryTag
+    (
+        TagKey
+    )
 
-CREATE UNIQUE NONCLUSTERED INDEX IX_ExtendedQueryTag_TagPath ON dbo.ExtendedQueryTag
-(
-    TagPath
-)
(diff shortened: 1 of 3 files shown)


test/Microsoft.Health.Dicom.Tests.Integration/Persistence/SqlServerSchemaUpgradeTests.cs
```diff
@@ -34,5 +36,13 @@ public async Task GivenTwoSchemaInitializationMethods_WhenCreatingTwoDatabases_B
             await snapshotFixture.DisposeAsync();
             await diffFixture.DisposeAsync();
         }
+
+        [Fact]
+        public async Task GivenASchemaVersion_WhenApplyingDiffTwice_ShouldSucceed()
+        {
+            SqlDataStoreTestsFixture snapshotFixture = new SqlDataStoreTestsFixture(SqlDataStoreTestsFixture.GenerateDatabaseName("SNAPSHOT"));
+            await snapshotFixture.InitializeAsync(forceIncrementalSchemaUpgrade: false);
+            await snapshotFixture.SchemaUpgradeRunner.ApplySchemaAsync(SchemaVersionConstants.Max, applyFullSchemaSnapshot: false, CancellationToken.None);
```
YazanMSFT, line 45: I think the test in FHIR is similar to this, but I think we can improve it by making this test a Theory that will the test multiple times, for each schema version between SchemaVersionConstants.Min and Max.

YazanMSFT (approved):
I'm signing off, but let's wait for signoff from a DICOM team member as well.

test/Microsoft.Health.Dicom.Tests.Integration/Persistence/SqlServerSchemaUpgradeTests.cs
```diff
@@ -37,12 +38,18 @@ public async Task GivenTwoSchemaInitializationMethods_WhenCreatingTwoDatabases_B
             await diffFixture.DisposeAsync();
         }
 
-        [Fact]
-        public async Task GivenASchemaVersion_WhenApplyingDiffTwice_ShouldSucceed()
+        [Theory]
+        [InlineData(SchemaVersionConstants.Max)]
```
YazanMSFT, line 42: Almost - what I'm aiming for is not having to update this test whenever a new schema version is added. If you use MemberData instead and have a field dynamically generating all the versions from min to max, it should accomplish that. So something like:

public static IEnumerable<object[]> SchemaDiffVersions = new List<object[]>(Enumerable.Range(SchemaVersionConstants.Min + 1, SchemaVersionConstants.Max).Select(x => new object[] { x }));

and then the test attribute would be [MemberData(nameof(SchemaDiffVersions)]

brandonpollett:
@pengchen0692 has updated the description with the context for why this change is being brought in. I'm not sure if we have any documentation that needs to be updated around this also.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
```diff
@@ -133,113 +168,143 @@ CREATE UNIQUE CLUSTERED INDEX IXC_ExtendedQueryTagDateTime ON dbo.ExtendedQueryT
...
-
-CREATE UNIQUE NONCLUSTERED INDEX IXC_ExtendedQueryTagPersonName_WatermarkAndTagKey ON dbo.ExtendedQueryTagPersonName
-(
-    WatermarkAndTagKey
-)
-
-CREATE FULLTEXT INDEX ON ExtendedQueryTagPersonName(TagValueWords LANGUAGE 1033)
-KEY INDEX IXC_ExtendedQueryTagPersonName_WatermarkAndTagKey
-WITH STOPLIST = OFF;
+IF NOT EXISTS (
```
smithago, line 171: Wrapping the if on table and index is not completely correct.

Table may exist and the index which is a different object maybe missing.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
```diff
@@ -6,24 +6,31 @@
...
+    CREATE TABLE dbo.ExtendedQueryTag (
+        TagKey                  INT                  NOT NULL, --PK
+        TagPath                 VARCHAR(64)          NOT NULL,
+        TagVR                   VARCHAR(2)           NOT NULL,
+        TagPrivateCreator       NVARCHAR(64)         NULL, 
+        TagLevel                TINYINT              NOT NULL,
+        TagStatus               TINYINT              NOT NULL
+    )
+
+    CREATE UNIQUE CLUSTERED INDEX IXC_ExtendedQueryTag ON dbo.ExtendedQueryTag
```
smithago, line 23: same here. Condition for index check is different than table check

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
```diff
@@ -6,24 +6,31 @@
     TagLevel can be 0, 1 or 2 to represent Instance, Series or Study level
```
smithago, line 6: Regarding the description on why this change. Why do we do initialization in transaction and upgrade not?

smithago (changes requested):

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
rbans96: Good catch, thanks! Fixed.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.diff.sql
rbans96: Good question! I think as per discussion with John and @YazanMSFT - we could run the schema initialization scripts in a transaction. Since schema initialization would happen when we are standing up a new account and the SQLServer database is empty which is not the case for migration.

But last friday, I found out that one of the dicom script(CREATE FULLTEXT CATALOG) can't run in a transaction as per t-sql so anyways we would need to remove logic of running the initialization scripts wrapped in a transaction from schema-manager tool and instead we would update the latest initialization script(2.sql) idempotent too. I will create another PR to address this soon.
Please let me know if you have any more thoughts/questions. I am happy to schedule a quick call. Thanks!

smithago: Thanks for the details Richa, can we sync when Brandon is back next week on this. Putting If condition on every SQL object, how do we test it?

For the Fulltext catalog, we can move it into another SQL, which can be configured in the code to mention it is transactional or not.

brandonpollett: Without wrapping these in transactions, we need to be very careful about the changes that we make to the schema. I think this leads to more strict requirements where the server has to be able to run when the schema update has been partially applied.

One example of this is where you add/alter a new "NOT NULL" constraint on a column but the script fails before the stored procedures are updated. There are a few other scenarios where this is the case. 

I hesitate to make everything not part of a transaction if the only reason is full text. Are there other reasons to not use a transaction?

(this is an old documentation, could quickly find updated about what is not allowed in a transaction - https://docs.microsoft.com/en-us/previous-versions/sql/sql-server-2008-r2/ms191544(v=sql.105)?redirectedfrom=MSDN).

rbans96: Thanks Brandon, I totally agree we would need to be careful while making changes to the schema. We have considered that the schema changes are additive and not destructive. And binaries are backward compatible at least for one lesser schema version.

Here are the two parts:
1. Schema Migration.
As John mentioned back in the day - the motivation behind removing the transaction was performance. 
Using his example:
BEGIN TRANSACTION
 
ALTER TABLE ...
 
CREATE INDEX ...
 
END TRANSACTION
 
The ALTER TABLE statement holds an exclusive lock on the table which will only be released at the end of the transaction. This if we have expansive index creation statements that follow, this can result in blocking for extended periods of time.
 
So we agreed to made scripts as idempotent instead of transactional. 
We could definitely add variant of having a transaction around the ALTER/CREATE TABLE statements, which should be relatively quick. Then we have one or more transactions around the slower CREATE INDEX or UPDATE statements

2. Schema Initialization:
The motivation behind removing the transaction from initialization is mainly non transactional script(FULL TEXT).
I am gearing towards a hybrid approach here i.e. update schema manager tool to not add transaction for initialization but add transaction in the x.sql scripts. So in dicom 2.sql(considering 2 is the latest schema version), just after FULL TEXT script, we could begin transaction and end at the last.

rbans96: As per yesterday's discussion, have put up a PR for initial guidelines for SQLServer migration scripts - https://github.com/microsoft/healthcare-shared-components/pull/240 and updated the 2.diff.sql migration script in this PR to add transaction based on the nature of the statements.

brandonpollett (approved):

smithago (approved):

PR #777, opened 2021-05-13
QIDO support for additional core tags (without handling UTC offset + additional DT formats)

snehabandla:
[AB#81725](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/81725)
[AB#81727](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/81727)


src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.TagValueParser.cs
```diff
@@ -76,5 +105,15 @@ private static DateTime ParseDate(string date, string tagName)
 
             return parsedDate;
         }
+
+        private static DateTime ParseDateTime(string dateTime, string tagName)
+        {
+            if (!DateTime.TryParseExact(dateTime, DateTimeTagValueFormat, null, System.Globalization.DateTimeStyles.None, out DateTime parsedDateTime))
```
pengchen0692, line 111: Based on standard, there are multiple formats we need to support, do we want to align with that?
BTW, seems FODicom already have all supported formats: https://github.com/fo-dicom/fo-dicom/blob/6be9d198d05913c00b3c2f866cd51471064417aa/FO-DICOM.Core/DicomElement.cs#L914

Standard: http://dicom.nema.org/dicom/2013/output/chtml/part05/sect_6.2.html

snehabandla: Yes, i'll add that as well with additional tests. This was mostly to show the base addition.

src/Microsoft.Health.Dicom.SqlServer/Features/Query/SqlQueryGenerator.cs
```diff
@@ -376,11 +381,13 @@ public override void Visit(DateSingleValueMatchCondition dateSingleValueMatchCon
                 .Append("AND ");
 
             AppendExtendedQueryTagKeyFilter(dicomTagSqlEntry, tableAlias, dateSingleValueMatchCondition);
+            bool? dateTime = queryTag.VR == DicomVR.DT ? true : (queryTag.VR == DicomVR.DA ? false : null);
+            var value = dateTime.Value ? dateSingleValueMatchCondition.Value.ToString(SqlDateTimeFormat) : dateSingleValueMatchCondition.Value.ToString(SqlDateFormat);
```
pengchen0692, line 385: * how we support multiple time format in this way?

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.TagValueParser.cs
snehabandla: Actually, I've spent more time reviewing other formats while testing and I have many concerns that I'd like to bring up during standup.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Model/V2.Generated.cs
```diff
@@ -0,0 +1,577 @@
+//------------------------------------------------------------------------------
+// <auto-generated>
+//     This code was generated by a tool.
+//
+//     Changes to this file may cause incorrect behavior and will be lost if
+//     the code is regenerated.
+// </auto-generated>
+//------------------------------------------------------------------------------
+namespace Microsoft.Health.Dicom.SqlServer.Features.Schema.Model
```
smithago, line 9: where is the SQL file change that generated this? Should'nt this be v3?

snehabandla: I switched this to just have V3 be latest!

snehabandla: -- I just realized this is just QIDO so V3 can't be latest. Reverting that now! Will change it in the STOW PR.

PR #780, opened 2021-05-17
Run end to end tests sequentially

pengchen0692:
## Description
Run end to end tests sequentially

## Related issues
Addresses AB#82173

## Testing
Test locally

### DIFF
--- a/test/Microsoft.Health.Dicom.Web.Tests.E2E/Properties/AssemblyInfo.cs
+++ b/test/Microsoft.Health.Dicom.Web.Tests.E2E/Properties/AssemblyInfo.cs
@@ -1,0 +1,9 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+
+using Xunit;
+
+[assembly: CollectionBehavior(DisableTestParallelization = true)]


test/Microsoft.Health.Dicom.Web.Tests.E2E/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,9 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+
+using Xunit;
+
+[assembly: CollectionBehavior(DisableTestParallelization = true)]
```
smithago, line 9: This will slow down builds. Can we fix the test instead

smithago (changes requested):

test/Microsoft.Health.Dicom.Web.Tests.E2E/Properties/AssemblyInfo.cs
pengchen0692: The test is to verify change feeds, some approaches I could think about:

1. Block other tests cases to upload/delete files during execution, but I don't know if there is (easy) way could do that
2. Update the logic to verify all Upload actions, but not require to be exactly match. E.g:  We upload file1 and file2 for this test, meanwhile another test case uploads file3,  we verify return from GetChangeFeed contains uploading file1 and file2 instead of only contains them.

Regarding the slow down, I executed all end to end tests locally and sequentially, it totally takes about 15 seconds, seems not a concern for now.

smithago (approved):

test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/ChangeFeedTests.cs
```diff
@@ -17,6 +17,7 @@
 
 namespace Microsoft.Health.Dicom.Web.Tests.E2E.Rest
 {
+    [CollectionDefinition("Non-Parallel Collection", DisableParallelization = true)]
```
wsugarman, line 20: nit: perhaps a more DICOM-specific collection name

wsugarman (approved):

PR #787, opened 2021-05-19
Reindex operation structure code

pengchen0692:
## Description
Reindex operation structure code

## Related issues
Addresses [AB#82252](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/82252)

## Testing
All unit tests pass


src/Microsoft.Health.Dicom.Core/Features/Indexing/IReindexService.cs
```diff
@@ -0,0 +1,13 @@
+﻿// -------------------------------------------------------------------------------------------------
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Health.Dicom.Core.Features.ExtendedQueryTag;
+
+namespace Microsoft.Health.Dicom.Core.Features.Reindex
+{
+    public interface IReindexService
```
pengchen0692, line 9: it's confusing between reindex operation and reindex service

pengchen0692: See if able to get rid of it.
Add ReindexInstance in IIndexDataStore,

src/Microsoft.Health.Dicom.Core/Features/Indexing/IReindexTagStore.cs
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Health.Dicom.Core.Features.Indexing
+{
+    public interface IReindexTagStore
```
pengchen0692, line 12: IQueryTagReindexOperationStore

src/Microsoft.Health.Dicom.Core/Features/Indexing/IReindexTagStore.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        Task<IReadOnlyList<ReindexTagStoreEntry>> GetTagsOnOperationAsync(
```
pengchen0692, line 14: GetStoreEntiesAsync(string operationId, CancellationToken cancellationToken = default)

src/Microsoft.Health.Dicom.Core/Features/Indexing/ReindexTagStatus.cs
```diff
@@ -0,0 +1,13 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Dicom.Core.Features.Indexing
+{
+    public enum ReindexTagStatus
```
pengchen0692, line 8: QueryTagOperationStatus

src/Microsoft.Health.Dicom.Core/Features/Indexing/ReindexTagStatus.cs
```diff
@@ -0,0 +1,13 @@
...
+    {
+        Running = 0,
```
pengchen0692, line 10: Processing

src/Microsoft.Health.Dicom.Core/Features/Indexing/ReindexTagStoreEntry.cs
```diff
@@ -0,0 +1,20 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using Microsoft.Health.Dicom.Core.Features.ExtendedQueryTag;
+
+namespace Microsoft.Health.Dicom.Core.Features.Indexing
+{
+    public class ReindexTagStoreEntry
```
pengchen0692, line 10: QueryTagReindexOperationEntry

src/Microsoft.Health.Dicom.Core/Features/Indexing/ExtendedQueryTagOperationStatus.cs
```diff
@@ -0,0 +1,13 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Dicom.Core.Features.Indexing
+{
+    public enum ExtendedQueryTagOperationStatus
```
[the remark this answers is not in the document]
pengchen0692: TagOperationStatus?

src/Microsoft.Health.Dicom.Core/Features/Indexing/ExtendedQueryTagReindexOperationEntry.cs
```diff
@@ -0,0 +1,22 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using Microsoft.Health.Dicom.Core.Features.ExtendedQueryTag;
+
+namespace Microsoft.Health.Dicom.Core.Features.Indexing
+{
+    public class ExtendedQueryTagReindexOperationEntry
```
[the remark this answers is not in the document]
pengchen0692: TagReindexOperationEntry

src/Microsoft.Health.Dicom.Core/Features/Indexing/IExtendedQueryTagReindexOperationStore.cs
```diff
@@ -0,0 +1,24 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Health.Dicom.Core.Features.Indexing
+{
+    public interface IExtendedQueryTagReindexOperationStore
```
pengchen0692, line 12: ITagReindexOperationStore

src/Microsoft.Health.Dicom.Core/Features/Indexing/IExtendedQueryTagReindexOperationStore.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        Task<IReadOnlyList<ExtendedQueryTagReindexOperationEntry>> GetEntriesAsync(
```
pengchen0692, line 14: GetEntriesOfOperation

src/Microsoft.Health.Dicom.Core/Features/Indexing/IExtendedQueryTagReindexOperationStore.cs
```diff
@@ -0,0 +1,24 @@
...
+            string operationId,
+            CancellationToken cancellationToken = default);
+
+        Task<IReadOnlyList<long>> GetNextWatermarks(string operationId, int count, CancellationToken cancellationToken = default);
```
pengchen0692, line 18: OfOperation

src/Microsoft.Health.Dicom.Core/Features/Indexing/IExtendedQueryTagReindexOperationStore.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        Task UpdateEndWatermarkAsync(string operationId, long endWatermark, CancellationToken cancellationToken = default);
```
pengchen0692, line 20: OfOperation

src/Microsoft.Health.Dicom.Core/Features/Indexing/IExtendedQueryTagReindexOperationStore.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        Task CompleteReindexOperationAsync(string operationId, CancellationToken cancellationToken = default);
```
pengchen0692, line 22: CompleteOperation

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.cs
```diff
@@ -0,0 +1,162 @@
...
+            EnsureArg.IsNotNull(instanceReindexer, nameof(instanceReindexer));
+            _reindexConfig = configOptions.Value.Add;
+            _tagOperationStore = tagOperationStore;
+            _instanceReindexer = instanceReindexer;
+        }
+
+        [FunctionName(nameof(RunOrchestrator))]
+        public async Task RunOrchestrator(
+            [OrchestrationTrigger] IDurableOrchestrationContext context,
+            ILogger log)
```
pengchen0692, line 48: don't need log in parameter

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.cs
```diff
@@ -0,0 +1,162 @@
...
+                IReadOnlyList<long> watermarks = await context.CallActivityAsync<IReadOnlyList<long>>(nameof(FetchWatarmarksAsync), operationId);
+                if (watermarks.Count == 0)
+                {
+                    break;
+                }
+
+                // TODO: what if failed here?
+                // Fetch Query Tags
+                IReadOnlyList<ExtendedQueryTagStoreEntry> queryTags = await context.CallActivityAsync<IReadOnlyList<ExtendedQueryTagStoreEntry>>(
+                    nameof(FetchQueryTagsAsync),
```
pengchen0692, line 70: GetExtendedQueryTagsAsync

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.cs
```diff
@@ -0,0 +1,162 @@
...
+            string operationId = context.InstanceId;
+
+            log = context.CreateReplaySafeLogger(log);
+
+            log.LogInformation("Beginning to re-index data");
+
+            // TODO: should start a new orchestraion instead of while loop
+            while (true)
+            {
+                IReadOnlyList<long> watermarks = await context.CallActivityAsync<IReadOnlyList<long>>(nameof(FetchWatarmarksAsync), operationId);
```
pengchen0692, line 61: GetWatermarksAsync

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.cs
```diff
@@ -0,0 +1,162 @@
...
+                     new ReindexActivityInput() { TagEntries = queryTags, Watermark = watermark });
+                }
+
+                await context.CallActivityAsync(nameof(UpdateProgressAsync),
+                    new ReindexProgress { NextWatermark = watermarks.Min() - 1, OperationId = operationId });
+
+            }
+
+            await context.CallActivityAsync(nameof(CompleteReindexAsync), operationId);
+            log.LogInformation($"Completed re-indexing of data on operation {operationId}");
```
pengchen0692, line 92: better logging

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.cs
```diff
@@ -0,0 +1,162 @@
...
+        {
+            EnsureArg.IsNotNull(log, nameof(log));
+
+            // TODO: make number of watermarks configurable
+            var result = await _tagOperationStore.GetNextWatermarks(operationKey, 1);
+            return result;
+        }
+
+        [FunctionName(nameof(UpdateProgressAsync))]
+        public Task UpdateProgressAsync([ActivityTrigger] ReindexProgress progress, ILogger log)
```
pengchen0692, line 127: use same name as internally called one

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.cs
```diff
@@ -0,0 +1,162 @@
...
+            [HttpTrigger(AuthorizationLevel.Anonymous, "get", Route = nameof(StartReindexOperation))] HttpRequest req,
+            [DurableClient] IDurableOrchestrationClient client,
+            ILogger log)
+        {
+            EnsureArg.IsNotNull(req, nameof(req));
+            EnsureArg.IsNotNull(client, nameof(client));
+            EnsureArg.IsNotNull(log, nameof(log));
+            string operationId = req.Query["operationId"];
+            log.LogInformation($"Start processing operation {operationId}");
+            await client.StartNewAsync(nameof(RunOrchestrator), instanceId: operationId);
```
pengchen0692, line 158: in AddSerivce add todo to add job and start operation

src/Microsoft.Health.Dicom.Core/Features/Indexing/IInstanceReindexer.cs
```diff
@@ -0,0 +1,27 @@
...
+    public interface IInstanceReindexer
+    {
+        /// <summary>
+        /// Reindex DICOM instance of watermark on extended query tags.
+        /// </summary>
+        /// <param name="entries">Extended query tag store entries.</param>
+        /// <param name="watermark">The watermark to DICOM instance.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns>The task.</returns>
+        Task ReindexInstanceAsync(IEnumerable<ExtendedQueryTagStoreEntry> entries, long watermark, CancellationToken cancellationToken = default);
```
wsugarman, line 25: nit: I think we can be a bit more restrictive and ask for `ICollection<ExtendedQueryTagStoreEntry>`

src/Microsoft.Health.Dicom.Functions/Indexing/Models/ReindexInstanceInput.cs
```diff
@@ -4,27 +4,30 @@
...
-        public long End { get; set; }
+        public long Watermark { get; set; }
 
         /// <summary>
         /// Gets or sets the tags upon which to index.
         /// </summary>
-        public IReadOnlyList<ExtendedQueryTagEntry> TagEntries { get; set; }
+        public IReadOnlyList<ExtendedQueryTagStoreEntry> TagEntries { get; set; }
+
+        public override string ToString()
```
wsugarman, line 27: What is this `ToString()` for?

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.Activity.cs
```diff
@@ -0,0 +1,136 @@
...
+        /// <param name="entries">The entries.</param>
+        /// <param name="log">The log.</param>
+        /// <returns>The store entries.</returns>
+        [FunctionName(nameof(AddExtendedQueryTagAsync))]
+        public async Task<IEnumerable<ExtendedQueryTagStoreEntry>> AddExtendedQueryTagAsync([ActivityTrigger] IEnumerable<AddExtendedQueryTagEntry> entries, ILogger log)
+        {
+            EnsureArg.IsNotNull(entries, nameof(entries));
+            EnsureArg.IsNotNull(log, nameof(log));
+            log.LogInformation("Adding extended query tags {entries}", entries);
+            var response = await _addExtendedQueryTagService.AddExtendedQueryTagAsync(entries);
```
wsugarman, line 68: This is just echoing the input?

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.cs
```diff
@@ -0,0 +1,141 @@
...
+        [FunctionName(nameof(AddExtendedQueryTagsOrchestrationAsync))]
+        public async Task AddExtendedQueryTagsOrchestrationAsync(
+            [OrchestrationTrigger] IDurableOrchestrationContext context,
+            ILogger logger)
+        {
+            EnsureArg.IsNotNull(context, nameof(context));
+            logger = context.CreateReplaySafeLogger(logger);
+            var input = context.GetInput<IEnumerable<AddExtendedQueryTagEntry>>();
+            var tagEntries = await context.CallActivityAsync<IEnumerable<ExtendedQueryTagStoreEntry>>(nameof(AddExtendedQueryTagAsync), input);
+            await context.CallSubOrchestratorAsync(nameof(ReindexExtendedQueryTagsOrchestrationAsync), context.InstanceId, tagEntries);
```
wsugarman, line 64: I am unsure of the validity of passing the same `InstanceId` to the sub-orchestration. I was thinking instead we would pass the "parent" `InstanceId` as part of the input

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.cs
```diff
@@ -0,0 +1,141 @@
...
+            _addExtendedQueryTagService = addExtendedQueryTagService;
+        }
+
+        /// <summary>
+        ///  The orchestration function to add extended query tags.
+        /// </summary>
+        /// <param name="context">The context.</param>
+        /// <param name="logger">The logger.</param>
+        /// <returns>The task.</returns>
+        [FunctionName(nameof(AddExtendedQueryTagsOrchestrationAsync))]
```
wsugarman, line 55: I think we can shorten these names a bit, like `AddExtendedQueryTagAsync` for the function names

src/Microsoft.Health.Dicom.Core/Features/Indexing/ITagReindexOperationStore.cs
```diff
@@ -0,0 +1,57 @@
...
+            CancellationToken cancellationToken = default);
+
+        /// <summary>
+        ///  Get next watermarks for operation.
+        /// </summary>
+        /// <param name="operationId">The operation id.</param>
+        /// <param name="count">The count of watermarks</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns>Watermark list.</returns>
+        Task<IReadOnlyList<long>> GetNextWatermarksOfOperationAsync(string operationId, int count, CancellationToken cancellationToken = default);
```
wsugarman, line 35: What is this method doing? It seems like it should't have to worry about "count" or returning a list

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.cs
```diff
@@ -0,0 +1,141 @@
...
+
+                // Reindex
+                // TODO: process them parallel
+                foreach (long watermark in watermarks)
+                {
+                    await context
+                        .CallActivityAsync(nameof(ReindexInstanceAsync), new ReindexInstanceInput() { TagEntries = queryTags, Watermark = watermark });
+                }
+
+                await context.CallActivityAsync(nameof(UpdateEndWatermarkOfOperationAsync),
```
wsugarman, line 109: How does Update differ from Complete? Can they be the same?

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.cs
```diff
@@ -0,0 +1,141 @@
...
+                IReadOnlyList<ExtendedQueryTagStoreEntry> queryTags = await context
+                    .CallActivityAsync<IReadOnlyList<ExtendedQueryTagStoreEntry>>(nameof(GetExtendedQueryTagsOfOperationAsync), context.InstanceId);
+
+                if (queryTags.Count == 0)
+                {
+                    break;
+                }
+
+                // Reindex
+                // TODO: process them parallel
```
wsugarman, line 102: Is there a particular reason we're not doing this now? Just for testing?

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.cs
```diff
@@ -0,0 +1,141 @@
...
+            while (true)
+            {
+                IReadOnlyList<long> watermarks = await context
+                    .CallActivityAsync<IReadOnlyList<long>>(nameof(GetNextWatermarksOfOperationAsync), context.InstanceId);
+                if (watermarks.Count == 0)
+                {
+                    break;
+                }
+
+                IReadOnlyList<ExtendedQueryTagStoreEntry> queryTags = await context
```
wsugarman, line 93: If we're always querying for the set of tags (to eventually support pause/cancel), should we even be passing them to the orchestration in the first place?

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.Activity.cs
```diff
@@ -0,0 +1,136 @@
...
+        /// <param name="logger">The log.</param>
+        /// <returns>The task</returns>
+        [FunctionName(nameof(ReindexInstanceAsync))]
+        public Task ReindexInstanceAsync([ActivityTrigger] ReindexInstanceInput input, ILogger logger)
+        {
+            EnsureArg.IsNotNull(input, nameof(input));
+            EnsureArg.IsNotNull(logger, nameof(logger));
+
+            logger.LogInformation("Reindexing with {input}", input);
+            return _instanceReindexer.ReindexInstanceAsync(input.TagEntries, input.Watermark);
```
wsugarman, line 132: How will you be able to re-index a batch of data with a single watermark? You'll need a range, right? Unless you plan for this method to only re-index 1 entry.

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.cs
```diff
@@ -0,0 +1,141 @@
...
+        /// <returns>The task.</returns>
+        [FunctionName(nameof(AddExtendedQueryTagsOrchestrationAsync))]
+        public async Task AddExtendedQueryTagsOrchestrationAsync(
+            [OrchestrationTrigger] IDurableOrchestrationContext context,
+            ILogger logger)
+        {
+            EnsureArg.IsNotNull(context, nameof(context));
+            logger = context.CreateReplaySafeLogger(logger);
+            var input = context.GetInput<IEnumerable<AddExtendedQueryTagEntry>>();
+            var tagEntries = await context.CallActivityAsync<IEnumerable<ExtendedQueryTagStoreEntry>>(nameof(AddExtendedQueryTagAsync), input);
```
wsugarman, line 63: What if the tag already exists and it cannot be added? I would think we'd just stop processing immediately with a "Failed" status

src/Microsoft.Health.Dicom.Functions/Indexing/Models/ReindexInstanceInput.cs
pengchen0692: Just for logging -- log the input when ReindexInstance activity is started.

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.Activity.cs
pengchen0692: sorry, not get you. It just pass to input to AddExtendedQueryTagService which previously proceed the Add.

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.cs
pengchen0692: How about OperationId in TagOperationStore? We always use the parent InstanceId?

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.cs
pengchen0692: AddExtendedQueryTagsAsync is occupied by the httpTrigger...
The naming always a bit headache for me, Orchestration is really long word -- maybe AddExtendedQueryTagsOrcAsync?

src/Microsoft.Health.Dicom.Core/Features/Indexing/ITagReindexOperationStore.cs
pengchen0692: The count will be `BatchSize` * `MaxParallelBatches`.
this method call a storage procedure which return next set of watermarks we need to proceed. The number of watermarks is defined by `count`

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.cs
pengchen0692: The Update update progress of operation, and underlying purely just update Endwatermark of operation. 
The Complete completes the operation. It deletes operation from TagOperationStore, and update Tag status to be `Ready` in TagStore.
They could be putting together, but I feel as for now it's more clear to separate them.

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.cs
pengchen0692: It's not hard to make it parallel now, just want to build the min system which could up and running at this moment.

I'm not worry about this part, once the system start working, I'll add the parallel.

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.cs
pengchen0692: Sorry, not get you.  If we passing them to orchestration, who should call  GetExtendedQueryTagsOfOperationAsync to get qualified tags?

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.Activity.cs
pengchen0692: The activity will accept mutiple watermark to support the Batch later.
To reindex multiple watermarks here, we could either call  `_instanceReindexer.ReindexInstanceAsync` in multi-thread way. Or we can make `_instanceReindexer.ReindexInstanceAsync` to take a bunch of watermarks and process within.

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.cs
pengchen0692: yes, that's right. eh, problem with it?

wsugarman: I was just wondering because it doesn't appear to be gracefully handled. Ideally we'd set a custom status such that it can be easily seen when a user queries the status of an operation.

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.Activity.cs
wsugarman: Yeah, I worry that a 1:1 mapping for activity to watermark isn't the best way to scale, as a given activity could itself be multi-threaded while the activities themselves scale across a cluster. This is fine for the immediate term, but we'll need to change this before we're done with the first release of this feature.

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.cs
wsugarman: I mean that the orchestration doesn't need the tags in its input since the orchestration is already querying the tags from SQL directly.

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.cs
wsugarman: I am unsure if we want the "complete" activity to delete the operation from the DB yet. We're already planning to have a job that regularly cleans up state from the history log. It seems reasonable it could perform this clean up too.

src/Microsoft.Health.Dicom.Core/Features/Indexing/ITagReindexOperationStore.cs
wsugarman: This is because we're returning the exact watermarks right now instead of the batches?

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.cs
wsugarman: Well, we could rename the HTTP trigger too. It's technically not "adding," but rather "starting." So perhaps the HTTP trigger is `StartExtendedQueryTagAdditionAsync` while the orchestration is `AddExtendedQueryTagAsync`?

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.cs
wsugarman: Yeah, I think that's appropriate here

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.Activity.cs
wsugarman: Yeah, so the method just returns its input. Eg. `response` is the same as `entries`?

src/Microsoft.Health.Dicom.Functions/Indexing/Models/ReindexInstanceInput.cs
wsugarman: Let's not use `ToString()` like this. Instead, we can either have a method explicitly for logging or perform this in-line when logging if it only happens in one place. `ToString()` is about creating a string representation of an object and this format seems non-obvious outside of a logging our specific messages.

pengchen0692: Will do.

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.Activity.cs
pengchen0692: * After query tag saved to database, a tag key is appended, the return should include that.
* If any of tags fails to add (e.g: already exist), the whole add fails.

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.cs
pengchen0692: or `StartAddingExtendedQueryTagAsync`?

src/Microsoft.Health.Dicom.Core/Features/Indexing/ITagReindexOperationStore.cs
pengchen0692: I think I'm using a different way from your design. we could talk about this tomorrow.

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.cs
pengchen0692: we can talk about this tomorrow.

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.cs
pengchen0692: that's right, updating code...

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.Activity.cs
pengchen0692: changed the input to watermark array. does it solve problem?

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexOperation.Activity.cs
pengchen0692: this should have been resolved.

src/Microsoft.Health.Dicom.Core/Features/ExtendedQueryTag/AddExtendedQueryTagService.cs
```diff
@@ -42,8 +42,6 @@ public async Task<AddExtendedQueryTagResponse> AddExtendedQueryTagAsync(IEnumera
             IEnumerable<AddExtendedQueryTagEntry> result = extendedQueryTags.Select(item => item.Normalize());
 
             await _extendedQueryTagStore.AddExtendedQueryTagsAsync(result, _maxAllowedCount, cancellationToken);
-
-            // Current solution is synchronous, no job uri is generated, so always return blank response.
             return new AddExtendedQueryTagResponse();
```
pengchen0692: revert changes here

src/Microsoft.Health.Dicom.Functions/Indexing/Models/StartReindexActivityInput.cs
```diff
@@ -8,23 +8,19 @@
...
         /// </summary>
-        public long Start { get; set; }
+        public string OperationId { get; set; }
 
         /// <summary>
-        /// Gets or sets the exclusive ending watermark.
+        /// Gets or sets the tag keys
         /// </summary>
-        public long End { get; set; }
+        public IEnumerable<int> TagKeys { get; set; }
```
wsugarman, line 23: Generally, let's try to use some list-like interfaces. `IReadOnlyList<T>` when the users doesn't need to/should not modify it, `IList<T>` when we don't care about the implementation, and `List<T>` when it's used internally between our own private or internal methods/classes/etc

src/Microsoft.Health.Dicom.Functions/Indexing/Models/ReindexInstanceActivityInput.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        /// <summary>
+        /// Gets or sets the inclusive start watermark.
+        /// </summary>
+        public long StartWatermark { get; set; }
+
+        /// <summary>
+        /// Gets or sets the inclusive end watermark.
+        /// </summary>
+        public long EndWatermark { get; set; }
```
wsugarman, line 24: If its easier to user, we can also use a tuple here

src/Microsoft.Health.Dicom.Core/Features/Retrieve/IInstanceStore.cs
```diff
@@ -26,5 +26,15 @@ public interface IInstanceStore
             string seriesInstanceUid,
             string sopInstanceUid,
             CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// Gets idenfiers of instances amoung startWatermark and endWatermark.
```
wsugarman, line 31: nit: perhaps `...within the given range of watermarks.`

src/Microsoft.Health.Dicom.Core/Features/Indexing/ReindexStatus.cs
```diff
@@ -0,0 +1,28 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Dicom.Core.Features.Indexing
+{
+    /// <summary>
+    /// The reindex status.
+    /// </summary>
+    public enum ReindexStatus
```
wsugarman, line 11: Should this be indexing-specific? Or more general for operations?

src/Microsoft.Health.Dicom.Core/Features/Retrieve/IInstanceStore.cs
```diff
@@ -26,5 +26,15 @@ public interface IInstanceStore
...
+        /// </summary>
+        /// <param name="startWatermark">The inclusive start watermark</param>
+        /// <param name="endWatermark">The inclusive end watermark</param>
+        /// <param name="cancellationToken">The cancellation token</param>
+        /// <returns>The instanceidentifiers</returns>
+        Task<IEnumerable<VersionedInstanceIdentifier>> GetInstanceIdentifierAsync(long startWatermark, long endWatermark, CancellationToken cancellationToken = default);
```
wsugarman, line 37: nit: probably should be plural: `GetInstanceIdentifiersAsync`

wsugarman (changes requested):

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Orchestration.cs
```diff
@@ -0,0 +1,127 @@
...
+using Microsoft.Health.Dicom.Core.Features.Indexing;
+using Microsoft.Health.Dicom.Functions.Indexing.Models;
+
+namespace Microsoft.Health.Dicom.Functions.Indexing
+{
+    /// <summary>
+    /// Represents the Azure Durable Functions that perform the re-indexing of previously added DICOM instances
+    /// based on new tags configured by the user.
+    /// </summary>
+    public partial class ReindexDurableFunction
```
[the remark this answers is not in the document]
pengchen0692: * Access modifier is required for each file, otherwise got build failure..

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Trigger.cs
```diff
@@ -0,0 +1,78 @@
...
+        [FunctionName(nameof(StartAddingExtendedQueryTagHttpTriggerAsync))]
+        public async Task StartAddingExtendedQueryTagHttpTriggerAsync(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "Post", Route = "extendedquerytags")] HttpRequestMessage request,
+            [DurableClient] IDurableOrchestrationClient client,
+            ILogger logger)
+        {
+            EnsureArg.IsNotNull(request, nameof(request));
+            EnsureArg.IsNotNull(client, nameof(client));
+            EnsureArg.IsNotNull(logger, nameof(logger));
+            var extendedQueryTags = await request.Content.ReadAsAsync<IEnumerable<AddExtendedQueryTagEntry>>();
```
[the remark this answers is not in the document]
pengchen0692: * As my understand, reason for `List` is it provides more capability while deserialization need same effort as `IEnumerable`? 
*  Do you mean we ultimately reserialize JSON by ourselve instead of utilize `ReadAsAsync`?

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Trigger.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        /// <summary>
+        /// The http trigger to add extended Query tags
+        /// </summary>
+        /// <param name="request">The http request.</param>
+        /// <param name="client">The client.</param>
+        /// <param name="logger">The logger.</param>
+        /// <returns>The task.</returns>
+        [FunctionName(nameof(StartAddingExtendedQueryTagHttpTriggerAsync))]
+        public async Task StartAddingExtendedQueryTagHttpTriggerAsync(
```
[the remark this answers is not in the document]
pengchen0692: I'm trying to build a pattern for durable azure function,  here are some ideas, feel free for any feedback. I have updated code based on these ideas.
* Should avoid (method/class) name which contains more than 4 words, it make code difficult to read (e.g: StartAddingExentededQueryTagHttpTriggerAsync contains 8 words). one idea is use 'Tag' to replace `ExtendedQueryTag`.
* Since Trigger/Activity/Orchestration method has special annotation, we don't need to append types(HttpTrigger e.g) in methods (imitating https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-classes-structs-and-interfaces#naming-enumerations)

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Trigger.cs
```diff
@@ -0,0 +1,78 @@
...
+            [HttpTrigger(AuthorizationLevel.Anonymous, "Post", Route = "extendedquerytags")] HttpRequestMessage request,
+            [DurableClient] IDurableOrchestrationClient client,
+            ILogger logger)
+        {
+            EnsureArg.IsNotNull(request, nameof(request));
+            EnsureArg.IsNotNull(client, nameof(client));
+            EnsureArg.IsNotNull(logger, nameof(logger));
+            var extendedQueryTags = await request.Content.ReadAsAsync<IEnumerable<AddExtendedQueryTagEntry>>();
+            logger.LogInformation("Start adding extended query tags {input}", extendedQueryTags);
+            await client.StartNewAsync(nameof(AddExtendedQueryTagsOrchestrationAsync), instanceId: null, extendedQueryTags);
```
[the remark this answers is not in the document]
pengchen0692: Just naively return instanceId as content for now, need to be updated based on contract to client

src/Microsoft.Health.Dicom.Functions/Indexing/Models/StartReindexActivityInput.cs
pengchen0692: Guess the reason is `List` kind types(`IList`, 'IReadonlyList','List') provides more capability, so makes code easier?
Also List is very commonly used when we don't know the size?

src/Microsoft.Health.Dicom.Functions/Indexing/Models/ReindexInstanceActivityInput.cs
pengchen0692: To use elements within tuple, need to call like `Tuple.Item1` `Tuple.Item2`, which I think unclear.

src/Microsoft.Health.Dicom.Core/Features/Indexing/ReindexStatus.cs
pengchen0692: I'm good with indexing-specific -- although so far I'm not sure if we need `Paused` and `Completed` for deindexing. :)
My concern with promoting to operation level  is that 'Paused' is special, it's on `Tag` instead of `Operation`

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Orchestration.cs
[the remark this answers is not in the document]
wsugarman: I can perhaps later fix the requirement for access modifiers per file through the .editorconfig if possible

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Orchestration.cs
```diff
@@ -0,0 +1,127 @@
...
+        /// <param name="logger">The logger.</param>
+        /// <returns>The task.</returns>
+        [FunctionName(nameof(AddExtendedQueryTagsOrchestrationAsync))]
+        public async Task AddExtendedQueryTagsOrchestrationAsync(
+            [OrchestrationTrigger] IDurableOrchestrationContext context,
+            ILogger logger)
+        {
+            EnsureArg.IsNotNull(context, nameof(context));
+            logger = context.CreateReplaySafeLogger(logger);
+            var input = context.GetInput<IEnumerable<AddExtendedQueryTagEntry>>();
```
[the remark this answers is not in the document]
wsugarman: More generally speaking, I think for data structures like the input types, list-like types are appropriate because it allows us to check its contents without enumerating the collection.

On the other hand, for methods and use cases where we simply need to enumerate the elements, regardless of size, then IEnumerable<T> is acceptable. (eg. where having 0 elements is valid). But when we start to want to verify the number of elements, or we want to do more like indexing, etc, then list starts to look much more attractive.

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Trigger.cs
```diff
@@ -0,0 +1,83 @@
...
+            EnsureArg.IsNotNull(request, nameof(request));
+            EnsureArg.IsNotNull(client, nameof(client));
+            EnsureArg.IsNotNull(logger, nameof(logger));
+            var extendedQueryTags = await request.Content.ReadAsAsync<List<AddExtendedQueryTagEntry>>();
+            logger.LogInformation("Start adding extended query tags {input}", extendedQueryTags);
+            string instanceId = await client.StartNewAsync(nameof(AddAndReindexTagsAsync), instanceId: null, extendedQueryTags);
+            logger.LogInformation("Started new orchestration with instanceId {instancId}", instanceId);
+
+            // TODO: these code need to be updated based on contract to client.
+            return new HttpResponseMessage() { Content = new StringContent(instanceId) };
```
wsugarman, line 79: nit: () is not necessary

src/Microsoft.Health.Dicom.Functions/Microsoft.Health.Dicom.Functions.csproj
```diff
@@ -11,15 +11,19 @@
     <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.27" />
     <PackageReference Include="Microsoft.Azure.WebJobs.Extensions" Version="4.0.1" />
     <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="2.4.3" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="3.1.14" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="3.1.14" />
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="3.1.14" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="3.1.14" />
-    <PackageReference Include="Microsoft.Extensions.Options" Version="3.1.14" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="3.1.15" />
```
wsugarman, line 14: I think after merging the the feature branch, these versions should be the MSBuild property instead of the version literal

wsugarman (approved):

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Orchestration.cs
```diff
@@ -0,0 +1,122 @@
+// -------------------------------------------------------------------------------------------------
```
wsugarman, line 1: There look to be a number of pending comments from my last iteration that still need review

Microsoft.Health.Dicom.sln
```diff
@@ -39,18 +39,24 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Health.Dicom.SqlS
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Health.Dicom.Client", "src\Microsoft.Health.Dicom.Client\Microsoft.Health.Dicom.Client.csproj", "{D100EA60-8DC8-4576-A177-56BC7193BF4A}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Health.Dicom.Functions", "src\Microsoft.Health.Dicom.Functions\Microsoft.Health.Dicom.Functions.csproj", "{BB88616C-6208-43D0-B9EF-79FC0652A151}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Health.Dicom.Functions", "src\Microsoft.Health.Dicom.Functions\Microsoft.Health.Dicom.Functions.csproj", "{BB88616C-6208-43D0-B9EF-79FC0652A151}"
```
wsugarman, line 42: Looks like we require a merge with the feature branch too

wsugarman (changes requested):

src/Microsoft.Health.Dicom.Core/Features/ExtendedQueryTag/IExtendedQueryTagStore.cs
```diff
@@ -40,5 +40,13 @@ public interface IExtendedQueryTagStore
...
         /// <returns>The task.</returns>
         Task DeleteExtendedQueryTagAsync(string tagPath, string vr, CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// Get extended query tags by keys.
+        /// </summary>
+        /// <param name="tagKeys">The tag keys.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns>The task.</returns>
+        Task<IReadOnlyList<ExtendedQueryTagStoreEntry>> GetExtendedQueryTagsByKeyAsync(IReadOnlyList<int> tagKeys, CancellationToken cancellationToken = default);
```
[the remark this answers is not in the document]
pengchen0692: I chose a lazy way -- without ByKey, I have to update a lot of unit tests. :)
I will fix it.

src/Microsoft.Health.Dicom.Core/Features/Indexing/IReindexStore.cs
```diff
@@ -0,0 +1,51 @@
...
+        Task<IEnumerable<ReindexEntry>> GetReindexEntriesAsync(string operationId, CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// Update reindex progress.
+        /// </summary>
+        /// <param name="operationId">The operation id.</param>
+        /// <param name="endWatermark">The end watermark.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns>The task.</returns>
+        Task UpdateReindexProgressAsync(string operationId, long endWatermark, CancellationToken cancellationToken = default);
```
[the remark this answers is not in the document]
pengchen0692: `UpdateReindexProgressAsync`:  Update `endwatermark`
`CompleteReindexAsync`: 
  * Update reindex status of all tags associated with this operation to be `Completed`
  * Update tag status in `TagStore` to be `Ready`

So they are doing different works, so I feel it's better to be in separate methods to keep `Single Responsibility`

src/Microsoft.Health.Dicom.Functions/Indexing/Models/ReindexInstanceInput.cs
```diff
@@ -4,27 +4,28 @@
...
     /// </summary>
-    public class ReindexBatch
+    public class ReindexInstanceInput
     {
         /// <summary>
-        /// Gets or sets the inclusive starting watermark.
+        /// Gets or sets the inclusive start watermark.
         /// </summary>
-        public long Start { get; set; }
+        public long StartWatermark { get; set; }
```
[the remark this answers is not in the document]
pengchen0692: Cool, never know this before, will update

src/Microsoft.Health.Dicom.Core/Features/Indexing/IInstanceReindexer.cs
pengchen0692: Changed to IReadonlyList

src/Microsoft.Health.Dicom.Core/Features/Indexing/IReindexStore.cs
```diff
@@ -0,0 +1,51 @@
...
+        Task CompleteReindexAsync(string operationId, CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// Start reindex.
+        /// </summary>
+        /// <param name="tagKeys">Key to tags</param>
+        /// <param name="operationId">The operation id.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns>The task.</returns>
+        Task<ReindexOperation> PrepareReindexingAsync(IEnumerable<int> tagKeys, string operationId, CancellationToken cancellationToken = default);
```
pengchen0692, line 49: Change this to IReadonlyList

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Orchestration.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+
+            // Reindex
+            long start = reindexOperation.StartWatermark;
+            long end = reindexOperation.EndWatermark;
+
+            List<Task> batches = new List<Task>();
+
+            // pickup next section and send task to activity
+            for (int parallel = 0; parallel < _reindexConfig.MaxParallelBatches; parallel++)
```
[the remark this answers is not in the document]
pengchen0692: When `end<start`, no tasks are added to batches.

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Trigger.cs
```diff
@@ -0,0 +1,78 @@
...
+        public async Task StartAddingExtendedQueryTagHttpTriggerAsync(
+            [HttpTrigger(AuthorizationLevel.Anonymous, "Post", Route = "extendedquerytags")] HttpRequestMessage request,
+            [DurableClient] IDurableOrchestrationClient client,
+            ILogger logger)
+        {
+            EnsureArg.IsNotNull(request, nameof(request));
+            EnsureArg.IsNotNull(client, nameof(client));
+            EnsureArg.IsNotNull(logger, nameof(logger));
+            var extendedQueryTags = await request.Content.ReadAsAsync<IEnumerable<AddExtendedQueryTagEntry>>();
+            logger.LogInformation("Start adding extended query tags {input}", extendedQueryTags);
```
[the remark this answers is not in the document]
pengchen0692: yeah, the log problem is not solved in this PR, I plan to discuss this with you and fix it with one commit -- you might notice not only just here , but also several other places have this issue.
 e.g `log.LogInformation("Updating reindex progress with {input}", input);`
The input type is `UpdateReindexProgressInput`, clearly won't log properly.

I remember you mention `ToString()` is not a good idea, but so far I didn't figure out a better way. Maybe we could discuss offline on this.

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Orchestration.cs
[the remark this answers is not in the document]
pengchen0692: I'm updating all `IEnumerable` to `IReadOnlyList`, if I miss any place, sorry about that. 😊

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Orchestration.cs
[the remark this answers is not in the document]
pengchen0692: Will do a check so that easier to understand

pengchen0692:
Overall, I'm thinking should we create structure `WatermarkRange` since it used everywhere.

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStoreV1.cs
```diff
@@ -330,5 +330,6 @@ SELECT MIN({VLatest.DeletedInstance.DeletedDateTime})
                 }
             }
         }
+
```
pengchen0692, line 333: Remove this line

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Trigger.cs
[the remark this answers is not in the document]
wsugarman: In these cases, we're trying to log built-in types/collections so ToString() cannot be used. So in this case we'll need to format them here, like `string.Join(...)`. For a given element, or when logging a user-defined type (UDT) depends. Often times we may not necessarily log an entire object. Perhaps we'd like to instead log IDs or something a bit smaller. Furthermore, perhaps we should be logging the raw input for APIs if it's text/json/xml/non-binary. That way we can dump the user input before parsing occurs, etc.

src/Microsoft.Health.Dicom.Core/Features/ExtendedQueryTag/IExtendedQueryTagStore.cs
```diff
@@ -40,5 +40,13 @@ public interface IExtendedQueryTagStore
...
         /// <returns>The task.</returns>
         Task DeleteExtendedQueryTagAsync(string tagPath, string vr, CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// Get extended query tags by keys.
+        /// </summary>
+        /// <param name="tagKeys">The tag keys.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <returns>The task.</returns>
+        Task<IReadOnlyList<ExtendedQueryTagStoreEntry>> GetExtendedQueryTagsAsync(IReadOnlyList<int> tagKeys, CancellationToken cancellationToken = default);
```
wsugarman, line 50: nit: let's move this next to the other overload

src/Microsoft.Health.Dicom.Core/Features/Indexing/ReindexOperation.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        /// <summary>
+        /// Gets or sets query tags store entries.
+        /// </summary>
+        public IReadOnlyList<ExtendedQueryTagStoreEntry> StoreEntries { get; set; }
+
+        /// <summary>
+        /// The inclusive start watermark.
+        /// </summary>
+        public long StartWatermark { get; set; }
```
wsugarman, line 29: If you decide to create a structure to represent this tuple, then for convenience please implement a deconstruct method detailed here: https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/deconstruct#user-defined-types

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Orchestration.cs
```diff
@@ -0,0 +1,124 @@
...
+        [FunctionName(nameof(AddAndReindexTagsAsync))]
+        public async Task AddAndReindexTagsAsync(
+            [OrchestrationTrigger] IDurableOrchestrationContext context,
+            ILogger logger)
+        {
+            EnsureArg.IsNotNull(context, nameof(context));
+            logger = context.CreateReplaySafeLogger(logger);
+            var input = context.GetInput<IReadOnlyList<AddExtendedQueryTagEntry>>();
+            var storeEntires = await context.CallActivityAsync<IReadOnlyList<ExtendedQueryTagStoreEntry>>(
+                nameof(AddTagsAsync),
```
wsugarman, line 37: nit: the indent doesn't look correct for these args. An extra space?

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Orchestration.cs
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        /// <summary>
+        /// The orchestration function to Reindex Extended Query Tags.
+        /// </summary>
+        /// <param name="context">the context.</param>
+        /// <param name="logger">The logger.</param>
+        /// <returns></returns>
+        [FunctionName(nameof(ReindexTagsAsync))]
+        public async Task ReindexTagsAsync([OrchestrationTrigger] IDurableOrchestrationContext context,
```
wsugarman, line 52: nit: put `[OrchestrationTrigger] IDurableOrchestrationContext context,` on the next line

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Orchestration.cs
```diff
@@ -0,0 +1,124 @@
...
+
+            ReindexOperation reindexOperation = context.GetInput<ReindexOperation>();
+
+            // EndWatermark < 0 means there are no instances.
+            if (reindexOperation.EndWatermark >= 0)
+            {
+                IReadOnlyList<ExtendedQueryTagStoreEntry> queryTags = await context
+                    .CallActivityAsync<IReadOnlyList<ExtendedQueryTagStoreEntry>>(nameof(GetProcessingTagsAsync), reindexOperation.OperationId);
+
+                if (queryTags.Any())
```
wsugarman, line 66: nit: use `Count > 0`

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Orchestration.cs
```diff
@@ -0,0 +1,124 @@
...
+                        };
+
+                        context.StartNewOrchestration(nameof(ReindexTagsAsync), input: newInput);
+                    }
+                }
+            }
+            await context.CallActivityAsync(nameof(CompleteReindexingTagsAsync), reindexOperation.OperationId);
+        }
+
+        private async Task<long> ReindexNextSegmentAsync(IDurableOrchestrationContext context,
```
wsugarman, line 86: nit: put `IDurableOrchestrationContext context,` on the next line

wsugarman (approved):
In a future PR I think we'll need to figure out how to do versioning. I think initially I proposed putting the version in the name of orchestration to isolate its state.

src/Microsoft.Health.Dicom.Core/Features/Indexing/ReindexOperation.cs
pengchen0692: will do this later, this PR has been there for too long time.

PR #794, opened 2021-05-24
Correlation ID / Request ID for Requests.

v-shaaal:
## Description
Currently, correlation id or request id is empty. Hence, there is no way to look at the logs and know about what request the customer is exactly complaining about. This user story should fix that issue.

Following is the pipeline that could be followed. Once the user's request fails, user will look into the logs exposed to them. The logs will have a correlation id or a request id. Customer would tell us the request id that failed. We will look at either `LinuxRequestTelemetry` table or `LinuxExceptionTelemetry` table to get more details about the request. There should be a column called operation id which should correlate to this correlation id or request id field.

## Related issues
[AB#82312](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/82312).

### DIFF
--- a/src/Microsoft.Health.Dicom.Api/Features/Context/DicomRequestContextMiddleware.cs
+++ b/src/Microsoft.Health.Dicom.Api/Features/Context/DicomRequestContextMiddleware.cs
@@ -45,6 +45,7 @@ namespace Microsoft.Health.Dicom.Api.Features.Context
                 method: request.Method,
                 uri,
                 baseUri,
+                correlationId: System.Diagnostics.Activity.Current?.RootId,
                 context.Request.Headers,
                 context.Response.Headers);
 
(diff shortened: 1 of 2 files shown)


smithago (approved):

src/Microsoft.Health.Dicom.Api/Features/Context/DicomRequestContextMiddleware.cs
```diff
@@ -45,6 +45,7 @@ public async Task Invoke(HttpContext context, IDicomRequestContextAccessor dicom
                 method: request.Method,
                 uri,
                 baseUri,
+                correlationId: System.Diagnostics.Activity.Current?.RootId,
```
wsugarman, line 48: What is this value? How is it set?

v-shaaal: Each request received by the application is treated as Diagnostic Activity and is assigned Activity ID for debugging purposes. 
I read about it from the following sources:
- https://docs.microsoft.com/en-us/dotnet/core/diagnostics/distributed-tracing
- https://docs.microsoft.com/en-us/dotnet/core/diagnostics/distributed-tracing-concepts

I also confirmed that the correlation id now correlates with the request id captured in LinuxRequestTelemetry table by application insights logs.

PR #802, opened 2021-05-27
Operation Id to KubernetesContainers

v-shaaal:
## Description
Added operation id to logs pushed to KubernetesContainers. This will help correlating the logs in LinuxRequestTelemetry.

`operationId` field is added as a property so won't be available as a separate field. Hence, a workaround will be required to join data from LinuxRequestTelemetry table to KubernetesContainers. Following is a Kusto query snippet that can be used:

> LinuxRequestTelemetry
> | where PreciseTimeStamp >= ago(15m)
> | where resultCode != 200
> | join (
> KubernetesContainers
> | where PodName contains "ali0427"
> | where PreciseTimeStamp >= ago(15m)
> | extend operation_Id = extract(@'operationId":"([a-z0-9]*)"', 1, log)
> ) on operation_Id

## Related issues
[AB#82459](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/82459).

### DIFF
--- a/src/Microsoft.Health.Dicom.Api/Features/Context/DicomRequestContextMiddleware.cs
+++ b/src/Microsoft.Health.Dicom.Api/Features/Context/DicomRequestContextMiddleware.cs
@@ -15,6 +15,7 @@ namespace Microsoft.Health.Dicom.Api.Features.Context
 {
     public class DicomRequestContextMiddleware
     {
+        private const string OperationIdName = "operationId";
         private readonly RequestDelegate _next;
 
         public DicomRequestContextMiddleware(RequestDelegate next)
@@ -53,7 +54,7 @@ namespace Microsoft.Health.Dicom.Api.Features.Context
             dicomRequestContextAccessor.RequestContext = dicomRequestContext;
 
             // Adding operation id for this request thread to the log context.
-            using (LogContext.PushProperty("operationId", System.Diagnostics.Activity.Current?.RootId))
+            using (LogContext.PushProperty(OperationIdName, System.Diagnostics.Activity.Current?.RootId))
             {
                 // Call the next delegate/middleware in the pipeline
                 await _next(context);
--- a/src/Microsoft.Health.Dicom.Api/Features/Context/DicomRequestContextMiddleware.cs
+++ b/src/Microsoft.Health.Dicom.Api/Features/Context/DicomRequestContextMiddleware.cs
@@ -9,6 +9,7 @@ using EnsureThat;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.Health.Dicom.Core.Features.Context;
+using Serilog.Context;
 
 namespace Microsoft.Health.Dicom.Api.Features.Context
 {
@@ -51,8 +52,12 @@ namespace Microsoft.Health.Dicom.Api.Features.Context
 
             dicomRequestContextAccessor.RequestContext = dicomRequestContext;
 
-            // Call the next delegate/middleware in the pipeline
-            await _next(context);
+            // Adding operation id for this request thread to the log context.
+            using (LogContext.PushProperty("operationId", System.Diagnostics.Activity.Current?.RootId))
+            {
+                // Call the next delegate/middleware in the pipeline
+                await _next(context);
+            }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Dicom.Api/Features/Context/DicomRequestContextMiddleware.cs
```diff
@@ -51,8 +52,12 @@ public async Task Invoke(HttpContext context, IDicomRequestContextAccessor dicom
 
             dicomRequestContextAccessor.RequestContext = dicomRequestContext;
 
-            // Call the next delegate/middleware in the pipeline
-            await _next(context);
+            // Adding operation id for this request thread to the log context.
+            using (LogContext.PushProperty("operationId", System.Diagnostics.Activity.Current?.RootId))
```
smithago, line 56: Did'nt your previous change already add this to the DicomRequestContext to show up in Exceptions. How come it shows up in Exceptions and not in Logs?

v-shaaal: That was added just for the audit logs which are different than usual trace logs.

v-shaaal: For clarification, change in https://github.com/microsoft/dicom-server/pull/794 puts correlation Id for logs and exception logs that end up in Audit Logs (IFx and Shoebox). They follow a separate path than this. This information, however, was missing for trace logs that end up in KubernetesContainers table which this pull request should fix.

smithago: It still does not make sense. If the linked PR was for audit logs how did it fix LinuxExceptionTelemetry? I though Audit logs are different. 
I would think Exceptions and normal logs would go through same pipeline.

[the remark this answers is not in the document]
v-shaaal: I agree, it's usually a good idea to have these property names as constants so I will make that change. However, this property will not be explicitly referenced elsewhere at least in this code.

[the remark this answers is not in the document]
v-shaaal: That was exactly my thought as well. I spent some time trying to find a way to get this done without making it serilog-specific but couldn't succeed so made this change for now. We can revisit this in future to make this better, and have it like this for now so that we get the operation id to correlate the logs.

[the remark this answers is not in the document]
v-shaaal: Updated.

v-shaaal: It did not :)
It fixed the audit logs, i.e. the IFx Audit Logs and the Azure Monitor Shoebox Logs. The reason I mentioned LinuxRequestTelemetry and LinuxExceptionTelemetry tables in the previous PR is that now customers will see CorrelationId in their logs i.e. the Audit Logs. When customers provide us with the correlation id, we can use that correlation id and get more details from LinuxRequestTelemetry and LinuxExceptionTelemetry table. So the previous PR had the change for both exceptions and normal audit logs.
This PR is focused on adding correlation id to the trace logs that end up in KubernetesContainers table. After this PR, we can use the same correlation id provided by the user and get further information using the trace logs in KubernetesContainers table.

src/Microsoft.Health.Dicom.Api/Features/Context/DicomRequestContextMiddleware.cs
```diff
@@ -51,8 +53,12 @@ public async Task Invoke(HttpContext context, IDicomRequestContextAccessor dicom
 
             dicomRequestContextAccessor.RequestContext = dicomRequestContext;
 
-            // Call the next delegate/middleware in the pipeline
-            await _next(context);
+            // Adding operation id for this request thread to the log context.
+            using (LogContext.PushProperty(OperationIdName, System.Diagnostics.Activity.Current?.RootId))
```
smithago, line 57: I don't see my previous comments, can we sync on this.

v-shaaal: I have reactivated the previous conversation and replied to that. Feel free to ping me whenever is feasible for you, and we can connect.

src/Microsoft.Health.Dicom.Api/Features/Context/DicomRequestContextMiddleware.cs
smithago: LinuxExceptionTelemetry does not even have Operation_id. How will be relate it? Was'nt the previous bug to allow joining of Request and Exception telemetry? How can we join if the Operation_id is not populated

v-shaaal: Actually LinuxExceptionTelemetry does have operation_id. Probably what you saw is data from IoT and FHIR. Let's get on a call. I have set up some custom exceptions so I can present a small demo to clarify these things.

PR #815, opened 2021-06-04
Add Function Test Boilerplate

wsugarman:
## Description
- Initialize test project
- Fix packages
- Multi-target common test project
- Disable sample timer task

## Related issues
N/A

## Testing
Build passes

### DIFF
--- a/src/Microsoft.Health.Dicom.Functions/Indexing/ReindexFunction.cs
+++ b/src/Microsoft.Health.Dicom.Functions/Indexing/ReindexFunction.cs
@@ -121,7 +121,7 @@ namespace Microsoft.Health.Dicom.Functions.Indexing
             return Task.CompletedTask;
         }
 
-        [FunctionName("ReindexV1_Timer")]
+        //[FunctionName("ReindexV1_Timer")]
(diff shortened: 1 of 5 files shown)


src/Microsoft.Health.Dicom.Functions/Indexing/ReindexFunction.cs
```diff
@@ -121,7 +121,7 @@ public static Task ProcessBatchAsync([ActivityTrigger] ReindexBatch batch, ILogg
             return Task.CompletedTask;
         }
 
-        [FunctionName("ReindexV1_Timer")]
+        //[FunctionName("ReindexV1_Timer")]
```
pengchen0692, line 124: remove this change?

pengchen0692 (approved):
one single minor comment

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexFunction.cs
wsugarman: I'll keep it for now as this timer may interfere with local debugging and is probably unnecessary in the end

PR #819, opened 2021-06-08
Durable Functions History Purge

yuedai0531:
## Description
Purges the history of durable functions that respects a set of conditions.

## Testing
Unit tests testing for when:
- Valid Query Conditions return a list of 0 elements, thus nothing to delete.
- Valid Query Conditions return a list of 3 elements to delete and a continuation token to go to next page.


src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
```diff
@@ -0,0 +1,55 @@
...
+                EnsureArg.IsNotNull(tokenSource, nameof(tokenSource));
+                CancellationToken token = tokenSource.Token;
+
+                OrchestrationStatusQueryResult listOfOrchestrators = await client.ListInstancesAsync(condition, token);
+
+                // Loop through the orchestration instances and purge them.
+                foreach (var orch in listOfOrchestrators.DurableOrchestrationState)
+                {
+                    await client.PurgeInstanceHistoryAsync(orch.InstanceId);
+                }
```
yuedai0531, line 47: As this:
```
        [Obsolete]
        Task<IList<DurableOrchestrationStatus>> GetStatusAsync(DateTime? createdTimeFrom = null, DateTime? createdTimeTo = null, IEnumerable<OrchestrationRuntimeStatus> runtimeStatus = null, CancellationToken cancellationToken = default);
```
is obsolete, we use this:
```
        Task<OrchestrationStatusQueryResult> ListInstancesAsync(OrchestrationStatusQueryCondition condition, CancellationToken cancellationToken);
```

However, the current purge instance history has similar parameters as the obsolete function to get orchestration instances.
```
Task<PurgeHistoryResult> PurgeInstanceHistoryAsync(DateTime createdTimeFrom, DateTime? createdTimeTo, IEnumerable<OrchestrationStatus> runtimeStatus);
```
This is why I loop through the orchestrators.

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
```diff
@@ -0,0 +1,55 @@
...
+            ILogger log)
+        {
+            log.LogInformation("Durable Functions cleanup started.");
+
+            EnsureArg.IsNotNull(req, nameof(req));
+
+            EnsureArg.IsNotNull(client, nameof(client));
+
+            // Specify conditions for Function Queries.
+            OrchestrationStatusQueryCondition condition = new OrchestrationStatusQueryCondition();
```
wsugarman, line 34: Don't we need to define some conditions for selecting instances? Because as of now we'll just purge whatever the default conditions return (which is a bit scary!)

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
```diff
@@ -0,0 +1,55 @@
...
+
+            using (CancellationTokenSource tokenSource = new CancellationTokenSource())
```
wsugarman, line 36: I think Azure functions allows you to have a token in the method signature that propagated from the framework. It makes less sense for there to be a `CancellationTokenSource` declared here because we're not cancelling the token under any condition.

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
```diff
@@ -0,0 +1,55 @@
...
+            {
+                EnsureArg.IsNotNull(tokenSource, nameof(tokenSource));
```
wsugarman, line 38: You don't need to `null`-check here as you just made the value with a constructor (sometimes abbreviated to "ctor")

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
```diff
@@ -0,0 +1,55 @@
...
+                CancellationToken token = tokenSource.Token;
```
wsugarman, line 39: You can skip these sorts of assignments if you're just going to forward the result to the method invocation below

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
```diff
@@ -0,0 +1,55 @@
...
+                OrchestrationStatusQueryResult listOfOrchestrators = await client.ListInstancesAsync(condition, token);
+
+                // Loop through the orchestration instances and purge them.
+                foreach (var orch in listOfOrchestrators.DurableOrchestrationState)
+                {
+                    await client.PurgeInstanceHistoryAsync(orch.InstanceId);
+                }
+            }
+
+            string responseMessage = $"Durable Functions cleanup executed successfully.";
```
wsugarman, line 50: nit: don't use string "interpolation" if you don't have any arguments to format

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
```diff
@@ -0,0 +1,55 @@
...
+
+            return new OkObjectResult(responseMessage);
```
wsugarman, line 52: I think you could even return a `string` from this method instead of `IActionResult`

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
wsugarman: You may want to retrieve some of these conditions from a configuration, which means the function may need to be an instance method (non-`static`).

wsugarman (changes requested):

src/Microsoft.Health.Dicom.Functions/Indexing/Configuration/ReindexConfiguration.cs
```diff
@@ -39,5 +41,13 @@ public class ReindexConfiguration
         ///// </summary>
         //[Range(0d, 1d)]
         //public double MaxErrorPercentage { get; set; } = 1;
+
+        public bool IsPurgeErrorEnabled { get; set; } = true;
+
+        public IEnumerable<OrchestrationRuntimeStatus> RuntimeStatus { get; } = new[] { OrchestrationRuntimeStatus.Completed };
+
+        public DateTime CreatedTimeFrom { get; set; } = DateTime.UtcNow.Subtract(TimeSpan.FromDays(7));
```
wsugarman, line 49: The configuration is resolved upon starting the function app -- not at the time of starting each function invocation. So something like the start and end times cannot be present here. Instead, you could have perhaps a "minimum age" for deletion, like 2 weeks or a month.

src/Microsoft.Health.Dicom.Functions/Indexing/Configuration/ReindexConfiguration.cs
```diff
@@ -39,5 +41,13 @@ public class ReindexConfiguration
         ///// </summary>
         //[Range(0d, 1d)]
         //public double MaxErrorPercentage { get; set; } = 1;
+
```
wsugarman, line 44: These values should be in a different configuration section as they do not necessarily concern "reindexing"

wsugarman: By creating a new section, you'll need to configure it like the existing `ReindexConfiguration` in *Startup.cs*

src/Microsoft.Health.Dicom.Functions/Indexing/Configuration/ReindexConfiguration.cs
```diff
@@ -39,5 +41,13 @@ public class ReindexConfiguration
...
+        public bool IsPurgeErrorEnabled { get; set; } = true;
+
+        public IEnumerable<OrchestrationRuntimeStatus> RuntimeStatus { get; } = new[] { OrchestrationRuntimeStatus.Completed };
```
wsugarman, line 47: This should probably be some sort of readonly collection, like `IEnumerable<T>`, but also have the `Count` proeprty available such as `IReadOnlyCollection<T>`. No reason to expose this as the "lazily" evaluated interface `IEnumerable<T>`

wsugarman, line 47: nit: RuntimeStatuses*

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
```diff
@@ -0,0 +1,73 @@
...
+        // Currently an HTTP trigger, TODO: change to the appropriate trigger.
+        [FunctionName("DurableFunctionsCleanup")]
+        public async Task<String> Run(
+            [HttpTrigger(AuthorizationLevel.Function, "get", "post", Route = null)] HttpRequest req,
+            [DurableClient] IDurableOrchestrationClient client,
+            ILogger log,
+            CancellationToken hostCancellationToken)
+        {
+            EnsureArg.IsNotNull(req, nameof(req));
+            using var cancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(hostCancellationToken, req.HttpContext.RequestAborted);
```
wsugarman, line 43: I had to look up why you did it this way, but it makes sense given how these tokens are canceled. We should probably abstract this into some idiomatic API for all HTTP triggered functions. That can be in another PR though

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
```diff
@@ -0,0 +1,73 @@
...
+
+            log.LogInformation("Durable Functions cleanup started.");
+
+            EnsureArg.IsNotNull(client, nameof(client));
+
+            Console.WriteLine(_clientConfig.IsPurgeErrorEnabled);
```
wsugarman, line 49: I know this is for testing, but the logger should help here too by printing to console

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
```diff
@@ -0,0 +1,73 @@
...
+
+            // Specify conditions for Function Queries.
+            OrchestrationStatusQueryCondition condition = new OrchestrationStatusQueryCondition()
```
wsugarman, line 52: nit: `()` are not needed for using the default ctor with an initializer list

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
```diff
@@ -0,0 +1,73 @@
...
+            {
+                RuntimeStatus = _clientConfig.RuntimeStatus,
+                CreatedTimeFrom = _clientConfig.CreatedTimeFrom,
+                CreatedTimeTo = _clientConfig.CreatedTimeTo
+            };
+
+            OrchestrationStatusQueryResult listOfOrchestrators = await client.ListInstancesAsync
+                (condition, cancellationTokenSource.Token);
```
wsugarman, line 60: nit: stylistically, either bring the method name onto the next line or place each parameter on their own line if you think it's too long. For a single method call, I'd prefer the 2nd option.

```csharp
OrchestrationStatusQueryResult listOfOrchestrators = await client
    .ListInstancesAsync(condition, cancellationTokenSource.Token);

// or

OrchestrationStatusQueryResult listOfOrchestrators = await client.ListInstancesAsync(
    condition, 
    cancellationTokenSource.Token);
```

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
```diff
@@ -0,0 +1,73 @@
...
+
+            // Loop through the orchestration instances and purge them.
+            foreach (var orch in listOfOrchestrators.DurableOrchestrationState)
+            {
+                await client.PurgeInstanceHistoryAsync(orch.InstanceId);
+            }
+
+            cancellationTokenSource.Dispose();
```
wsugarman, line 68: This is unnecessary with the `using` statement above. It won't hurt anything though

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
```diff
@@ -0,0 +1,73 @@
...
+
+            return "Durable Functions cleanup executed successfully.";
```
wsugarman, line 70: For the non-HTTP trigger version of this, I'd log the number of instances we deleted (or even list the IDs + names?)

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
```diff
@@ -0,0 +1,73 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Threading.Tasks;
+using Microsoft.Azure.WebJobs;
+using Microsoft.Azure.WebJobs.Extensions.Http;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
+using EnsureThat;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask;
```
wsugarman, line 12: nit: sort these using statements alphabetically with `System` on top. There should be an option in Visual Studio to do this for you under Edit > IntelliSense > Sort Usings

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
```diff
@@ -0,0 +1,73 @@
...
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Health.Dicom.Functions
+{
+    /// <summary>
+    /// Represents the Azure Functions that perform the cleanup of completed Durable Functions used for re-indexing.
+    /// </summary>
+    public class DurableFunctionsCleanup
+    {
+
```
wsugarman, line 25: nit: extra line

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
```diff
@@ -0,0 +1,73 @@
...
+        private readonly ReindexConfiguration _clientConfig;
```
wsugarman, line 26: nit: newline after a private member before the ctor

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
yuedai0531: I think this was done automatically, but will change it!

src/Microsoft.Health.Dicom.Functions/Indexing/Configuration/ReindexConfiguration.cs
yuedai0531: is this what you mean?
`public IReadOnlyCollection<OrchestrationRuntimeStatus> RuntimeStatus { get; } = new[] { OrchestrationRuntimeStatus.Unknown };`

wsugarman: You'll want to keep `set` in there too, otherwise I don't think the binder will be able to set it

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
```diff
@@ -0,0 +1,86 @@
...
+        private readonly ReindexErrorConfiguration _cleanupConfig;
+
+        // TODO: Leverages services for fulfilling the various activies
+        public DurableFunctionsCleanup(IOptions<IndexingConfiguration> configOptions, IOptions<ReindexErrorConfiguration> cleanupOptions)
+        {
+            EnsureArg.IsNotNull(configOptions, nameof(configOptions));
+            _clientConfig = configOptions.Value.Add;
+
+            EnsureArg.IsNotNull(cleanupOptions, nameof(cleanupOptions));
```
wsugarman, line 36: nit: `cleanupOptions?.Value`

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
```diff
@@ -0,0 +1,86 @@
...
+        {
+            EnsureArg.IsNotNull(req, nameof(req));
+            using var cancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(hostCancellationToken, req.HttpContext.RequestAborted);
+
+            log.LogInformation("Durable Functions cleanup started.");
+
+            EnsureArg.IsNotNull(client, nameof(client));
+
+
+            EnsureArg.IsNotNull(_clientConfig, nameof(_clientConfig));
```
wsugarman, line 56: Don't need to re-check the config's nullability. You checked this in the ctor

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
```diff
@@ -0,0 +1,86 @@
...
+            [DurableClient] IDurableOrchestrationClient client,
+            ILogger log,
+            CancellationToken hostCancellationToken)
+        {
+            EnsureArg.IsNotNull(req, nameof(req));
+            using var cancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(hostCancellationToken, req.HttpContext.RequestAborted);
+
+            log.LogInformation("Durable Functions cleanup started.");
+
+            EnsureArg.IsNotNull(client, nameof(client));
```
wsugarman, line 53: nit: perform null checks as the first statement(s)

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
```diff
@@ -0,0 +1,86 @@
...
+            // Specify conditions for Function Queries.
+            OrchestrationStatusQueryCondition condition = new OrchestrationStatusQueryCondition
+            {
+                RuntimeStatus = _cleanupConfig.RuntimeStatus,
+                CreatedTimeFrom = DateTime.MinValue,
+                CreatedTimeTo = DateTime.UtcNow.AddDays(-_cleanupConfig.DaysCompleted)
+            };
+
+            OrchestrationStatusQueryResult listOfOrchestrators =
+                await client.ListInstancesAsync(new OrchestrationStatusQueryCondition(), cancellationTokenSource.Token);
```
wsugarman, line 68: You'll want to pass the `OrchestrationStatusQueryCondition` you just created

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
```diff
@@ -0,0 +1,86 @@
...
+
+            //await client
+            //    .ListInstancesAsync(condition, cancellationTokenSource.Token);
+
+            IEnumerable<String> nameList = listOfOrchestrators.DurableOrchestrationState.Select(i => i.Name);
+
+            Console.WriteLine(String.Join(", ", nameList));
+
+            // Loop through the orchestration instances and purge them.
+            //foreach (var orch in listOfOrchestrators.DurableOrchestrationState)
```
wsugarman, line 78: You'll want to uncomment this

src/Microsoft.Health.Dicom.Functions/Startup.cs
```diff
@@ -27,6 +27,13 @@ public override void Configure(IFunctionsHostBuilder builder)
                     .GetSection(HostSectionName)
                     .GetSection(IndexingConfiguration.SectionName)
                     .Bind(sectionObj));
+
+            builder.Services
+                .AddOptions<ReindexErrorConfiguration>()
+                .Configure<IConfiguration>((sectionObj, config) => config
+                    .GetSection(HostSectionName)
+                    .GetSection("orchestrationCleanup")
```
wsugarman, line 35: nit: create a member in the config type for this section name

wsugarman (changes requested):

src/Microsoft.Health.Dicom.Functions.UnitTests/DurableFunctionsCleanupTests.cs
```diff
@@ -0,0 +1,21 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Health.Dicom.Functions.UnitTests
+{
+    public class DurableFunctionsCleanupTests
+    {
+
+        public void CleanupCompletedDurableFunctions_GivenValidInput_ReturnsStringListOfOrchestrations()
```
pengchen0692, line 15: I think our unit tests following `Given-When-Then` pattern (https://en.wikipedia.org/wiki/Given-When-Then).
Example: https://github.com/microsoft/dicom-server/blob/ebaf5e2594c99646de3dc6e980919d5b0de0dc8f/src/Microsoft.Health.Dicom.Core.UnitTests/Serialization/JsonDicomConverterExtendedTests.cs#L20

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
```diff
@@ -0,0 +1,86 @@
...
+            {
+                RuntimeStatus = _cleanupConfig.RuntimeStatus,
+                CreatedTimeFrom = DateTime.MinValue,
+                CreatedTimeTo = DateTime.UtcNow.AddDays(-_cleanupConfig.DaysCompleted)
+            };
+
+            OrchestrationStatusQueryResult listOfOrchestrators =
+                await client.ListInstancesAsync(new OrchestrationStatusQueryCondition(), cancellationTokenSource.Token);
+
+            //await client
```
pengchen0692, line 70: remove lines if commented out.  This could keep the code clean.

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
```diff
@@ -0,0 +1,86 @@
...
+            //    .ListInstancesAsync(condition, cancellationTokenSource.Token);
+
+            IEnumerable<String> nameList = listOfOrchestrators.DurableOrchestrationState.Select(i => i.Name);
+
+            Console.WriteLine(String.Join(", ", nameList));
```
pengchen0692, line 75: Use log.LogInformation instead?

yuedai0531:
Changes in progress, do not review.

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
yuedai0531: oh yeah forgot to take it out
it was complaining that I wasn't using it so I added this line.

src/Microsoft.Health.Dicom.Functions/DurableFunctionsCleanup.cs
yuedai0531: yup! this was for testing purposes

src/Microsoft.Health.Dicom.Operations.Functions/DurableFunctionsCleanup.cs
```diff
@@ -0,0 +1,76 @@
...
+
+            OrchestrationStatusQueryResult listOfOrchestrators =
+                await client.ListInstancesAsync(condition, cancellationTokenSource.Token);
+
+            EnsureArg.IsNotNull(listOfOrchestrators, nameof(listOfOrchestrators));
+
+            IEnumerable<String> nameList = listOfOrchestrators.DurableOrchestrationState.Select(i => i.Name);
+
+            // Loop through the orchestration instances and purge them.
+            foreach (var orch in listOfOrchestrators.DurableOrchestrationState)
```
wsugarman, line 68: You actually need to use the resulting `OrchestrationStatusQueryResult` a bit more. If you inspect the [type](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.webjobs.extensions.durabletask.orchestrationstatusqueryresult?view=azure-dotnet), you'll see it has a property called `ContinuationToken` which is used to paginate the results. So if the Continuation is not `null`, you'll need to call `ListInstancesAsync` again with the token to get the next set of results.

Luckily, there are some new abstractions added to .NET, for operating on what we could consider to be an "asynchronous" enumeration (because we'd asynchronously fetch chunks of the elements we'd enumerate over):

```csharp
private async IAsyncEnumerable<DurableOrchestrationStatus> ListAllInstanceAsync(CancellationToken cancellationToken = default)
{
    OrchestrationStatusQueryCondition condition = new OrchestrationStatusQueryCondition  { ... };
    do
    {
        // TODO: Ensure the condition's ContinuationToken is updated based on the last batch
        OrchestrationStatusQueryResult batch = listOfOrchestrators = await client.ListInstancesAsync(condition, cancellationToken);

        // TODO: Yield the next batch of elements
    } while (batch.ContinuationToken != null);
}
```

Then, you can enumerate over this `IAsyncEnumerable<T>` using the new `await foreach` syntax.

wsugarman: If this takes long enough, it may even warrant an orchestration!

wsugarman (changes requested):

src/Microsoft.Health.Dicom.Operations.Functions.UnitTests/DurableFunctionsCleanupTests.cs
```diff
@@ -0,0 +1,54 @@
...
+namespace Microsoft.Health.Dicom.Functions.UnitTests
+{
+    public class DurableFunctionsCleanupTests
+    {
+        private readonly ReindexErrorConfiguration _cleanupConfig = new ReindexErrorConfiguration();
+
+        [Fact]
+        public async Task CleanupCompletedDurableFunctions_GivenValidInput_ReturnsStringListOfOrchestrationsAsync()
+        {
+            Console.WriteLine("Hello");
```
wsugarman, line 31: nit: probably don't want this here 😛

wsugarman: Nor the assertion below

src/Microsoft.Health.Dicom.Operations.Functions.UnitTests/DurableFunctionsCleanupTests.cs
```diff
@@ -0,0 +1,54 @@
...
+            Assert.Equal(0, 0);
+            var request = GenerateHttpRequest();
+
+            //ILogger logger = NullLoggerFactory.Instance.CreateLogger("test");
```
wsugarman, line 35: My rule of thumb is: if you're asserting logic, then substitute. Otherwise use the null or dummy implementation for "pass-through"

src/Microsoft.Health.Dicom.Functions/DurableFunctionsTimedCleanup.cs
```diff
@@ -0,0 +1,81 @@
...
+            ILogger log,
+            CancellationToken hostCancellationToken)
+        {
+            List<string> orchestrationInstanceIdList = new List<string>();
+            log.LogInformation($"C# Timer trigger function executed at: {DateTime.Now}");
+
+            EnsureArg.IsNotNull(client, nameof(client));
+            EnsureArg.IsNotNull(myTimer, nameof(myTimer));
+
+            log.LogInformation("Durable Functions cleanup started at: " + myTimer.ToString());
```
wsugarman, line 48: Same comment about structured logging. In fact, you probably want to combine this log with the first one in order to convey we're executing a timer for some time `t1` at the current time `t2`. I am unsure `TimerInfo` has a non default `ToString()` method, but I think you'll have more luck with like `myTimer.ScheduleStatus.Next` maybe? Hopefully some ad-hoc testing will reveal their behavior.

src/Microsoft.Health.Dicom.Functions/DurableFunctionsTimedCleanup.cs
```diff
@@ -0,0 +1,81 @@
...
+            string isPastDue = myTimer.IsPastDue ? "Current function invocation is later than scheduled." : "Function invocation is on schedule.";
+            log.LogInformation(isPastDue);
```
wsugarman, line 50: nit: We can perhaps simply log a `Warning` if the timer is past due. Otherwise we don't log anything

src/Microsoft.Health.Dicom.Functions/DurableFunctionsTimedCleanup.cs
```diff
@@ -0,0 +1,81 @@
...
+
+            EnsureArg.IsNotNull(client, nameof(client));
```
wsugarman, line 52: You've already validated this

src/Microsoft.Health.Dicom.Functions/DurableFunctionsTimedCleanup.cs
```diff
@@ -0,0 +1,81 @@
...
+
+            // Specify conditions for Function Queries.
```
wsugarman, line 54: nit: `for orchestration instances`

src/Microsoft.Health.Dicom.Functions/DurableFunctionsTimedCleanup.cs
```diff
@@ -0,0 +1,81 @@
...
+                   await client.ListInstancesAsync(condition, hostCancellationToken);
+                condition.ContinuationToken = listOfOrchestrators.ContinuationToken;
+
+                // Loop through the orchestration instances and purge them.
+                foreach (var orchestration in listOfOrchestrators.DurableOrchestrationState)
+                {
+                    orchestrationInstanceIdList.Add(orchestration.InstanceId);
+                    await client.PurgeInstanceHistoryAsync(orchestration.InstanceId);
+                }
+
```
wsugarman, line 75: nit: extra new line

src/Microsoft.Health.Dicom.Functions/DurableFunctionsTimedCleanup.cs
```diff
@@ -0,0 +1,81 @@
...
+            } while (condition.ContinuationToken != null);
+
+            log.LogInformation($"Durable Functions cleaned up successfully:\n{string.Join("\n", orchestrationInstanceIdList)}");
```
wsugarman, line 78: Update this to leverage structured logging. And perhaps we can have two elements:
1. The number of instances purged
2. The instance IDs

src/Microsoft.Health.Dicom.Functions.UnitTests/DurableFunctionsTimedCleanupTests.cs
```diff
@@ -0,0 +1,101 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+//using System;
```
wsugarman, line 6: Remove these

src/Microsoft.Health.Dicom.Functions.UnitTests/DurableFunctionsTimedCleanupTests.cs
```diff
@@ -0,0 +1,101 @@
...
+        private readonly DurableFunctionsTimedCleanup _durableFunctionsCleanup;
+        private readonly IDurableOrchestrationClient _durableOrchestrationClientMock;
+
+        public DurableFunctionsTimedCleanupTests()
+        {
+            _cleanupConfig = new ReindexErrorConfiguration();
+            _cleanupConfig.RuntimeStatuses = new OrchestrationRuntimeStatus[] { OrchestrationRuntimeStatus.Completed };
+            _timer = Substitute.For<TimerInfo>(default, default, default);
+            _logger = Substitute.For<ILogger>();
+            _cancellationTokenSource = new CancellationTokenSource();
```
wsugarman, line 41: I'd say create the `CancellationTokenSource` per test with a `using` statement

src/Microsoft.Health.Dicom.Functions.UnitTests/DurableFunctionsTimedCleanupTests.cs
```diff
@@ -0,0 +1,101 @@
...
+                    Arg.Any<CancellationToken>())
+                .Returns(Task.FromResult(new OrchestrationStatusQueryResult { DurableOrchestrationState = Array.Empty<DurableOrchestrationStatus>() }));
+
+            await _durableFunctionsCleanup.Run(_timer, _durableOrchestrationClientMock, _logger, _cancellationTokenSource.Token);
+        }
+
+        [Fact]
+        public async Task GivenValidInput_CleanupCompletedDurableFunctions__ShouldPurgeMultiPageListOfOrchestrationsAsync()
+        {
+            IEnumerable<DurableOrchestrationStatus> durableOrchestrationStatuses = new List<DurableOrchestrationStatus>() {
```
wsugarman, line 68: nit: keep this as `List` on the left-hand side. There's no need to eagerly treat it as an `IEnumerable<T>`

wsugarman, line 68: nit: curly braces always go on the next line

src/Microsoft.Health.Dicom.Functions.UnitTests/DurableFunctionsTimedCleanupTests.cs
```diff
@@ -0,0 +1,101 @@
...
+        public async Task GivenValidInput_CleanupCompletedDurableFunctions__ShouldPurgeListOfOrchestrationsAsync()
+        {
+            _durableOrchestrationClientMock
+                .ListInstancesAsync(
+                    Arg.Is<OrchestrationStatusQueryCondition>(givenCondition =>
+                        givenCondition.RuntimeStatus.SequenceEqual(_cleanupConfig.RuntimeStatuses)
+                        && givenCondition.CreatedTimeFrom == DateTime.MinValue
+                        && givenCondition.CreatedTimeTo == _definedNow.AddDays(-_cleanupConfig.MinimumAgeDays)
+                    ),
+                    Arg.Any<CancellationToken>())
```
wsugarman, line 59: nit: check the token

src/Microsoft.Health.Dicom.Functions.UnitTests/DurableFunctionsTimedCleanupTests.cs
```diff
@@ -0,0 +1,101 @@
...
+            _cancellationTokenSource = new CancellationTokenSource();
+            _definedNow = DateTime.UtcNow;
+            _durableFunctionsCleanup = new DurableFunctionsTimedCleanup(
+                Options.Create(_cleanupConfig),
+                () => _definedNow);
+            _durableOrchestrationClientMock = Substitute.For<IDurableOrchestrationClient>();
+        }
+
+        [Fact]
+        public async Task GivenValidInput_CleanupCompletedDurableFunctions__ShouldPurgeListOfOrchestrationsAsync()
```
wsugarman, line 50: These methods are quite validating, as they simply run and don't check to see whether the methods were called as expected. So you'll need to leverage the `Received(int)` method in NSubstitute.

To that end, I think you'll find it easier to simply make a helper function for the `OrchestrationStatusQueryCondition` predicate, like:

```csharp
private bool IsCondition(OrchestrationStatusQueryCondition actual, string token)
{
    return actual.RuntimeStatus.SequenceEqual(_cleanupConfig.RuntimeStatuses)
        && actual.CreatedTimeFrom == DateTime.MinValue
        && actual.CreatedTimeTo == _definedNow.AddDays(-_cleanupConfig.MinimumAgeDays)
        && actual.Token == token;
}
```

Then you can say things like:

```csharp
_durableOrchestrationClientMock
    .ListInstancesAsync(
        Arg.Is<OrchestrationStatusQueryCondition>(x => IsCondition(x, null)),
        Arg.Is(tokenSource.Token))
    .Returns(Task.FromResult(...));

await _durableFunctionsCleanup.Run(...);

_durableOrchestrationClientMock
    .Received(1)
    .ListInstancesAsync(
        Arg.Is<OrchestrationStatusQueryCondition>(x => IsCondition(x, null)),
        Arg.Is(tokenSource.Token));
```

src/Microsoft.Health.Dicom.Functions.UnitTests/DurableFunctionsTimedCleanupTests.cs
```diff
@@ -0,0 +1,101 @@
...
+                .ListInstancesAsync(
+                    Arg.Is<OrchestrationStatusQueryCondition>(givenCondition =>
+                        givenCondition.RuntimeStatus.SequenceEqual(_cleanupConfig.RuntimeStatuses)
+                        && givenCondition.CreatedTimeFrom == DateTime.MinValue
+                        && givenCondition.CreatedTimeTo == _definedNow.AddDays(-_cleanupConfig.MinimumAgeDays)
+                    ),
+                    Arg.Any<CancellationToken>())
+                .Returns(Task.FromResult(new OrchestrationStatusQueryResult { DurableOrchestrationState = Array.Empty<DurableOrchestrationStatus>() }));
+
+            await _durableFunctionsCleanup.Run(_timer, _durableOrchestrationClientMock, _logger, _cancellationTokenSource.Token);
```
wsugarman, line 62: We also need to check if the proper instance IDs were passed to `IDurableOrchestrationClient.PurgeInstanceHistory` async. We'll need to collect the arguments passed to this method and check the number of calls versus the number of pages.

Eg.
```csharp
var actual = new List<string>();
var expected = new List<string> { "foo", "bar", "baz" };

_durableOrchestrationClientMock
    .When(x => x.PurgeInstanceHistory(Arg.Any<string>())
    .Do(x => actual.Add(x.ArgAt<string>(0));

await _durableFunctionsCleanup.Run(...);

_durableOrchestrationClientMock
    .Received(3)
    .PurgeInstanceHistory(Arg.Is<string>(x => expected.Contains(x));
Assert.Equal(expected, actual);
```

wsugarman (changes requested):

src/Microsoft.Health.Dicom.Functions/Indexing/Configuration/ReindexErrorConfiguration.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.ComponentModel.DataAnnotations;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask;
+
+namespace Microsoft.Health.Dicom.Functions.Indexing.Configuration
+{
+    public class ReindexErrorConfiguration
+    {
+        public const string SectionName = "OrchestrationsCleanup";
+
+        [MinLength(1)]
```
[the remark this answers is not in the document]
yuedai0531: oh oops I removed it at the wrong place

image

Microsoft.Health.Dicom.sln
```diff
@@ -140,20 +140,16 @@ Global
 		{D100EA60-8DC8-4576-A177-56BC7193BF4A} = {176641B3-297C-4E04-A83D-8F80F80485E8}
 		{E15123F6-5A28-4D86-A28D-30FCB699B9EE} = {176641B3-297C-4E04-A83D-8F80F80485E8}
 		{FA0484A7-AA0C-4CC6-A75F-1D6B23DD847D} = {176641B3-297C-4E04-A83D-8F80F80485E8}
+		{BB88616C-6208-43D0-B9EF-79FC0652A151} = {176641B3-297C-4E04-A83D-8F80F80485E8}
```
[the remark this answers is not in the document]
yuedai0531: yep

src/Microsoft.Health.Dicom.Functions.UnitTests/DurableFunctionsTimedCleanupTests.cs
yuedai0531: ah good old Cs

src/Microsoft.Health.Dicom.Functions/Management/PurgeOrchestrationInstancesHistory.cs
```diff
@@ -0,0 +1,75 @@
...
+
+                // Loop through the orchestration instances and purge them.
+                foreach (DurableOrchestrationStatus orchestration in listOfOrchestrators.DurableOrchestrationState)
+                {
+                    orchestrationInstanceIdList.Add(orchestration.InstanceId);
+                    await client.PurgeInstanceHistoryAsync(orchestration.InstanceId);
+                }
+            } while (condition.ContinuationToken != null);
+
+            log.LogInformation("{Count} Durable Functions cleaned up successfully. List of cleaned instance IDs:\n{list}", orchestrationInstanceIdList.Count, String.Join(",\n", orchestrationInstanceIdList));
```
wsugarman, line 72: nit: I think we can skip the newlines

src/Microsoft.Health.Dicom.Functions.UnitTests/Management/PurgeOrchestrationInstancesHistoryTests.cs
```diff
@@ -0,0 +1,132 @@
...
+        private readonly PurgeOrchestrationInstancesHistoryConfiguration _purgeConfig;
+        private readonly TimerInfo _timer;
+        private readonly ILogger _logger;
+        private readonly DateTime _definedNow;
+        private readonly PurgeOrchestrationInstancesHistory _purgeOrchestrationInstancesHistory;
+        private readonly IDurableOrchestrationClient _durableOrchestrationClientMock;
+
+        public PurgeOrchestrationInstancesHistoryTests()
+        {
+            _purgeConfig = new PurgeOrchestrationInstancesHistoryConfiguration();
```
wsugarman, line 32: nit: use the initializer syntax:
```csharp
_purgeConfig = new PurgeOrchestrationInstancesHistoryConfiguration
{
    RuntimeStatuses = new OrchestrationRuntimeStatus[] { OrchestrationRuntimeStatus.Completed }
};
```

src/Microsoft.Health.Dicom.Functions.UnitTests/Management/PurgeOrchestrationInstancesHistoryTests.cs
```diff
@@ -0,0 +1,132 @@
...
+            _logger = Substitute.For<ILogger>();
+            _definedNow = DateTime.UtcNow;
+            _purgeOrchestrationInstancesHistory = new PurgeOrchestrationInstancesHistory(
+                Options.Create(_purgeConfig),
+                () => _definedNow);
+            _durableOrchestrationClientMock = Substitute.For<IDurableOrchestrationClient>();
+        }
+
+        [Fact]
+        public async Task GivenValidInput_PurgeCompletedDurableFunctionsHistory_ShouldNotPurgeAnythingOrchestrationsAsync()
```
wsugarman, line 44: nit: a more typical name in our repository would be using Given, When, Then:

Eg. `GivenNoInstances_WhenPurgingHistory_ThenSkipPurge`

wsugarman: This comment goes for the rest: the "given" scenario is not valid input, but rather something more specific like no instances, no continuation, or multiple page

src/Microsoft.Health.Dicom.Functions.UnitTests/Management/PurgeOrchestrationInstancesHistoryTests.cs
```diff
@@ -0,0 +1,132 @@
...
+            using CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
+
+            List<DurableOrchestrationStatus> durableOrchestrationStatuses = new List<DurableOrchestrationStatus>()
+            {
+                new DurableOrchestrationStatus{ InstanceId = "1" },
+                new DurableOrchestrationStatus{ InstanceId = "2" },
+                new DurableOrchestrationStatus{ InstanceId = "3" }
+            };
+
+            OrchestrationStatusQueryResult orchestrationStatusQueryResult = new OrchestrationStatusQueryResult();
```
wsugarman, line 73: nit: use the initializer syntax

src/Microsoft.Health.Dicom.Functions.UnitTests/Management/PurgeOrchestrationInstancesHistoryTests.cs
```diff
@@ -0,0 +1,132 @@
...
+                .DidNotReceiveWithAnyArgs()
+                .PurgeInstanceHistoryAsync(Arg.Any<string>());
+        }
+
+        [Fact]
+        public async Task GivenValidInput_PurgeCompletedDurableFunctionsHistory_ShouldPurgeMultiPageListOfOrchestrationsAsync()
+        {
+            using CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
+
+            List<DurableOrchestrationStatus> durableOrchestrationStatuses = new List<DurableOrchestrationStatus>()
```
wsugarman, line 66: nit: be consistent with `var` with its usage below. Pick one style or the other 😛

src/Microsoft.Health.Dicom.Functions.UnitTests/Management/PurgeOrchestrationInstancesHistoryTests.cs
```diff
@@ -0,0 +1,132 @@
...
+            _durableOrchestrationClientMock
+                .ListInstancesAsync(
+                    Arg.Is<OrchestrationStatusQueryCondition>(
+                            givenCondition => IsCondition(givenCondition, null)),
+                    Arg.Is<CancellationToken>(cancellationTokenSource.Token))
+                // First page, should have 3 orchestrations, then change continuation token.
+                .Returns(_ =>
+                    {
+                        orchestrationStatusQueryResult.ContinuationToken = null;
+                        Assert.True(orchestrationStatusQueryResult.DurableOrchestrationState.SequenceEqual(durableOrchestrationStatuses));
```
wsugarman, line 88: This isn't an appropriate place to `Assert` behavior as you're configuring it here. Instead, you should return a new `orchestrationStatusQueryResult` object with the desired state

src/Microsoft.Health.Dicom.Functions.UnitTests/Management/PurgeOrchestrationInstancesHistoryTests.cs
wsugarman: Move this into the first `Returns` method:
```csharp
.Returns(Task.FromResult(
    new OrchestrationStatusQueryResult
    {
        DurableOrchestrationState = durableOrchestrationStatuses,
        ContinuationToken = "fake_cancellation_token",
    }));
```

src/Microsoft.Health.Dicom.Functions.UnitTests/Management/PurgeOrchestrationInstancesHistoryTests.cs
```diff
@@ -0,0 +1,132 @@
...
+
+            _durableOrchestrationClientMock
+                .ListInstancesAsync(
+                    Arg.Is<OrchestrationStatusQueryCondition>(
+                            givenCondition => IsCondition(givenCondition, orchestrationStatusQueryResult.ContinuationToken)),
+                    Arg.Is<CancellationToken>(cancellationTokenSource.Token))
+                // Second page, should have no orchestrations.
+                .Returns(_ =>
+                    {
+                        Assert.True(orchestrationStatusQueryResult.DurableOrchestrationState.SequenceEqual(null));
```
wsugarman, line 100: Similarly, this assertion is unnecessary

src/Microsoft.Health.Dicom.Functions.UnitTests/Management/PurgeOrchestrationInstancesHistoryTests.cs
```diff
@@ -0,0 +1,132 @@
...
+                    {
+                        orchestrationStatusQueryResult.ContinuationToken = null;
+                        Assert.True(orchestrationStatusQueryResult.DurableOrchestrationState.SequenceEqual(durableOrchestrationStatuses));
+                        return Task.FromResult(orchestrationStatusQueryResult);
+                    });
+
+            _durableOrchestrationClientMock
+                .ListInstancesAsync(
+                    Arg.Is<OrchestrationStatusQueryCondition>(
+                            givenCondition => IsCondition(givenCondition, orchestrationStatusQueryResult.ContinuationToken)),
```
wsugarman, line 95: nit: too large of an indentation. Some changes may help save space. For example, I think you could use a shorter parameter name like `x` here as it's fairly obvious what the argument is. Furthermore, I would copy the expected continuation token inline instead of referencing this value

src/Microsoft.Health.Dicom.Functions.UnitTests/Management/PurgeOrchestrationInstancesHistoryTests.cs
```diff
@@ -0,0 +1,132 @@
...
+
+            await _purgeOrchestrationInstancesHistory.Run(_timer, _durableOrchestrationClientMock, _logger, cancellationTokenSource.Token);
+
+            await _durableOrchestrationClientMock
+                .Received(1)
+                .ListInstancesAsync(
+                    Arg.Is<OrchestrationStatusQueryCondition>(x => IsCondition(x, null)),
+                    Arg.Is<CancellationToken>(cancellationTokenSource.Token));
+
+            var expected = durableOrchestrationStatuses.Select(i => i.InstanceId);
```
wsugarman, line 116: Perhaps allocate this as a `List<T>` with `ToList<T>()`?

src/Microsoft.Health.Dicom.Functions/Management/PurgeOrchestrationInstancesHistory.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        public PurgeOrchestrationInstancesHistory(IOptions<OrchestrationsHistoryConfiguration> cleanupOptions)
+            : this(cleanupOptions, () => DateTime.UtcNow)
+        { }
+
+        internal PurgeOrchestrationInstancesHistory(IOptions<OrchestrationsHistoryConfiguration> cleanupOptions, Func<DateTime> getDateTimeUtcNow)
+        {
+            EnsureArg.IsNotNull(cleanupOptions?.Value, nameof(cleanupOptions));
+            EnsureArg.IsNotNull(getDateTimeUtcNow, nameof(getDateTimeUtcNow));
+            _purgeConfig = cleanupOptions?.Value;
```
wsugarman, line 31: nit: you don't need the [`null`-conditional operator](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators#null-conditional-operators--and-) (`?.`) as you `null`-checked above

wsugarman (approved):

src/Microsoft.Health.Dicom.Functions/Management/PurgeOrchestrationInstancesHistory.cs
```diff
@@ -0,0 +1,88 @@
...
+using System.Threading.Tasks;
+using EnsureThat;
+using Microsoft.Azure.WebJobs;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Health.Dicom.Functions.Management
+{
+    public class PurgeOrchestrationInstancesHistory
```
wsugarman, line 18: nit: let's call it `PurgeOrchestrationInstanceHistory` instead (singular `Instance`)

PR #830, opened 2021-06-10
Able to run Azure functions with faked implementation

pengchen0692:
## Description
Able to run Azure functions with faked implementation

## Related issues
Addresses [AB#82723](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/82723)

## Notes
Need to create user stories for future change once approved
* Need to move AddInitializedSqlServerBase into healthcare-shared-component once code is solid (e.g: merging back to main)
* Need to improvement performance of UpdateSchemaVersionAsync -- not need to get version for each request


src/Microsoft.Health.Dicom.Core/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Health.Extensions.DependencyInjection;
+
+namespace Microsoft.Health.Dicom.Core.Extensions
+{
+    public static class IServiceCollectionExtensions
+    {
+        public static IServiceCollection AddSingletonDefault<T>(this IServiceCollection services)
```
pengchen0692, line 13: Add summaries

src/Microsoft.Health.Dicom.Core/Features/ExtendedQueryTag/AddExtendedQueryTagService.cs
```diff
@@ -17,7 +18,7 @@ namespace Microsoft.Health.Dicom.Core.Features.ExtendedQueryTag
 {
     public class AddExtendedQueryTagService : IAddExtendedQueryTagService
     {
-        private readonly IExtendedQueryTagStore _extendedQueryTagStore;
+        private readonly Lazy<IExtendedQueryTagStore> _extendedQueryTagStore;
```
pengchen0692, line 21: Explain why Lazy

src/Microsoft.Health.Dicom.Core/Registration/DicomFunctionsBuilderCoreRegistrationExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+        public static IDicomFunctionsBuilder AddCoreComponents(
+            this IDicomFunctionsBuilder builder)
+        {
+            EnsureArg.IsNotNull(builder, nameof(builder));
+            var services = builder.Services;
+            services.AddScopedDefault<InstanceReindexer>();
+            services.AddScopedDefault<AddExtendedQueryTagService>();
+            services.AddSingletonDefault<DicomTagParser>();
+            services.AddSingletonDefault<ExtendedQueryTagEntryValidator>();
+
```
pengchen0692, line 25: remove additional lines

src/Microsoft.Health.Dicom.Core/Registration/IDicomFunctionsBuilder.cs
```diff
@@ -0,0 +1,17 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Health.Dicom.Core.Registration
+{
+    /// <summary>
+    /// A builder type for configuring DICOM functions.
```
pengchen0692, line 11: Update summary

src/Microsoft.Health.Dicom.Functions.UnitTests/Indexing/ReindexDurableFunctionTests.Orchestration.cs
```diff
@@ -31,16 +31,26 @@ public async Task GivenValidTags_WhenAddExtendedQueryTagsOrchestrationAsync_Shou
             IDurableOrchestrationContext context = Substitute.For<IDurableOrchestrationContext>();
             context.GetInput<IReadOnlyList<AddExtendedQueryTagEntry>>().Returns(tagsEntries);
             await _reindexDurableFunction.AddAndReindexTagsAsync(context, NullLogger.Instance);
+
+            // Verify   UpdateSchemaVersionAsync is called
```
pengchen0692, line 35: remove addtional space

src/Microsoft.Health.Dicom.Functions/Configs/DicomFunctionsConfiguration.cs
```diff
@@ -0,0 +1,23 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.ComponentModel.DataAnnotations;
+
+namespace Microsoft.Health.Dicom.Functions.Configs
+{
+    /// <summary>
+    /// Represents configuration settings related to the indexing of data.
```
pengchen0692, line 11: update summary

src/Microsoft.Health.Dicom.Functions/Configs/SqlServerConfiguration.cs
```diff
@@ -5,19 +5,19 @@
...
 using System.ComponentModel.DataAnnotations;
 
-namespace Microsoft.Health.Dicom.Functions.Indexing.Configuration
+namespace Microsoft.Health.Dicom.Functions.Configs
 {
     /// <summary>
     /// Represents configuration settings related to the indexing of data.
     /// </summary>
-    public class IndexingConfiguration
+    public class SqlServerConfiguration
```
pengchen0692, line 13: remove this

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Activity.cs
```diff
@@ -128,5 +128,18 @@ public async Task ReindexInstancesAsync([ActivityTrigger] ReindexInstanceInput i
...
+        [FunctionName(nameof(UpdateSchemaVersionAsync))]
+        public async Task UpdateSchemaVersionAsync([ActivityTrigger] IDurableActivityContext context, ILogger log)
+        {
+            // TODO: performance improvement, don't need to call service for every call.
+            EnsureArg.IsNotNull(context, nameof(context));
+            EnsureArg.IsNotNull(log, nameof(log));
+            log.LogInformation("Update schema version");
+            int version = await _schemaManagerDataStore.GetCurrentSchemaVersionAsync(default);
+            _schemaInformation.Current = version;
+            log.LogInformation("Current schema version is set to {version}", version);
```
pengchen0692, line 142: Remove current

src/Microsoft.Health.Dicom.Functions/Registration/DicomFunctionsServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+{
+    public static class DicomFunctionsServiceCollectionExtensions
+    {
+        /// <summary>
+        /// Add services for DICOM background workers.
+        /// </summary>
+        /// <param name="services">The DICOM server builder instance.</param>
+        /// <param name="configuration"></param>
+        /// <returns>The DICOM server builder instance.</returns>
+        public static IDicomFunctionsBuilder AddDicomFunctions(this IServiceCollection services, IConfiguration configuration)
```
pengchen0692, line 23: update summary

src/Microsoft.Health.Dicom.Functions/host.json
```diff
@@ -1,17 +1,22 @@
...
-                "excludedTypes": "Request"
-            }
-        }
-    },
-    "indexing": {
-        "add": {
-            "batchSize": 1000,
-            "parallelBatches": 10
-        }
+  "version": "2.0",
```
pengchen0692, line 2: Formatting should be consistent with Dicom server

src/Microsoft.Health.Dicom.SqlServer/Microsoft.Health.Dicom.SqlServer.csproj
```diff
@@ -39,6 +39,7 @@
     </EmbeddedResource>
   </ItemGroup>
   <ItemGroup>
+    <Folder Include="Configs\" />
```
pengchen0692, line 42: remove this

src/Microsoft.Health.Dicom.SqlServer/Registration/DicomServerBuilderSqlServerRegistrationExtensions.cs
```diff
@@ -104,6 +105,11 @@ public static class DicomServerBuilderSqlServerRegistrationExtensions
               .AsSelf()
               .AsImplementedInterfaces();
 
+            services.Add<SqlReindexStore>()
```
pengchen0692, line 108: AddScopedDefault

src/Microsoft.Health.Dicom.Core/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Health.Extensions.DependencyInjection;
+
+namespace Microsoft.Health.Dicom.Core.Extensions
+{
+    public static class IServiceCollectionExtensions
```
smithago, line 12: They don't add a lot of value, do we really need these wrapper methods?

src/Microsoft.Health.Dicom.Core/Features/ExtendedQueryTag/AddExtendedQueryTagService.cs
```diff
@@ -17,7 +18,9 @@ namespace Microsoft.Health.Dicom.Core.Features.ExtendedQueryTag
 {
     public class AddExtendedQueryTagService : IAddExtendedQueryTagService
     {
-        private readonly IExtendedQueryTagStore _extendedQueryTagStore;
+        // Different from Asp.net Core, we couldn't get schemaVersion before AzureFunction execution.
```
smithago, line 21: I don't understand this comment. Maybe AzureFunction needs to fixed on when it gets initialized

src/Microsoft.Health.Dicom.Core/Features/Indexing/InstanceReindexer.cs
```diff
@@ -0,0 +1,23 @@
...
+
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Health.Dicom.Core.Features.ExtendedQueryTag;
+
+namespace Microsoft.Health.Dicom.Core.Features.Indexing
+{
+    /// <summary>
+    /// Reindexer which reindexes DICOM instance.
```
smithago, line 14: reindexes DICOM tags

src/Microsoft.Health.Dicom.Core/Features/Indexing/InstanceReindexer.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// </summary>
+    public class InstanceReindexer : IInstanceReindexer
+    {
+        public Task ReindexInstanceAsync(IReadOnlyList<ExtendedQueryTagStoreEntry> entries, long watermark, CancellationToken cancellationToken = default)
```
smithago, line 18: QueueReindexAsync()?

src/Microsoft.Health.Dicom.Core/Features/Indexing/InstanceReindexer.cs
```diff
@@ -0,0 +1,23 @@
...
+        {
```
smithago, line 19: IReadOnlyCollection? Can we be consistent on which collection we use?

src/Microsoft.Health.Dicom.Core/Registration/DicomFunctionsBuilderCoreRegistrationExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+using EnsureThat;
+using Microsoft.Health.Dicom.Core.Extensions;
+using Microsoft.Health.Dicom.Core.Features.Common;
+using Microsoft.Health.Dicom.Core.Features.ExtendedQueryTag;
+using Microsoft.Health.Dicom.Core.Features.Indexing;
+
+namespace Microsoft.Health.Dicom.Core.Registration
+{
+    public static class DicomFunctionsBuilderCoreRegistrationExtensions
```
smithago, line 14: We need naming guidlines :-)
We don't predix DICOM any where in this solution. Because everything is dicom.

I don't understand this class name. Is this to consolidate all ExtendedQueryTag related services? We should name accordingly

src/Microsoft.Health.Dicom.Core/Registration/IDicomFunctionsBuilder.cs
```diff
@@ -0,0 +1,17 @@
...
+    /// </summary>
+    public interface IDicomFunctionsBuilder
```
smithago, line 13: Same issue with the name

src/Microsoft.Health.Dicom.Functions/Configs/DicomFunctionsConfiguration.cs
```diff
@@ -0,0 +1,23 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System.ComponentModel.DataAnnotations;
+
+namespace Microsoft.Health.Dicom.Functions.Configs
+{
+    /// <summary>
+    /// Represents configuration settings related to the Dicom functions.
+    /// </summary>
+    public class DicomFunctionsConfiguration
```
smithago, line 13: We should not use DicomFunction. It also looks quite dummy, why do we need this additional configuration

src/Microsoft.Health.Dicom.Functions/Configs/ReindexConfiguration.cs
```diff
@@ -6,13 +6,14 @@
 using System;
 using System.ComponentModel.DataAnnotations;
 
-namespace Microsoft.Health.Dicom.Functions.Indexing.Configuration
+namespace Microsoft.Health.Dicom.Functions.Configs
 {
     /// <summary>
     /// Represents the configuration for a "reindex" function.
     /// </summary>
     public class ReindexConfiguration
```
smithago, line 14: ReIndexJobConfiguration?

src/Microsoft.Health.Dicom.Functions/Configs/SqlServerConfiguration.cs
smithago: +1, we should not have store specific things in generic projects.

Why do you need connectionString?

smithago: If you need SQL, you should be able to reference DICOM.core

src/Microsoft.Health.Dicom.SqlServer/Features/Indexing/SqlReindexStore.cs
```diff
@@ -0,0 +1,36 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Health.Dicom.Core.Features.Indexing;
+
+namespace Microsoft.Health.Dicom.SqlServer.Features.Indexing
+{
+    internal class SqlReindexStore : IReindexStore
```
smithago, line 14: Do we need this class if nothing is implemented?

smithago: Also, these don't look specific to Reindex, can these be more generic like GetJobEntriesAsync(), UpdateJobEntriesAsync()...

src/Microsoft.Health.Dicom.SqlServer/Registration/DicomFunctionsBuilderSqlServerRegistrationExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+using Microsoft.Health.SqlServer;
+using Microsoft.Health.SqlServer.Features.Schema;
+using Microsoft.Health.Dicom.SqlServer.Features.Schema;
+using Microsoft.Health.SqlServer.Features.Schema.Manager;
+using Microsoft.Health.SqlServer.Configs;
+using Microsoft.Health.Dicom.Core.Extensions;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    public static class DicomFunctionsBuilderSqlServerRegistrationExtensions
```
smithago, line 24: Name needs to be fixed

src/Microsoft.Health.Dicom.SqlServer/Registration/DicomFunctionsBuilderSqlServerRegistrationExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+    {
+        public static IDicomFunctionsBuilder AddSqlServer(
```
smithago, line 26: These look duplicate. Is there a opportunity to create common SQL component extensions that we can reuse in both places?

src/Microsoft.Health.Dicom.SqlServer/Registration/DicomServerBuilderSqlServerRegistrationExtensions.cs
```diff
@@ -104,6 +106,8 @@ public static class DicomServerBuilderSqlServerRegistrationExtensions
               .AsSelf()
               .AsImplementedInterfaces();
 
+            services.AddScopedDefault<SqlReindexStore>();
```
smithago, line 109: Same here, we should not duplicate these

smithago, line 109: Does this go to separate registrationExtension called by functions?

src/Microsoft.Health.Dicom.Functions/Configs/SqlServerConfiguration.cs
pengchen0692: Azure Function is running on different host than Dicom service, so not able to access the [config](https://github.com/microsoft/dicom-server/blob/main/src/Microsoft.Health.Dicom.Web/appsettings.json) file , instead it should read config from [host.json](https://github.com/microsoft/dicom-server/blob/features/ExtendedQueryTagsV2/src/Microsoft.Health.Dicom.Functions/host.json), the same connection string will be duplicated to host.json and consumed by Azure Function.

src/Microsoft.Health.Dicom.Core/Extensions/IServiceCollectionExtensions.cs
pengchen0692: I feel it will, so far as I know these are required:
1. Core layer
   * AddExtendedQueryTagService
   * DicomTagParser
   * ExtendedQueryTagEntryValidator
2. Sql Layer
   * SqlInstanceStore
   * SqlExtendedQueryTagStoreV1
   * SqlExtendedQueryTagStoreV2
   * SqlExtendedQueryTagStoreV3
   * SqlStoreFactory<ISqlExtendedQueryTagStore, IExtendedQueryTagStore>>
   * SqlReindexStore

Once we add support to log exception, probably need components to access table store. 
I think the existing pattern for DicomServer is good, and following that could have a easy understanding code.

src/Microsoft.Health.Dicom.Core/Features/ExtendedQueryTag/AddExtendedQueryTagService.cs
pengchen0692: In DicomServer scenario, we get schema version through [SchemaJobWorkerBackgroundService](https://github.com/microsoft/healthcare-shared-components/blob/master/src/Microsoft.Health.SqlServer.Api/Schema/SchemaJobWorkerBackgroundService.cs) before other Dicom components (e.g AddExtendedQueryTagService), so that the components can initialize correctly.
But I didn't find similar things in Azure Function -- [HostedService is not part of Azure Function Today](https://github.com/Azure/azure-functions-host/issues/5447#issuecomment-575368316).

src/Microsoft.Health.Dicom.SqlServer/Registration/DicomServerBuilderSqlServerRegistrationExtensions.cs
pengchen0692: yes, SqlReindexStore should only be used by Dicom functions, this line should be removed.

src/Microsoft.Health.Dicom.Core/Features/Indexing/InstanceReindexer.cs
pengchen0692: This class reindex tags on single DICOM instance, which is called by [`ReindexTagsAsync`](https://github.com/microsoft/dicom-server/blob/15315868594b0df61e3d9d345b0185ad3e917748/src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Orchestration.cs#L52). So I feel it makes sense for saying this one reindex DICOM instance, while the other reindex DICOM tags?

src/Microsoft.Health.Dicom.Core/Features/Indexing/InstanceReindexer.cs
pengchen0692: Currently we don't Queue tags for reindexing, instead simple have [HttpTrigger ](https://github.com/microsoft/dicom-server/blob/15315868594b0df61e3d9d345b0185ad3e917748/src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Trigger.cs#L65) which add tags to database then start reindexing.

src/Microsoft.Health.Dicom.Core/Features/Indexing/InstanceReindexer.cs
pengchen0692: Any reason for `IReadOnlyCollection` > `IReadOnlyList`?
Actually I don't have favor between them, use `IReadOnlyList` just because it provides more capability and we can easily get it from `List<>`

src/Microsoft.Health.Dicom.Core/Registration/DicomFunctionsBuilderCoreRegistrationExtensions.cs
pengchen0692: It imitates [`DicomServerBuilderSqlServerRegistrationExtensions`](https://github.com/microsoft/dicom-server/blob/features/ExtendedQueryTagsV2/src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Trigger.cs#L65), the difference is this one is for services/components in `Core`, while `DicomServerBuilderSqlServerRegistrationExtensions` is for `SqlServer`. 

I feel we probably should refactor how we build DicomServer/DicomFunctions, but I don't have a good idea now, will think about that.

src/Microsoft.Health.Dicom.Core/Registration/IDicomFunctionsBuilder.cs
pengchen0692: No we have Dicom AzureFunction and DicomServer, they share some components. `IDicomFunctionsBuilder` imitate [`IDicomServerBuilder`](https://github.com/microsoft/dicom-server/blob/main/src/Microsoft.Health.Dicom.Core/Registration/IDicomServerBuilder.cs) to build all dependencies.

src/Microsoft.Health.Dicom.Functions/Configs/DicomFunctionsConfiguration.cs
pengchen0692: It imitate [`DicomServerConfiguration`](https://github.com/microsoft/dicom-server/blob/main/src/Microsoft.Health.Dicom.Api/Configs/DicomServerConfiguration.cs)  to provide configuration for Dicom Functions.

src/Microsoft.Health.Dicom.SqlServer/Features/Indexing/SqlReindexStore.cs
pengchen0692: The fake implementation is for other people to be able to run the application without injection failure, so that they won't be blocked.

This table stores `Reindex` status instead of Job information. It has `TagKey`, `StartWatermark`,`EndWatermark`,`OperationId`, `Failures`, (we may add more), which describe current Reindex status on this tag.

BTW, we use `Operation` to replace `Job`, since feel it make more sense for customer. 

We are discussing merging `Reindex` status and `Deindex` status into one single table (maybe call it `SqlIndexStore` or `SqlIndexStatusStore`?), but I have some concerns since `Reindex` and `Deindex` have different set of parameters -- e.g  `Reindex` need `StartWatermark` and `EndWatermark` which are useless for `Deindex`.

src/Microsoft.Health.Dicom.Functions/Configs/ReindexConfiguration.cs
pengchen0692: I will change this to `ReindexOperationConfiguration`, since we decided to call it `ReindexOperation` instead of `ReindexJob` in design doc.

src/Microsoft.Health.Dicom.SqlServer/Registration/DicomFunctionsBuilderSqlServerRegistrationExtensions.cs
pengchen0692: Working on...

src/Microsoft.Health.Dicom.SqlServer/Registration/DicomFunctionsBuilderSqlServerRegistrationExtensions.cs
pengchen0692: Working on...

src/Microsoft.Health.Dicom.Functions/Configs/DicomFunctionsConfiguration.cs
pengchen0692: I will contains configuration for other operations like DeindexOperation

src/Microsoft.Health.Dicom.SqlServer/Registration/DicomFunctionsBuilderSqlServerRegistrationExtensions.cs
pengchen0692: I put both extension method (`AddSqlServer`) into same class `ServiceCollectionRegistrationExtensions`, so that can easily extract common part.
So far I don't find much things in common. As we add more and more into Azure Functions, that could change.
I'm thinking of keep refactoring as we go,  how do you think?

src/Microsoft.Health.Dicom.Core/Extensions/IServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,82 @@
...
+            return services;
+        }
+        public static IServiceCollection AddScopedDefault<T1, T2, T3, T4>(this IServiceCollection services)
+        {
+            services.AddScopedDefault<T1>();
+            services.AddScopedDefault<T2>();
+            services.AddScopedDefault<T3>();
+            services.AddScopedDefault<T4>();
+            return services;
+        }
```
wsugarman, line 65: It seems unnecessary to support these `T1, T2, ... Tn` overloads in that is it really necessary to group together services in a single call.

src/Microsoft.Health.Dicom.Core/Features/ExtendedQueryTag/AddExtendedQueryTagService.cs
```diff
@@ -30,7 +34,7 @@ public class AddExtendedQueryTagService : IAddExtendedQueryTagService
             EnsureArg.IsNotNull(extendedQueryTagEntryValidator, nameof(extendedQueryTagEntryValidator));
             EnsureArg.IsNotNull(extendedQueryTagConfiguration?.Value, nameof(extendedQueryTagConfiguration));
 
-            _extendedQueryTagStore = extendedQueryTagStoreFactory.GetInstance();
+            _extendedQueryTagStore = new Lazy<IExtendedQueryTagStore>(() => extendedQueryTagStoreFactory.GetInstance());
```
wsugarman, line 37: With this service probably being "scoped," it seems strange to also have a nested `Lazy<T>`. Couldn't we resolve the instance via DI every time this service is resolved?

src/Microsoft.Health.Dicom.Core/Features/Indexing/InstanceReindexer.cs
wsugarman, line 14: nit: A lot of classes being with `Represents ...`

src/Microsoft.Health.Dicom.Core/Registration/ServiceCollectionRegistrationExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <summary>
+        /// Add core components for Dicom Operations.
+        /// </summary>
+        /// <param name="builder">the DicomOperationsBuilder.</param>
+        /// <returns>The DicomOperationsBuilder</returns>
+        public static IDicomFunctionsBuilder AddDicomOperationsCore(
+            this IDicomFunctionsBuilder builder)
+        {
+            EnsureArg.IsNotNull(builder, nameof(builder));
+            var services = builder.Services;
```
wsugarman, line 25: nit: you could chain these, right?

```csharp
builder.Services
    .AddScopedDefault<InstanceReindexer>()
    .AddScopedDefault<AddExtendedQueryTagService>()
    .AddSingletonDefault<DicomTagParser>()
    .AddSingletonDefault<ExtendedQueryTagEntryValidator>()
```

src/Microsoft.Health.Dicom.Core/Registration/ServiceCollectionRegistrationExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    public static class ServiceCollectionRegistrationExtensions
+    {
+        /// <summary>
+        /// Add core components for Dicom Operations.
+        /// </summary>
+        /// <param name="builder">the DicomOperationsBuilder.</param>
+        /// <returns>The DicomOperationsBuilder</returns>
+        public static IDicomFunctionsBuilder AddDicomOperationsCore(
+            this IDicomFunctionsBuilder builder)
```
wsugarman, line 22: nit: this could go on the previous line. I think there's enough room

src/Microsoft.Health.Dicom.Core/Microsoft.Health.Dicom.Core.csproj
```diff
@@ -27,6 +26,7 @@
     <PackageReference Include="Microsoft.Health.Extensions.DependencyInjection" Version="$(HealthcareSharedPackageVersion)" />
     <PackageReference Include="Microsoft.IO.RecyclableMemoryStream" Version="2.1.1" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+    <PackageReference Include="Newtonsoft.Json.Schema" Version="3.0.14" />
```
wsugarman, line 29: Where is this used?

src/Microsoft.Health.Dicom.Core/Microsoft.Health.Dicom.Core.csproj
```diff
@@ -16,7 +16,6 @@
     <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="$(SdkPackageVersion)" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="$(SdkPackageVersion)" />
     <PackageReference Include="Microsoft.Extensions.Diagnostics.HealthChecks" Version="$(SdkPackageVersion)" />
-    <PackageReference Include="Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions" Version="$(SdkPackageVersion)" />
```
wsugarman, line 19: I think we actually depend on this, don't we?

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Activity.cs
```diff
@@ -128,5 +128,23 @@ public async Task ReindexInstancesAsync([ActivityTrigger] ReindexInstanceInput i
 
             await Task.WhenAll(tasks);
         }
+
+        /// <summary>
+        ///  Fetch schema version.
```
wsugarman, line 133: nit: looks like there is an extra space here

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Orchestration.cs
```diff
@@ -32,6 +32,7 @@ public partial class ReindexDurableFunction
         {
             EnsureArg.IsNotNull(context, nameof(context));
             logger = context.CreateReplaySafeLogger(logger);
+            await context.CallActivityAsync(nameof(FetchSchemaVersionAsync), null);
```
wsugarman, line 35: We just making the call for  now without checking it?

src/Microsoft.Health.Dicom.Functions/Registration/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// Add services for DICOM functions.
+        /// </summary>
+        /// <param name="services">The DICOM function builder instance.</param>
+        /// <param name="configuration">The configuration</param>
+        /// <returns>The DICOM function builder instance.</returns>
+        public static IDicomFunctionsBuilder AddDicomFunctions(this IServiceCollection services, IConfiguration configuration)
+        {
+            EnsureArg.IsNotNull(services, nameof(services));
+            DicomFunctionsConfiguration dicomOperationsConfig = new DicomFunctionsConfiguration();
+            configuration?.GetSection(DicomFunctionsConfiguration.SectionName).Bind(dicomOperationsConfig);
```
wsugarman, line 28: nit: we can use `Get<T>` instead of `Bind(object)`.

Eg.
```csharp
configuration.GetSection(DicomFunctionsConfiguration.SectionName).Get<DicomFunctionsConfiguration>();
```

src/Microsoft.Health.Dicom.Functions/Startup.cs
```diff
@@ -3,36 +3,26 @@
...
-                    .GetSection(HostSectionName)
-                    .GetSection(IndexingConfiguration.SectionName)
-                    .Bind(sectionObj));
-
-            builder.Services
-                .AddMvcCore()
-                .AddNewtonsoftJson(x => x.SerializerSettings.Converters
-                    .Add(new StringEnumConverter()));
+            EnsureArg.IsNotNull(builder, nameof(builder));
+            IConfiguration configuration = builder.GetContext().Configuration?.GetSection(AzureFunctionsJobHostSection);
```
wsugarman, line 22: Should we be so permissive for missing configurations?

src/Microsoft.Health.Dicom.Functions/Registration/ServiceCollectionExtensions.cs
wsugarman, line 28: Should we be so permissive for missing configurations?

src/Microsoft.Health.Dicom.Core/Registration/ServiceCollectionRegistrationExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace Microsoft.Health.Dicom.Core.Registration
+{
+    public static class ServiceCollectionRegistrationExtensions
+    {
+        /// <summary>
+        /// Add core components for Dicom Operations.
+        /// </summary>
+        /// <param name="builder">the DicomOperationsBuilder.</param>
+        /// <returns>The DicomOperationsBuilder</returns>
+        public static IDicomFunctionsBuilder AddDicomOperationsCore(
```
wsugarman, line 21: Perhaps just `AddDicomOperations` or `AddDicomOperationServices`?

src/Microsoft.Health.Dicom.Functions/Startup.cs
```diff
@@ -3,36 +3,26 @@
...
+            builder.Services.AddDicomFunctions(configuration)
```
wsugarman, line 23: Should `AddDicomFunctions` just call `AddDicomOperationsCore`?

src/Microsoft.Health.Dicom.SqlServer/Registration/ServiceCollectionRegistrationExtensions.cs
```diff
@@ -0,0 +1,145 @@
...
+
+            services.AddSingletonDefault<DefaultSqlConnectionFactory>();
+
+            return services;
+        }
+
+        private static SqlServerDataStoreConfiguration GetConfiguration(IConfiguration configurationRoot)
+        {
+            var config = new SqlServerDataStoreConfiguration();
+            configurationRoot?.GetSection("SqlServer").Bind(config);
```
wsugarman, line 141: nit: use `Get<T>` here instead of `Bind(object)`

src/Microsoft.Health.Dicom.SqlServer/Registration/ServiceCollectionRegistrationExtensions.cs
```diff
@@ -0,0 +1,145 @@
...
+            services.AddScopedDefault<SqlChangeFeedStore>();
+
+            services.AddSqlExtendedQueryTagStores();
+
+            return dicomServerBuilder;
+        }
+
+        private static void AddSqlExtendedQueryTagStores(this IServiceCollection services)
+        {
+            services.AddScopedDefault<
```
wsugarman, line 106: Now I see why you want this method overload, but I think it's best just to separate this out into multiple calls. Actually, I think perhaps in the future we should refactor these "versioned" services so we can dynamically register all of the versions we find in the assembly.

But given we'll need more method overloads as we add new versions, I think it's best just to have separate method calls.

src/Microsoft.Health.Dicom.SqlServer/Registration/ServiceCollectionRegistrationExtensions.cs
```diff
@@ -0,0 +1,145 @@
...
+        private static IServiceCollection AddInitializedSqlServerBase(
+           this IServiceCollection services,
+           SqlServerDataStoreConfiguration configuration)
+        {
+            //  SqlServerDataStoreConfiguration is consumed by DefaultSqlConnectionStringProvider
+            services.AddSingleton(configuration);
+
+            // TODO: consider moving these logic into healthcare-shared-components (https://github.com/microsoft/healthcare-shared-components/)
+            // once code becomes solid (e.g: merging back to main branch).                 
+            services.AddScopedDefault<SqlTransactionHandler>();
```
wsugarman, line 122: nit: we should just chain these method calls, right?

src/Microsoft.Health.Dicom.SqlServer/Registration/ServiceCollectionRegistrationExtensions.cs
```diff
@@ -0,0 +1,145 @@
...
+            services.AddScopedDefault<
+                SqlIndexDataStoreV1,
+                SqlIndexDataStoreV2,
+                SqlIndexDataStoreV3,
+                SqlStoreFactory<ISqlIndexDataStore, IIndexDataStore>>();
+
+            // TODO: Ideally, the logger can be registered in the API layer since it's agnostic to the implementation.
+            // However, the current implementation of the decorate method requires the concrete type to be already registered,
+            // so we need to register here. Need to some more investigation to see how we might be able to do this.
+            services.Decorate<ISqlIndexDataStore, SqlLoggingIndexDataStore>();
```
wsugarman, line 91: nit: chain these methods together

src/Microsoft.Health.Dicom.SqlServer/Registration/ServiceCollectionRegistrationExtensions.cs
```diff
@@ -0,0 +1,145 @@
...
+        {
+            EnsureArg.IsNotNull(builder, nameof(builder));
+            IServiceCollection services = builder.Services;
+            SqlServerDataStoreConfiguration config = GetConfiguration(configurationRoot);
+            services.AddSingleton(Options.Options.Create(config));
+
+            SchemaInformation schemaInformation = new SchemaInformation(SchemaVersionConstants.Min, SchemaVersionConstants.Max);
+            services.AddSingleton(schemaInformation);
+
+            AddInitializedSqlServerBase(services, config);
```
wsugarman, line 47: nit: use the extension syntax as opposed to calling this using the `static` syntax

src/Microsoft.Health.Dicom.SqlServer/Registration/ServiceCollectionRegistrationExtensions.cs
```diff
@@ -0,0 +1,145 @@
...
+        private static void AddSqlExtendedQueryTagStores(this IServiceCollection services)
+        {
+            services.AddScopedDefault<
+                SqlExtendedQueryTagStoreV1,
+                SqlExtendedQueryTagStoreV2,
+                SqlExtendedQueryTagStoreV3,
+                SqlStoreFactory<ISqlExtendedQueryTagStore, IExtendedQueryTagStore>>();
+        }
+
+        private static IServiceCollection AddInitializedSqlServerBase(
```
wsugarman, line 113: Maybe `AddCommonSqlServices` or `AddSqlConnectionServices`?

wsugarman (approved):

src/Microsoft.Health.Dicom.Core/Features/ExtendedQueryTag/AddExtendedQueryTagService.cs
pengchen0692: I didn't find a different way out so far.
In the context of Dicom Server, these components (`AddExtendedQueryTagService` etc.) are initialized after the host service which initialize database and update schemaInformation correctly. In Azure Function , we don't have such things.

wsugarman: Ultimately, can we instead call the factory in the methods to resolve the proper store instead of in the ctor? Sort of like the `HttpClientFactory`?

pengchen0692: I haven't figure out a good way for this.
The underlying reason is that DicomServer and DicomFunctions(the durable function) initialize ExtendedQueryTagStore in different ways -- DicomServer has a hostedService which initialize SQL before `AddExtendedQueryTagService` is initialized, so `SchemaInformation.CurrentVersion` already have the version. For DicomFunctions, we have to set up the version as part of orchestration, so that value always be `Null` when DI.
That means `SqlStoreFactory` doesn't fit for both case, we may need a different implementation for DicomFunctions.

src/Microsoft.Health.Dicom.Core/Features/ExtendedQueryTag/AddExtendedQueryTagService.cs
smithago: Not true, we need to initialize the SqlServer components before initializing the backgroundworkers.
You just need to do the same in Azure functions. The order in how you initialize things matter

src/Microsoft.Health.Dicom.Core/Registration/IDicomFunctionsBuilder.cs
smithago: Does this class go to Azure Functions project then?

src/Microsoft.Health.Dicom.Core/Registration/ServiceCollectionRegistrationExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace Microsoft.Health.Dicom.Core.Registration
+{
+    public static class ServiceCollectionRegistrationExtensions
+    {
+        /// <summary>
+        /// Add core components for Dicom Operations.
+        /// </summary>
+        /// <param name="builder">the DicomOperationsBuilder.</param>
+        /// <returns>The DicomOperationsBuilder</returns>
+        public static IDicomFunctionsBuilder AddDicomOperationsCore(this IDicomFunctionsBuilder builder)
```
smithago, line 21: Same here.

smithago, line 21: Dicom core should not know anything about functions or Operations

src/Microsoft.Health.Dicom.Functions/Configs/DicomFunctionsConfiguration.cs
smithago: got it

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Orchestration.cs
```diff
@@ -32,6 +32,7 @@ public partial class ReindexDurableFunction
...
             var input = context.GetInput<IReadOnlyList<AddExtendedQueryTagEntry>>();
             var storeEntires = await context.CallActivityAsync<IReadOnlyList<ExtendedQueryTagStoreEntry>>(
```
smithago, line 37: spelling

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Orchestration.cs
```diff
@@ -32,6 +32,7 @@ public partial class ReindexDurableFunction
         {
             EnsureArg.IsNotNull(context, nameof(context));
```
smithago, line 33: Should we rethink this method. This is supposed to just reindex not add. 
It can change the status to "processing". But it should be added in Dicom server

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Activity.cs
```diff
@@ -128,5 +128,23 @@ public async Task ReindexInstancesAsync([ActivityTrigger] ReindexInstanceInput i
...
+        /// <param name="log">The log</param>
+        /// <returns>The task.</returns>
+        [FunctionName(nameof(FetchSchemaVersionAsync))]
+        public async Task FetchSchemaVersionAsync([ActivityTrigger] IDurableActivityContext context, ILogger log)
+        {
+            // TODO: performance improvement, don't need to call service for every call.
+            EnsureArg.IsNotNull(context, nameof(context));
+            EnsureArg.IsNotNull(log, nameof(log));
+            log.LogInformation("Fetching schema version");
+            int version = await _schemaManagerDataStore.GetCurrentSchemaVersionAsync(default);
```
smithago, line 145: This is odd. This is done as part of SchemaInitializer.Start, we should not be doing this here

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Orchestration.cs
smithago: Will is there a way to initialize services like in web apps in Azure functions?

src/Microsoft.Health.Dicom.SqlServer/Features/Indexing/SqlReindexStore.cs
smithago: Check with Will again, we discussed to simplify this. You should maybe add a todoapi and wait for the service implementor to add the right methods

src/Microsoft.Health.Dicom.SqlServer/Registration/ServiceCollectionRegistrationExtensions.cs
```diff
@@ -0,0 +1,131 @@
...
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Health.Extensions.DependencyInjection;
+using Microsoft.Health.Dicom.SqlServer.Features.Store;
+using Microsoft.Health.Dicom.Core.Features.Store;
+using Microsoft.Health.Dicom.SqlServer.Features.Query;
+using Microsoft.Health.Dicom.SqlServer.Features.ChangeFeed;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    public static class ServiceCollectionRegistrationExtensions
```
smithago, line 33: Has discussed we should reuse the DicomServer one

src/Microsoft.Health.Dicom.Functions.UnitTests/Indexing/ReindexDurableFunctionTests.Orchestration.cs
```diff
@@ -31,16 +31,26 @@ public async Task GivenValidTags_WhenAddExtendedQueryTagsOrchestrationAsync_Shou
             IDurableOrchestrationContext context = Substitute.For<IDurableOrchestrationContext>();
             context.GetInput<IReadOnlyList<AddExtendedQueryTagEntry>>().Returns(tagsEntries);
             await _reindexDurableFunction.AddAndReindexTagsAsync(context, NullLogger.Instance);
+
```
smithago, line 34: Can we be consistent with the names use Reindexing everywhere?

Microsoft.Health.Dicom.Functions.UnitTests.Indexing, is called indexing

src/Microsoft.Health.Dicom.SqlServer/Features/Indexing/SqlReindexStore.cs
pengchen0692: removed.

src/Microsoft.Health.Dicom.SqlServer/Registration/ServiceCollectionRegistrationExtensions.cs
pengchen0692: removed

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Activity.cs
pengchen0692: Will could explain more on this. :)

src/Microsoft.Health.Dicom.SqlServer/Registration/DicomServerBuilderSqlServerRegistrationExtensions.cs
```diff
@@ -17,34 +17,53 @@
...
+        {
             var config = new SqlServerDataStoreConfiguration();
             configurationRoot?.GetSection("SqlServer").Bind(config);
 
             services.Add(provider =>
-                {
-                    configureAction?.Invoke(config);
-                    return config;
-                })
+            {
```
pengchen0692, line 63: fix indents

src/Microsoft.Health.Dicom.Functions/host.json
```diff
@@ -8,10 +8,21 @@
       }
     }
   },
-  "indexing": {
-    "add": {
-      "batchSize": 1000,
-      "parallelBatches": 10
+  "DicomFunctions": {
```
smithago, line 11: This project/runtime needs to be a dummy worker. All feature settings need to happen in one place in Dicom web app.

pengchen0692: How we could do that? I mean one is host.json, the other is appsetting.json.

src/Microsoft.Health.Dicom.Functions.UnitTests/Indexing/ReindexDurableFunctionTests.cs
```diff
@@ -7,37 +7,46 @@
...
             _reindexStore = Substitute.For<IReindexStore>();
             _instanceReindexer = Substitute.For<IInstanceReindexer>();
             _addExtendedQueryTagService = Substitute.For<IAddExtendedQueryTagService>();
             _instanceStore = Substitute.For<IInstanceStore>();
             _extendedQueryTagStore = Substitute.For<IExtendedQueryTagStore>();
+            _schemaManagerDataStore = Substitute.For<ISchemaManagerDataStore>();
+            _schemaInformation = new SchemaInformation(SchemaVersionConstants.Min, SchemaVersionConstants.Max);
             _reindexDurableFunction = new ReindexDurableFunction(
-                Options.Create(new IndexingConfiguration() { Add = _reindexConfig }),
+                Options.Create(new DicomFunctionsConfiguration() { Reindex = _reindexConfig }),
```
brandonpollett, line 42: Any particular reason we're not just Substituting for `IOptions<DicomFunctionsConfiguration>`?

brandonpollett: Also, wouldn't it be `DicomFunctionsConfiguration.ServicesConfiguration.Reindex` or am I missing something??

src/Microsoft.Health.Dicom.Functions/Configs/DicomFunctionsConfiguration.cs
```diff
@@ -0,0 +1,21 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using Microsoft.Health.Dicom.Core.Configs;
+
+namespace Microsoft.Health.Dicom.Operations.Functions.Configs
+{
+    /// <summary>
+    /// Represents configuration settings related to the Dicom functions.
```
brandonpollett, line 11: nit: `Dicom` is redundant in the name and we chose not to prefix most of our classes with it.

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Activity.cs
```diff
@@ -128,5 +128,23 @@ public async Task ReindexInstancesAsync([ActivityTrigger] ReindexInstanceInput i
...
+            _schemaInformation.Current = version;
```
brandonpollett, line 146: We most likely need to register our use of this schema version in the database also. What happens if this is running and the schema version is upgraded under the hood? Ideally we would inform the database of our use of the schema along with the min/max that it supports.

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.cs
```diff
@@ -18,33 +21,41 @@ namespace Microsoft.Health.Dicom.Functions.Indexing
...
             EnsureArg.IsNotNull(configOptions, nameof(configOptions));
             EnsureArg.IsNotNull(reindexStore, nameof(reindexStore));
             EnsureArg.IsNotNull(instanceReindexer, nameof(instanceReindexer));
             EnsureArg.IsNotNull(addExtendedQueryTagService, nameof(addExtendedQueryTagService));
             EnsureArg.IsNotNull(instanceStore, nameof(instanceStore));
             EnsureArg.IsNotNull(extendedQueryTagStore, nameof(extendedQueryTagStore));
-            _reindexConfig = configOptions.Value.Add;
+            EnsureArg.IsNotNull(schemaManagerDataStore, nameof(schemaManagerDataStore));
+            EnsureArg.IsNotNull(schemaInformation, nameof(schemaInformation));
+            _reindexConfig = configOptions.Value.Services.Reindex;
```
brandonpollett, line 51: We have generally injected something like `IOptions<ReindexOperationConfiguration>` elsewhere.

src/Microsoft.Health.Dicom.Functions/Registration/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+            services.AddSingleton(Options.Create(dicomOperationsConfig));
+
+            services.AddMvcCore()
+                .AddNewtonsoftJson(x => x.SerializerSettings.Converters
+                .Add(new StringEnumConverter()));
+
+            services.RegisterAssemblyModules(typeof(ServiceModule).Assembly, dicomOperationsConfig.Features, dicomOperationsConfig.Services);
+            return new DicomServerBuilder(services);
+        }
+        private class DicomServerBuilder : IDicomServerBuilder
```
brandonpollett, line 40: nit: missing line break

src/Microsoft.Health.Dicom.Functions/Registration/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+        {
+            public DicomServerBuilder(IServiceCollection services)
+            {
+                EnsureArg.IsNotNull(services, nameof(services));
+                Services = services;
+            }
+
+            public IServiceCollection Services { get; }
+        }
+
```
brandonpollett, line 50: nit: extra line

src/Microsoft.Health.Dicom.SqlServer/Registration/DicomServerBuilderSqlServerRegistrationExtensions.cs
```diff
@@ -17,34 +17,53 @@
...
         {
-            EnsureArg.IsNotNull(dicomServerBuilder, nameof(dicomServerBuilder));
-            IServiceCollection services = dicomServerBuilder.Services;
-
-            services.AddSqlServerBase<SchemaVersion>(configurationRoot);
-            services.AddSqlServerApi();
+            EnsureArg.IsNotNull(builder, nameof(builder));
+            IServiceCollection services = builder.Services;
+            services.AddSqlServerCommon(configurationRoot, configureAction);
+            if (initializeSchema)
```
brandonpollett, line 42: This name seems off. It's deciding way more than whether to initialize the schema or not.

brandonpollett: I wonder if some sort of broken down composition makes more sense.

src/Microsoft.Health.Dicom.Functions/host.json
pengchen0692: Talked with Will, there will be another PR for moving AddTags logic back to DicomServer, in this way, all feature configs should be removed from here.
For this PR, I just remove them from `host.json`

src/Microsoft.Health.Dicom.Core/Registration/DicomFunctionsBuilderCoreRegistrationExtensions.cs
pengchen0692: The class has been removed.

src/Microsoft.Health.Dicom.Core/Features/ExtendedQueryTag/AddExtendedQueryTagService.cs
pengchen0692: Will's PR https://github.com/microsoft/dicom-server/pull/843 should solve this problem.

src/Microsoft.Health.Dicom.Functions.UnitTests/Indexing/ReindexDurableFunctionTests.cs
pengchen0692: * just moved to Substitution
* I'm not sure if we want to add another layer (`Services`) because so far we don't have other things like `Security`, `Features` etc.

src/Microsoft.Health.Dicom.Functions/Configs/DicomFunctionsConfiguration.cs
pengchen0692: It simulates [`DicomServerConfiguration`](https://github.com/microsoft/dicom-server/blob/main/src/Microsoft.Health.Dicom.Api/Configs/DicomServerConfiguration.cs)...

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Activity.cs
pengchen0692: Yes, that's something Will and I discuss about. 
Reindex job could take very long time, what if database schema is upgrade in the between? 
In Dicom Web Server, we always have latest binary, while in Azure Function, we could have old version. 
How should we handle with that is still an open question. 
So I'm thinking maybe we should start from simplest and slowly improve it.
Btw, Will has another PR to simplify the way to get version https://github.com/microsoft/dicom-server/pull/843

src/Microsoft.Health.Dicom.SqlServer/Registration/DicomServerBuilderSqlServerRegistrationExtensions.cs
pengchen0692: The problem comes from Azure Function doesn't support `IHostedService` (https://github.com/Azure/azure-functions-host/issues/5447#issuecomment-575368316) -- `SchemaInitializer` is an hosted service. Because of this, we can not use `AddSqlServerBase`.

One thought is Azure Function don't need to upgrade Database schema, the ability to connect is good enough. so the solution would be only initialize those services/components required for connection. The best place the made the change should be in shared component, but as this moment, the feature has not completed yet, these code could change dramatically. Thus I duplicated code from  https://github.com/microsoft/healthcare-shared-components/blob/af69cf5660398ffe2980dd31cca6443e1ee7d433/src/Microsoft.Health.SqlServer/Registration/SqlServerBaseRegistrationExtensions.cs#L22 as a start point.

I don't have a good solution for this, one possible solution could be: separate `AddSqlServerBase` into 2 parts -- utilities for SqlConnection (`PollyRetryLoggerFactory`, `RetrySqlCommandWrapperFactory` etc) and schemaUpgrade (`SchemaInitializer`, `SchemaJobWorker` etc.)

src/Microsoft.Health.Dicom.Functions/Registration/ServiceCollectionExtensions.cs
pengchen0692: removed

smithago (approved):

src/Microsoft.Health.Dicom.Core/Features/Indexing/InstanceReindexer.cs
brandonpollett: They're basically the same. (maybe there are more extensions on `IReadOnlyList`, but didn't compare them all) 

I believe the biggest difference is the addition of an indexer on `IReadOnlyList`. You can also obtain a an `IReadOnlyCollection` from a `List` via https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.asreadonly?view=net-5.0

brandonpollett (approved):

PR #836, opened 2021-06-15
Add ApiVersion v1.0-prerelease

jnlycklama:
## Description
- Created v1.0-prerelease as the first version of the API. Routes are like: https://localhost:63838/v1.0-prerelease/studies
- Un-versioned routes are still valid, like: https://localhost:63838/studies
- "prerelease" was used instead of "preview" because of https://github.com/microsoft/aspnet-api-versioning/issues/750
- The Retrieve route returned in the Study response will have the version in the URL if Study was called with a version. Ex:                         "https://localhost:63838/v1.0-prerelease/studies/1.2.826.0.1.3680043.8.498.13230779778012324449356534479549187420/series/1.2.826.0.1.3680043.8.498.45787841905473114233124723359129632652/instances/1.2.826.0.1.3680043.8.498.47359123102728459884412887463296905395"

## Related issues
Addresses [issue [#82658](https://microsofthealth.visualstudio.com/Health/_workitems/edit/82658)].

## Testing
- E2E / unit tests added
- Ran project locally & tested using Postman


src/Microsoft.Health.Dicom.Api/Features/Routing/KnownRoutes.cs
```diff
@@ -12,6 +12,7 @@ public static class KnownRoutes
         private const string InstancesRouteSegment = "instances";
         private const string MetadataSegment = "metadata";
         private const string ExtendedQueryTagsRouteSegment = "extendedquerytags";
+        private const string VersionSegment = "v{version:apiVersion}";
```
abdullah248, line 15: Wondering if there are any thoughts on if we want it to be 
service_url/api/v1
or
service_url/v1

src/Microsoft.Health.Dicom.Web/Startup.cs
```diff
@@ -38,6 +41,19 @@ public virtual void ConfigureServices(IServiceCollection services)
                 .AddSqlServer(Configuration)
                 .AddBackgroundWorkers();
 
+            services.AddMvc()
```
abdullah248, line 44: Should the versioning be configured here or in the API project?

src/Microsoft.Health.Dicom.Api/Features/Routing/KnownRoutes.cs
jnlycklama: I hadn't thought about that. I'm ok with either way, although the non-versioned url would not have the /api in it since it needs to stay unchanged

src/Microsoft.Health.Dicom.Web/Startup.cs
jnlycklama: It has to be in the Startup class I think. I'm not sure if there's a way to move this to the API project

src/Microsoft.Health.Dicom.Api/Features/Routing/KnownRoutes.cs
abdullah248: Yea I've seen APIs with and without the /api/. 
@smithago is there any preference for either.

src/Microsoft.Health.Dicom.Web/Startup.cs
wsugarman: Our APIs after this change will leverage verisoning as part of their respective routes. As such, we should make API verisoning a requirement of our API project and in turn register the appropriate API versioning services at the same time we register our "MVC" services. Currently this registration happens in the `AddDicomServer` method used above, so I think we should add any versioning service registration there too.

Furthermore, it would have the added benefit of re-using the same code between OSS and PAAS.

src/Microsoft.Health.Dicom.Api/Features/Routing/KnownRoutes.cs
wsugarman: Personally, since we have the OHIF viewer optionally at the root, it may make sense to keep `/api` in the path to direct users to our REST APIs.

src/Microsoft.Health.Dicom.Api/Controllers/ChangeFeedController.cs
```diff
@@ -19,6 +19,8 @@
 
 namespace Microsoft.Health.Dicom.Api.Controllers
 {
+    [ApiController]
```
wsugarman, line 22: This includes a couple other features detailed [here](https://docs.microsoft.com/en-us/aspnet/core/web-api/?view=aspnetcore-5.0#apicontroller-attribute). Is this a necessary change to use this API versioning library? Perhaps it also has functionality we want?

I see you also already disable the problem details, which I think is reasonable given we may not want to change our schemas

src/Microsoft.Health.Dicom.Api/Controllers/ChangeFeedController.cs
```diff
@@ -19,6 +19,8 @@
...
+    [ApiVersion("1.0-prerelease")]
```
wsugarman, line 23: If we wanted, could we make the prerelease monitor another segment in the path, like `/api/v1.0/prerelease/`?

src/Microsoft.Health.Dicom.Api/Features/Routing/KnownRoutes.cs
```diff
@@ -21,29 +22,48 @@ public static class KnownRoutes
         private const string ExtendedQueryTagPathRouteSegment = "{" + KnownActionParameterNames.TagPath + "}";
 
         public const string StoreRoute = StudiesRouteSegment + "/{" + KnownActionParameterNames.StudyInstanceUid + "?}";
+        public const string VersionedStoreRoute = VersionSegment + "/" + StoreRoute;
```
wsugarman, line 25: Gosh, thank you for adding all of these. But this also look like a big pain to write lol. I tried something that I think should work here; we should extend the `RouteAttribute` class to automatically prepend the version. Ideally we'd define a new `Attribute` that extends `RouteAttribute` and annotate our controller actions with this new attribute too.

Eg.

```csharp
[Route("Foo/Bar")]
[VersionedRoute("Foo/Bar")]
public async Task<IActionResult> GetResponse()
{
    // ...
}
```

Furthermore, it'd be even cooler if we could combine them. Perhaps there is a template syntax that could make the verison optional which we could control via an argument `[VersionedRoute("Foo/Bar", optionalVersion = true)]`

src/Microsoft.Health.Dicom.Api/Features/Routing/UrlResolver.cs
```diff
@@ -46,9 +46,10 @@ public sealed class UrlResolver : IUrlResolver
...
         {
             EnsureArg.IsNotNull(studyInstanceUid, nameof(studyInstanceUid));
+            var hasVersion = _httpContextAccessor.HttpContext.Request.RouteValues.ContainsKey("version");
 
             return RouteUri(
-                KnownRouteNames.RetrieveStudy,
+                hasVersion ? KnownRouteNames.VersionedRetrieveStudy : KnownRouteNames.RetrieveStudy,
                 new RouteValueDictionary()
                 {
                     { KnownActionParameterNames.StudyInstanceUid, studyInstanceUid },
```
wsugarman, line 55: We would want to propagate the version if it was passed, right? Here and below

wsugarman (changes requested):

src/Microsoft.Health.Dicom.Web/Startup.cs
jnlycklama: Ah thanks! I didn't think to look inside `AddDicomServer ` to see what it was doing. I agree it makes sense to move the configuration there. 

& to clarify - the API project is used in PAAS, and the Web project is not? Is that correct?

wsugarman: Yeah, OSS and PAAS have their own respective "web" projects that both leverage the API project for controllers, services, etc

src/Microsoft.Health.Dicom.Api/Controllers/ChangeFeedController.cs
jnlycklama: I don't think this would work. In the ApiVersion class, the status (prerelease) must be appended as "-status". Doing "/status" will throw an error.
We could put it as part of the route, like "/api/v{version:apiVersion}/prerelease/" but then the version would be reported as just 1.0, and to add a new version we would need to register a new route instead of using the same one.

src/Microsoft.Health.Dicom.Api/Features/Routing/UrlResolver.cs
jnlycklama: By getting the route name, it will be able to grab the version used. Just double checked that it works with with multiple versions registered (at least, it works in the E2E tests)

src/Microsoft.Health.Dicom.Api/Features/Routing/KnownRoutes.cs
jnlycklama: Makes sense to me! I've added /api to the versioned route

src/Microsoft.Health.Dicom.Web/Startup.cs
jnlycklama: I've moved the configuration to AddDicomServer

src/Microsoft.Health.Dicom.Api/Controllers/ChangeFeedController.cs
jnlycklama: Ah, I had assumed that it was required! Adding ` c.UseApiBehavior = false;` to the setup allows us to use ApiVersion without adding the controller attribute

src/Microsoft.Health.Dicom.Api/Features/Routing/KnownRoutes.cs
jnlycklama: Great idea - I've added a VersionedRoute attribute.
I didn't add the optionalVersion param though - I think the idea is to get all customers to move off of the non-versioned route & then get rid of it. So routes without versions should go away soon-ish

src/Microsoft.Health.Dicom.Api/Features/Routing/KnownRoutes.cs
jnlycklama: Actually I have removed /api. We discussed at standup, and decided that since PAAS is only APIs and the OHIF viewer is just for testing purposes, it doesn't add value to have a distinct API route

src/Microsoft.Health.Dicom.Api/Features/Routing/UrlResolver.cs
wsugarman: Oh, would `UrlHelper.RouteUrl(...)` return a URL with version we recevied already populated if you pass `KnownRouteNames.VersionedRetrieveStudy` as the name? I wonder how it would do that lol. I assume it would grab the corresponding template `"v{version:apiVersion}/study/{studyInstaneUid}"`, then we pass it a mapping of route values to resolve the template.

If that works, does it work for non-default versions?

src/Microsoft.Health.Dicom.Api/Features/Routing/VersionedRouteAttribute.cs
```diff
@@ -0,0 +1,16 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using Microsoft.AspNetCore.Mvc;
+
+namespace Microsoft.Health.Dicom.Api.Features.Routing
+{
+    public sealed class VersionedRouteAttribute : RouteAttribute
+    {
+        public VersionedRouteAttribute(string template) : base("v{version:apiVersion}/" + template)
```
wsugarman, line 12: nit: I think the style in the repository is to place the nested ctor on the next line and indented:
```csharp
public VersionedRouteAttribute(string template)
    : base("v{version:apiVersion}/" + template)
```

src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
```diff
@@ -94,6 +95,14 @@ public static IDicomServerBuilder AddBackgroundWorkers(this IDicomServerBuilder
                 jsonOptions.JsonSerializerOptions.Converters.Add(new JsonStringEnumConverter());
             });
 
+            services.AddApiVersioning(c =>
+            {
+                c.AssumeDefaultVersionWhenUnspecified = true;
+                c.DefaultApiVersion = new ApiVersion(1, 0, "prerelease");
+                c.ReportApiVersions = true;
```
wsugarman, line 102: We'll probably want to remember to document this new header we'll return

test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/ExtendedQueryTagTests.cs
```diff
@@ -24,15 +24,20 @@ public class ExtendedQueryTagTests : IClassFixture<HttpIntegrationTestFixture<St
...
             EnsureArg.IsNotNull(fixture, nameof(fixture));
             EnsureArg.IsNotNull(fixture.Client, nameof(fixture.Client));
             _client = fixture.Client;
         }
 
-        [Fact]
-        public async Task GivenValidExtendedQueryTags_WhenGoThroughEndToEndScenario_ThenShouldSucceed()
+        [Theory]
+        [InlineData("")]
+        [InlineData(PrereleaseV1VersionPath)]
```
wsugarman, line 39: Instead of inline data, perhaps we should store these centrally for use across all of the tests with `[ClassData]`?

wsugarman: Or are the tests varied enough we cannot really do it? Is there a way to minimize the effort of testing each new version?

src/Microsoft.Health.Dicom.Api/Features/Routing/UrlResolver.cs
jnlycklama: Yeah I tested non-default versions by having [ApiVersion("1.0")] and [ApiVersion("1.0-prerelease")] on the StoreController (with prerelease being the default). It seems to just magically known which version was used without anything needing to be passed in. 
The URL returned in the postman response is:
 - `baseUrl/studies/...` for the non-versioned route (KnownRouteNames.RetrieveStudy)
 - `baseUrl/v1.0-prerelease/studies/...` for 1.0-prerelease
 - `baseUrl/v1.0/studies/...` for 1.0

I've updated the tests in `StoreTransactionTests` that call `ValidateReferencedSopSequence` to ensure that the URLs match their version

src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
jnlycklama: Good call. Where should this documentation go? I didn't see any obvious place in the docs folder

src/Microsoft.Health.Dicom.Api/Features/Routing/VersionedRouteAttribute.cs
jnlycklama: Thanks! Updated

test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/ExtendedQueryTagTests.cs
jnlycklama: Thanks, I'm new to xUnit so just took a look at ClassData & MemberData. Let me know what you think of the updates - it should be much easier to update to new versions now, but I had to hide some of the other inline data.

src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
wsugarman: I imagine when this feature is done we'll have a doc like `api-versioning.md` or something where we'll talk about the URLs, how long a feature is supported ,etc

test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/VersionAPIData.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    public class VersionAPIData : IEnumerable<object[]>
+    {
+        private static readonly List<object[]> VersionSegmentData = new List<object[]>
+        {
+            new object[] { "" },
+            new object[] { "v1.0-prerelease" }
+        };
+
+        public static IEnumerable<object[]> GetVersionData() => VersionSegmentData;
```
wsugarman, line 19: nit: you can combine these like:
```csharp
public static IReadOnlyList<object[]> VersionSegmentData { get; } = new List<object[]> { ... };
```

wsugarman (approved):

test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/VersionAPIData.cs
jnlycklama: Updated!

PR #847, opened 2021-06-22
Reindex Basic implementation - Structure Code

pengchen0692:
## Description
Implement Reindex Operation. 
As a basic implemention, it allows reindex operation be triggered and execute until completion. 

## Related issues
Addresses [AB#83018](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/83018)

## Testing
All tests pass


src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/4.diff.sql
```diff
@@ -0,0 +1,168 @@
...
+/*************************************************************
+    Reindex State Table. 
+    Reindex state on each extended query tag.
+**************************************************************/
+IF NOT EXISTS (
+    SELECT * 
+    FROM sys.tables
+    WHERE name = 'ReindexState')
+BEGIN
+    CREATE TABLE dbo.ReindexState (
```
smithago, line 18: While discussing with Will, we decided we do not need this. All we need is Tag table index status.

smithago: Think like a framework who's concern is to handle the indexing internal state. Externaly all you care is passed or failed. To use it with QIDO.

Job framework also provides job status and error messages if failed

smithago: We should just add Operation ID To existing tag table

smithago (changes requested):
why do we need the new table? Can you follow up with Will

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/4.diff.sql
pengchen0692: Thanks Smitha!
Discussed with Will. 
* How to store internal state is still in discussion.
* This PR would start with a simple solution , which will be incrementally improved later

pengchen0692:
Due to https://microsofthealth.visualstudio.com/Health/_workitems/edit/81200, created new PR https://github.com/microsoft/dicom-server/pull/864

pengchen0692:
Due to https://microsofthealth.visualstudio.com/Health/_workitems/edit/81200, created new PR #864

PR #849, opened 2021-06-22
Add swagger for API versioning

abdullah248:
## Description
Add code to generate swagger and make it available at an endpoint.

When run locally can view swagger ui:  https://localhost:63838/swagger/index.html
and the JSON for v1 can be viewed: https://localhost:63838/swagger/v1-prerelease/swagger.json

Also check in the base swagger for v1-prerelease. This will be used as a base to compare with swagger generated at each pr to make sure no breaking change. This part has not been implemented yet.

## Related issues
[AB#82659](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/82659)

## Testing
Locally made sure that swagger ui worked. Also locally created a different version to see that the swagger json was created for both versions.


src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
```diff
@@ -103,6 +106,16 @@ public static IDicomServerBuilder AddBackgroundWorkers(this IDicomServerBuilder
                 c.UseApiBehavior = false;
             });
 
+            services.AddApiVersioning();
```
jnlycklama, line 109: `AddApiVersioning()` has already been added above - do we need it again?

abdullah248: removed, thanks

wsugarman:
Are multiple versions shown in the swagger UI?

src/Microsoft.Health.Dicom.Api/Registration/Swagger/ConfigureSwaggerOptions.cs
```diff
@@ -0,0 +1,35 @@
...
+using Microsoft.OpenApi.Models;
+using Swashbuckle.AspNetCore.SwaggerGen;
+
+namespace Microsoft.Health.Dicom.Api.Registration.Swagger
+{
+    public class ConfigureSwaggerOptions : IConfigureOptions<SwaggerGenOptions>
+    {
+        private readonly IApiVersionDescriptionProvider _provider;
+
+        public ConfigureSwaggerOptions(IApiVersionDescriptionProvider provider) =>
```
wsugarman, line 18: nit: place the `=>` on the next line, like how `{` would be on the next line: `=>_provider = provider;`

src/Microsoft.Health.Dicom.Api/Registration/Swagger/ConfigureSwaggerOptions.cs
```diff
@@ -0,0 +1,35 @@
...
+          _provider = provider;
```
wsugarman, line 19: nit: `null`-check?

src/Microsoft.Health.Dicom.Api/Registration/Swagger/ConfigureSwaggerOptions.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        public void Configure(SwaggerGenOptions options)
+        {
+            foreach (var description in _provider.ApiVersionDescriptions)
```
wsugarman, line 23: nit: use type name instead of `var` here

src/Microsoft.Health.Dicom.Api/Registration/Swagger/ConfigureSwaggerOptions.cs
```diff
@@ -0,0 +1,35 @@
...
+            {
+                options.SwaggerDoc(
+                  description.GroupName,
+                    new OpenApiInfo()
```
wsugarman, line 27: nit: no `()`

src/Microsoft.Health.Dicom.Api/Registration/Swagger/ConfigureSwaggerOptions.cs
```diff
@@ -0,0 +1,35 @@
...
+                    {
+                        Title = $"API {description.ApiVersion}",
```
wsugarman, line 29: I don't think we need version in the title as there's a separate field for version. So perhaps we should call this something more formal like `"Medical Imaging Server for DICOM"`? We should also fill out the other fields like description, contact, license, etc

src/Microsoft.Health.Dicom.Api/Registration/Swagger/ConfigureSwaggerOptions.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        public ConfigureSwaggerOptions(IApiVersionDescriptionProvider provider) =>
+          _provider = provider;
+
+        public void Configure(SwaggerGenOptions options)
+        {
+            foreach (var description in _provider.ApiVersionDescriptions)
+            {
+                options.SwaggerDoc(
+                  description.GroupName,
```
wsugarman, line 26: I think this alignment is off

src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
```diff
@@ -103,6 +106,15 @@ public static IDicomServerBuilder AddBackgroundWorkers(this IDicomServerBuilder
                 c.UseApiBehavior = false;
             });
 
+            services.AddVersionedApiExplorer(options =>
+            {
+                options.GroupNameFormat = "'v'VVV";
```
wsugarman, line 111: lol what am I looking at here? Is this their format string where like `v` and `V` indicate different objects like `y` is year and `H` is hour for `DateTime`?

abdullah248:
> Are multiple versions shown in the swagger UI?

Yes, when there are multipl versions there is a drop down in the top right that lets you switch between the different versions and you can also acccess the different jsons like this https://localhost:63838/swagger/<version_number>/swagger.json

src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
```diff
@@ -168,6 +182,16 @@ public Action<IApplicationBuilder> Configure(Action<IApplicationBuilder> next)
 
                     app.UseRequestContextAfterAuthentication<IDicomRequestContext>();
 
+                    app.UseSwagger();
+
+                    app.UseSwaggerUI(options =>
+                    {
+                        foreach (var description in provider.ApiVersionDescriptions)
+                        {
+                            options.SwaggerEndpoint($"/swagger/{description.GroupName}/swagger.json", description.GroupName.ToUpperInvariant());
```
wsugarman, line 191: I think after their rename (rebranding?), "Swagger" refers to the tooling whereas "OpenAPI" refers to the specification.

Should we move the version-specific OpenAPI documents to existing versioned routes as opposed to their own routes like `"/{description.GroupName}/api.json"`? I suppose a better question too is: is there a standard convention for exposing these documents and swagger ui?

src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
```diff
@@ -168,6 +182,16 @@ public Action<IApplicationBuilder> Configure(Action<IApplicationBuilder> next)
 
                     app.UseRequestContextAfterAuthentication<IDicomRequestContext>();
 
+                    app.UseSwagger();
+
+                    app.UseSwaggerUI(options =>
+                    {
+                        foreach (var description in provider.ApiVersionDescriptions)
```
wsugarman, line 189: nit: use type name instead of `var`

src/Microsoft.Health.Dicom.Api/Registration/Swagger/SwaggerDefaultValues.cs
```diff
@@ -0,0 +1,45 @@
...
+            EnsureArg.IsNotNull(operation, nameof(operation));
+            EnsureArg.IsNotNull(context, nameof(context));
+
+            var apiDescription = context.ApiDescription;
+            operation.Deprecated |= apiDescription.IsDeprecated();
+
+            if (operation.Parameters == null)
+                return;
+
+            foreach (var parameter in operation.Parameters)
```
wsugarman, line 28: nit: use type name instead of `var`

src/Microsoft.Health.Dicom.Api/Registration/Swagger/SwaggerDefaultValues.cs
```diff
@@ -0,0 +1,45 @@
...
+            {
+                var description = apiDescription.ParameterDescriptions.First(p => p.Name == parameter.Name);
```
wsugarman, line 30: Do we need to match on more than `Name`, like a query string parameter vs path?

src/Microsoft.Health.Dicom.Api/Registration/Swagger/SwaggerDefaultValues.cs
```diff
@@ -0,0 +1,45 @@
...
+                if (parameter.Description == null)
```
wsugarman, line 31: Looks like this code is about backfilling from the data model section of the document if it's not populated in the method's metadata? Is this for the Swagger UI, the generated doc, or both?

swagger/v1-prerelease/swagger.json
```diff
@@ -0,0 +1,5792 @@
...
+            }
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "Success",
+            "content": {
+              "text/plain": {
+                "schema": {
+                  "$ref": "#/components/schemas/JsonResult"
```
wsugarman, line 47: I think things like this we'll want to change in our attributes lol. It's not very helpful to say something's schema is "JSON" haha. Perhaps that is a separate PR to clean up these schemas/contracts.

wsugarman (changes requested):

src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
abdullah248: Sorry, its in this format: 'v'major[.minor][-status]
Ill add a comment detailing it.

swagger/v1-prerelease/swagger.json
abdullah248: Updated in this pr: https://github.com/microsoft/dicom-server/pull/858

src/Microsoft.Health.Dicom.Api/Registration/Swagger/SwaggerDefaultValues.cs
abdullah248: So this code is due to a bug that exists in in swagger generator. I actually copied it over from the solution they said to use in their sample repo: https://github.com/microsoft/aspnet-api-versioning/blob/master/samples/aspnetcore/SwaggerSample/SwaggerDefaultValues.cs

src/Microsoft.Health.Dicom.Api/Registration/Swagger/SwaggerDefaultValues.cs
abdullah248: I believe it is for both, there were some open github issues and this was the suggested solution. I updated the file to match the solution provided in the repo: https://github.com/microsoft/aspnet-api-versioning/blob/master/samples/aspnetcore/SwaggerSample/SwaggerDefaultValues.cs

src/Microsoft.Health.Dicom.Api/Registration/Swagger/ConfigureSwaggerOptions.cs
abdullah248: I updated the title and added the license. I didn't know what to put for contact or description so I havent added them, but if you have any ideas I can add them!

src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
abdullah248: We looked into it a bit, couldnt find any standard and seems like most apis just have their own preference. I left it how it was for now but if you have any preference I can update it.

src/Microsoft.Health.Dicom.Api/Controllers/QueryController.cs
```diff
@@ -43,6 +43,7 @@ public QueryController(IMediator mediator, ILogger<QueryController> logger)
 
         [HttpGet]
         [AcceptContentFilter(new[] { KnownContentTypes.ApplicationDicomJson }, allowSingle: true, allowMultiple: false)]
+        [ProducesAttribute(KnownContentTypes.ApplicationDicomJson)]
```
abdullah248, line 46: @wsugarman I added these because in the generated swagger it was showing other accept types. Adding this fixes it but I want to make sure there arent any unintended consequences.

pengchen0692: `Produces` instead of `ProducesAttribute`?

abdullah248: I moved adding these to a seperate pr

swagger/v1-prerelease/swagger.json
```diff
@@ -0,0 +1,3455 @@
...
+                "schema": {
+                  "type": "string"
+                }
+              }
+            }
+          }
+        }
+      }
+    },
+    "/changefeed": {
```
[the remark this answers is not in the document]
wsugarman: In fact, should we separate out the unspecified into its own doc? I am unsure that'd be possible

swagger/v1-prerelease/swagger.json
```diff
@@ -0,0 +1,3455 @@
+{
```
wsugarman, line 1: At some point we should invest in better examples in the document for our data models

swagger/v1-prerelease/swagger.json
```diff
@@ -0,0 +1,3455 @@
...
+        }
+      },
+      "post": {
+        "tags": [
+          "Store"
+        ],
+        "parameters": [
+          {
+            "name": "studyInstanceUid",
+            "in": "query",
```
wsugarman, line 369: This is not a query-string parameter but a path parameter. Presumably this generated incorrectly because OpenAPI does not support optional path parameters and the method makes this argument optional. To that end, we probably need to have two methods in the `StoreController`: one for each route.

swagger/v1-prerelease/swagger.json
```diff
@@ -0,0 +1,3455 @@
...
+            "name": "sopInstanceUid",
+            "in": "path",
+            "required": true,
+            "schema": {
+              "type": "string"
+            }
+          },
+          {
+            "name": "frames",
+            "in": "query",
```
wsugarman, line 2808: `frames` is also generating incorrectly as a query-string parameter

src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
wsugarman: Perhaps:
- `/api/swagger` - the Swagger UI
- `/api/{version}.yaml` - the OpenAPI specification for the given `version`

It seems this UI is powered by a static file, but I wish it was something whose format, either JSON or YAML, could be converted at runtime. Oh well!

swagger/v1-prerelease/swagger.json
```diff
@@ -0,0 +1,3455 @@
+{
+  "openapi": "3.0.1",
```
wsugarman, line 2: By the way, I see most docs for OpenAPI at least on their website use YAML instead. Should we by default be using YAML?

src/Microsoft.Health.Dicom.Api/Registration/Swagger/ConfigureSwaggerOptions.cs
```diff
@@ -0,0 +1,41 @@
...
+            _provider = provider;
+        }
+
+        public void Configure(SwaggerGenOptions options)
+        {
+            foreach (ApiVersionDescription description in _provider.ApiVersionDescriptions)
+            {
+                options.SwaggerDoc(
+                    description.GroupName,
+                    new OpenApiInfo
```
wsugarman, line 32: Perhaps given the wide breadth of use-cases and customers, it makes sense to leave the description and contact info out

wsugarman (changes requested):

swagger/v1-prerelease/swagger.json
```diff
@@ -0,0 +1,3455 @@
...
+                "schema": {
+                  "type": "string"
+                }
+              }
+            }
+          }
+        }
+      }
+    },
+    "/_schema/versions": {
```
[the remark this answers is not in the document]
abdullah248: So these are coming from healthcare-shared-components: https://github.com/microsoft/healthcare-shared-components/blob/1e0dc8532113a5f7fdd0692c922e91b8273c240e/src/Microsoft.Health.SqlServer.Api/Controllers/SchemaController.cs

What is the best way to handle this?

swagger/v1-prerelease/swagger.json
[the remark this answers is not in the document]
abdullah248: As discussed offline, will leave in this version for now. In future versions we would not want the ones without routes

swagger/v1-prerelease/swagger.json
```diff
@@ -0,0 +1,3455 @@
...
+        ],
+        "responses": {
+          "200": {
+            "description": "Success",
+            "content": {
+              "application/dicom": {
+                "schema": {
+                  "type": "array",
+                  "items": {
+                    "$ref": "#/components/schemas/Stream"
```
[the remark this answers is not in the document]
abdullah248: Should now be either application/dicom or multipart related

swagger/v1-prerelease/swagger.json
```diff
@@ -0,0 +1,3455 @@
...
+          {
+            "name": "includeMetadata",
+            "in": "query",
+            "schema": {
+              "type": "boolean",
+              "default": true
+            }
+          },
+          {
+            "name": "api-version",
```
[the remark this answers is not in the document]
abdullah248: Removed

swagger/v1-prerelease/swagger.json
abdullah248: Added a separate controller

swagger/v1-prerelease/swagger.json
abdullah248: Fixed, should be path now

swagger/v1-prerelease/swagger.json
abdullah248: Changed it to yaml, for the path comment, swashbuckle only supportst he paths I have currently, so 
ui: `/swagger`
yaml:` /swagger/{version}/swagger.yaml`

src/Microsoft.Health.Dicom.Api/Registration/Swagger/ConfigureSwaggerOptions.cs
abdullah248: Left them out

src/Microsoft.Health.Dicom.Api/Registration/Swagger/ConfigureSwaggerOptions.cs
```diff
@@ -0,0 +1,41 @@
...
+                    {
+                        Title = $"Medical Imaging Server for DICOM",
```
wsugarman, line 34: nit: remove `$` as this is not an interpolated string

src/Microsoft.Health.Dicom.Api/Registration/Swagger/ErrorCodeOperationFilter.cs
```diff
@@ -0,0 +1,46 @@
...
+
+                    OpenApiResponse response = operation.Responses[responseKey];
+
+                    foreach (string contentType in response.Content.Keys)
+                    {
+                        if (response.Content.Count == 1)
+                        {
+                            OpenApiMediaType value = response.Content[contentType];
+                            response.Content.Remove(contentType);
+                            response.Content.Add("text/json", value);
```
wsugarman, line 34: The official MIME type is `application/json` as far as I'm aware

src/Microsoft.Health.Dicom.Api/Registration/Swagger/ErrorCodeOperationFilter.cs
```diff
@@ -0,0 +1,46 @@
...
+                            break;
+                        }
+
+                        response.Content.Remove(contentType);
```
wsugarman, line 38: Should this be within an `else` clause? What case is this?

src/Microsoft.Health.Dicom.Api/Registration/Swagger/RetrieveOperationFilter.cs
```diff
@@ -0,0 +1,44 @@
...
+                        OpenApiResponse response = operation.Responses[responseKey];
+
+                        response.Content.Clear();
+
+                        if (operation.OperationId.EndsWith("Instance", StringComparison.OrdinalIgnoreCase))
+                        {
+                            response.Content.Add("application/dicom", new OpenApiMediaType());
+                        }
+
+                        response.Content.Add("multipart/related", new OpenApiMediaType());
```
wsugarman, line 38: Should this media type eventually contain the inner types like `application/dicom`, etc?

src/Microsoft.Health.Dicom.Web/Microsoft.Health.Dicom.Web.csproj
```diff
@@ -40,4 +40,4 @@
     </VisualStudio>
   </ProjectExtensions>
 
-</Project>
+</Project>
```
wsugarman, line 43: can you undo this?

swagger/v1-prerelease/swagger.yaml
```diff
@@ -0,0 +1,1950 @@
...
+          nullable: true
+      additionalProperties: false
+    QueryTagLevel:
+      enum:
+        - 0
+        - 1
+        - 2
+      type: integer
+      format: int32
+    Stream:
```
wsugarman, line 1922: This pesky type is still here lol. Perhaps left over after the operation filters cleaned up its references?

src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
```diff
@@ -168,6 +188,16 @@ public Action<IApplicationBuilder> Configure(Action<IApplicationBuilder> next)
 
                     app.UseRequestContextAfterAuthentication<IDicomRequestContext>();
 
+                    app.UseSwagger();
+
+                    app.UseSwaggerUI(options =>
+                    {
+                        foreach (ApiVersionDescription description in provider.ApiVersionDescriptions)
+                        {
+                            options.SwaggerEndpoint($"/swagger/{description.GroupName}/swagger.yaml", description.GroupName.ToUpperInvariant());
```
wsugarman, line 197: Why is this the path it supports btw? Can we simply change where these files are output because this is presumably a static file that is served?

wsugarman:
Can we also update the default landing page to be the swagger UI?

src/Microsoft.Health.Dicom.Api/Registration/Swagger/ErrorCodeOperationFilter.cs
abdullah248: No, I want to keep one value so I can preserve the schema and remove all the other ones. So i remove all of them until only 1 remaining and then change that ones type

src/Microsoft.Health.Dicom.Api/Registration/Swagger/RetrieveOperationFilter.cs
abdullah248: We could perhaps add it in the future, I think we need to come up with a way to handle the inner types and transfer syntax in the future.

abdullah248:
> Can we also update the default landing page to be the swagger UI?

The accessability team still hasnt gotten back to us so I removed the swagger ui. The yaml file can now be found at `{version}/api.yaml`

src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
abdullah248: updated path

wsugarman (approved):

PR #851, opened 2021-06-23
Add Experimental Data Provider APIs

wsugarman:
## Description
- Add new experimental builder APIs for SQL and Blob for re-use in Azure Functions
- Add blob and SQL services to Azure Functions
- Update config with placeholder

## Related issues
[AB#83001](https://microsofthealth.visualstudio.com/Health/_workitems/edit/83001)

## Testing
Pass CI and tested locally inside of example Azure Functions


src/Microsoft.Health.Dicom.Blob/Registration/Experimental/IBlobServiceBuilder.cs
```diff
@@ -0,0 +1,22 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using EnsureThat;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    public interface IBlobServiceBuilder
```
pengchen0692, line 10: I'm thinking if the `IxxBuilder` is a good pattern.
* Pros:
One pros I see is it prevents the long extension list for `IServiceCollection`
image

* Cons:
1. Extensibility: 
The first extension method which returns `IxxBuilder` has to be called, if you don't want it, you have to create your own `IxxBuilder` implemention (e.g: this duplicate [code](https://github.com/microsoft/dicom-server/blob/03a281fe62a7a46bf7498ccc80ce10501e2ec723/src/Microsoft.Health.Dicom.Functions/Registration/ServiceCollectionExtensions.cs#L44)), which I'm really not a favor.

Actually the project we need is `Microsoft.Health.Dicom.Metadata` instead of `Microsoft.Health.Dicom.Blob`, in this case, how should we make this builder pattern extended to that project?

* Propose
 Not like  [IFunctionsHostBuilder](https://github.com/Azure/azure-functions-dotnet-extensions/blob/main/src/Extensions/DependencyInjection/IFunctionsHostBuilder.cs), whose implementation is [here](https://github.com/Azure/azure-functions-dotnet-extensions/blob/main/src/Extensions/DependencyInjection/FunctionsHostBuilder.cs), our builder is a simply wrapper on `IServiceCollection`. 
So maybe we could get rid of this builder pattern.

src/Microsoft.Health.Dicom.Functions/AzureFunctionsConfiguration.cs
```diff
@@ -0,0 +1,12 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Dicom.Functions
+{
+    internal static class AzureFunctionsConfiguration
```
pengchen0692, line 8: do we need this to be static?

src/Microsoft.Health.Dicom.Functions/Microsoft.Health.Dicom.Functions.csproj
```diff
@@ -18,11 +18,13 @@
     <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="2.4.3" />
     <PackageReference Include="Microsoft.NET.Sdk.Functions" Version="3.0.11" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
-    <PackageReference Include="System.Text.Encodings.Web" Version="4.5.1" PrivateAssets="All" />
+    <PackageReference Include="System.Text.Encodings.Web" Version="4.7.2" PrivateAssets="All" />
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\Microsoft.Health.Dicom.Blob\Microsoft.Health.Dicom.Blob.csproj" />
```
pengchen0692, line 25: we need `Microsoft.Health.Dicom.Metadata.csproj` instead. 
I made this same mistake before. :)

src/Microsoft.Health.Dicom.Functions/host.json
```diff
@@ -13,5 +13,29 @@
...
+    "AllowDatabaseCreation": "false",
+    "ConnectionString": "server=(local);Initial Catalog=Dicom;Integrated Security=true",
+    "TransientFaultRetryPolicy": {
+      "InitialDelay": "00:00:00.100",
+      "RetryCount": 3,
+      "Factor": 2,
+      "FastFirst": true
+    },
+    "SchemaOptions": {
+      "AutomaticUpdatesEnabled": false
```
pengchen0692, line 38: Do we need SchemaOptions?

src/Microsoft.Health.Dicom.Functions/host.json
```diff
@@ -13,5 +13,29 @@
...
+      "ServerTimeoutInMinutes": 2,
+      "DownloadMaximumConcurrency": 5,
+      "UploadMaximumConcurrency": 5
+    }
+  },
+  "SqlServer": {
+    "Initialize": "false",
+    "AllowDatabaseCreation": "false",
+    "ConnectionString": "server=(local);Initial Catalog=Dicom;Integrated Security=true",
+    "TransientFaultRetryPolicy": {
```
pengchen0692, line 31: my best guess is we only need `ConnectionString` and `TransientFaultRetryPolicy` (could be useful later) here, since we don't initialize database.

src/Microsoft.Health.Dicom.SqlServer/Registration/Experimental/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,177 @@
...
+                    services.AddSingleton<ISqlConnectionFactory, DefaultSqlConnectionFactory>();
+                    break;
+            }
+
+            return services;
+        }
+
+        #region Shared Component Copies (TODO: Remove)
+
+        private sealed class RetrySqlCommandWrapper : SqlCommandWrapper
```
pengchen0692, line 94: do we want to move these class into separate partial file so that this file is more clean?

pengchen0692 (commented):
I have some high level thoughts 
1. Is `IxxxBuilder` a good pattern that we should follow? 
2. Should builder classes in  `Microsoft.Health.Dicom.SqlServer` only focus on components it has, and move components from heathcare-share-component into `Microsoft.Health.Dicom.Functions`? I feel the logic is more clean in this way.

src/Microsoft.Health.Dicom.Functions/AzureFunctionsConfiguration.cs
wsugarman: This is meant to be a static class that holds some constants about configuration. So I don't mean for this to be instantiated. Perhaps there is a better name for this, but hopefully `AzureFunctionsConfiguration` is too generic for our own usage as a config object

src/Microsoft.Health.Dicom.Functions/host.json
wsugarman: Good call. We don't

src/Microsoft.Health.Dicom.Functions/host.json
wsugarman: True!

src/Microsoft.Health.Dicom.SqlServer/Registration/Experimental/ServiceCollectionExtensions.cs
wsugarman: This is all temporary and is only used in this class for set-up. I'll keep them here for now, and of course delete them before merging into main

src/Microsoft.Health.Dicom.Metadata/Registration/Experimental/BlobServiceBuilderExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+            IServiceCollection services = EnsureArg.IsNotNull(builder?.Services, nameof(builder));
+
+            services.Add<BlobMetadataStore>()
+                .Scoped()
+                .AsSelf()
+                .AsImplementedInterfaces();
+
+            // TODO: Ideally, the logger can be registered in the API layer since it's agnostic to the implementation.
+            // However, the current implementation of the decorate method requires the concrete type to be already registered,
+            // so we need to register here. Need to some more investigation to see how we might be able to do this.
```
yuedai0531, line 28: nit: `Need to do some`

yuedai0531: wouldn't LoggerFactory work here?

src/Microsoft.Health.Dicom.Metadata/Registration/Experimental/IBlobServiceBuilder.cs
```diff
@@ -0,0 +1,30 @@
...
+        BlobDataStoreConfiguration Configuration { get; }
+    }
+
+    internal sealed class BlobServiceBuilder : IBlobServiceBuilder
+    {
+        public IServiceCollection Services { get; }
+
+        public BlobDataStoreConfiguration Configuration { get; }
+
+        public BlobServiceBuilder(IServiceCollection serviceCollection, BlobDataStoreConfiguration config)
```
yuedai0531, line 24: is it a c# thing to have a constructor for a normal class inside of an Interface file?

src/Microsoft.Health.Dicom.Metadata/Registration/Experimental/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+
+            services.Configure<BlobContainerConfiguration>(
+                Constants.ContainerConfigurationName,
+                containerConfiguration => configurationRoot
+                    .GetSection("DicomWeb:MetadataStore")
+                    .Bind(containerConfiguration));
+
+            // TODO: Leverage Shared Components
+            var config = new BlobDataStoreConfiguration();
+            configurationRoot.GetSection("BlobStore").Bind(config);
```
yuedai0531, line 34: Question: how do you decide when to use Object.SectionName vs having a hard coded string?
I would guess that having it as an attribute would be better in most of the cases?

src/Microsoft.Health.Dicom.Metadata/Registration/Experimental/IBlobServiceBuilder.cs
pengchen0692: may be we should put this builder into a separated file.

src/Microsoft.Health.Dicom.Metadata/Registration/Experimental/ServiceCollectionExtensions.cs
pengchen0692: I think SectionName is better, but BlobDataStoreConfiguration is defined in another repository, so...

src/Microsoft.Health.Dicom.Metadata/Registration/Experimental/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+using Microsoft.Extensions.Configuration;
+using Microsoft.Health.Blob.Configs;
+using Microsoft.Health.Dicom.Metadata;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    public static class ServiceCollectionExtensions
+    {
+        public static IBlobServiceBuilder AddAzureBlobServiceClient(
+            this IServiceCollection services,
```
pengchen0692, line 20: IBlobServiceBuilder instead of IServiceCollection?

pengchen0692: I feel leaving only configuration setup in the first build function and move other logic into BlobServiceBuilderExtensions is more clean, but I'm OK with leaving  the logic to create BlobServiceClient here now , since both WebServer and AzureFunction need it.

src/Microsoft.Health.Dicom.SqlServer/Registration/Experimental/ISqlServiceBuilder.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace Microsoft.Extensions.DependencyInjection
+{
+    public interface ISqlServiceBuilder
+    {
+        IServiceCollection Services { get; }
+
+        SqlServerDataStoreConfiguration Configuration { get; }
+    }
+
+    internal sealed class SqlServiceBuilder : ISqlServiceBuilder
```
pengchen0692, line 18: Move this class into separate file?

src/Microsoft.Health.Dicom.SqlServer/Registration/Experimental/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,174 @@
...
+            configurationRoot.GetSection(SqlServerSectionName).Bind(config);
+
+            services
+                .AddSingleton(config)
+                .AddConnectionServices(config);
+
+            return new SqlServiceBuilder(services, config);
+        }
+
+        private static IServiceCollection AddConnectionServices(
```
pengchen0692, line 47: Same as Blob, I feel changing `AddConnectionServices` to base on `ISqlServiceBuilder` is more clean, but I'm ok with it right now.

src/Microsoft.Health.Dicom.SqlServer/Registration/Experimental/SqlServiceBuilderExtensions.cs
```diff
@@ -0,0 +1,156 @@
...
+{
+    public static class SqlServiceBuilderExtensions
+    {
+        public static ISqlServiceBuilder AddIndexDataStores(this ISqlServiceBuilder builder)
+        {
+            IServiceCollection services = EnsureArg.IsNotNull(builder?.Services, nameof(builder));
+
+            services.Add<SqlIndexDataStoreV1>()
+                .Scoped()
+                .AsImplementedInterfaces();
```
pengchen0692, line 33: Add `.AsSelf()` before `.AsImplementedInterfaces()`, there is an Assert checking this https://github.com/microsoft/healthcare-shared-components/blob/af69cf5660398ffe2980dd31cca6443e1ee7d433/src/Microsoft.Health.Extensions.DependencyInjection/TypeRegistrationBuilder.cs#L125.

src/Microsoft.Health.Dicom.SqlServer/Registration/Experimental/SqlServiceBuilderExtensions.cs
```diff
@@ -0,0 +1,156 @@
...
+
+            services.Add<SqlIndexDataStoreV2>()
+                .Scoped()
+                .AsImplementedInterfaces();
```
pengchen0692, line 37: same as above.

src/Microsoft.Health.Dicom.SqlServer/Registration/Experimental/SqlServiceBuilderExtensions.cs
```diff
@@ -0,0 +1,156 @@
...
+
+            services.Add<SqlIndexDataStoreV3>()
+                .Scoped()
```
pengchen0692, line 40: same as above

pengchen0692 (approved):

src/Microsoft.Health.Dicom.Metadata/Registration/Experimental/BlobServiceBuilderExtensions.cs
wsugarman: This is all copied from the shared components until I check in these new APIS there

src/Microsoft.Health.Dicom.Metadata/Registration/Experimental/ServiceCollectionExtensions.cs
wsugarman: Section name is definitely better. Unfortunately this PR is not a collection of best practices, as I have temporarily copied chunks of code from the shared components to test this API. Once we get it to a place where we like it, then I'll merge it back

src/Microsoft.Health.Dicom.Metadata/Registration/Experimental/ServiceCollectionExtensions.cs
wsugarman: I think the only problem with leaving it out would be the possibility that it's never called (and presumably it's needed by anyone configuring it)

src/Microsoft.Health.Dicom.SqlServer/Registration/Experimental/SqlServiceBuilderExtensions.cs
wsugarman: I see. We're actually missing this for our dicom server registration path, so I'll add it there too

src/Microsoft.Health.Dicom.SqlServer/Registration/Experimental/ServiceCollectionExtensions.cs
wsugarman: I think these sort of things are necessary, so it may not make sense to separate out this core component unlike the others

PR #858, opened 2021-06-28
Fix Produces attribute in controllers

abdullah248:
## Description
There were some routes tagged with ProducesResponseType(typeof(JsonResult) instead of the correct type. Fix types so swagger can be generated correctly.

## Related issues
[AB#82659](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/82659)

## Testing


src/Microsoft.Health.Dicom.Api/Controllers/ChangeFeedController.cs
```diff
@@ -37,7 +38,7 @@ public ChangeFeedController(IMediator mediator, ILogger<ChangeFeedController> lo
         }
 
         [HttpGet]
-        [ProducesResponseType(typeof(JsonResult), (int)HttpStatusCode.OK)]
+        [ProducesResponseType(typeof(IEnumerable<ChangeFeedEntry>), (int)HttpStatusCode.OK)]
```
wsugarman, line 41: Below in `GetChangeFeedLatest`, it says it returns `string`, but that's not quite correct either, right? It is really `ChangeFeedEntry`. Are there more examples of this?

wsugarman (changes requested):

src/Microsoft.Health.Dicom.Api/Controllers/ChangeFeedController.cs
wsugarman: Oh, that's actually for the `BadRequest` below. Whoops!

wsugarman (approved):

PR #866, opened 2021-07-01
Service to startup without schema initialized

rbans96:
## Description
This PR has couple of changes to enable the dicom-service to startup(serve schema apis) even with no schema initialization

## Related issues
Addresses [#AB82101](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Resolute/Stories/?workitem=82101)

## Testing
Tested manually


src/Microsoft.Health.Dicom.Core/Features/HealthCheck/BackgroundServiceHealthCheck.cs
```diff
@@ -41,13 +41,20 @@ public class BackgroundServiceHealthCheck : IHealthCheck
 
         public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context, CancellationToken cancellationToken = default)
         {
-            Task<DateTimeOffset> oldestWaitingToBeDeleated = _backgroundServiceHealthCheckCache.GetOrAddOldestTimeAsync(_indexDataStore.GetOldestDeletedAsync, cancellationToken);
-            Task<int> numReachedMaxedRetry = _backgroundServiceHealthCheckCache.GetOrAddNumExhaustedDeletionAttemptsAsync(
-                t => _indexDataStore.RetrieveNumExhaustedDeletedInstanceAttemptsAsync(_deletedInstanceCleanupConfiguration.MaxRetries, t),
-                cancellationToken);
+            try
```
rbans96, line 44: CheckHealthAsync method doesn't seems to be used as of now, still made change to catch any exception in the background service and report health as unhealthy. Thoughts?

wsugarman: This health check should be in use, but it is initialized via DI and triggered at the endpoint `/health/check`.

While you should indeed catch exceptions that you know can be thrown, we should use "catchalls" very sparingly. Note that if an `IHealthCheck` throws an exception that it result in a health check failure automatically:

See [here](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/health-checks?view=aspnetcore-5.0#create-health-checks):
> If CheckHealthAsync throws an exception during the check, a new HealthReportEntry is returned with its HealthReportEntry.Status set to the FailureStatus, which is defined by AddCheck (see the Register health check services section) and includes the inner exception that initially caused the check failure. The Description is set to the exception's message.

rbans96: Thanks Will for the information. I have pushed the commit to catch specific exception instead of generic one.

src/Microsoft.Health.Dicom.Core/Features/HealthCheck/BackgroundServiceHealthCheck.cs
```diff
@@ -41,13 +42,20 @@ public class BackgroundServiceHealthCheck : IHealthCheck
...
+                    cancellationToken);
 
-            _telemetryClient.GetMetric("Oldest-Requested-Deletion").TrackValue((await oldestWaitingToBeDeleated).ToUnixTimeSeconds());
-            _telemetryClient.GetMetric("Count-Deletions-Max-Retry").TrackValue(await numReachedMaxedRetry);
+                _telemetryClient.GetMetric("Oldest-Requested-Deletion").TrackValue((await oldestWaitingToBeDeleated).ToUnixTimeSeconds());
+                _telemetryClient.GetMetric("Count-Deletions-Max-Retry").TrackValue(await numReachedMaxedRetry);
+            }
+            catch (DataStoreException e) // This is expected when service is starting up without schema initialization
+            {
+                return HealthCheckResult.Unhealthy("Unhealthy service." + e.Message);
```
brandonpollett, line 57: I think we can log the exception message, but we probably shouldn't put it in the health check result. That could lead to data leakage. I think generally we would have a message like "Couldn't connect to the data store."

brandonpollett (approved):

src/Microsoft.Health.Dicom.Core/Features/HealthCheck/BackgroundServiceHealthCheck.cs
rbans96: Thanks, updated!

PR #869, opened 2021-07-06
upgrade healthcare nugets version 3.0.0 and make initialization script transactional and idempotent

rbans96:
## Description
This PR bumps the healthcare-shared nugets version which has the change in schema-manager tool to remove the wrapping up of the initialization scripts under a transaction.
So, this PR also adds multiple transactions and make latest schema initialization script(2.sql) idempotent

## Related issues
Addresses #AB82137

## Testing
Manually tested and existing tests continue to pass.


pengchen0692:
Close for cleaning up resources and will reopen again

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -1,336 +1,503 @@
...
-    PatientName
-)
-INCLUDE
-(
-    StudyKey
-)
-WITH (DATA_COMPRESSION = PAGE)
+IF NOT EXISTS (
+    SELECT * 
+	FROM sys.indexes
```
YazanMSFT, line 240: Nit: there are tabs instead of spaces in a few spots like htis

YazanMSFT (approved):
Looks good to me, would be good to get approval from someone in DICOM team as well.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -1,336 +1,503 @@
--- NOTE: This script DROPS AND RECREATES all database objects.
 -- Style guide: please see: https://github.com/ktaranov/sqlserver-kit/blob/master/SQL%20Server%20Name%20Convention%20and%20T-SQL%20Programming%20Style.md
 
 /*************************************************************
-    Configure database
```
brandonpollett, line 5: Why was this section removed?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
```diff
@@ -1542,3 +1873,6 @@ AS
         
     COMMIT TRANSACTION
 GO
+
+COMMIT TRANSACTION
```
brandonpollett, line 1877: Just for my understanding, what's the purpose of the nested transactions?

brandonpollett (approved):

Directory.Build.props
```diff
@@ -9,7 +9,7 @@
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
     <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <HealthcareSharedPackageVersion>2.1.40</HealthcareSharedPackageVersion>
+    <HealthcareSharedPackageVersion>3.0.6</HealthcareSharedPackageVersion>
```
brandonpollett, line 12: Can you update this to the latest of the shared package version? (I believe 3.0.10)

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
rbans96: This configuration is already part of [baseSchema.sql](https://github.com/microsoft/healthcare-shared-components/blob/master/src/Microsoft.Health.SqlServer/Features/Schema/Migrations/BaseSchema.sql#L5) which is executed before executing the initialization schema script so its redundant over here.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql
rbans96: So I have wrapped up all the SQL statements in a transaction(given the statements are transactional) for lesser recovery points. So the inner transaction is for DeleteExtendedQueryTag stored proc and while outer transaction is for multiple statements.

Directory.Build.props
rbans96: I tried that locally, but looks like there had been some changes from 3.0.6 to 3.0.10 which needs some more changes. I would prefer to have a different PR for that. Thoughts?

PR #874, opened 2021-07-08
Fixes produces and consumes attributes in API Routes

abdullah248:
## Description
Add produces and consumes attributes to the existing routes.

## Related issues
[AB#83343](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/83343)

## Testing

### DIFF
--- a/src/Microsoft.Health.Dicom.Api/Controllers/RetrieveController.cs
+++ b/src/Microsoft.Health.Dicom.Api/Controllers/RetrieveController.cs
@@ -37,6 +37,7 @@ namespace Microsoft.Health.Dicom.Api.Controllers
         private readonly IMediator _mediator;
         private readonly ILogger<RetrieveController> _logger;
         private const string IfNoneMatch = "If-None-Match";
+        private const string TypeSeperator = "=";
 
         public RetrieveController(IMediator mediator, ILogger<RetrieveController> logger)
         {
@@ -47,7 +48,7 @@ namespace Microsoft.Health.Dicom.Api.Controllers
             _logger = logger;
         }
 
-        //Todo: need to add produces for the multipart ones from here: https://github.com/microsoft/dicom-server/blob/main/docs/resources/conformance-statement.md#retrieve-instances-within-study-or-series
+        [Produces(KnownContentTypes.MultipartRelated + TypeSeperator + KnownContentTypes.ApplicationDicom)]
         [ProducesResponseType(typeof(IEnumerable<Stream>), (int)HttpStatusCode.OK)]
         [ProducesResponseType(typeof(string), (int)HttpStatusCode.BadRequest)]
         [ProducesResponseType(typeof(string), (int)HttpStatusCode.NotFound)]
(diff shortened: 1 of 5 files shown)


src/Microsoft.Health.Dicom.Api/Controllers/RetrieveController.cs
```diff
@@ -123,6 +128,7 @@ public async Task<IActionResult> GetSeriesMetadataAsync([FromHeader(Name = IfNon
             return CreateResult(response);
         }
 
+        [Produces(KnownContentTypes.ApplicationDicom, new[] { KnownContentTypes.MultipartRelated + TypeSeperator + KnownContentTypes.ApplicationDicom })]
```
wsugarman, line 131: Can this produce multipart responses?

src/Microsoft.Health.Dicom.Api/Controllers/RetrieveController.cs
```diff
@@ -47,6 +48,7 @@ public RetrieveController(IMediator mediator, ILogger<RetrieveController> logger
             _logger = logger;
         }
 
+        [Produces(KnownContentTypes.MultipartRelated + TypeSeperator + KnownContentTypes.ApplicationDicom)]
```
wsugarman, line 51: Should we just create a `KnownContentTypes` for this combination?

src/Microsoft.Health.Dicom.Api/Controllers/StoreController.cs
```diff
@@ -40,6 +41,8 @@ public StoreController(IMediator mediator, ILogger<StoreController> logger)
         }
 
         [AcceptContentFilter(new[] { KnownContentTypes.ApplicationDicomJson }, allowSingle: true, allowMultiple: false)]
+        [Produces(KnownContentTypes.ApplicationDicomJson)]
+        [Consumes(KnownContentTypes.ApplicationDicom, new[] { KnownContentTypes.MultipartRelated + TypeSeperator + KnownContentTypes.ApplicationDicom })]
```
wsugarman, line 45: nit: the second parameter is `params`, so the allocation of an array should be unnecessary (as it will happen behind the scenes).

Also, this `public ProducesAttribute (string contentType, params string[] additionalContentTypes)` ctor reminds me of lisp lol

wsugarman, line 45: Also, do we consume just JSON or do we also consume images too?

abdullah248: Fixed

abdullah248: https://github.com/microsoft/dicom-server/blob/main/docs/resources/conformance-statement.md#store-stow-rs
I think we only support consumption of application/dicom and multipart/realted

src/Microsoft.Health.Dicom.Api/Controllers/RetrieveController.cs
abdullah248: Just made them only return multipart

src/Microsoft.Health.Dicom.Api/Controllers/RetrieveController.cs
abdullah248: https://github.com/microsoft/dicom-server/blob/main/docs/resources/conformance-statement.md#retrieve-an-instance

From here it looks we do produce a multipart/related

wsugarman (approved):

PR #881, opened 2021-07-12
Support starting DICOM server with explicit user assigned MI

YazanMSFT:
## Description
Add a config to allow specifying which user assigned app ID the DICOM server should authenticate as

## Related issues
Addresses [issue #].

## Testing
Deployed to personal OSS deployment and verified:

- Server starts up when config is null/empty
- Server starts up when a user assigned App ID is specified in DicomServer:ServerIdentity:UserAssignedAppId (and the user is granted KeyVault access)
- Server fails to start up when an invalid user assigned App ID is specified


rbans96 (approved):

src/Microsoft.Health.Dicom.Web/Program.cs
```diff
@@ -23,9 +23,17 @@ public static void Main(string[] args)
...
+
+                    if (!string.IsNullOrEmpty(userAssignedAppId))
+                    {
+                        tokenProviderConnectionString = $"RunAs=App;AppId={userAssignedAppId}";
+                    }
+
                     if (!string.IsNullOrEmpty(keyVaultEndpoint))
                     {
-                        var azureServiceTokenProvider = new AzureServiceTokenProvider();
+                        var azureServiceTokenProvider = new AzureServiceTokenProvider(connectionString: tokenProviderConnectionString);
```
brandonpollett, line 36: This looks to only change the usage of the keyvault provider if it's being used. It doesn't seem like it would have an affect on the [SQL Provider](https://github.com/microsoft/healthcare-shared-components/blob/57e8f06dde0e27dfccbc9281e78665492d6c203f/src/Microsoft.Health.SqlServer/Registration/SqlServerBaseRegistrationExtensions.cs#L84) or the [Storage Provider](https://github.com/microsoft/healthcare-shared-components/blob/ab91e7a841b23b10ea5a611715528892d59fcdc1/src/Microsoft.Health.Blob/Features/Storage/BlobClientFactory.cs#L27)

YazanMSFT: Resolved by https://github.com/microsoft/healthcare-shared-components/pull/272

src/Microsoft.Health.Dicom.Web/appsettings.json
```diff
@@ -27,7 +27,10 @@
         },
         "Audit": {
             "CustomAuditHeaderPrefix": "X-MS-AZUREDICOM-AUDIT-"
-        }
+        },
+      "ServerIdentity": {
```
brandonpollett, line 31: nit: spacing.

brandonpollett (approved):

PR #896, opened 2021-07-19
Refactor Re-indexing Operation

wsugarman:
## Description
Update the implementation based on the previous design discussions. In particular:
- Gracefully handle race conditions when multiple clients attempt to add the same tag
- Allow tags to be deleted in the middle of re-indexing
- Leverage `ContinueAsNew` to reduce the storage requirements
- Add new SQL table for mapping an extended query tag to an ongoing operation
- Add new stored procedures:
  - Confirm which tags are being indexed by a given operation
  - Mark the tags' re-indexing as complete
- Adjust names
- Update unit tests for the durable function lifecycle

## Related issues
[AB#83601](https://microsofthealth.visualstudio.com/Health/_workitems/edit/83601)

## Testing
Unit tests and integration tests


src/Microsoft.Health.Dicom.Core/Features/Model/WatermarkRange.cs
```diff
@@ -4,63 +4,61 @@
...
+            => Start == other.Start && Count == other.Count;
 
         public void Deconstruct(out long start, out long end)
         {
             start = Start;
             end = End;
         }
+
+        public override string ToString()
+            => "[" + Start + ", " + (Start + Count) + ")";
```
wsugarman, line 62: nit: should be `End`

src/Microsoft.Health.Dicom.Core/Features/ExtendedQueryTag/AddExtendedQueryTagService.cs
```diff
@@ -47,17 +48,32 @@ public class AddExtendedQueryTagService : IAddExtendedQueryTagService
...
+                ready: false,
+                cancellationToken: cancellationToken);
+
+            // Start re-indexing
+            string operationId = await _client.StartQueryTagIndexingAsync(expected, cancellationToken);
+
+            // Associate the tags to the operation and confirm their processing
+            if ((await extendedQueryTagStore.ConfirmReindexingAsync(expected, operationId, cancellationToken)).Count == 0)
+            {
+                throw new ExtendedQueryTagsAlreadyExistsException();
```
wsugarman, line 75: Will add test

src/Microsoft.Health.Dicom.Core/Features/Model/WatermarkRange.cs
```diff
@@ -4,63 +4,61 @@
...
-        public long End { get; set; }
+        public long End { get; }
+
+        /// <summary>
+        /// Gets the maximum number of instances within this range.
+        /// </summary>
+        /// <remarks>
+        /// Some instances may be missing in the range due to previous deletion operations.
+        /// </remarks>
+        public long Count => End - Start;
```
wsugarman, line 38: I have to test the serialization for this

src/Microsoft.Health.Dicom.Functions/Indexing/Models/QueryTagLease.cs
```diff
@@ -0,0 +1,16 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+
+namespace Microsoft.Health.Dicom.Functions.Indexing.Models
+{
+    public class QueryTagLease
```
wsugarman, line 10: I can remove this class.

src/Microsoft.Health.Dicom.Functions/Indexing/Models/ReindexProgress.cs
```diff
@@ -0,0 +1,14 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Dicom.Functions.Indexing.Models
+{
+    internal class ReindexProgress
```
wsugarman, line 8: I can remove this

src/Microsoft.Health.Dicom.Functions/Microsoft.Health.Dicom.Functions.csproj
```diff
@@ -20,6 +20,7 @@
     <PackageReference Include="Microsoft.Health.SqlServer" Version="$(HealthcareSharedPackageVersion)" />
     <PackageReference Include="Microsoft.NET.Sdk.Functions" Version="3.0.11" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+    <PackageReference Include="System.Data.SqlClient" Version="4.8.2" PrivateAssets="All" />
```
wsugarman, line 23: I will re-assess the dependencies before check-in.

src/Microsoft.Health.Dicom.Functions.UnitTests/Indexing/ReindexDurableFunctionTests.Orchestration.cs
```diff
@@ -23,86 +22,395 @@ namespace Microsoft.Health.Dicom.Functions.UnitTests.Indexing
...
-            context.Received().StartNewOrchestration(nameof(ReindexDurableFunction.SubReindexTagsAsync),
-                Arg.Is<ReindexOperation>(
-                    x => x.WatermarkRange.Start == reindexOperation.WatermarkRange.Start
-                    && x.WatermarkRange.End == reindexOperation.WatermarkRange.End - _reindexConfig.BatchSize * _reindexConfig.MaxParallelBatches
-                    && x.StoreEntries == storeEntries));
+            context
+                .CallActivityAsync<long>(nameof(ReindexDurableFunction.GetMaxInstanceWatermarkAsync), input: null)
+                .Returns(Task.FromResult<long>(49));
+
+            context
```
wsugarman, line 55: I can coalesce these method calls

src/Microsoft.Health.Dicom.Functions.UnitTests/Microsoft.Health.Dicom.Functions.UnitTests.csproj
```diff
@@ -11,6 +11,8 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.10.0" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
```
wsugarman, line 14: I need to update these dependencies

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/SqlExtendedQueryTagStoreV4.cs
```diff
@@ -64,5 +64,72 @@ internal class SqlExtendedQueryTagStoreV4 : SqlExtendedQueryTagStoreV3
                 };
             }
         }
+
+        public override async Task<IReadOnlyCollection<ExtendedQueryTagStoreEntry>> ConfirmReindexingAsync(
```
wsugarman, line 68: I need tests for this and below

src/Microsoft.Health.Dicom.SqlServer/Features/Retrieve/SqlInstanceStoreV4.cs
```diff
@@ -52,15 +54,32 @@ public SqlInstanceStoreV4(SqlConnectionWrapperFactory sqlConnectionWrapperFactor
...
+        public override async Task<long> GetMaxInstanceWatermarkAsync(CancellationToken cancellationToken)
+        {
+            using SqlConnectionWrapper sqlConnectionWrapper = await SqlConnectionWrapperFactory.ObtainSqlConnectionWrapperAsync(cancellationToken);
+            using SqlCommandWrapper sqlCommandWrapper = sqlConnectionWrapper.CreateSqlCommand();
+
+            VLatest.GetMaxInstanceWatermark.PopulateCommand(sqlCommandWrapper);
+
+            try
+            {
+                return ((long?)await sqlCommandWrapper.ExecuteScalarAsync(cancellationToken)).GetValueOrDefault();
```
wsugarman, line 77: Need a test

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/4.diff.sql
```diff
@@ -23,7 +80,7 @@ GO
 -- RETURN VALUE
 --     The instance identifiers.
 ------------------------------------------------------------------------
-CREATE OR ALTER PROCEDURE dbo.GetInstancesByWatermarkRange(
+ALTER PROCEDURE dbo.GetInstancesByWatermarkRange(
```
wsugarman, line 83: Whoops! Will add back

src/Microsoft.Health.Dicom.Core/Features/ExtendedQueryTag/AddExtendedQueryTagService.cs
```diff
@@ -47,17 +48,32 @@ public class AddExtendedQueryTagService : IAddExtendedQueryTagService
...
             var normalized = extendedQueryTags
                 .Select(item => item.Normalize())
                 .ToList();
 
             // TODO: Handle tags that have already been added
+            // Add the extended query tags to the DB
             IExtendedQueryTagStore extendedQueryTagStore = await _extendedQueryTagStoreFactory.GetInstanceAsync(cancellationToken);
-            IReadOnlyList<int> keys = await extendedQueryTagStore.AddExtendedQueryTagsAsync(normalized, _maxAllowedCount, ready: false, cancellationToken: cancellationToken);
-            string operationId = await _client.StartQueryTagIndexingAsync(keys, cancellationToken);
+            IReadOnlyCollection<int> expected = await extendedQueryTagStore.AddExtendedQueryTagsAsync(
```
yuedai0531, line 63: would it be better to call this statuses?

wsugarman: These aren't status, like `ExtendedQueryTagStatus`. Instead these are the echoed tag keys for the new entries

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Orchestration.cs
```diff
@@ -21,105 +20,109 @@ namespace Microsoft.Health.Dicom.Functions.Indexing
...
-            ReindexOperation reindexOperation = context.GetInput<ReindexOperation>();
+            logger.LogInformation(
+                "Found {Count} extended query tag paths to re-index {{{TagPaths}}}.",
+                queryTags.Count,
+                string.Join(", ", queryTags.Select(x => x.Path)));
 
-            if (reindexOperation.WatermarkRange.Start >= 0
-                && reindexOperation.WatermarkRange.End >= 0)
+            List<int> queryTagKeys = queryTags.Select(x => x.Key).ToList();
+            if (queryTags.Count > 0)
```
pengchen0692, line 52: I'm confused on the logic, could we have a chat?

pengchen0692 (commented):
could we have a chat? I'm confused on the reindex logic.

src/Microsoft.Health.Dicom.Core/Features/ExtendedQueryTag/AddExtendedQueryTagService.cs
```diff
@@ -47,17 +48,32 @@ public class AddExtendedQueryTagService : IAddExtendedQueryTagService
...
+                normalized,
+                _maxAllowedCount,
+                ready: false,
+                cancellationToken: cancellationToken);
+
+            // Start re-indexing
+            string operationId = await _client.StartQueryTagIndexingAsync(expected, cancellationToken);
+
+            // Associate the tags to the operation and confirm their processing
+            if ((await extendedQueryTagStore.ConfirmReindexingAsync(expected, operationId, cancellationToken)).Count == 0)
```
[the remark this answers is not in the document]
wsugarman: On the client-side, this is used to:
- Ensure the state representation of each tag is correct before returning
- Confirm that some other client didn't override our tag mid-execution

src/Microsoft.Health.Dicom.Core/Features/Model/WatermarkRange.cs
```diff
@@ -4,63 +4,66 @@
...
         /// </summary>
-        public long Start { get; set; }
+        public long Start { get; }
 
         /// <summary>
-        /// Gets or sets inclusive end watermark.
+        /// Gets exclusive ending instance watermark.
         /// </summary>
-        public long End { get; set; }
+        public long End { get; }
```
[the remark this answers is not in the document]
wsugarman: Yup, I've updated the method!

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Activity.cs
```diff
@@ -20,84 +21,117 @@ namespace Microsoft.Health.Dicom.Functions.Indexing
...
+        /// </remarks>
+        /// <param name="context">The context for the activity.</param>
+        /// <param name="logger">A diagnostic logger.</param>
+        /// <returns>
+        /// A task representing the <see cref="GetMaxInstanceWatermarkAsync"/> operation.
+        /// The value of its <see cref="Task{TResult}.Result"/> property contains the maximum watermark value if found;
+        /// otherwise, <c>0</c>.
+        /// </returns>
+        [FunctionName(nameof(GetMaxInstanceWatermarkAsync))]
+        public Task<long> GetMaxInstanceWatermarkAsync(
```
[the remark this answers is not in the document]
wsugarman: That's a good idea, but I think it can go in another PR

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Trigger.cs
```diff
@@ -4,42 +4,60 @@
...
+            if (extendedQueryTags == null || extendedQueryTags.Count == 0)
+            {
+                return new HttpResponseMessage { StatusCode = HttpStatusCode.BadRequest };
+            }
+
+            logger.LogInformation("Received request to index {tagCount} extended query tag keys {{{extendedQueryTagKeys}}}.",
+                extendedQueryTags.Count,
+                string.Join(", ", extendedQueryTags));
+
+            string instanceId = await client.StartNewAsync(nameof(ReindexInstancesAsync), new ReindexInput { QueryTagKeys = extendedQueryTags });
```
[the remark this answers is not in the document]
wsugarman: Sure, I can add a constant! Conveniently, it is also the default value

src/Microsoft.Health.Dicom.Core/Features/Model/WatermarkRange.cs
```diff
@@ -4,63 +4,66 @@
...
         public void Deconstruct(out long start, out long end)
         {
             start = Start;
             end = End;
         }
+
+        public override string ToString()
+            => "[" + Start + ", " + End + ")";
+
+        public static WatermarkRange Between(long start, long end)
```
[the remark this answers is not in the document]
wsugarman: Unlike the ctor which takes in the count, this is a convenience method for specifying the start and end

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/4.diff.sql
```diff
@@ -83,5 +163,108 @@ AS
...
+CREATE OR ALTER PROCEDURE dbo.CompleteReindexing (
+    @extendedQueryTagKeys dbo.ExtendedQueryTagKeyTableType_1 READONLY
+)
+AS
+    SET NOCOUNT     ON
+    SET XACT_ABORT  ON
+
+    BEGIN TRANSACTION
+
+        DECLARE @updatedKeys AS dbo.ExtendedQueryTagKeyTableType_1
```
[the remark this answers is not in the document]
wsugarman: This cannot happen, as tags can only be assigned in two scenarios:
1. The client received an operation ID back from the functions and performed the assignment itself
2. The orchestration assigned all of the IDs it received as input

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/4.diff.sql
```diff
@@ -83,5 +163,108 @@ AS
...
+
+        -- Update the TagStatus of all rows to Completed (1)
+        UPDATE XQT
+        SET TagStatus = 1
+        OUTPUT INSERTED.TagKey
+        INTO @updatedKeys
+        FROM dbo.ExtendedQueryTag AS XQT
+        INNER JOIN @extendedQueryTagKeys input
+        ON XQT.TagKey = input.TagKey
+        INNER JOIN dbo.ExtendedQueryTagOperation AS XQTO
```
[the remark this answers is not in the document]
wsugarman: This is part of an `UPDATE` which should escalate exclusive row-level locks for the duration of the transaction

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/4.sql
```diff
@@ -399,6 +399,22 @@ CREATE UNIQUE NONCLUSTERED INDEX IX_ExtendedQueryTag_TagPath ON dbo.ExtendedQuer
...
+**************************************************************/
+CREATE TABLE dbo.ExtendedQueryTagOperation (
+    TagKey                  INT                  NOT NULL, --PK
+    OperationId             VARCHAR(36)          NOT NULL -- TODO: Use uniqueidentifier
+)
+
+CREATE UNIQUE CLUSTERED INDEX IXC_ExtendedQueryTagOperation ON dbo.ExtendedQueryTagOperation
+(
+    TagKey
+)
```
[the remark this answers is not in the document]
wsugarman: Sure!

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/4.sql
```diff
@@ -1502,16 +1546,36 @@ AS
...
+        -- Because the web client may fail between the addition of the tag and the invoking of re-indexing,
+        -- the stored procedure allows tags unassociated with an operation to be overwritten
+        DECLARE @existingTags TABLE(TagKey INT, TagStatus TINYINT, OperationId VARCHAR(36) NULL)
+
+        INSERT INTO @existingTags
+            (TagKey, TagStatus, OperationId)
+        SELECT dbo.ExtendedQueryTag.TagKey, TagStatus, OperationId
+        FROM dbo.ExtendedQueryTag
+        INNER JOIN @extendedQueryTags input ON input.TagPath = dbo.ExtendedQueryTag.TagPath
+        LEFT OUTER JOIN dbo.ExtendedQueryTagOperation ON dbo.ExtendedQueryTag.TagKey = dbo.ExtendedQueryTagOperation.TagKey
```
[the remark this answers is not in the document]
wsugarman: I think it is unnecessary given that an exclusive lock will be retrieved automatically:

>## Exclusive Locks
> Exclusive (X) locks prevent access to a resource by concurrent transactions. With an exclusive (X) lock, no other transactions can modify data; read operations can take place only with the use of the NOLOCK hint or read uncommitted isolation level.
>
> Data modification statements, such as INSERT, UPDATE, and DELETE combine both modification and read operations. The statement first performs read operations to acquire data before performing the required modification operations. Data modification statements, therefore, typically request both shared locks and exclusive locks. For example, an UPDATE statement might modify rows in one table based on a join with another table. In this case, the UPDATE statement requests shared locks on the rows read in the join table in addition to requesting exclusive locks on the updated rows.

[Docs](https://docs.microsoft.com/en-us/sql/relational-databases/sql-server-transaction-locking-and-row-versioning-guide?view=sql-server-ver15#exclusive)

Given that the [lock escalates](https://docs.microsoft.com/en-us/sql/relational-databases/sql-server-transaction-locking-and-row-versioning-guide?view=sql-server-ver15#lock-escalation) to the transaction, I would suspect that my shared lock (and in turn exclusive lock) would persist for the entire transaction. My reading is that we could lose the lock if we didn't have a transaction here, but I can confirm with Smitha next week.

> Each escalation event operates primarily at the level of a single Transact-SQL statement. When the event starts, the Database Engine attempts to escalate all the locks owned by the current transaction in any of the tables that have been referenced by the active statement provided it meets the escalation threshold requirements. If the escalation event starts before the statement has accessed a table, no attempt is made to escalate the locks on that table. If lock escalation succeeds, any locks acquired by the transaction in a previous statement and still held at the time the event starts will be escalated if the table is referenced by the current statement and is included in the escalation event.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/4.sql
```diff
@@ -1587,6 +1691,102 @@ AS
...
+
+        DECLARE @updatedKeys AS dbo.ExtendedQueryTagKeyTableType_1
+
+        -- Update the TagStatus of all rows to Completed (1)
+        UPDATE XQT
+        SET TagStatus = 1
+        OUTPUT INSERTED.TagKey
+        INTO @updatedKeys
+        FROM dbo.ExtendedQueryTag AS XQT
+        INNER JOIN @extendedQueryTagKeys AS input ON XQT.TagKey = input.TagKey
```
pengchen0692, line 1777: do we want to add lock here so that `TagStatus` cannot be changed by other transactions during this one?
Imaging TagStatus is changed or the tag is removed before the second statement (DELETE), would DELETE still be valid ?  Although I cannot image a way this could happen, but I think make entire transaction really atomic would be a good idea.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/4.diff.sql
```diff
@@ -40,12 +112,42 @@ AS
...
+/***************************************************************************************/
+CREATE OR ALTER PROCEDURE dbo.GetExtendedQueryTagsByOperation (
+    @operationId VARCHAR(32)
+)
+AS
+BEGIN
+    SET NOCOUNT     ON
+    SET XACT_ABORT  ON
+
+    SELECT XQT.*
```
yuedai0531, line 137: XQT 😊
nice nice

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/4.sql
```diff
@@ -1587,6 +1691,102 @@ AS
...
+        INNER JOIN dbo.ExtendedQueryTagOperation AS XQTO ON input.TagKey = XQTO.TagKey
```
pengchen0692, line 1778: is it possible that `@updatedKeys` is  not same as `@extendedQueryTagKeys`?

wsugarman: Not today, as delete will be synchronous and there is no pausing. But perhaps in the future, the status may have changed to something like `Removing` and it wouldn't be correct to then complete. Put another way, this is like we're confirming the state of the tags before changing their state.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/4.sql
```diff
@@ -1502,16 +1586,36 @@ AS
...
+        -- Because the web client may fail between the addition of the tag and the starting of re-indexing operation,
+        -- the stored procedure allows tags that are not assigned to an operation to be overwritten
+        DECLARE @existingTags TABLE(TagKey INT, TagStatus TINYINT, OperationId VARCHAR(32) NULL)
+
+        INSERT INTO @existingTags
+            (TagKey, TagStatus, OperationId)
+        SELECT XQT.TagKey, TagStatus, OperationId
+        FROM dbo.ExtendedQueryTag AS XQT
+        INNER JOIN @extendedQueryTags AS input ON input.TagPath = XQT.TagPath
+        LEFT OUTER JOIN dbo.ExtendedQueryTagOperation AS XQTO ON XQT.TagKey = XQTO.TagKey
```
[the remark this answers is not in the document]
wsugarman: `INSERT` will obtain the lock

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Orchestration.cs
```diff
@@ -21,105 +20,109 @@ namespace Microsoft.Health.Dicom.Functions.Indexing
...
-            ILogger logger)
-        {
-            EnsureArg.IsNotNull(context, nameof(context));
             logger = context.CreateReplaySafeLogger(logger);
+            ReindexInput input = context.GetInput<ReindexInput>();
+
+            // Determine the set of query tags that should be indexed and only continue if there is at least 1.
+            // For the first time this orchestration executes, assign all of the tags in the input to the operation,
+            // otherwise simply fetch the tags from the database for this operation.
+            IReadOnlyList<ExtendedQueryTagStoreEntry> queryTags = input.Completed.Count == 0
```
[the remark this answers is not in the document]
wsugarman: I kind of like being consistent with `Count` because if for some reason someone passed the initial progress like [12, 12) that wouldn't contain any data, but it would have `Count == 0`

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/4.diff.sql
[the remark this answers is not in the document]
pengchen0692: in this way, can we get rid of @updatedKeys?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/4.diff.sql
```diff
@@ -183,139 +313,240 @@ AS
...
+        DECLARE @updatedKeys AS dbo.ExtendedQueryTagKeyTableType_1
+
+        -- Update the TagStatus of all rows to Completed (1)
+        UPDATE XQT
+        SET TagStatus = 1
+        OUTPUT INSERTED.TagKey
+        INTO @updatedKeys
+        FROM dbo.ExtendedQueryTag AS XQT
+        INNER JOIN @extendedQueryTagKeys AS input ON XQT.TagKey = input.TagKey
+        INNER JOIN dbo.ExtendedQueryTagOperation AS XQTO ON input.TagKey = XQTO.TagKey
```
pengchen0692, line 539: do we need to join with `dbo.ExtendedQueryTagOperation`?

pengchen0692 (approved):

PR #898, opened 2021-07-21
Refactor Versioned SQL APIs

wsugarman:
> **Reviewers**
>
> Many test files had to change in this pull request.
> Feel free to browse those, but the most important logic is in the SQL Server project

## Description
Instead of relying explicitly on `StoreFactory<T>` to resolve the correct value in the middle of the request handler, we can hide this factory by creating thin wrappers around each of the respective interfaces, like `IInstanceStore`. In this PR, I made the following changes:
- Replace `StoreFactory<T>` with `VersionedCache<T>` which resolve the proper underlying store once before caching it
- Create new interface implementations for our SQL-based data stores that pass calls down to the proper implementation based on the current version
- Refactor the usage of `StoreFactory<T>`
- Register the cache as `Scoped` so that the same version is used throughout a single request

## Related issues
N/A

## Testing
Ran tests locally


src/Microsoft.Health.Dicom.SqlServer/Features/Schema/PassthroughSchemaVersionResolver.cs
```diff
@@ -14,18 +14,19 @@ namespace Microsoft.Health.Dicom.SqlServer.Features.Schema
...
-        /// Initializes a new instance of the <see cref="BackgroundSchemaVersionResolver"/> class.
+        /// Initializes a new instance of the <see cref="PassthroughSchemaVersionResolver"/> class.
         /// </summary>
         /// <param name="schemaInformation">The information updated in the background.</param>
         /// <exception cref="ArgumentNullException">
         /// <paramref name="schemaInformation"/> is <see langword="null"/>.
         /// </exception>
-        public BackgroundSchemaVersionResolver(SchemaInformation schemaInformation)
+        public PassthroughSchemaVersionResolver
+            (SchemaInformation schemaInformation)
```
wsugarman, line 29: Whoops, used a newline here

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/VersionedCache.cs
```diff
@@ -0,0 +1,69 @@
...
+using Microsoft.Health.Dicom.SqlServer.Exceptions;
+
+namespace Microsoft.Health.Dicom.SqlServer.Features.Schema
+{
+    internal sealed class VersionedCache<T> : IDisposable where T : class, IVersioned
+    {
+        private readonly ISchemaVersionResolver _schemaVersionResolver;
+        private readonly Dictionary<SchemaVersion, T> _entities;
+        private readonly SemaphoreSlim _semaphore = new SemaphoreSlim(1, 1);
+        private volatile T _cachedEntity;
```
wsugarman, line 22: I have to read up more on the use `volatile` for the double-checked lock pattern. Outside of the memory barrier, I am unsure if it's needed

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/VersionedCache.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Globalization;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using EnsureThat;
+using Microsoft.Health.Dicom.SqlServer.Exceptions;
+
+namespace Microsoft.Health.Dicom.SqlServer.Features.Schema
+{
+    internal sealed class VersionedCache<T> : IDisposable where T : class, IVersioned
```
pengchen0692, line 17: * Do we want to cache the version? We don't need to handle the schema upgrade scenario now, but when we need to, we couldn't cache it.
* Could we use `Lazy<T>` to simplify the logic to handle with cached entity?

wsugarman: I would really like to, so I don't have to think about the locking semantics, but we also need the ability to:
1. Invoke the factory "asynchronously"
2. Cancel the factory invocation.

There are some examples of doing it asynchronously, but cancellation poses a problem because of how you have to capture the cancellation token. So I've kept my double-checked locking with some changes based on the source code for [Lazy\<T\>](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Lazy.cs).

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/VersionedCache.cs
wsugarman: It appears it is indeed needed, as `Lazy<T>` itself leverages a volatile write for this purpose

pengchen0692 (approved):

PR #905, opened 2021-07-23
OpenApi Diff for breaking changes

abdullah248:
## Description
Add OpenApi diff for versions. In ci and pr pipeline generates the OpenApi doc based on the latest build and then compares it to the baseline that has been checked in previously. Fails the pipeline if there are breaking changes.

## Related issues
[AB#83069](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/83069)

## Testing
Ran in the pipeline, also introduced a breaking change to ensure that the pipeline breaks if there is a breaking change.

### DIFF
--- a/build/versioning.yml
+++ b/build/versioning.yml
@@ -1,5 +1,5 @@
 jobs:
-- job: SetupAndRun
+- job: OpenApiDiff
   pool:
     vmImage: 'ubuntu-latest'
   steps:
@@ -10,11 +10,16 @@ jobs:
       downloadPath: '$(System.ArtifactsDirectory)'
       artifactName: 'deploy'
 
+  - task: ExtractFiles@1
+    inputs:
+      archiveFilePatterns: '$(System.ArtifactsDirectory)/deploy/Microsoft.Health.Dicom.Web.zip'
+      destinationFolder: '$(System.ArtifactsDirectory)/deploy/webArtifacts'
+
   - task: PowerShell@2
     displayName: 'Detect API Breaking Changes'
     inputs:
       filepath: '$(System.DefaultWorkingDirectory)/samples/scripts/PowerShell/GenerateOpenApiYaml.ps1'
       arguments: >
-        -WorkingDir '$(System.WorkingDir)'
-        -AssemblyDir '$(build.artifactStagingDirectory)/web/net5.0/Microsoft.Health.Dicom.Web.dll'
+        -SwaggerDir '$(System.DefaultWorkingDirectory)/swagger/'
+        -AssemblyDir '$(System.ArtifactsDirectory)/deploy/webArtifacts/Microsoft.Health.Dicom.Web.dll'
         -Version 'v1-prerelease'
--- a/build/versioning.yml
+++ b/build/versioning.yml
@@ -16,5 +16,5 @@ jobs:
       filepath: '$(System.DefaultWorkingDirectory)/samples/scripts/PowerShell/GenerateOpenApiYaml.ps1'
       arguments: >
         -WorkingDir '$(System.WorkingDir)'
-        -AssemblyDir '$(System.ArtifactsDirectory)/src/Microsoft.Health.Dicom.Web/bin/$(buildConfiguration)/net5.0/Microsoft.Health.Dicom.Web.dll'
+        -AssemblyDir '$(build.artifactStagingDirectory)/web/net5.0/Microsoft.Health.Dicom.Web.dll'
         -Version 'v1-prerelease'
--- a/build/versioning.yml
+++ b/build/versioning.yml
@@ -1,6 +1,6 @@
 steps:
 - task: PowerShell@2
-  displayName: "generate v1-prerelease swagger"
+  displayName: "Validate Api has no breaking changes"
   inputs:
     filepath: '$(System.DefaultWorkingDirectory)/samples/scripts/PowerShell/GenerateOpenApiYaml.ps1'
     arguments: >
--- a/build/versioning.yml
+++ b/build/versioning.yml
@@ -3,6 +3,12 @@ jobs:
   pool:
     vmImage: 'ubuntu-latest'
   steps:
+
+  - task: UseDotNet@2
+    displayName: 'Use .NET Core sdk'
+    inputs:
+      useGlobalJson: true
+
   - task: DownloadBuildArtifacts@0
     inputs:
       buildType: 'current'
@@ -11,6 +17,7 @@ jobs:
       artifactName: 'deploy'
 
   - task: ExtractFiles@1
+    displayName: 'Extract Web zip'
     inputs:
       archiveFilePatterns: '$(System.ArtifactsDirectory)/deploy/Microsoft.Health.Dicom.Web.zip'
       destinationFolder: '$(System.ArtifactsDirectory)/deploy/webArtifacts'
@@ -20,6 +27,6 @@ jobs:
     inputs:
       filepath: '$(System.DefaultWorkingDirectory)/samples/scripts/PowerShell/GenerateOpenApiYaml.ps1'
       arguments: >
-        -SwaggerDir '$(System.DefaultWorkingDirectory)/swagger/'
+        -SwaggerDir '$(System.DefaultWorkingDirectory)/swagger'
         -AssemblyDir '$(System.ArtifactsDirectory)/deploy/webArtifacts/Microsoft.Health.Dicom.Web.dll'
         -Version 'v1-prerelease'
--- a/build/versioning.yml
+++ b/build/versioning.yml
@@ -1,4 +1,8 @@
-steps:
+jobs:
+- job: SetupAndRun
+  pool:
+    vmImage: 'ubuntu-latest'
+  steps:
   - task: DownloadBuildArtifacts@0
     inputs:
       buildType: 'current'
--- a/build/versioning.yml
+++ b/build/versioning.yml
@@ -1,10 +1,16 @@
 steps:
-- task: PowerShell@2
-  displayName: "Validate Api has no breaking changes"
-  inputs:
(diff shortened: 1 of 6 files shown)


samples/scripts/PowerShell/GenerateOpenApiYaml.ps1
```diff
@@ -0,0 +1,33 @@
...
+)
+foreach($Version in $Versions){
+    write-host "Generating Yaml file for $Version"
+
+    dotnet new tool-manifest --force
+    dotnet tool install --version 6.1.4 Swashbuckle.AspNetCore.Cli
+    dotnet swagger tofile --yaml --output $WorkingDir/$OutputPathOfOpenApiDoc/$Version.yaml $WorkingDir/$OutputPathofDll $Version
+
+    write-host "Running comparison with baseline for version $Version"
+    docker run --rm -t -v ${WorkingDir}:/dicom-server openapitools/openapi-diff:latest /dicom-server/swagger/$Version/swagger.yaml /dicom-server/$OutputPathOfOpenApiDoc/$version.yaml --fail-on-incompatible
```
jnlycklama, line 31: So for versions that are still in development / unreleased, /dicom-server/swagger/$Version/swagger.yaml can be changed whenever. As soon as the version gets released, that file shouldn't be changed at all so this comparison diff will work. Is that right?

jnlycklama: If so I wonder if we can lock the file somehow after release, so it isn't accidentally updated

abdullah248: Yes that is correct. I am not sure if there is anyway to lock it, but I was thinking that is something that the people reviewing the pr should notice? Or do you have any ideas on some kind of automatic check

wsugarman: I think we had chatted about this, and I think there are probably two ways to perform this check:
1. Generate each version's OpenAPI YAML specification dynamically from the assembly and compare that against a versioned control. If it's at all changed, I'd say we fail the build because it is out-of-sync, and instead force the developer to re-generate it through some script such that the change is reviewed in PR.
2. We never generate the YAML manually. Instead, we always auto-generate it as part of the build and we compare the auto-generated files from the current build to the auto-generated files from the previous build. We can fail the build if there is a breaking change, while using monikers like `prerelease` to determine when they're allowed.

wsugarman, line 31: Should we pin to a version of the docker image that isn't latest, or to a specific commit? We don't want to necessarily pull in bad changes into our pipeline.

samples/scripts/PowerShell/GenerateOpenApiYaml.ps1
```diff
@@ -0,0 +1,33 @@
...
+.PARAMETER OutputPathofDll
+Path for the web projects dll
+.PARAMETER OutputPathOfOpenApiDoc
+Path to output the OpenApi Doc (in yaml)
+.PARAMETER Versions
+Api versions to generate the OpenApiDoc for and compare with baseline
+#>
+
+param(
+    [string]$WorkingDir,
```
wsugarman, line 15: Do we need `$WorkingDir` or can the `$OutputPathofDll` always be absolute?

samples/scripts/PowerShell/GenerateOpenApiYaml.ps1
```diff
@@ -0,0 +1,33 @@
...
+
+    [string]$OutputPathofDll,
+
+    [string]$OutputPathOfOpenApiDoc,
+
+    [String[]] $Versions
+)
+foreach($Version in $Versions){
```
wsugarman, line 23: Can these versions be auto-detected?

wsugarman (changes requested):

samples/scripts/PowerShell/GenerateOpenApiYaml.ps1
abdullah248: It doesnt look like they can be with the CLI tool. It requires that a specific version is passed in.

samples/scripts/PowerShell/GenerateOpenApiYaml.ps1
```diff
@@ -0,0 +1,33 @@
...
+.PARAMETER Versions
+Api versions to generate the OpenApiDoc for and compare with baseline
+#>
+
+param(
+    [string]$WorkingDir,
+
+    [string]$OutputPathofDll,
+
+    [string]$OutputPathOfOpenApiDoc,
```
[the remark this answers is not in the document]
abdullah248: Removed it

samples/scripts/PowerShell/GenerateOpenApiYaml.ps1
abdullah248: Changed to be swagger path to use to get swagger and output files

samples/scripts/PowerShell/GenerateOpenApiYaml.ps1
```diff
@@ -0,0 +1,33 @@
...
+
+    [String[]] $Versions
+)
+foreach($Version in $Versions){
+    write-host "Generating Yaml file for $Version"
+
+    dotnet new tool-manifest --force
+    dotnet tool install --version 6.1.4 Swashbuckle.AspNetCore.Cli
```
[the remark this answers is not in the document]
abdullah248: This one is for the local install

samples/scripts/PowerShell/GenerateOpenApiYaml.ps1
abdullah248: As discussed, we are following approach 1 currently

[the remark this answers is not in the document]
abdullah248: Changed to be its own stage in linux so it is guaranteed to have docker

abdullah248: Pinned to a sha

samples/scripts/PowerShell/GenerateOpenApiYaml.ps1
```diff
@@ -0,0 +1,34 @@
...
+
+    [String[]]$Versions,
+
+    [string]$SwashbuckleCLIVersion = '6.1.4'
+)
+foreach ($Version in $Versions)
+{
+    write-host "Generating Yaml file for $Version"
+
+    dotnet new tool-manifest --force
```
wsugarman, line 27: Are we guaranteed to have `dotnet` available? Otherwise we should install it with a task

samples/scripts/PowerShell/GenerateOpenApiYaml.ps1
```diff
@@ -0,0 +1,34 @@
...
+    dotnet tool install --version $SwashbuckleCLIVersion Swashbuckle.AspNetCore.Cli
+    dotnet swagger tofile --yaml --output "$SwaggerDir/$Version.yaml" "$AssemblyDir" $Version
```
wsugarman, line 29: nit: I think there's an extra slash here, at least based on the pipeline stdout. If you want to, you can use `Join-Path` to properly join these together

samples/scripts/PowerShell/GenerateOpenApiYaml.ps1
```diff
@@ -0,0 +1,34 @@
...
+    [String[]]$Versions,
+
+    [string]$SwashbuckleCLIVersion = '6.1.4'
+)
+foreach ($Version in $Versions)
+{
+    write-host "Generating Yaml file for $Version"
+
+    dotnet new tool-manifest --force
+    dotnet tool install --version $SwashbuckleCLIVersion Swashbuckle.AspNetCore.Cli
```
wsugarman, line 28: We should probably only install the tool once and not for every version, right?

samples/scripts/PowerShell/GenerateOpenApiYaml.ps1
```diff
@@ -0,0 +1,34 @@
...
+    dotnet swagger tofile --yaml --output "$SwaggerDir/$Version.yaml" "$AssemblyDir" $Version
+
+    write-host "Running comparison with baseline for version $Version"
+    docker run --rm -t -v ${SwaggerDir}:/swagger openapitools/openapi-diff:latest@sha256:5da8291d3947414491e4c62de74f8fc1ee573a88461fb2fb09979ecb5ea5eb02 "/swagger/$Version/swagger.yaml" "/swagger/$version.yaml" --fail-on-incompatible
```
wsugarman, line 32: Should we build the docker image once before all of the versions run? That was the stdout is more organized?

wsugarman (approved):

build/versioning.yml
```diff
@@ -0,0 +1,25 @@
...
+  steps:
+  - task: DownloadBuildArtifacts@0
+    inputs:
+      buildType: 'current'
+      downloadType: 'single'
+      downloadPath: '$(System.ArtifactsDirectory)'
+      artifactName: 'deploy'
+
+  - task: ExtractFiles@1
+    inputs:
```
wsugarman, line 14: nit: add display name

samples/scripts/PowerShell/GenerateOpenApiYaml.ps1
abdullah248: Used create to have them docker image pulled before each version run

samples/scripts/PowerShell/GenerateOpenApiYaml.ps1
abdullah248: Added the task

samples/scripts/PowerShell/GenerateOpenApiYaml.ps1
abdullah248: Added the task

PR #908, opened 2021-07-25
Refactor V3 and V4 to be Transactional

wsugarman:
## Description
- Update v3 and v4 SQL scripts to be within a transaction

## Related issues
[AB#82137](https://microsofthealth.visualstudio.com/Health/_workitems/edit/82137/)

## Testing
Test locally and in unit tests

### DIFF
--- a/src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/3.sql
+++ b/src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/3.sql
@@ -1,344 +1,538 @@
 -- Style guide: please see: https://github.com/ktaranov/sqlserver-kit/blob/master/SQL%20Server%20Name%20Convention%20and%20T-SQL%20Programming%20Style.md
+
+/*************************************************************
+Full text catalog creation
+**************************************************************/
+IF NOT EXISTS (
+    SELECT * 
+    FROM sys.fulltext_catalogs
+    WHERE name = 'Dicom_Catalog')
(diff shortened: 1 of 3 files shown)


pengchen0692 (commented):
looks like there are different way to make it transational: 
https://github.com/microsoft/dicom-server/blob/main/src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/3.sql and  
https://github.com/microsoft/dicom-server/blob/main/src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/2.sql, I'm not sure which one we should pickup

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/3.sql
```diff
@@ -1,344 +1,538 @@
 -- Style guide: please see: https://github.com/ktaranov/sqlserver-kit/blob/master/SQL%20Server%20Name%20Convention%20and%20T-SQL%20Programming%20Style.md
+
+/*************************************************************
+Full text catalog creation
+**************************************************************/
+IF NOT EXISTS (
```
smithago, line 6: These changes are not needed. It is already re-runnable and more readable and cleaner. We should follow similar model in 4.sql

wsugarman: Ah, my mistaken! I'll abandoned this one then

PR #909, opened 2021-07-27
Use Guid for Operation ID

wsugarman:
## Description
- Enforce `Guid` for the Operation ID (and the Instance ID used within Azure Functions)
    - Even though the platform generates Instance IDs using `Guid.NewGuid.ToString("N")`, we now explicitly create these values ourselves and pass it to `StartNewAsync`
- Ensure `Guid` for Operation ID is serialized as a string using the `"N"` format specifier consistently

## Related issues
[AB#83611](https://microsofthealth.visualstudio.com/Health/_workitems/edit/83611)

## Testing
Test locally and via build pipelines


src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/4.sql
```diff
@@ -407,7 +407,7 @@ CREATE UNIQUE NONCLUSTERED INDEX IX_ExtendedQueryTag_TagPath ON dbo.ExtendedQuer
 **************************************************************/
 CREATE TABLE dbo.ExtendedQueryTagOperation (
     TagKey                  INT                  NOT NULL, --PK
-    OperationId             VARCHAR(32)          NOT NULL -- TODO: Use uniqueidentifier
+    OperationId             uniqueidentifier     NOT NULL
```
yuedai0531, line 410: Should we also change {study-series-sop}instanceUids to be uniqueidentifier instead of varchar?

wsugarman: While these are "Unique Identifiers" (UI) according to the DICOM spec, their representation cannot be reconciled with the .NET `Guid` type. While the DICOM UI uses at most 64 bytes, the .NET GUID (or UUID) uses 16 bytes (128 bits).

yuedai0531 (approved):

PR #911, opened 2021-07-27
Validate tagValue when Reindex

pengchen0692:
## Description
Validate tagValue when Reindex

## Related issues
Addresses [AB#83811](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/83811)

## Testing
All automated tests pass.


src/Microsoft.Health.Dicom.Core/DicomCoreResource.resx
```diff
@@ -131,7 +131,7 @@
     <value>The data store operation failed.</value>
   </data>
   <data name="DicomElementValidationFailed" xml:space="preserve">
-    <value>Dicom element '{0}' with value '{1}' does not validate VR '{2}': {3}.</value>
+    <value>Dicom element '{0}' does not validate VR '{1}': {2}.</value>
```
yuedai0531, line 134: should we say something along the lines of given VR does not match expected VR type?

src/Microsoft.Health.Dicom.Core/Features/Validation/ValidationErrorCode.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// </summary>
+        PatientNameGroupIsTooLong = 1001,
+
+        /// <summary>
+        ///  Group of patient name contains invalid characters.
+        /// </summary>
+        PatientNameGroupContainsInvalidCharacters = 1002,
+
+        /// <summary>
+        /// Patient name has too many components
```
yuedai0531, line 56: super super nit:
```suggestion
        /// Patient name has too many components.
```

yuedai0531 (approved):
LGTM 👍

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMinimumValidator.cs
```diff
@@ -3,67 +3,53 @@
...
-            else if (_byteBufferValidators.TryGetValue(dicomVR, out Action<IByteBuffer, string> byteBufferValidator))
+            if (Validations.TryGetValue(vr, out IDicomElementValidation validationRule))
             {
-                byteBufferValidator(element.Buffer, GetName(element.Tag));
+                validationRule.Validate(dicomElement);
             }
             else
             {
-                Debug.Fail($"Missing validation action for for VR :{dicomVR.Code}, add a new validation and register in the constructor.");
+                Debug.Fail($"Validating VR {vr?.Code} is not supported.");
```
wsugarman, line 50: I see we are using a lot of `Debug.Fail` in these APIs. Are these pre-conditions? Otherwise, should we be throwing errors instead?

src/Microsoft.Health.Dicom.Core/Features/Validation/IDicomElementMinimumValidator.cs
```diff
@@ -9,6 +9,6 @@ namespace Microsoft.Health.Dicom.Core.Features.Validation
 {
     public interface IDicomElementMinimumValidator
     {
-        void Validate(DicomElement element);
+        void Validate(DicomElement dicomElement);
```
wsugarman, line 12: We should probably add some XML doc here that indicates it throws `DicomElementValidationException`

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementValidation.cs
```diff
@@ -0,0 +1,41 @@
...
+                }
+            }
+            else
+            {
+                Debug.Fail($"Validating VR {vr.Code} is not supported.");
+            }
+        }
+
+        protected static bool IsControlExceptESC(char c)
+        => char.IsControl(c) && (c != '\u001b');
```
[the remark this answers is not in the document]
wsugarman: Furthermore, instead of defining this as a predicate, it seems like it is always used to check for characters in a `string`. Should this not simply be defined like:

```csharp
protected static bool ContainsControlExceptESC(string s)
```

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomUidValidation.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Text.RegularExpressions;
+using Dicom;
+using Microsoft.Health.Dicom.Core.Exceptions;
+using Microsoft.Health.Dicom.Core.Extensions;
+
+namespace Microsoft.Health.Dicom.Core.Features.Validation
+{
+    internal class DicomUidValidation : DicomElementValidation
+    {
+        private static readonly Regex ValidIdentifierCharactersFormat = new Regex("^[0-9\\.]*$", RegexOptions.Compiled);
```
wsugarman, line 15: We can end with a period?

src/Microsoft.Health.Dicom.Core/Features/Validation/IDicomElementValidation.cs
```diff
@@ -3,21 +3,12 @@
...
-        {
-            if (string.IsNullOrWhiteSpace(value))
-            {
-                throw new InvalidIdentifierException(value, name);
-            }
-
-            // validate the value
-            DicomElementMinimumValidation.ValidateUI(value, name);
-        }
+        void Validate(DicomElement dicomElement);
```
wsugarman, line 12: Let's add some XML doc

src/Microsoft.Health.Dicom.Core/Features/Validation/ValidationErrorCode.cs
```diff
@@ -0,0 +1,65 @@
...
+    /// Validation Error Code.
+    /// Error Code is  4 letter number. 
+    /// First 2 letters indicate VR:
+    ///  00 - General error for all VR
+    ///  01 - Error for PN
+    /// Last 2 letters indicate specific errors for VR.
+    /// </summary>
+    public enum ValidationErrorCode
+    {
+        None = 0,
```
wsugarman, line 18: nit: missing XML doc for none

src/Microsoft.Health.Dicom.Core/Features/Validation/ValidationErrorCode.cs
```diff
@@ -0,0 +1,65 @@
...
+
+namespace Microsoft.Health.Dicom.Core.Features.Validation
+{
+    /// <summary>
+    /// Validation Error Code.
+    /// Error Code is  4 letter number. 
+    /// First 2 letters indicate VR:
+    ///  00 - General error for all VR
+    ///  01 - Error for PN
+    /// Last 2 letters indicate specific errors for VR.
```
wsugarman, line 14: I think it makes sense to try to group error codes together by VR, but I think if we want to do this we need to rephrase it. Ultimately, the values we operate upon will be integral numbers and not strings. Therefore, there won't be a concept of leading 0s. I think the more correct structure, if we so choose to do so, would be to organize them into ranges. Eg [0, 100)  are general errors, [100, 101) are errors for VR type 1, etc. (Or however large we want the ranges to be)

src/Microsoft.Health.Dicom.Core/DicomCoreResource.resx
pengchen0692: As my understanding, "does not match expected VR type" means customer provided wrong PR. And in our case, the validation could failure a bunch of different reason, e.g: provided value is too long.

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomUidValidation.cs
pengchen0692: I copied from https://github.com/microsoft/dicom-server/blob/44acba0557256d7fd3a23894b8c7e0267dbed689/src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMinimumValidation.cs#L18.
But you are right. `The UID is a series of numeric components separated by the period "." character.` (http://dicom.nema.org/dicom/2013/output/chtml/part05/sect_6.2.html). I'll fix it.

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMaxLengthValidation.cs
```diff
@@ -0,0 +1,46 @@
...
+using Microsoft.Health.Dicom.Core.Exceptions;
+using Microsoft.Health.Dicom.Core.Extensions;
+
+namespace Microsoft.Health.Dicom.Core.Features.Validation
+{
+    internal class DicomElementMaxLengthValidation : DicomElementValidation
+    {
+        public DicomElementMaxLengthValidation(int maxLength)
+        {
+            MaxLength = maxLength;
```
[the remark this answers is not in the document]
pengchen0692: Added a Debug.Assert.

src/Microsoft.Health.Dicom.Core/Exceptions/DicomElementValidationException.cs
```diff
@@ -5,30 +5,35 @@
...
+            Value = value;
         }
 
-        public override string Message => string.Format(DicomCoreResource.DicomElementValidationFailed, Name, Value, VR.Code, base.Message);
+        public ValidationErrorCode ErrorCode { get; }
 
-        public string Name { get; }
+        public override string Message => Value == null ?
+                string.Format(DicomCoreResource.DicomElementValidationFailed, Name, VR.Code, base.Message)
+                : string.Format(DicomCoreResource.DicomElementValidationFailedWithValue, Name, Value, VR.Code, base.Message);
```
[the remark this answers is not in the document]
pengchen0692: For every DicomElementValidationException, we provide meaningful message.

src/Microsoft.Health.Dicom.Core/Exceptions/DicomElementValidationException.cs
```diff
@@ -5,30 +5,34 @@
...
 using Dicom;
 using EnsureThat;
+using Microsoft.Health.Dicom.Core.Features.Validation;
 
 namespace Microsoft.Health.Dicom.Core.Exceptions
 {
     public class DicomElementValidationException : ValidationException
     {
-        public DicomElementValidationException(string name, string value, DicomVR vr, string message)
+        public DicomElementValidationException(ValidationErrorCode errorCode, string name, DicomVR vr, string message, string value = null)
```
smithago, line 14: can we have overloaded constructor, instead of figuring our which param results in which message?

src/Microsoft.Health.Dicom.Core/Exceptions/DicomElementValidationException.cs
```diff
@@ -5,30 +5,39 @@
...
 {
     public class DicomElementValidationException : ValidationException
     {
-        public DicomElementValidationException(string name, string value, DicomVR vr, string message)
+        public DicomElementValidationException(ValidationErrorCode errorCode, string name, DicomVR vr, string message)
+           : this(errorCode, name, vr, message, null)
+        {
+        }
+
+        public DicomElementValidationException(ValidationErrorCode errorCode, string name, DicomVR vr, string message, string value)
```
smithago, line 19: Is it fixed? From the caller, there is till this constructor that takes some params that can be null.

Why don't you make each constructor set message explictly instead

src/Microsoft.Health.Dicom.Core/Features/Delete/DeleteHandler.cs
```diff
@@ -61,16 +61,16 @@ public async Task<DeleteResourcesResponse> Handle(DeleteResourcesRequest request
 
         private static void ValidateDeleteResourcesRequest(DeleteResourcesRequest request)
         {
-            UidValidator.Validate(request.StudyInstanceUid, nameof(request.StudyInstanceUid));
+            DicomUidValidation.Validate(request.StudyInstanceUid, nameof(request.StudyInstanceUid));
```
smithago, line 64: We had removed the Dicom prefix before. Do we need it?

src/Microsoft.Health.Dicom.Core/Features/ExtendedQueryTag/ExtendedQueryTagEntryValidator.cs
```diff
@@ -171,7 +171,7 @@ private static void ValidatePrivateCreator(DicomTag tag, string privateCreator,
 
             try
             {
-                DicomElementMinimumValidation.ValidateLO(privateCreator, nameof(privateCreator));
+                DicomLongStringValidation.Validate(privateCreator, nameof(privateCreator));
```
smithago, line 174: Was'nt the previous way of calling better? What needs to be validated for a VR type was a concern in one API, now it is callers responsibility to figure this out.

src/Microsoft.Health.Dicom.Core/Features/Indexing/IDicomDatasetReindexValidator.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Collections.Generic;
+using Dicom;
+using Microsoft.Health.Dicom.Core.Features.ExtendedQueryTag;
+
+namespace Microsoft.Health.Dicom.Core.Features.Indexing
+{
+    /// <summary>
+    /// Validator that validates DicomDataset for reindexing.
+    /// </summary>
+    public interface IDicomDatasetReindexValidator
```
smithago, line 15: Why do we need another class to validate this? Is'nt this same as STOW validation which we already do with ExtendedQueryTag.

src/Microsoft.Health.Dicom.Core/Features/Indexing/DicomDatasetReindexValidator.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Collections.Generic;
+using Dicom;
+using EnsureThat;
+using Microsoft.Health.Dicom.Core.Exceptions;
+using Microsoft.Health.Dicom.Core.Features.ExtendedQueryTag;
+using Microsoft.Health.Dicom.Core.Features.Validation;
+
+namespace Microsoft.Health.Dicom.Core.Features.Indexing
+{
+    public class DicomDatasetReindexValidator : IDicomDatasetReindexValidator
```
smithago, line 15: FYI, don;t use DICOM prefix.

src/Microsoft.Health.Dicom.Core/Features/Indexing/DicomDatasetReindexValidator.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        private readonly IDicomElementMinimumValidator _minimumValidator;
+
+        public DicomDatasetReindexValidator(IDicomElementMinimumValidator minimumValidator)
+        {
+            _minimumValidator = EnsureArg.IsNotNull(minimumValidator, nameof(minimumValidator));
+        }
+        public IReadOnlyCollection<QueryTag> Validate(DicomDataset dicomDataset, IReadOnlyCollection<QueryTag> queryTags)
```
smithago, line 23: Left a comment on the interface, if possible we should avoid these multiple validator classes. Logically they are just indexing like STOW, you may want to just control some behaviors through params

src/Microsoft.Health.Dicom.Core/Features/Indexing/InstanceReindexer.cs
```diff
@@ -23,19 +23,29 @@ public class InstanceReindexer : IInstanceReindexer
...
             _metadataStore = EnsureArg.IsNotNull(metadataStore, nameof(metadataStore));
             _indexDataStore = EnsureArg.IsNotNull(indexDataStore, nameof(indexDataStore));
+            _dicomDatasetReindexValidator = EnsureArg.IsNotNull(dicomDatasetReindexValidator, nameof(dicomDatasetReindexValidator));
         }
 
         public async Task ReindexInstanceAsync(IReadOnlyCollection<ExtendedQueryTagStoreEntry> entries, VersionedInstanceIdentifier versionedInstanceId, CancellationToken cancellationToken)
         {
             EnsureArg.IsNotNull(entries, nameof(entries));
             EnsureArg.IsNotNull(versionedInstanceId, nameof(versionedInstanceId));
+            if (entries.Count == 0)
```
smithago, line 39: When would we be in ReIndexing and not have any tags to re-index?

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Validation/DicomDateValidationTests.cs
```diff
@@ -0,0 +1,34 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using Dicom;
+using Microsoft.Health.Dicom.Core.Exceptions;
+using Microsoft.Health.Dicom.Core.Features.Validation;
+using Xunit;
+
+namespace Microsoft.Health.Dicom.Core.UnitTests.Features.Validation
+{
+    public class DicomDateValidationTests
```
smithago, line 13: Brandon removed Dicom from all class name prefix, because as a team we thought it was reduandent to use Dicom in Dicom project.

Please remove Dicom prefix in all classed

src/Microsoft.Health.Dicom.Core/Features/Validation/ValidationLimits.cs
```diff
@@ -0,0 +1,47 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using System.Linq;
+using Dicom;
+using EnsureThat;
+
+namespace Microsoft.Health.Dicom.Core.Features.Validation
+{
+    internal static class ValidationLimits
```
smithago, line 13: Can you pick a name that is more specific to VRs?

src/Microsoft.Health.Dicom.Core/Features/Validation/ValidationErrorCode.cs
```diff
@@ -0,0 +1,69 @@
...
+    /// [100,200) => Error for PN.
+    /// [200,300) => Error for DA.
+    /// Each VR could have up to 100 error code.
+    /// </summary>
+    public enum ValidationErrorCode
+    {
+        /// <summary>
+        /// No error.
+        /// </summary>
+        None = 0,
```
smithago, line 22: Is this even valid? ErrorCode enum and none

src/Microsoft.Health.Dicom.Core/Features/Validation/ValidationErrorCode.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        /// <summary>
+        /// Dicom element has multiple values -- we only support indexing single value dicom element.
+        /// </summary>
+        ElementHasMultipleValues = 1,
```
smithago, line 27: If we really need to use errorCodes, we need strongly typed exceptions with the error code in them

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMinimumValidation.cs
```diff
@@ -1,223 +0,0 @@
...
-
-namespace Microsoft.Health.Dicom.Core.Features.Validation
-{
-    public static class DicomElementMinimumValidation
-    {
-        private static readonly Regex ValidIdentifierCharactersFormat = new Regex("^[0-9\\.]*$", RegexOptions.Compiled);
-        private const string DateFormatDA = "yyyyMMdd";
-        private const string BinaryDataPlaceHolder = "<BinaryData>";
-
-        internal static void ValidateAE(string value, string name)
```
smithago, line 22: I like the generalization, but we still need type specific validation calling those.

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMinimumValidator.cs
```diff
@@ -3,67 +3,53 @@
...
-        }
-
-        // only works for single value dicom element
-        public void Validate(DicomElement element)
+            { DicomVR.AE, new DicomElementMaxLengthValidation(16) },
+            { DicomVR.AS, new DicomElementRequiredLengthValidation(4) },
+            { DicomVR.CS, new DicomElementMaxLengthValidation(16) },
+            { DicomVR.DA, new DicomDateValidation() },
+            { DicomVR.DS, new DicomElementMaxLengthValidation(16) },
+            { DicomVR.FL, new DicomElementRequiredLengthValidation(4) },
```
smithago, line 23: How do we handle more than one type validation for a specific VR? We still need VR specific validation encapsulated in one place for all callers.

src/Microsoft.Health.Dicom.Core/Exceptions/DicomElementValidationException.cs
pengchen0692: sure, solved.

src/Microsoft.Health.Dicom.Core/Features/Delete/DeleteHandler.cs
pengchen0692: Yeah, Brandon mentioned similar things before -- we should remove prefix `Dicom`. But today I still see some inconsistency -- e.g: `DicomElementMinimumValidator`.

 I don't have good idea on this. Check the classes under Validation namespace: 
image
If `UidValidator` is used, should `DicomDateValidation` to be `DateValidation` ? And if `DateValidation` is too generic?

src/Microsoft.Health.Dicom.Core/Features/ExtendedQueryTag/ExtendedQueryTagEntryValidator.cs
pengchen0692: If I understand correctly, what you mean is previously caller only need to know `DicomElementMinimumValidation`, and now need to figure out which class (it's `DicomLongStringValidation`) in this case for validation?

yeah, I think that's the Cons of this approach. 

In the other side, I feel this approach make `DicomElementMinimumValidator` more clean and easier to extended. 
I feel `DicomElementMinimumValidator` is the "main scenario" while  validating LongString in `ExtendedQueryTagEntryValidator` is kind of "edge case", so  I feel this way is better.

src/Microsoft.Health.Dicom.Core/Features/Indexing/IDicomDatasetReindexValidator.cs
pengchen0692: there are some difference between them. 
1. STOW validates all 3 Uids (exists and different), and some required tag (patientId, SopClassUid), while reindex don't need to
2. STOW support full validation, while reindex don't need to
3. STOW validate all query tags while reindex is on subset (this could be solved by refactoring)
4. STOW throws exception on failure, while reindex log the error (this could also be solved by refactoring)

src/Microsoft.Health.Dicom.Core/Features/Indexing/InstanceReindexer.cs
pengchen0692: The intention is to avoid Blob read when there is no query tags (query tag could be deleted during reindex).
But I checked the logic, it's impossible `entries.Count ==0`
will update.

src/Microsoft.Health.Dicom.Core/Features/Validation/ValidationErrorCode.cs
pengchen0692: it for code analysis CA1008
image

src/Microsoft.Health.Dicom.Core/Features/Validation/ValidationErrorCode.cs
pengchen0692: The error code will be stored into SQL database.  we won't throw exceptions based on them.

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMinimumValidator.cs
pengchen0692: could you share a case when need need such validation?

src/Microsoft.Health.Dicom.Core/Features/Validation/ValidationErrorCode.cs
pengchen0692: solved

src/Microsoft.Health.Dicom.Core/Features/Indexing/IDicomDatasetReindexValidator.cs
pengchen0692: I was trying merge these 2 validators together, but find the code is fragile, so use a different way to simplify.
* Encapsulate STOW validation into 3 classes -- `DatasetCoreTagsValidation`, `DatasetFullValidation` and `DatasetQueryTagsValidation`
* Reindex validation also calls `DatasetQueryTagsValidation`
* `DatasetQueryTagsValidation` accept custom `Function` to decided whether throws exceptions or not ( STOW throws exception while Reindex not throw but write into database)
* Also rename:  `DicomDatasetValidator` => `StoreDatasetValidator` , `DicomDatasetReindexValidator`=>`ReindexDatasetValidator` , which I feel more clear.

Please let me know if you have any concerns.

src/Microsoft.Health.Dicom.Core/Features/Indexing/DicomDatasetReindexValidator.cs
pengchen0692: removed

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementMinimumValidation.cs
pengchen0692: resolve as we discussed

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementValidation.cs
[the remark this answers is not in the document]
pengchen0692: will be resolved.

src/Microsoft.Health.Dicom.Core/Features/Validation/ValidationLimits.cs
```diff
@@ -0,0 +1,47 @@
...
+    {
+        private static readonly HashSet<DicomVR> StringVrs = new HashSet<DicomVR>()
```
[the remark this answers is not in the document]
pengchen0692: resolved.

src/Microsoft.Health.Dicom.Core/Features/Validation/ValidationLimits.cs
```diff
@@ -0,0 +1,47 @@
...
+            DicomVR.FD,
+            DicomVR.SL,
+            DicomVR.SS,
+            DicomVR.UL,
+            DicomVR.US,
+        };
+
+        public static readonly HashSet<DicomVR> SupportedVRs = new HashSet<DicomVR>(StringVrs.Union(BinaryVrs));
+
+        public static bool CanGetAsString(this DicomVR vr)
```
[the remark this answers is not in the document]
pengchen0692: should be resolved

src/Microsoft.Health.Dicom.Core/Features/Validation/DicomElementRequiredLengthValidation.cs
```diff
@@ -0,0 +1,61 @@
...
+
+namespace Microsoft.Health.Dicom.Core.Features.Validation
+{
+    internal class DicomElementRequiredLengthValidation : DicomElementValidation
+    {
+        public int RequiredLength { get; }
+
+        public DicomElementRequiredLengthValidation(int requiredLength)
+        {
+            RequiredLength = requiredLength;
```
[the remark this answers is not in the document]
pengchen0692: should be resolved.

src/Microsoft.Health.Dicom.Core/DicomCoreResource.resx
wsugarman: Perhaps then just `Dicom element '{0}' failed validation for VR '{1}': {2}` ?

src/Microsoft.Health.Dicom.Core/Exceptions/DicomElementValidationException.cs
```diff
@@ -10,25 +10,35 @@ namespace Microsoft.Health.Dicom.Core.Exceptions
...
+            EnsureArg.IsNotNull(value, nameof(value));
 
             Name = name;
-            Value = value;
             VR = vr;
+            Value = value;
         }
 
-        public override string Message => string.Format(DicomCoreResource.DicomElementValidationFailed, Name, Value, VR.Code, base.Message);
+        public override string Message => Value == null
```
wsugarman, line 34: I think we should check the message too. If the user passed their own message, I don't think we need to preface it with another one.

src/Microsoft.Health.Dicom.Core/Features/Validation/ElementMinimumValidator.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.Diagnostics;
+using Dicom;
+using EnsureThat;
+
+namespace Microsoft.Health.Dicom.Core.Features.Validation
+{
+    public class ElementMinimumValidator : IElementMinimumValidator
+    {
+
+        private static readonly IReadOnlyDictionary<DicomVR, IElementValidation> Validations = new Dictionary<DicomVR, IElementValidation>
```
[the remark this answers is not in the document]
pengchen0692: like how we do with `Store`?

src/Microsoft.Health.Dicom.Core/Exceptions/DicomElementValidationException.cs
pengchen0692: This is existing behavior.  I think the idea is to given a general error message, along with detailed information. I'm not sure if should change it or not.

src/Microsoft.Health.Dicom.Core/Features/Indexing/IDicomDatasetReindexValidator.cs
pengchen0692: Discussed with Will, the 3 validation classes (DatasetCoreTagsValidation, DatasetFullValidation and DatasetQueryTagsValidation) is not appropriate based on their usage (new class , call function and discard), method would be  better.
Updated code based on the discussion. Get rid of the 3 classes.

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -160,5 +163,40 @@ public static void AddValueIfNotNull<T>(this DicomDataset dicomDataset, DicomTag
...
+            EnsureArg.IsNotNull(dataset, nameof(dataset));
+            EnsureArg.IsNotNull(queryTag, nameof(queryTag));
+            EnsureArg.IsNotNull(minimumValidator, nameof(minimumValidator));
+            DicomElement dicomElement = dataset.GetDicomItem<DicomElement>(queryTag.Tag);
+
+            if (dicomElement != null)
+            {
+                if (dicomElement.ValueRepresentation != queryTag.VR)
+                {
+
```
wsugarman, line 184: nit: extra newline here and below

src/Microsoft.Health.Dicom.Core/Features/Indexing/ReindexDatasetValidator.cs
```diff
@@ -0,0 +1,45 @@
...
+{
+    public class ReindexDatasetValidator : IReindexDatasetValidator
+    {
+        private readonly IElementMinimumValidator _minimumValidator;
+
+        public ReindexDatasetValidator(IElementMinimumValidator minimumValidator)
+        {
+            _minimumValidator = EnsureArg.IsNotNull(minimumValidator, nameof(minimumValidator));
+        }
+        public IReadOnlyCollection<QueryTag> Validate(DicomDataset dataset, IReadOnlyCollection<QueryTag> queryTags)
```
wsugarman, line 24: nit: missing newline

wsugarman (approved):

PR #912, opened 2021-07-28
Fix Re-indexing Dependency Injection

wsugarman:
## Description
- Fix dependency injection service container to include missing services like the SQL index stores, JSON.NET serializers, and RecycleableMemoryStreamManager
- Normalize serialization between web server and client to use System.Text.Json over JSON.NET in preparation for fo-dicom 5 and .NET 6

## Related issues
N/A

## Testing
Test locally

### DIFF
--- a/src/Microsoft.Health.Dicom.Functions.Client.UnitTests/DicomAzureFunctionsHttpClientTests.cs
+++ b/src/Microsoft.Health.Dicom.Functions.Client.UnitTests/DicomAzureFunctionsHttpClientTests.cs
@@ -8,8 +8,10 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
+using System.Net.Http.Json;
 using System.Net.Mime;
-using System.Text;
+using System.Text.Json;
+using System.Text.Json.Serialization;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Options;
@@ -17,14 +19,15 @@ using Microsoft.Health.Dicom.Core.Exceptions;
 using Microsoft.Health.Dicom.Core.Messages.Operations;
 using Microsoft.Health.Dicom.Core.Models.Operations;
 using Microsoft.Health.Dicom.Functions.Client.Configs;
-using Newtonsoft.Json;
 using Xunit;
 
 namespace Microsoft.Health.Dicom.Functions.Client.UnitTests
 {
     public class DicomAzureFunctionsHttpClientTests
     {
-        private static readonly IOptions<FunctionsClientOptions> DefaultConfig = Options.Create(
+        private readonly IOptions<JsonSerializerOptions> _jsonSerializerOptions;
+
+        private static readonly IOptions<FunctionsClientOptions> DefaultOptions = Options.Create(
             new FunctionsClientOptions
             {
                 BaseAddress = new Uri("https://dicom.core/unit/tests/", UriKind.Absolute),
@@ -35,19 +38,30 @@ namespace Microsoft.Health.Dicom.Functions.Client.UnitTests
                 }
             });
 
+        public DicomAzureFunctionsHttpClientTests()
+        {
+            var options = new JsonSerializerOptions();
+            options.Converters.Add(new JsonStringEnumConverter());
+            _jsonSerializerOptions = Options.Create(options);
+        }
+
         [Fact]
         public void GivenNullArguments_WhenConstructing_ThenThrowArgumentNullException()
         {
             var handler = new MockMessageHandler(new HttpResponseMessage(HttpStatusCode.NotFound));
             Assert.Throws<ArgumentNullException>(
-                () => new DicomAzureFunctionsHttpClient(null, DefaultConfig));
+                () => new DicomAzureFunctionsHttpClient(null, _jsonSerializerOptions, DefaultOptions));
 
             Assert.Throws<ArgumentNullException>(
-                () => new DicomAzureFunctionsHttpClient(new HttpClient(handler), null));
+                () => new DicomAzureFunctionsHttpClient(new HttpClient(handler), null, DefaultOptions));
+
+            Assert.Throws<ArgumentNullException>(
+                () => new DicomAzureFunctionsHttpClient(new HttpClient(handler), _jsonSerializerOptions, null));
 
             Assert.Throws<ArgumentNullException>(
                 () => new DicomAzureFunctionsHttpClient(
                     new HttpClient(handler),
+                    _jsonSerializerOptions,
                     Options.Create<FunctionsClientOptions>(null)));
         }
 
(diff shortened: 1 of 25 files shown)


src/Microsoft.Health.Dicom.Functions.Client.UnitTests/DicomAzureFunctionsHttpClientTests.cs
```diff
@@ -35,19 +38,30 @@ public class DicomAzureFunctionsHttpClientTests
                 }
             });
 
+        public DicomAzureFunctionsHttpClientTests()
+        {
+            var options = new JsonSerializerOptions();
+            options.Converters.Add(new JsonStringEnumConverter());
+            _jsonSerializerOptions = Options.Create(options);
+        }
```
yuedai0531, line 46: is it normal that this method is not used anywhere?

src/Microsoft.Health.Dicom.Functions.Client.UnitTests/DicomAzureFunctionsHttpClientTests.cs
```diff
@@ -8,23 +8,26 @@
...
 using Xunit;
 
 namespace Microsoft.Health.Dicom.Functions.Client.UnitTests
 {
     public class DicomAzureFunctionsHttpClientTests
     {
-        private static readonly IOptions<FunctionsClientOptions> DefaultConfig = Options.Create(
+        private readonly IOptions<JsonSerializerOptions> _jsonSerializerOptions;
+
+        private static readonly IOptions<FunctionsClientOptions> DefaultOptions = Options.Create(
```
yuedai0531, line 30: going forward are we changing all Config to be Options?

wsugarman: I think generally we should create a distinction between the concept of "options," a collection of settings, and a "config," a collection of settings sourced from a particular storage medium such as a file. In .NET Core, it appears these settings are often called "options" and are loaded into the DI container via the `IOptions<T>`, `IOptionsSnapshot<T>`, and `IOptionsMonitor<T>`. It also just so happens that the data in an `IOptions<T>` instance is sourced from an `IConfiguration` object, but it is not a requirement.

src/Microsoft.Health.Dicom.Functions.Client.UnitTests/DicomAzureFunctionsHttpClientTests.cs
```diff
@@ -110,16 +124,16 @@ public async Task GivenSuccessfulResponse_WhenGettingStatus_ThenReturnStatus()
 @$"
 {{
   ""OperationId"": ""{id}"",
-  ""Type"": ""Reindex"",
-  ""CreatedTime"": ""{createdDateTime}"",
-  ""LastUpdatedTime"": ""{createdDateTime.AddMinutes(15)}"",
-  ""Status"": ""Running""
+  ""Type"": ""{OperationType.Reindex}"",
+  ""CreatedTime"": ""{createdDateTime:O}"",
+  ""LastUpdatedTime"": ""{createdDateTime.AddMinutes(15):O}"",
```
yuedai0531, line 129: what is this for? `:O`

yuedai0531: also is it normal to indent json like this?

src/Microsoft.Health.Dicom.Functions.Client.UnitTests/DicomAzureFunctionsHttpClientTests.cs
yuedai0531: should we rename this then?
image

src/Microsoft.Health.Dicom.Functions.Client.UnitTests/DicomAzureFunctionsHttpClientTests.cs
wsugarman: That is the [format specifier used by `DateTime`](https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings#table-of-format-specifiers) to indicate "Round Trip." .NET string interpolation uses the [following syntax](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated#structure-of-an-interpolated-string):

```
{<interpolationExpression>[,<alignment>][:<formatString>]}
```

src/Microsoft.Health.Dicom.Functions.Client.UnitTests/DicomAzureFunctionsHttpClientTests.cs
yuedai0531: oh oops nevermind

wsugarman: Yeah, I think that makes sense

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Trigger.cs
```diff
@@ -60,17 +57,18 @@ public partial class ReindexDurableFunction
...
+                    throw new JsonException("Expected at least 1 element in array.");
+                }
             }
-
-            if (extendedQueryTagKeys == null || extendedQueryTagKeys.Count == 0)
+            catch (JsonException e)
             {
+                logger.LogError(e, "Cannot deserialize extended query tag keys.");
                 return new HttpResponseMessage { StatusCode = HttpStatusCode.BadRequest };
             }
```
yuedai0531, line 73: https://media.giphy.com/media/mGK1g88HZRa2FlKGbz/giphy.gif

src/Microsoft.Health.Dicom.Functions/Management/DurableClientProxy.cs
```diff
@@ -69,16 +80,20 @@ public static class DurableClientProxyFunctions
 
             if (string.IsNullOrWhiteSpace(instanceId))
             {
-                return new BadRequestResult();
+                return new HttpResponseMessage { StatusCode = HttpStatusCode.BadRequest };
```
yuedai0531, line 83: what is the reason for the change here?

yuedai0531: I think the majority of our controllers return an IActionResult? for consistency shouldn't we keep it like that?

yuedai0531 (approved):
j'ai fait mon possible :)

src/Microsoft.Health.Dicom.Functions/Management/DurableClientProxy.cs
wsugarman: So ASP.NET Core relies on middleware to perform serialization. So typically APIs return `IActionResult` which typically encapsulate an HTTP Status Code and optionally some payload. Middleware then takes that payload and serializes it into the proper content type, like JSON or XML. However, Azure Functions are special. It seems they've custom-built a bunch of it including the DI service provider. So unfortunately while ASP.NET Core 3.1 uses the new `System.Text.Json` lib for serialization that is part of the BCL, Azure Functions relies on Json.NET (aka `Newtonsoft.Json`). Furthermore, you can seem to swap it out, so I am forced to serialize it myself and directly return `HttpResponseMessage` with the serialized payload.

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Trigger.cs
wsugarman: lol I didn't want two different conditions to return `BadRequest`, so the deserialization may throw the exception... or I may do it directly 😛

PR #922, opened 2021-07-30
Retry Activities on Failure

wsugarman:
## Description
- Pass `RetryPolicy` when invoking activities
- Fix `IGuidFactory` service descriptor
- Add retry policy to configuration
- Add option validation

## Related issues
[AB#84035](https://microsofthealth.visualstudio.com/Health/_workitems/edit/84035)

## Testing
Unit test and local run

### DIFF
--- a/src/Microsoft.Health.Dicom.Functions/Durable/RetryOptionsTemplate.cs
+++ b/src/Microsoft.Health.Dicom.Functions/Durable/RetryOptionsTemplate.cs
@@ -11,7 +11,7 @@ namespace Microsoft.Health.Dicom.Functions.Durable
     {
         public TimeSpan FirstRetryInterval { get; set; } = TimeSpan.Zero;
 
-        public TimeSpan MaxRetryInterval { get; set; } = TimeSpan.FromDays(6.0);
+        public TimeSpan MaxRetryInterval { get; set; } = TimeSpan.FromDays(6.0); // Durable Function Maximum
 
         public double BackoffCoefficient { get; set; } = 1;
 
--- a/src/Microsoft.Health.Dicom.Functions/Durable/RetryOptionsTemplate.cs
+++ b/src/Microsoft.Health.Dicom.Functions/Durable/RetryOptionsTemplate.cs
@@ -1,0 +1,22 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.Health.Dicom.Functions.Durable
+{
+    internal sealed class RetryOptionsTemplate
+    {
+        public TimeSpan FirstRetryInterval { get; set; } = TimeSpan.Zero;
+
+        public TimeSpan MaxRetryInterval { get; set; } = TimeSpan.FromDays(6.0);
+
+        public double BackoffCoefficient { get; set; } = 1;
+
+        public TimeSpan RetryTimeout { get; set; } = TimeSpan.MaxValue;
+
+        public int MaxNumberOfAttempts { get; set; } = 1;
+    }
+}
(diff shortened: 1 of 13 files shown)


pengchen0692 (commented):
I'm not sure if retrying on DataStoreException is good idea. It could contains errors that not retriable.

src/Microsoft.Health.Dicom.Functions/Durable/RetryOptionsTemplate.cs
```diff
@@ -0,0 +1,22 @@
...
+
+using System;
+
+namespace Microsoft.Health.Dicom.Functions.Durable
+{
+    internal sealed class RetryOptionsTemplate
+    {
+        public TimeSpan FirstRetryInterval { get; set; } = TimeSpan.Zero;
+
+        public TimeSpan MaxRetryInterval { get; set; } = TimeSpan.FromDays(6.0);
```
pengchen0692, line 14: would 6 days too big?
Imaging the activity is not executed until 6 days later...

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/SqlExtendedQueryTagStoreV2.cs
```diff
@@ -94,23 +94,30 @@ public override async Task<IReadOnlyList<ExtendedQueryTagStoreEntry>> GetExtende
                 VLatest.GetExtendedQueryTag.PopulateCommand(sqlCommandWrapper, path);
 
                 var executionTimeWatch = Stopwatch.StartNew();
-                using (var reader = await sqlCommandWrapper.ExecuteReaderAsync(CommandBehavior.SequentialAccess, cancellationToken))
+                try
```
pengchen0692, line 97: Do we need to catch any other **unknown** exceptions as DataStoreException?

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/SqlExtendedQueryTagStoreV4.cs
```diff
@@ -41,21 +41,28 @@ public override async Task<IReadOnlyList<ExtendedQueryTagStoreEntry>> GetExtende
             {
                 VLatest.GetExtendedQueryTagsByOperation.PopulateCommand(sqlCommandWrapper, operationId);
 
-                using (SqlDataReader reader = await sqlCommandWrapper.ExecuteReaderAsync(CommandBehavior.SequentialAccess, cancellationToken))
+                try
```
pengchen0692, line 44: Do we need to catch any other **unknown** exceptions as DataStoreException?

src/Microsoft.Health.Dicom.Functions/Durable/RetryOptionsTemplate.cs
wsugarman: I should add a comment. This is their internal max lol and the default max retry interval. This is to serve as a ceiling for the retry delay. Consider the following piecewise function, based on my reading of their [documentation](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-error-handling?tabs=csharp#automatic-retry-on-failure)

Let D(i) be the delay before the i-th retry where 1 <= i <= *MaxNumberOfAttempts* and *MaxNumberOfAttempts* > 1. Note that if *MaxNumberOfAttempts* is 1, then there is no retrying the operation.

```
       { MAX(MaxRetryInterval, FirstRetryInterval), if i == 1
D(i) = {
       { MAX(MaxRetryInterval, FirstRetryInterval * BackoffCoefficient * (i - 1)), if i > 1
```

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/SqlExtendedQueryTagStoreV2.cs
```diff
@@ -94,23 +94,30 @@ public override async Task<IReadOnlyList<ExtendedQueryTagStoreEntry>> GetExtende
...
+                                VLatest.ExtendedQueryTag.TagVR,
+                                VLatest.ExtendedQueryTag.TagPrivateCreator,
+                                VLatest.ExtendedQueryTag.TagLevel,
+                                VLatest.ExtendedQueryTag.TagStatus);
+
+                            results.Add(new ExtendedQueryTagStoreEntry(tagKey, tagPath, tagVR, tagPrivateCreator, (QueryTagLevel)tagLevel, (ExtendedQueryTagStatus)tagStatus));
+                        }
+
+                        executionTimeWatch.Stop();
+                        Logger.LogInformation(executionTimeWatch.ElapsedMilliseconds.ToString());
```
jovinson-ms, line 115: Should this log message have more context?

pengchen0692 (approved):

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/SqlExtendedQueryTagStoreV2.cs
wsugarman: Moved this change out of the PR

PR #924, opened 2021-08-01
Make DicomWebClient apiversion aware

smithago:
## Description
- Move apiversion from callers or tests DicomWebClient. This allows some tests that where not using apiversion to always use it.
- Make all URI taking APIs private and expose more friendlier APIs
- This change also means E2E will target the latest Web APIs. In the future if we need to test other versions we may need routing unit test and tests targeting and validating specific versions.

## Related issues
[AB#84041](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/84041)

## Testing
E2E existing tests pass


tools/scale-testing/QidoQueryGenerator/Program.cs
```diff
@@ -34,15 +34,15 @@ public static void Main(string[] args)
                 string namePrefix = pI.Name.Split('^')[0].Substring(0, 3);
                 string startDate = DateTime.Parse(pI.PerformedProcedureStepStartDate).AddMonths(-1).ToString("yyyyMMdd");
                 string endDate = DateTime.Parse(pI.PerformedProcedureStepStartDate).AddMonths(1).ToString("yyyyMMdd");
-                string studyGenericVal = $"/Studies?PatientName={namePrefix}&StudyDate={startDate}-{endDate}&fuzzyMatching=true";
-                string studySpecificVal = $"/Studies?PatientID={pI.PatientId}";
-                string seriesGenericVal = $"/Series?Modality={pI.Modality}";
-                string seriesSpecificVal = $"/Series?Modality={pI.Modality}&PatientID={pI.PatientId}";
+                string studyGenericVal = $"PatientName={namePrefix}&StudyDate={startDate}-{endDate}&fuzzyMatching=true";
+                string studySpecificVal = $"PatientID={pI.PatientId}";
+                //string seriesGenericVal = $"/Series?Modality={pI.Modality}";
```
jnlycklama, line 39: Should these paths be commented out?

jnlycklama (approved):

tools/scale-testing/QidoQueryGenerator/Program.cs
smithago: Yes, because the url was targeting the generic QueryAsync(). So for now the scale test will target Study level query.

PR #935, opened 2021-08-06
Publish unit test coverage

jovinson-ms:
## Description
Adds argument to publish test coverage for unit tests.

## Testing
Manual testing.

### DIFF
--- a/build/build.yml
+++ b/build/build.yml
@@ -25,7 +25,7 @@ steps:
     inputs:
       command: test
       projects: '**/*UnitTests/*.csproj'
-      arguments: '--configuration $(buildConfiguration) --no-build'
+      arguments: '--configuration $(buildConfiguration) --no-build --publishTestCoverage'
 
   - ${{ if eq(parameters.packageArtifacts, 'true') }}:
     - template: package.yml


build/build.yml
```diff
@@ -25,7 +25,7 @@ steps:
     inputs:
       command: test
       projects: '**/*UnitTests/*.csproj'
-      arguments: '--configuration $(buildConfiguration) --no-build'
+      arguments: '--configuration $(buildConfiguration) --no-build --publishTestResults'
```
smithago, line 28: Good idea, how much more time is it adding to the build times?

wsugarman, line 28: I am unaware of this switch. I know Visual Studio enterprise has code coverage included, but I think we'll want to integetrate something like Coverlet into VSTest. and run its corresponding data collector.

MS Docs: https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-code-coverage?tabs=windows#tooling

I've also done this before in my own repo, and you can see below for some examples:
- [ADO Pipeline Template](https://github.com/wsugarman/Sweetener/blob/79bc4dcea6de1091be81b6a4368fa6fb39958b73/eng/Templates/Steps.Validate.yml#L28)
    - There is some nuance with it in that I run a PS script afterwards to ensure some percentage of line and branch coverage. This "threshold validation" is only available when via MSBuild last time I checked. See [here](https://github.com/coverlet-coverage/coverlet/blob/master/Documentation/VSTestIntegration.md#coverlet-options-supported-by-vstest-integration) for details
- [CodeCoverage.runsettings](https://github.com/wsugarman/Sweetener/blob/main/src/CodeCoverage.runsettings)

PR #937, opened 2021-08-10
Add HTTP Authentication for DICOM Function Web Triggers

wsugarman:
## Description
- Add authentication for web triggers
- Minor refactor

## Related issues
[AB#84167](https://microsofthealth.visualstudio.com/Health/_workitems/edit/84167)

## Testing
Test locally and via build pipelines


src/Microsoft.Health.Dicom.Functions/Registration/ServiceCollectionExtensions.cs
```diff
@@ -101,5 +108,43 @@ public static IMvcBuilder AddJsonSerializerOptions(this IMvcBuilder builder, Act
             builder.Services.Configure(configure);
             return builder;
         }
+
+        private static IServiceCollection AddTriggerAuthentication(this IServiceCollection services, SecurityConfiguration config)
```
wsugarman, line 112: Unfortunately, we don't yet today have a package for shared ASP.NET Core dependencies, so I copied a subset of the logic from the API project here.

jovinson-ms (approved):

PR #945, opened 2021-08-13
PR Pipeline optimizations to improve time

smithago:
## Description
- Make build and publish in linux
- Make windows build for security scan which is slow and make it not depend on environment setup
- Make nuget generation optional and skip it in PR
- Make integration tests run parallel and reduce the SQL schema version supported

## Related issues
[AB#84346](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/84346)

## Testing
Attached PR build is successfull

### DIFF
--- a/src/Microsoft.Health.Dicom.SqlServer/Features/Schema/SchemaVersionConstants.cs
+++ b/src/Microsoft.Health.Dicom.SqlServer/Features/Schema/SchemaVersionConstants.cs
@@ -7,7 +7,7 @@ namespace Microsoft.Health.Dicom.SqlServer.Features.Schema
 {
     public static class SchemaVersionConstants
     {
-        public const int Min = (int)SchemaVersion.V1;
+        public const int Min = (int)SchemaVersion.V2;
         public const int Max = (int)SchemaVersion.V3;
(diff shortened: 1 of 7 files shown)


src/Microsoft.Health.Dicom.SqlServer/Features/Schema/SchemaVersionConstants.cs
```diff
@@ -7,7 +7,7 @@ namespace Microsoft.Health.Dicom.SqlServer.Features.Schema
 {
     public static class SchemaVersionConstants
     {
-        public const int Min = (int)SchemaVersion.V1;
+        public const int Min = (int)SchemaVersion.V2;
```
brandonpollett, line 10: Do we need to roll versions when we do this? 

This change implies that anyone that wants to take this upgrade they have to migrate from schema 1 if that's where they still are.

smithago: I can bump minor version

jovinson-ms (approved):

PR #951, opened 2021-08-13
Update swagger generation

brandonpollett:
## Description
This PR updates the swagger generation in the following ways:
* Moves strings to be config settings, settable via configuration
* Allows for generation of both `yaml` and `json` swagger files.
* Allows for setting a server url in the generated swagger
* Moves swagger classes to `features` folder.

## Related issues
Addresses [AB#84341](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/84341).

## Testing
Manual testing of generated swagger files.

### DIFF
--- a/src/Microsoft.Health.Dicom.Api/Configs/SwaggerConfiguration.cs
+++ b/src/Microsoft.Health.Dicom.Api/Configs/SwaggerConfiguration.cs
@@ -1,0 +1,19 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using Microsoft.OpenApi.Models;
+
+namespace Microsoft.Health.Dicom.Api.Configs
+{
+    public class SwaggerConfiguration
+    {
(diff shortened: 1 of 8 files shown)


wsugarman (approved):

src/Microsoft.Health.Dicom.Api/Configs/SwaggerConfiguration.cs
```diff
@@ -0,0 +1,19 @@
...
+using System;
+using Microsoft.OpenApi.Models;
+
+namespace Microsoft.Health.Dicom.Api.Configs
+{
+    public class SwaggerConfiguration
+    {
+        public Uri ServerUri { get; set; }
+
+        public string Title { get; } = "Medical Imaging Server for DICOM";
```
abdullah248, line 15: Just wondering why we have the title set here and the license information in the appsettings.

brandonpollett: I thought this made sense as the default as it wouldn't likely change. Additionally, in looking at the underlying `OpenApiInfo`, the `title` is a required field and the `license` doesn't appear to be. 

Essentially, this should allow our service to startup without specifying anything for the `SwaggerConfiguration` and it should still spit out a valid swagger file.

Issue #932, opened 2021-08-04
DICOM web standard in Python Notebook broken


SimonBiggs:
Hi All,

I was just taking a squiz at this repo and as I was "driving by" I saw a few issues, figured I'd flag them so that you can be aware :slightly_smiling_face: .

---

The following notebook:

https://github.com/microsoft/dicom-server/blob/main/docs/resources/use-dicom-web-standard-apis-with-python.ipynb

Gives the following error message:

```
File Load Error for use-dicom-web-standard-apis-with-python.ipynb
Unreadable Notebook: /home/simon/Downloads/use-dicom-web-standard-apis-with-python.ipynb NotJSONError('Notebook does not appear to be JSON: \'{\\n "cells": [\\n {\\n "cell_type": "m...')
```

Potentially it might be worth including the notebook examples within the testing suite with something like https://github.com/nteract/testbook?

---

Also, the link to the notebook from the markdown file gives a 404:

https://github.com/microsoft/dicom-server/blame/main/docs/tutorials/use-dicom-web-standard-apis-with-python.md#L3

Cheers :slightly_smiling_face:,
Simon

abdullah248:
Hi @SimonBiggs,
Thanks for pointing these out! The jupyter notebook should work now and the link the documentation has been updated. 

Thanks!

SimonBiggs:
Thanks Abdullah :), and my pleasure :)

PR #952, opened 2021-08-13
Jupyter Notebook and Doc Update

abdullah248:
## Description
Fixes jupyter notebook and a broken link.

## Related issues
#932


docs/resources/use-dicom-web-standard-apis-with-python.ipynb
```diff
@@ -89,7 +89,7 @@
     "dicom_server_name = \"{server-name}\"\n",
     "path_to_dicoms_dir = \"{path to the folder that includes green-square.dcm and other dcm files}\"\n",
     "\n",
-    "base_url = f\"https://{dicom_server_name}.azurewebsites.net\v{version}\"\n",
+    "base_url = f\"https://{dicom_server_name}.azurewebsites.net/v{version}\"\n",
```
wsugarman, line 92: We should probably define this `version` variable too, right? Like we did for `dicom_server_name`?

wsugarman (approved):

PR #957, opened 2021-08-18
Fix Extended Query Tag end to end tests

pengchen0692:
## Description
* Fix Extended Query Tag end to end tests
* Split DicomWebClient.cs and IDicomWebClient.cs into multiple files based on purpose

## Related issues
Addresses [AB#84410](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/84410).

## Testing
All tests pass.


src/Microsoft.Health.Dicom.Client/Models/OperationReference.cs
```diff
@@ -0,0 +1,27 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.Health.Dicom.Client.Models
+{
+    /// <summary>
+    /// Represents a reference to an existing long-running oepration.
+    /// </summary>
+    public class OperationReference
```
wsugarman, line 13: At some point we should definitely have a separate shared lib that contains our API contracts. That is way too big of an ask to do in this PR but something to remember.

wsugarman (approved):

PR #960, opened 2021-08-19
Record user error during indexing validation

yuedai0531:
## Description
Record user error during indexing validation.
Inject ErrorStore to ReindexDatasetValidator and record user error in error store.

## Related issues
Addresses [User story 82179](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Medical%20Imaging/Stories/?workitem=82179).

## Testing

Was manually tested to work and demonstrated during my demo


src/Microsoft.Health.Dicom.Core/Features/Indexing/ReindexDatasetValidator.cs
```diff
@@ -35,9 +37,12 @@ public IReadOnlyCollection<QueryTag> Validate(DicomDataset dataset, long waterma
                     dataset.ValidateQueryTag(queryTag, _minimumValidator);
                     validTags.Add(queryTag);
                 }
-                catch (DicomElementValidationException)
+                catch (DicomElementValidationException e)
                 {
-                    // TODO: log failure
+                    _extendedQueryTagErrorStore.AddExtendedQueryTagErrorAsync(
+                        queryTag.ExtendedQueryTagStoreEntry.Key,
+                        e.Message,
```
wsugarman, line 44: We do have a max size here. How will we enforce this?

pengchen0692:
As discussed we use error code instead.

PR #962, opened 2021-08-20
Remove docker project

pengchen0692:
## Description
Remove docker project from solution

## Related issues
Addresses [AB#82178](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/82178)


wsugarman:
Can we simply remove it from the solution until we fix the functions usage? Or simply remove functions from it?

wsugarman:
We also need to keep the SQL dockerfile regardless of what we do

wsugarman (approved):

PR #979, opened 2021-08-30
Refactor Re-Index Batching to Skip Unused Watermarks

wsugarman:
## Description
- Add new stored procedure for retrieving batches of DICOM instances from SQL
- Remove unnecessary properties from `WatermarkRange` and refactor to make `End` inclusive
- Add new tests

## Related issues
[AB#84771](https://microsofthealth.visualstudio.com/Health/_workitems/edit/84771/)

## Testing
Tested locally and via pipeline


src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Orchestration.cs
```diff
@@ -56,11 +55,15 @@ public partial class ReindexDurableFunction
             List<int> queryTagKeys = queryTags.Select(x => x.Key).ToList();
             if (queryTags.Count > 0)
             {
-                List<WatermarkRange> batches = await GetBatchesAsync(context, input.Completed, logger);
+                IReadOnlyList<WatermarkRange> batches = await context.CallActivityWithRetryAsync<IReadOnlyList<WatermarkRange>>(
```
pengchen0692, line 58: this logic is much more cleaner than the other one https://github.com/microsoft/dicom-server/pull/976, maybe we should check this one in at first.

wsugarman: Looking at my other PR, I think that makes sense too

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/4.diff.sql
```diff
@@ -151,6 +134,50 @@ AS
...
+
+    SELECT
+        MIN(Watermark) AS MinWatermark,
+        MAX(Watermark) AS MaxWatermark
+    FROM
+    (
+        SELECT TOP (@batchSize * @batchCount)
+            Watermark,
+            (ROW_NUMBER() OVER(ORDER BY Watermark DESC) - 1) / @batchSize AS Batch
+        FROM dbo.Instance
```
pengchen0692, line 173: Status  = 1 ?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/4.diff.sql
```diff
@@ -151,6 +134,50 @@ AS
...
+        WHERE @maxWatermark IS NULL or Watermark < @maxWatermark
```
pengchen0692, line 174: Do we want to make @maxWatermark inclusive,  since everywhere is Inclusive?
Or add comments here explicit tells @maxWatermark is exclusive?

pengchen0692: nit: or =>OR

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/4.diff.sql
wsugarman: Ah yes! Good catch

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/4.diff.sql
wsugarman: Here, it's a bit different, as we often pass in the last "start" value which we've already processed. So we could instead pass something like "start - 1" which would be fine too. It would also make the nullable value easier to process because would could do something like `Watermark <= ISNULL(@maxWatermark, Watermark)`

pengchen0692 (approved):

PR #980, opened 2021-08-30
DicomElementValidationException now include ErrorCode

pengchen0692:
## Description
DicomElementValidationException now include ErrorCode.
Move error related information (name,value, vr) into ElementValidationError class.

## Related issues
Addresses [AB#83998](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/83998)

## Testing
All tests passed


src/Microsoft.Health.Dicom.Core/Features/Validation/Errors/ElementValidationError.cs
```diff
@@ -0,0 +1,46 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using Dicom;
+using EnsureThat;
+
+namespace Microsoft.Health.Dicom.Core.Features.Validation.Errors
+{
+    public abstract class ElementValidationError
```
wsugarman, line 11: Should all of these "error" classes simply be derived `Exceptions` from `DicomElementValidationException`? In other words, should we merge `DicomElementValidationException` and `ElementValidationError`?

src/Microsoft.Health.Dicom.Core/Features/Validation/PersonNameValidation.cs
```diff
@@ -27,23 +29,34 @@ public override void Validate(DicomElement dicomElement)
...
             }
 
             foreach (var group in groups)
             {
-                ElementMaxLengthValidation.Validate(group, 64, $"{name} Group", dicomElement.ValueRepresentation);
+                try
+                {
+                    ElementMaxLengthValidation.Validate(group, 64, name, dicomElement.ValueRepresentation);
+                }
+                catch (DicomElementValidationException ex)
```
wsugarman, line 41: nit: you can move this predicate up into the catch expression:

```suggestion
                catch (DicomElementValidationException ex) when (ex.Error is ExceedMaxLengthError)
```

src/Microsoft.Health.Dicom.Core/Features/Validation/ValidationErrorCode.cs
```diff
@@ -0,0 +1,84 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Dicom.Core.Features.Validation
+{
+    /// <summary>
```
wsugarman, line 8: This is quite a bit for the `<summary/>`. Remember that this content is part of intellisense and may be rendered as part of docs. So I'd keep the summary to be the brief description of what it is. Details can be structured inside of `<remarks/>` and any notes for our own devs can be standard code comments like `//` or `/* ... */`

src/Microsoft.Health.Dicom.Core/Extensions/ValidationErrorCodeExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using Microsoft.Health.Dicom.Core.Features.Validation;
+using static Microsoft.Health.Dicom.Core.Features.Validation.ValidationErrorCode;
+
+namespace Microsoft.Health.Dicom.Core.Extensions
+{
+    internal static class ValidationErrorCodeExtensions
+    {
+        private static readonly IReadOnlyDictionary<ValidationErrorCode, string> MessageMap = new Dictionary<ValidationErrorCode, string>()
```
wsugarman, line 15: should we just use a `switch` statement or expression in the method?

src/Microsoft.Health.Dicom.Core.UnitTests/Extensions/ValidationErrorCodeExtensionsTests.cs
```diff
@@ -0,0 +1,26 @@
...
+using System;
+using Microsoft.Health.Dicom.Core.Extensions;
+using Microsoft.Health.Dicom.Core.Features.Validation;
+using Xunit;
+
+namespace Microsoft.Health.Dicom.Core.UnitTests.Extensions
+{
+    public class ValidationErrorCodeExtensionsTests
+    {
+
```
wsugarman, line 15: nit: extra newline

src/Microsoft.Health.Dicom.Core/Features/Validation/Errors/NotRequiredVRError.cs
```diff
@@ -0,0 +1,34 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Globalization;
+using Dicom;
+using EnsureThat;
+
+namespace Microsoft.Health.Dicom.Core.Features.Validation.Errors
+{
+    public class UnexpectedVRError : ElementValidationError
```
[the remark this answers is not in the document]
jovinson-ms: I've been using the [.NET guidelines on this](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions#capitalization-rules-for-identifiers) that call out capitalizing two letter acronyms as a special case of PascalCasing. If we have been following the convention you're suggesting, let's stick to that.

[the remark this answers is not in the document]
wsugarman: Oh, that makes sense. Let's stick to capitalizing here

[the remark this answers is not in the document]
pengchen0692: <=2 acronyms should be both capitalized. :)  according to https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions

src/Microsoft.Health.Dicom.Core/Features/Validation/Errors/ElementValidationError.cs
pengchen0692: I'm trying to encapsulate error information into a separated class, so that in the future, if we want both Store and Reindex have same error message, can easily jsonify into string, and store into some store( like SQL or table storage).

From another point, this might be overkill for now.

src/Microsoft.Health.Dicom.Core/Extensions/ValidationErrorCodeExtensions.cs
pengchen0692: There are a bunch of error codes, could add more in the future. I feel a long switch may not be a good idea.

BTW, find something for `switch` performance https://stackoverflow.com/questions/44905/c-sharp-switch-statement-limitations-why,  looks like C# compiler could optimize `switch` with jump table or binary search tree when it's going to big.

wsugarman: Ah yes. You can see the optimizations yourself by creating different sizes of switch statements and viewing them in ILSpy!

src/Microsoft.Health.Dicom.Core/Exceptions/Validation/DateIsInvalidException.cs
```diff
@@ -0,0 +1,18 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using Dicom;
+using Microsoft.Health.Dicom.Core.Features.Validation;
+
+namespace Microsoft.Health.Dicom.Core.Exceptions.Validation
+{
+    public class DateIsInvalidException : ElementValidationException
```
pengchen0692, line 11: Not sure if good idea to create Exception for each error code. The intention is to organize error code and associated error message together.

src/Microsoft.Health.Dicom.Core/Exceptions/Validation/ExceedMaxLengthException.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Globalization;
+using Dicom;
+using Microsoft.Health.Dicom.Core.Features.Validation;
+
+namespace Microsoft.Health.Dicom.Core.Exceptions.Validation
+{
+    public class ExceedMaxLengthException : ElementValidationException
+    {
+        public ExceedMaxLengthException(string name, DicomVR vr, string value, int maxlength) :
+            base(
```
wsugarman, line 15: If you do decide to keep these exceptions, a couple formatting guidelines:
1. If the call to the base ctor is too long for the line, move `: base` to the next line.
2. Ensure each line is only idented by 4 spaces and do not attempt to align it vertically (most of the time)
3. Typically, the last `)` is kept on the same line.

Eg.
```csharp
public ExceedMaxLengthException(string name, DicomVR vr, string value, int maxlength)
    : base(
        name,
        vr,
        value,
        ValidationErrorCode.ExceedMaxLength,
        string.Format(CultureInfo.InvariantCulture, DicomCoreResource.ErrorMessageExceedMaxLength, maxlength))
{
    // TODO: ...
}
```
Hopefully we can also keep the call to the base ctor on one line at most lol

src/Microsoft.Health.Dicom.Core/Exceptions/Validation/DateIsInvalidException.cs
pengchen0692: I think single `ElementValidationException` might be a cleaner solution. 
I created several static methods to generate exceptions for different error code, and encapsulate them into static class. in this way, don't need so many classes, also maintain the connection between error code and message, so that can easily create `ElementValidationException` for specific error code.

src/Microsoft.Health.Dicom.Core/Exceptions/ElementValidationException.cs
```diff
@@ -0,0 +1,47 @@
...
+    public class ElementValidationException : ValidationException
+    {
+        public ElementValidationException(string name, DicomVR vr, ValidationErrorCode errorCode, string message) : base(message)
+        {
+            Name = EnsureArg.IsNotNull(name, nameof(name));
+            VR = EnsureArg.IsNotNull(vr, nameof(vr));
+            ErrorCode = errorCode;
+        }
+
+        public ElementValidationException(string name, DicomVR vr, string value, ValidationErrorCode errorCode, string message) : base(message)
```
wsugarman, line 21: Do we actually care if `Value` is `null` or not? If not, we could simplify the first ctor to something like:
```csharp
public ElementValidationException(string name, DicomVR vr, ValidationErrorCode errorCode, string message)
    : this(name, vr, null , errorCode, message)
{
}
```

src/Microsoft.Health.Dicom.Core/Exceptions/ElementValidationExceptions.cs
```diff
@@ -0,0 +1,100 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System.Globalization;
+using Dicom;
+using EnsureThat;
+using Microsoft.Health.Dicom.Core.Features.Validation;
+
+namespace Microsoft.Health.Dicom.Core.Exceptions
+{
+    internal static class ElementValidationExceptions
```
wsugarman, line 13: nit: these are all methods, so they should probably start with a verb, like `CreateDateIsInvalidException`. If we do something like this, then maybe this class name should end with "Factory," like `ValidationExceptionFactory`?

src/Microsoft.Health.Dicom.Core/Features/Indexing/ReindexDatasetValidator.cs
```diff
@@ -35,7 +35,7 @@ public IReadOnlyCollection<QueryTag> Validate(DicomDataset dataset, long waterma
                     dataset.ValidateQueryTag(queryTag, _minimumValidator);
                     validTags.Add(queryTag);
                 }
-                catch (DicomElementValidationException)
+                catch (ElementValidationException)
                 {
                     // TODO: log failure
```
wsugarman, line 40: I take it we'll resolve this in another PR?

wsugarman (approved):

src/Microsoft.Health.Dicom.Core/Exceptions/ElementValidationException.cs
pengchen0692: There are some cases we log value into error message, while other cases don't want. They distinguished by given value or not.
e.g: When DA value is invalid, we log tag value. When there are multiple values in one tag, we don't log.

src/Microsoft.Health.Dicom.Core/Features/Indexing/ReindexDatasetValidator.cs
pengchen0692: I remember we  need to remove `Dicom` prefix from all classes in this repository, but don't remember if I did this in another PR. :)

PR #1001, opened 2021-09-12
Update Blob Service Client Registration APIs and Configuration

wsugarman:
## Description
- Update blob service client registration for the following reasons:
  1. Integrate blob client and underlying http request pipeline with existing DI service container
  2. Reduce boilerplate
  3. Streamline configuration whose types integrate with existing APIs
- Normalize JSON to 2 spaces for indent
- Bump version of Azurite image such that we can leverage latest blob client API

## Related issues
[AB#84923](https://microsofthealth.visualstudio.com/Health/_workitems/edit/84923)

## Testing
PR pipeline and local unit test runs

### DIFF
--- a/src/Microsoft.Health.Dicom.Metadata/Registration/DicomServerBuilderMetadataRegistrationExtensions.cs
+++ b/src/Microsoft.Health.Dicom.Metadata/Registration/DicomServerBuilderMetadataRegistrationExtensions.cs
@@ -5,10 +5,7 @@
 
 using EnsureThat;
 using Microsoft.Extensions.Configuration;
-using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Options;
 using Microsoft.Health.Blob.Configs;
-using Microsoft.Health.Blob.Features.Storage;
 using Microsoft.Health.Dicom.Core.Features.Common;
 using Microsoft.Health.Dicom.Core.Registration;
 using Microsoft.Health.Dicom.Metadata;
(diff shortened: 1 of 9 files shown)


smithago:
Should we rev the major version, since this is a breaking change in appsettings for blob?

smithago (approved):

src/Microsoft.Health.Dicom.Metadata/Registration/DicomServerBuilderMetadataRegistrationExtensions.cs
```diff
@@ -34,33 +31,23 @@ public static IDicomServerBuilder AddMetadataStorageDataStore(this IDicomServerB
...
-                        .AddMetadataHealthCheck();
+                .AddMetadataPersistence(configuration)
+                .AddMetadataHealthCheck();
         }
 
         private static IDicomServerBuilder AddMetadataPersistence(this IDicomServerBuilder serverBuilder, IConfiguration configuration)
         {
             IServiceCollection services = serverBuilder.Services;
 
-            services.AddBlobDataStore();
```
brandonpollett, line 45: I assume that most of what was in `AddBlobDataStore` is generally now considered to be called directly. Should we mark it as deprecated or provide guidance in the documentation for it?

brandonpollett (approved):

src/Microsoft.Health.Dicom.Metadata/Registration/DicomServerBuilderMetadataRegistrationExtensions.cs
wsugarman: It differs a bit from `AddBlobDataStore` in that the previous method combined:
1. Registration of a `BlobServiceClient`
2. Registration of `BlobDataStoreConfiguration`
3. Registration of blob initialization hosted service.

Whereas today I've split them out so provide some more idiomatic APIs and customization options. I've also tried to remove the dependency on `BlobDataStoreConfiguration` and instead use configurations that `BlobServiceClient` can directly interface with. That way we don't need to map every setting ourselves. You're right though. I should probably move to deprecate the old API if everything works out fine with this for us in DICOM.

PR #1011, opened 2021-09-14
Paginate Extended Query Tag Errors

wsugarman:
## Description
- Enable pagination for extended query tag errors with same `limit` and `offset` parameters as QIDO

## Related issues
[AB#85156](https://microsofthealth.visualstudio.com/Health/_workitems/edit/85156/)

## Testing
PR pipeline

### DIFF
--- a/src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/4.sql
+++ b/src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/4.sql
@@ -420,6 +420,17 @@ CREATE UNIQUE CLUSTERED INDEX IXC_ExtendedQueryTagError ON dbo.ExtendedQueryTagE
     Watermark
 )
 
+CREATE NONCLUSTERED INDEX IX_ExtendedQueryTagError_CreatedTime_Watermark_TagKey ON dbo.ExtendedQueryTagError
+(
+    CreatedTime,
+    Watermark,
+    TagKey
+)
+INCLUDE
+(
+    ErrorCode
+)
+
 /*************************************************************
     Extended Query Tag Operation Table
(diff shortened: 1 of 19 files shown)


src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/4.sql
```diff
@@ -1960,6 +1967,9 @@ BEGIN
     INNER JOIN dbo.Instance AS I
     ON XQTE.Watermark = I.Watermark
     WHERE XQTE.TagKey = @tagKey
+    ORDER BY CreatedTime ASC
```
pengchen0692, line 1970: 2 concerns:
1. should we index `CreatedTime` if sort by?
2. What if there are multiple instances with same `CreatedTime`, would that return inconsistent result?

pengchen0692: Looks like we should paginate on unique column https://stackoverflow.com/questions/57688411/sql-pagination-duplicates-with-ordering-by-non-unique-value, https://stackoverflow.com/questions/62133543/sql-server-paging-using-limit-and-offset-with-non-unique-ordering-column

pengchen0692: Propose1: use `Watermark` instead, as it always unique for single tag
Propose2: still use `CreatedTime` but combination with `Watermark`, in this way, we still need to add index for them.

wsugarman: I added the index after all!

pengchen0692 (approved):

PR #1018, opened 2021-09-14
Add default value for backward compatibility

pengchen0692:
## Description
Add default value to SP AddExtendedQueryTag for backward compatibility -- allow old binaries work with new SQL schema

## Related issues
Addresses [AB#85211](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/85211)

## Testing
All tests pass.


src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/4.diff.sql
```diff
@@ -535,7 +535,7 @@ GO
 /***************************************************************************************/
 CREATE OR ALTER PROCEDURE dbo.AddExtendedQueryTags (
     @extendedQueryTags dbo.AddExtendedQueryTagsInputTableType_1 READONLY,
-    @maxAllowedCount INT,
+    @maxAllowedCount INT = 128, -- Default value for backwards compatibility
```
wsugarman, line 538: Ah, v2 and v3 didn't have this. That makes sense.

wsugarman (approved):

PR #1046, opened 2021-09-21
Update middleware to add transcode and egress data for billing

jnlycklama:
## Description
We are responsible for calculating billing charges for: # api requests, bytes transcoded, and egress.

I've updated the middleware to add the necessary data to the response, so it can be used for billing in another middleware in workspace platform.

Code under Api/Features/ByteCounter is from health-pass, and is used by FHIR to calculate their egress. I don't know if there's a better way to consume that code other than copying it.

## Related issues
Addresses 85307

## Testing
Ran project locally & unit tested

### DIFF
--- a/src/Microsoft.Health.Dicom.Api/Features/ByteCounter/ByteCountingStream.cs
+++ b/src/Microsoft.Health.Dicom.Api/Features/ByteCounter/ByteCountingStream.cs
@@ -1,0 +1,134 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using EnsureThat;
+
+namespace Microsoft.Health.Dicom.Api.Features.ByteCounter
+{
+    /// <summary>
+    /// A stream to wrap an underlying stream and counts the number of bytes passed while operating with it.
+    /// </summary>
+    public class ByteCountingStream : Stream
+    {
+        private readonly Stream _stream;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ByteCountingStream"/> class.
+        /// </summary>
+        /// <param name="stream">An underlying stream.</param>
+        public ByteCountingStream(Stream stream)
+        {
+            EnsureArg.IsNotNull(stream, nameof(stream));
+
+            _stream = stream;
+        }
+
+        /// <inheritdoc />
+        public override bool CanRead => _stream.CanRead;
+
+        /// <inheritdoc />
+        public override bool CanSeek => _stream.CanSeek;
+
+        /// <inheritdoc />
+        public override bool CanWrite => _stream.CanWrite;
+
+        /// <inheritdoc />
+        public override long Length => _stream.Length;
+
+        /// <inheritdoc />
+        public override long Position
+        {
+            get { return _stream.Position; }
+            set { _stream.Position = value; }
+        }
+
+        /// <summary>
+        /// The number of bytes that have been written.
+        /// </summary>
+        public long WrittenByteCount { get; private set; }
+
+        /// <inheritdoc />
+        public override void Flush()
+        {
+            _stream.Flush();
+        }
+
+        /// <inheritdoc />
+        public override Task FlushAsync(CancellationToken cancellationToken)
+        {
+            return _stream.FlushAsync(cancellationToken);
+        }
+
+        /// <inheritdoc />
+        public override int Read(byte[] buffer, int offset, int count)
+        {
+            return _stream.Read(buffer, offset, count);
+        }
+
+        /// <inheritdoc />
+        public override long Seek(long offset, SeekOrigin origin)
+        {
+            return _stream.Seek(offset, origin);
+        }
+
+        /// <inheritdoc />
+        public override void SetLength(long value)
+        {
+            _stream.SetLength(value);
+        }
+
+        /// <inheritdoc />
+        public override void WriteByte(byte value)
+        {
+            _stream.WriteByte(value);
+            WrittenByteCount++;
+        }
+
+        /// <inheritdoc />
+        public override void Write(byte[] buffer, int offset, int count)
+        {
+            _stream.Write(buffer, offset, count);
+            WrittenByteCount += count;
+        }
+
+        /// <summary>
+        /// Asynchronously writes a sequence of bytes to the current stream and advances the current position within this stream by the number of bytes written.
+        /// </summary>
+        /// <param name="buffer">The buffer to write data from.</param>
+        /// <param name="offset">The zero-based byte offset in buffer from which to begin copying bytes to the stream.</param>
+        /// <param name="count">The maximum number of bytes to write.</param>
+        /// <returns>A task that represents the asynchronous write operation.</returns>
+        public new Task WriteAsync(byte[] buffer, int offset, int count)
+        {
+            WrittenByteCount += count;
+            return _stream.WriteAsync(buffer, offset, count);
+        }
+
+        /// <inheritdoc />
+        public override Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            WrittenByteCount += count;
+            return _stream.WriteAsync(buffer, offset, count, cancellationToken);
+        }
+
+        /// <inheritdoc />
+        public override IAsyncResult BeginWrite(byte[] buffer, int offset, int count, AsyncCallback callback, object state)
+        {
+            WrittenByteCount += count;
+            return _stream.BeginWrite(buffer, offset, count, callback, state);
+        }
+
+        /// <inheritdoc />
+        protected override void Dispose(bool disposing)
+        {
+            _stream.Dispose();
+            base.Dispose(disposing);
+        }
+    }
+}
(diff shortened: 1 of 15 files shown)


src/Microsoft.Health.Dicom.Api/Features/ByteCounter/ByteCountingStream.cs
```diff
@@ -0,0 +1,134 @@
...
+        public override long Position
+        {
+            get { return _stream.Position; }
+            set { _stream.Position = value; }
+        }
+
+        /// <summary>
+        /// The number of bytes that have been written.
+        /// </summary>
+        public long WrittenByteCount { get; private set; }
```
brandonpollett, line 54: Do we need to worry about thread safety here?

src/Microsoft.Health.Dicom.Api/Features/Context/DicomRequestContextMiddleware.cs
```diff
@@ -51,8 +57,24 @@ public async Task Invoke(HttpContext context, IDicomRequestContextAccessor dicom
 
             dicomRequestContextAccessor.RequestContext = dicomRequestContext;
 
-            // Call the next delegate/middleware in the pipeline
-            await _next(context);
+            ByteCountingStream byteCountingStream = _responseLogStreamFactory.CreateByteCountingResponseLogStream(context.Response.Body);
```
brandonpollett, line 60: Do we need to dispose this?

src/Microsoft.Health.Dicom.Api/Features/Context/DicomRequestContextMiddleware.cs
```diff
@@ -51,8 +57,24 @@ public async Task Invoke(HttpContext context, IDicomRequestContextAccessor dicom
...
+            {
+                // Call the next delegate/middleware in the pipeline
+                await _next(context);
+            }
+            finally
+            {
+                long responseBodySize = byteCountingStream.WrittenByteCount;
+                long responseHeaderSize = HeaderUtility.GetTotalHeaderLength(context.Response.Headers);
+                long totalResponseSize = responseBodySize + responseHeaderSize;
+                context.Items["ResponseSize"] = totalResponseSize;
```
brandonpollett, line 73: Should we use a constant for the item keys?

src/Microsoft.Health.Dicom.Api/Features/Context/DicomRequestContextMiddleware.cs
```diff
@@ -51,8 +57,24 @@ public async Task Invoke(HttpContext context, IDicomRequestContextAccessor dicom
...
+
+                context.Items["TranscodedSize"] = dicomRequestContextAccessor.RequestContext.BytesTranscoded;
+                context.Items["IsTranscodeRequested"] = dicomRequestContextAccessor.RequestContext.IsTranscodeRequested;
```
brandonpollett, line 76: Any particular reason to put these in a property bag vs the `dicomRequestContext`?

jnlycklama: I thought about using the dicomRequestContext, but I left it directly on context only so I wouldn't need to pull in dicomRequestContext in workspaceplatform. I could go either way though

src/Microsoft.Health.Dicom.Api/Features/ByteCounter/ByteCountingStream.cs
```diff
@@ -0,0 +1,134 @@
+﻿// -------------------------------------------------------------------------------------------------
```
jovinson-ms, line 1: It does seem that the ByteCounter code would belong in the healthcare-shared-components repo - seems pretty self-contained.

src/Microsoft.Health.Dicom.Api/Features/Context/DicomRequestContextMiddleware.cs
brandonpollett: Kinda depends on how it gets coded on the consuming side, but if it's just additional middleware in the same program, then it already has access to the `DicomRequestContext` through the `DicomRequestContextAccessor`

src/Microsoft.Health.Dicom.Api/Features/ByteCounter/ByteCountingStream.cs
jnlycklama: Yeah that makes sense, especially since this code is originally from FHIR, so it should be sharable. Created PR https://github.com/microsoft/healthcare-shared-components/pull/314
& will remove the files here

src/Microsoft.Health.Dicom.Api/Features/ByteCounter/ByteCountingStream.cs
jnlycklama: Hm yeah seems like we would need to worry about thread safety. I could update it to only be incremented by Interlocked.Add or Interlocked.Increment

Although this code is used in health-paas https://microsofthealth.visualstudio.com/Health/_git/health-paas?path=/src/Common/ServicePlatform/ByteCounting/ByteCountingStream.cs&_a=contents&version=GBmaster
So I'm curious why they didn't add anything like that. Maybe I should follow up with them.

Also moving this to healthcare-shared-components per josiahs suggestion - https://github.com/microsoft/healthcare-shared-components/pull/314

jnlycklama: Followed up with Eric, who originally wrote this file. He said there was no need for them to keep it thread safe but he sees no reason why we can't. 
I've tested & updated it in the other PR. I'll resolve this thread and we can continue the discussion in the other PR if necessary

src/Microsoft.Health.Dicom.Api/Features/Context/DicomRequestContextMiddleware.cs
jnlycklama: Makes sense. BytesTranscoded and IsTranscodeRequest are already on there too so I can see how using DicomRequestContext is better. I've updated the code

src/Microsoft.Health.Dicom.Api/Features/Context/DicomRequestContextMiddleware.cs
jnlycklama: Removed

src/Microsoft.Health.Dicom.Api/Features/Context/DicomRequestContextMiddleware.cs
brandonpollett: I couldn't find the actual documentation, but I found at least one example where this is done with a `using` - https://weblog.west-wind.com/posts/2020/Mar/29/Content-Injection-with-Response-Rewriting-in-ASPNET-Core-3x#aspnet-core-3x-response-rewriting-with-ihttpresponsestreamfeature

src/Microsoft.Health.Dicom.Api/Features/ByteCounter/IResponseLogStreamFactory.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace Microsoft.Health.Dicom.Api.Features.ByteCounter
+{
+    public interface IResponseLogStreamFactory
+    {
+        /// <summary>
+        /// Creates a <see cref="ByteCountingStream"/>
+        /// </summary>
+        /// <param name="stream">An underlying stream.</param>
+        /// <returns>A byte counting stream that wraps the specified underlying stream.</returns>
+        ByteCountingStream CreateByteCountingResponseLogStream(Stream stream);
```
brandonpollett, line 17: Any reason we're deviating from the factory pattern we have elsewhere in the code base? [IGuidFactory](https://github.com/microsoft/dicom-server/blob/main/src/Microsoft.Health.Dicom.Functions/Durable/IGuidFactory.cs) or [IMultiPartReadFactory](https://github.com/microsoft/dicom-server/blob/main/src/Microsoft.Health.Dicom.Core/Web/IMultipartReaderFactory.cs)? It seems like this would be called a `IByteCountingStreamFactory` with a method `IByteCountingStream Create(Stream stream)`

src/Microsoft.Health.Dicom.Api/Features/Context/DicomRequestContextMiddleware.cs
```diff
@@ -51,8 +57,24 @@ public async Task Invoke(HttpContext context, IDicomRequestContextAccessor dicom
...
+                await _next(context);
+            }
+            finally
+            {
+                long responseBodySize = byteCountingStream.WrittenByteCount;
+                long responseHeaderSize = HeaderUtility.GetTotalHeaderLength(context.Response.Headers);
+                long totalResponseSize = responseBodySize + responseHeaderSize;
+
+                dicomRequestContext.ResponseSize = totalResponseSize;
+                dicomRequestContextAccessor.RequestContext = dicomRequestContext;
```
brandonpollett, line 76: Could you just do `dicomRequestContextAccessor.RequestContext.ResponseSize = totalResponseSize`?

src/Microsoft.Health.Dicom.Api/Features/Context/DicomRequestContextMiddleware.cs
jnlycklama: Ah thanks. I was worried that disposing it would dispose the response before it was meant to be disposed. Updated.

src/Microsoft.Health.Dicom.Api/Features/ByteCounter/IResponseLogStreamFactory.cs
jnlycklama: I was just keeping the same names used for fhir but this makes sense. I made the rename changes in this PR if you could take a look: https://github.com/microsoft/healthcare-shared-components/pull/314

I will delete the ByteCounter/ files from this PR once that is merged in and I can reference those files from here

jnlycklama: Code has been checked in to healthcare-shared-components and deleted here.

PR #1049, opened 2021-09-21
DB Changes - Introducing partitionId

bcarthic:
## Description

## Related issues
Addresses [issue #].

## Testing

### DIFF
--- a/src/Microsoft.Health.Dicom.SqlServer/Features/ChangeFeed/SqlChangeFeedStoreV1.cs
+++ b/src/Microsoft.Health.Dicom.SqlServer/Features/ChangeFeed/SqlChangeFeedStoreV1.cs
@@ -1,0 +1,119 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Data;
+using System.Threading;
+using System.Threading.Tasks;
+using EnsureThat;
+using Microsoft.Data.SqlClient;
+using Microsoft.Health.Dicom.Core.Features.ChangeFeed;
+using Microsoft.Health.Dicom.SqlServer.Features.Schema;
+using Microsoft.Health.Dicom.SqlServer.Features.Schema.Model;
+using Microsoft.Health.SqlServer.Features.Client;
+using Microsoft.Health.SqlServer.Features.Storage;
+
+namespace Microsoft.Health.Dicom.SqlServer.Features.ChangeFeed
+{
+    public class SqlChangeFeedStoreV1 : ISqlChangeFeedStore
+    {
+        public virtual SchemaVersion Version => SchemaVersion.V1;
+
+        protected SqlConnectionWrapperFactory SqlConnectionWrapperFactory { get; }
+
+        public SqlChangeFeedStoreV1(SqlConnectionWrapperFactory sqlConnectionWrapperFactory)
+        {
+            SqlConnectionWrapperFactory = EnsureArg.IsNotNull(sqlConnectionWrapperFactory, nameof(sqlConnectionWrapperFactory));
+        }
+
+        public virtual async Task<ChangeFeedEntry> GetChangeFeedLatestAsync(CancellationToken cancellationToken)
+        {
+            using SqlConnectionWrapper sqlConnectionWrapper = await SqlConnectionWrapperFactory.ObtainSqlConnectionWrapperAsync(cancellationToken);
+            using SqlCommandWrapper sqlCommandWrapper = sqlConnectionWrapper.CreateSqlCommand();
+
+            V4.GetChangeFeedLatest.PopulateCommand(sqlCommandWrapper);
+
+            using SqlDataReader reader = await sqlCommandWrapper.ExecuteReaderAsync(CommandBehavior.SequentialAccess, cancellationToken);
+            if (await reader.ReadAsync(cancellationToken))
+            {
+                (long rSeq, DateTimeOffset rTimestamp, int rAction, string rStudyInstanceUid, string rSeriesInstanceUid, string rSopInstanceUid, long oWatermark, long? cWatermark) = reader.ReadRow(
+                    V4.ChangeFeed.Sequence,
+                    V4.ChangeFeed.Timestamp,
+                    V4.ChangeFeed.Action,
+                    V4.ChangeFeed.StudyInstanceUid,
+                    V4.ChangeFeed.SeriesInstanceUid,
+                    V4.ChangeFeed.SopInstanceUid,
+                    V4.ChangeFeed.OriginalWatermark,
+                    V4.ChangeFeed.CurrentWatermark);
+
+                return new ChangeFeedEntry(
+                        rSeq,
+                        rTimestamp,
+                        (ChangeFeedAction)rAction,
+                        rStudyInstanceUid,
+                        rSeriesInstanceUid,
+                        rSopInstanceUid,
+                        oWatermark,
+                        cWatermark,
+                        ConvertWatermarkToCurrentState(oWatermark, cWatermark));
+            }
+
+            return null;
+        }
+
+        public virtual async Task<IReadOnlyCollection<ChangeFeedEntry>> GetChangeFeedAsync(long offset, int limit, CancellationToken cancellationToken)
+        {
+            var results = new List<ChangeFeedEntry>();
+
+            using SqlConnectionWrapper sqlConnectionWrapper = await SqlConnectionWrapperFactory.ObtainSqlConnectionWrapperAsync(cancellationToken);
+            using SqlCommandWrapper sqlCommandWrapper = sqlConnectionWrapper.CreateSqlCommand();
+
+            V4.GetChangeFeed.PopulateCommand(sqlCommandWrapper, limit, offset);
+
+            using SqlDataReader reader = await sqlCommandWrapper.ExecuteReaderAsync(CommandBehavior.SequentialAccess, cancellationToken);
+            while (await reader.ReadAsync(cancellationToken))
+            {
+                (long rSeq, DateTimeOffset rTimestamp, int rAction, string rStudyInstanceUid, string rSeriesInstanceUid, string rSopInstanceUid, long oWatermark, long? cWatermark) = reader.ReadRow(
+                    V4.ChangeFeed.Sequence,
+                    V4.ChangeFeed.Timestamp,
+                    V4.ChangeFeed.Action,
+                    V4.ChangeFeed.StudyInstanceUid,
+                    V4.ChangeFeed.SeriesInstanceUid,
+                    V4.ChangeFeed.SopInstanceUid,
+                    V4.ChangeFeed.OriginalWatermark,
+                    V4.ChangeFeed.CurrentWatermark);
+
+                results.Add(new ChangeFeedEntry(
+                        rSeq,
+                        rTimestamp,
+                        (ChangeFeedAction)rAction,
+                        rStudyInstanceUid,
+                        rSeriesInstanceUid,
+                        rSopInstanceUid,
+                        oWatermark,
+                        cWatermark,
+                        ConvertWatermarkToCurrentState(oWatermark, cWatermark)));
+            }
+
+            return results;
+        }
+
+        protected static ChangeFeedState ConvertWatermarkToCurrentState(long originalWatermark, long? currentWatermak)
+        {
+            if (currentWatermak == null)
+            {
+                return ChangeFeedState.Deleted;
+            }
+
+            if (currentWatermak != originalWatermark)
+            {
+                return ChangeFeedState.Replaced;
+            }
+
+            return ChangeFeedState.Current;
+        }
+    }
+}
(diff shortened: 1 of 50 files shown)


src/Microsoft.Health.Dicom.SqlServer/Features/ChangeFeed/SqlChangeFeedStoreV1.cs
```diff
@@ -0,0 +1,119 @@
...
+using Microsoft.Data.SqlClient;
+using Microsoft.Health.Dicom.Core.Features.ChangeFeed;
+using Microsoft.Health.Dicom.SqlServer.Features.Schema;
+using Microsoft.Health.Dicom.SqlServer.Features.Schema.Model;
+using Microsoft.Health.SqlServer.Features.Client;
+using Microsoft.Health.SqlServer.Features.Storage;
+
+namespace Microsoft.Health.Dicom.SqlServer.Features.ChangeFeed
+{
+    public class SqlChangeFeedStoreV1 : ISqlChangeFeedStore
```
smithago, line 21: The supported Schema is V3-v4 before. Why do we need to implement all the historical versions?

Can we reuse classes for versions that are not changing anything?

This smells to me

src/Microsoft.Health.Dicom.SqlServer/Features/Partition/SqlPartitionStoreV1.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Health.Dicom.Core.Exceptions;
+using Microsoft.Health.Dicom.Core.Features.Partition;
+using Microsoft.Health.Dicom.SqlServer.Features.Schema;
+
+namespace Microsoft.Health.Dicom.SqlServer.Features.Partition
+{
+    internal class SqlPartitionStoreV1 : ISqlPartitionStore
```
smithago, line 15: same here

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.diff.sql
```diff
@@ -0,0 +1,1613 @@
...
+
+---BEGIN - CREATE/ALTER Statements----------
+
+IF NOT EXISTS (
+    SELECT *
+    FROM sys.tables
+    WHERE name = 'Partition')
+BEGIN
+    CREATE TABLE dbo.Partition (
+        PartitionKey            BIGINT                 NOT NULL, --PK
```
smithago, line 16: nit: should we had some comments on difference between PartitionKey and PartitionId

smithago: Also is BIGINT necessary? Maybe INT? Will save four bytes on each row

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.diff.sql
```diff
@@ -0,0 +1,1613 @@
...
+        PartitionId             VARCHAR(64)            NOT NULL    DEFAULT 'Microsoft.Default',
```
smithago, line 17: Should we control the default from Compute. Here it does not fully understand we to default and when to error

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.diff.sql
```diff
@@ -0,0 +1,1613 @@
...
+GO
+
+IF NOT EXISTS (
+    SELECT *
+    FROM sys.indexes
+    WHERE name='IXC_Partition_PartitionKey_PartitionId' AND object_id = OBJECT_ID('dbo.Partition'))
+BEGIN
+    CREATE UNIQUE NONCLUSTERED INDEX IXC_Partition_PartitionKey_PartitionId ON dbo.Partition
+    (
+        PartitionKey,
```
smithago, line 46: Both the indexes are same, You are filtering on PartitionKey first. 

Don't you need a index to find the Key based on used specified PartitionId

smithago: You can also use to enforce Unique PartitionId

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.diff.sql
smithago: Also see if another name is less confusing

bcarthic: When you say another naming, you mean to change `Microsoft.Default` ?

bcarthic: Also thinking twice we dont need a default here since we are already handling by inserting Microsoft.Default always.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.diff.sql
bcarthic: Sure, INT makes sense

src/Microsoft.Health.Dicom.SqlServer/Features/ChangeFeed/SqlChangeFeedStoreV1.cs
bcarthic: Is that possible that someone will be in V1 or V2. Should we migrate them?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.diff.sql
```diff
@@ -0,0 +1,1613 @@
...
+    SET @newWatermark = NEXT VALUE FOR dbo.WatermarkSequence
+    SET @instanceKey = NEXT VALUE FOR dbo.InstanceKeySequence
+
+    -- Insert Study
+    SELECT @studyKey = StudyKey
+    FROM dbo.Study WITH(UPDLOCK)
+    WHERE StudyInstanceUid = @studyInstanceUid
+    AND PartitionKey = @partitionKey
+
+    IF @@ROWCOUNT = 0
```
smithago, line 241: It is very hard to review this file. I will see if 4.sql diff with 5.sql is better.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.diff.sql
```diff
@@ -0,0 +1,1613 @@
...
+
+    -- The instance does not exist, insert it.
+    SET @newWatermark = NEXT VALUE FOR dbo.WatermarkSequence
+    SET @instanceKey = NEXT VALUE FOR dbo.InstanceKeySequence
+
+    -- Insert Study
+    SELECT @studyKey = StudyKey
+    FROM dbo.Study WITH(UPDLOCK)
+    WHERE StudyInstanceUid = @studyInstanceUid
+    AND PartitionKey = @partitionKey
```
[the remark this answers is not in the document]
smithago: How did you test your changes

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.sql
```diff
@@ -0,0 +1,2803 @@
...
+    SopInstanceUid
+)
+INCLUDE
+(
+    Status,
+    Watermark
+)
+WITH (DATA_COMPRESSION = PAGE)
+
+CREATE NONCLUSTERED INDEX IX_Instance_StudyInstanceUid_Status on dbo.Instance
```
smithago, line 94: Why does not this index have partitionKey?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.sql
```diff
@@ -0,0 +1,2803 @@
...
+CREATE TABLE dbo.ExtendedQueryTagString (
+    TagKey                  INT                  NOT NULL, --PK
+    TagValue                NVARCHAR(64)         NOT NULL,
+    StudyKey                BIGINT               NOT NULL, --FK
+    SeriesKey               BIGINT               NULL,     --FK
+    InstanceKey             BIGINT               NULL,     --FK
+    Watermark               BIGINT               NOT NULL
+) WITH (DATA_COMPRESSION = PAGE)
+
+CREATE UNIQUE CLUSTERED INDEX IXC_ExtendedQueryTagString ON dbo.ExtendedQueryTagString
```
smithago, line 520: Same here, how will QIDO work?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.sql
```diff
@@ -0,0 +1,2803 @@
...
+--         * Double extended query tag data
+--     @dateTimeExtendedQueryTags
+--         * DateTime extended query tag data
+--     @personNameExtendedQueryTags
+--         * PersonName extended query tag data
+-- RETURN VALUE
+--     The watermark (version).
+------------------------------------------------------------------------
+CREATE OR ALTER PROCEDURE dbo.AddInstance
+    @partitionId                        VARCHAR(64)    = 'Microsoft.Default',
```
smithago, line 810: Do we need default here? We should be explicit and send it from compute layer

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.sql
```diff
@@ -0,0 +1,2803 @@
...
+
+    SELECT @partitionKey = PartitionKey
+    FROM dbo.Partition WITH(UPDLOCK)
+    WHERE PartitionId = @partitionId
+
+    IF @@ROWCOUNT = 0
+    BEGIN
+        SET @partitionKey = NEXT VALUE FOR dbo.PartitionKeySequence
+
+        INSERT INTO dbo.Partition
```
smithago, line 852: Can we do this in a separate procedure? Before calling AddInstance and pass in a PartitionKey instead?

src/Microsoft.Health.Dicom.SqlServer/Features/ChangeFeed/SqlChangeFeedStoreV1.cs
jovinson-ms: We were trying to follow the existing pattern but agree these files are not necessary. We will address this in a future refactor.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.diff.sql
jovinson-ms: Added comments and changed the type to INT.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.diff.sql
jovinson-ms: Moving default values to compute and changing the column name to `PartitionName` to make the usage clearer.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.diff.sql
jovinson-ms: Changing the index on `PartitionName` to only index that column.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.diff.sql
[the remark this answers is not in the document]
jovinson-ms: Regarding ensuring `PartitionName` being present for every operation - that is not currently the case since we made the decision to only require partition to be specified on STOW, WADO, QIDO, and delete. There are many queries related to change feed and extended query tags, for example, which are not affected.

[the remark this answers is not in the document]
jovinson-ms: We tested a clean install with and without the partition feature enabled, and upgrade from the previous version with and without the partition feature enabled. We ran the SQL integration and E2E test suite on each variation.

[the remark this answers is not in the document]
jovinson-ms: After conversation, we need to consider performance testing since the changes are extensive. We can do ad-hoc manual testing, and potentially invest in benchmarking tests.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.sql
```diff
@@ -0,0 +1,2803 @@
...
+    PartitionKey
+)
+
+CREATE UNIQUE NONCLUSTERED INDEX IXC_Partition_PartitionKey_PartitionId on dbo.Partition
+(
+    PartitionKey,
+    PartitionId
+)
+
+IF NOT EXISTS (SELECT 1 FROM dbo.Partition WHERE PartitionId = 'Microsoft.Default')
```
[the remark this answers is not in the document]
jovinson-ms: Removed.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.sql
jovinson-ms: For the Instance table, the new version of the schema will not add any reference to partition. This is because the table already has a clustered index on `StudyKey`, which is partition-specific.

PR #1052, opened 2021-09-22
Refactor QIDO Input to Rely on ASP.NET Core for Parsing

wsugarman:
## Description
- Use ASP.NET Core model binding for parsing the following:
  - `fuzzyMatching`
  - `includeField`
  - `limit`
  - `offset`
- Remove usage of `StringValues` and data cleaning through `Trim()` from Core layer; check for duplicate strings ahead-of-time
- Avoid passing Mediatr request directly to `IQueryTagService`
- Create generic `CsvModelBinder` for parsing comma-separated values
- Refactor `QueryIncludeField` with static property representing all fields

## Related issues
N/A

## Testing
Test pipeline and ad-hoc


src/Microsoft.Health.Dicom.Api/Features/ModelBinders/CsvModelBinder.cs
```diff
@@ -0,0 +1,41 @@
...
+
+using System;
+using System.Threading.Tasks;
+using EnsureThat;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Health.Dicom.Api.Features.ModelBinders
+{
+    internal class CsvModelBinder : IModelBinder
```
wsugarman, line 14: Need tests

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.cs
```diff
@@ -68,63 +57,54 @@ public QueryParser(IDicomTagParser dicomTagPathParser)
...
-                _parsedQuery.FuzzyMatch,
-                _parsedQuery.Limit,
-                _parsedQuery.Offset,
-                _parsedQuery.FilterConditions);
+                parameters.QueryResourceType,
+                ParseIncludeFields(parameters.IncludeField),
+                parameters.FuzzyMatching,
+                parameters.Limit,
+                parameters.Offset,
+                _parsedQuery.FilterConditions.Values);
```
wsugarman, line 107: I don't think the order of the filters matter, do they?

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.cs
```diff
@@ -231,26 +192,35 @@ private static QueryTag GetSupportedQueryTag(DicomTag dicomTag, string attribute
             return queryTag;
         }
 
-        private class QueryExpressionImp
+        private QueryIncludeField ParseIncludeFields(IReadOnlyList<string> includeFields)
         {
-            public QueryExpressionImp()
+            // Check if `all` is present as one of the values in IncludeField parameter.
```
wsugarman, line 197: omg, I tried to use markdown lol

src/Microsoft.Health.Dicom.Api/Features/ModelBinders/CsvModelBinder.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        protected virtual object DefaultValue { get; } = Array.Empty<string>();
+
+        public Task BindModelAsync(ModelBindingContext bindingContext)
```
pengchen0692, line 18: I find when binding failed, execution doesn't fail out but continue.

src/Microsoft.Health.Dicom.Api/Features/ModelBinders/CsvModelBinder.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            EnsureArg.IsNotNull(bindingContext, nameof(bindingContext));
+            StringValues values = bindingContext.ValueProvider.GetValue(bindingContext.ModelName).Values;
+
+            if (values.Count == 0)
+            {
+                bindingContext.Result = ModelBindingResult.Success(DefaultValue);
+            }
+            else if (values.Count > 1)
```
pengchen0692, line 27: We previously support this scenario, this could cause regression issue. 
I don't think such request (`/studies?includeField=PatientName&IncludeField=PatientAge`) is very common, but it's better to still support.

pengchen0692: https://github.com/microsoft/dicom-server/blob/b8e8ec5202b1778415e05d9f4c3397e47a9d7695/src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.ParameterParser.cs#L18

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.cs
```diff
@@ -25,106 +23,85 @@ public partial class QueryParser : IQueryParser
...
 
         private const string IncludeFieldValueAll = "all";
-        private const StringComparison QueryParameterComparision = StringComparison.OrdinalIgnoreCase;
-        private QueryExpressionImp _parsedQuery;
-        private readonly Dictionary<string, Action<KeyValuePair<string, StringValues>>> _paramParsers =
-            new Dictionary<string, Action<KeyValuePair<string, StringValues>>>(StringComparer.OrdinalIgnoreCase);
 
-        private readonly Dictionary<DicomVR, Func<QueryTag, string, QueryFilterCondition>> _valueParsers =
-            new Dictionary<DicomVR, Func<QueryTag, string, QueryFilterCondition>>();
+        private readonly static Dictionary<DicomVR, Func<QueryTag, string, QueryFilterCondition>> ValueParsers = new Dictionary<DicomVR, Func<QueryTag, string, QueryFilterCondition>>
```
pengchen0692, line 27: ValueParsers could be move out of this class to further simplify.

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.cs
```diff
@@ -25,106 +23,85 @@ public partial class QueryParser : IQueryParser
         private readonly IDicomTagParser _dicomTagPathParser;
```
pengchen0692, line 23: `IDicomTagParser` could be make as static, and remove supporting sequence dicom tag  to further simplify

pengchen0692: I can do this in a separated PR

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.cs
pengchen0692: I can do this in a separated PR

src/Microsoft.Health.Dicom.Api/Features/ModelBinders/CsvModelBinder.cs
wsugarman: Fixed by added errors

src/Microsoft.Health.Dicom.Api/Features/ModelBinders/CsvModelBinder.cs
wsugarman: This does strangely seem allowed by the DICOMweb standard: http://dicom.nema.org/medical/dicom/current/output/html/part18.html#sect_8.3.4.3

> The request may contain one or more include parameters; however, if a parameter with the value of "all" is present, then other includefield parameters shall not be present.

I also now throw an error if "all" is specify with other fields

src/Microsoft.Health.Dicom.Api/DicomApiResource.Designer.cs
```diff
@@ -78,6 +78,15 @@ internal class DicomApiResource {
             }
         }
         
+        /// <summary>
+        ///   Looks up a localized string similar to Cannot specify value more than once..
```
pengchen0692, line 82: In the context, multiple values in `StringValues` means the query key is specified multiple times in query string, so `DuplicateValues` may not be a good present.
How about `DuplicateQueryKey`? also include key name so that customer can easily identify the problematic key.

src/Microsoft.Health.Dicom.Api/Features/Filters/QueryModelStateValidatorAttribute.cs
```diff
@@ -17,7 +18,8 @@ public override void OnActionExecuting(ActionExecutingContext context)
             EnsureArg.IsNotNull(context, nameof(context));
             if (!context.ModelState.IsValid)
             {
-                throw new InvalidQueryStringValuesException(context.ModelState.FirstOrDefault(x => x.Value.Errors.Any()).Key);
+                (string key, ModelStateEntry value) = context.ModelState.Where(x => x.Value.Errors.Count > 0).First();
+                throw new InvalidQueryStringValuesException(key, value.Errors[0].ErrorMessage);
```
pengchen0692, line 22: I like this change, it make result more clear to customer.

pengchen0692 (approved):

src/Microsoft.Health.Dicom.Api/DicomApiResource.Designer.cs
wsugarman: That makes sense

src/Microsoft.Health.Dicom.Api/Features/Filters/QueryModelStateValidatorAttribute.cs
wsugarman: I see you did something similar haha

PR #1057, opened 2021-09-22
[DRAFT][PLEASE NOT REVIEW]Add API to update extended query tag

pengchen0692:
## Description
Add API to update extended query tag

## Related issues
Addresses AB#84773

## Testing
All tests pass

### DIFF
--- a/src/Microsoft.Health.Dicom.Api/Controllers/ExtendedQueryTagController.cs
+++ b/src/Microsoft.Health.Dicom.Api/Controllers/ExtendedQueryTagController.cs
@@ -179,7 +179,7 @@ namespace Microsoft.Health.Dicom.Api.Controllers
             return StatusCode((int)HttpStatusCode.OK, response.ExtendedQueryTagErrors);
         }
 
-        [HttpGet]
+        [HttpPatch]
         [Produces(KnownContentTypes.ApplicationJson)]
         [ProducesResponseType(typeof(IEnumerable<ExtendedQueryTagError>), (int)HttpStatusCode.OK)]
         [ProducesResponseType(typeof(string), (int)HttpStatusCode.BadRequest)]
@@ -187,12 +187,12 @@ namespace Microsoft.Health.Dicom.Api.Controllers
         [VersionedRoute(KnownRoutes.UpdateExtendedQueryTagQueryStatusRoute, Name = KnownRouteNames.VersionedUpdateExtendedQueryTagQueryStatus)]
         [Route(KnownRoutes.UpdateExtendedQueryTagQueryStatusRoute, Name = KnownRouteNames.UpdateExtendedQueryTagQueryStatus)]
         [AuditEventType(AuditEventSubType.UpdateExtendedQueryTagQueryStatus)]
-        public async Task<IActionResult> UpdateTagQueryStatusAsync([FromRoute] string tagPath, QueryStatus queryStatus)
+        public async Task<IActionResult> UpdateTagQueryStatusAsync([FromRoute] string tagPath, [Required][FromBody] UpdateExtendedQueryTagEntry updateEntry)
         {
-            _logger.LogInformation("DICOM Web Update Extended Query Tag Query Status request received for extended query tag {tagPath} and query status {queryStatus}", tagPath, queryStatus);
+            _logger.LogInformation("DICOM Web Update Extended Query Tag Query Status request received for extended query tag {tagPath} and new value {updateEntry}", tagPath, updateEntry);
 
             EnsureFeatureIsEnabled();
-            var response = await _mediator.UpdateExtendedQueryTagQueryStatusAsync(tagPath, queryStatus, HttpContext.RequestAborted);
+            var response = await _mediator.UpdateExtendedQueryTagAsync(tagPath, updateEntry, HttpContext.RequestAborted);
 
             return StatusCode((int)HttpStatusCode.OK, response.TagEntry);
         }
--- a/src/Microsoft.Health.Dicom.Api/Controllers/ExtendedQueryTagController.cs
+++ b/src/Microsoft.Health.Dicom.Api/Controllers/ExtendedQueryTagController.cs
@@ -179,6 +179,24 @@ namespace Microsoft.Health.Dicom.Api.Controllers
             return StatusCode((int)HttpStatusCode.OK, response.ExtendedQueryTagErrors);
         }
 
+        [HttpGet]
+        [Produces(KnownContentTypes.ApplicationJson)]
+        [ProducesResponseType(typeof(IEnumerable<ExtendedQueryTagError>), (int)HttpStatusCode.OK)]
+        [ProducesResponseType(typeof(string), (int)HttpStatusCode.BadRequest)]
+        [ProducesResponseType(typeof(string), (int)HttpStatusCode.NotFound)]
+        [VersionedRoute(KnownRoutes.UpdateTagQueryStatusRoute, Name = KnownRouteNames.VersionedUpdateTagQueryStatus)]
+        [Route(KnownRoutes.UpdateTagQueryStatusRoute, Name = KnownRouteNames.UpdateTagQueryStatus)]
+        [AuditEventType(AuditEventSubType.UpdateTagQueryStatus)]
+        public async Task<IActionResult> UpdateTagQueryStatusAsync([FromRoute] string tagPath, QueryStatus queryStatus)
+        {
+            _logger.LogInformation("DICOM Web Update Extended Query Tag Query Status request received for extended query tag {tagPath} and query status {queryStatus}", tagPath, queryStatus);
+
+            EnsureFeatureIsEnabled();
+            var response = await _mediator.UpdateTagQueryStatusAsync(tagPath, queryStatus, HttpContext.RequestAborted);
+
+            return StatusCode((int)HttpStatusCode.OK, response.TagEntry);
+        }
+
         private void EnsureFeatureIsEnabled()
         {
             if (!_featureEnabled)
--- a/src/Microsoft.Health.Dicom.Api/Controllers/ExtendedQueryTagController.cs
+++ b/src/Microsoft.Health.Dicom.Api/Controllers/ExtendedQueryTagController.cs
@@ -184,15 +184,15 @@ namespace Microsoft.Health.Dicom.Api.Controllers
         [ProducesResponseType(typeof(IEnumerable<ExtendedQueryTagError>), (int)HttpStatusCode.OK)]
         [ProducesResponseType(typeof(string), (int)HttpStatusCode.BadRequest)]
         [ProducesResponseType(typeof(string), (int)HttpStatusCode.NotFound)]
-        [VersionedRoute(KnownRoutes.UpdateTagQueryStatusRoute, Name = KnownRouteNames.VersionedUpdateTagQueryStatus)]
-        [Route(KnownRoutes.UpdateTagQueryStatusRoute, Name = KnownRouteNames.UpdateTagQueryStatus)]
-        [AuditEventType(AuditEventSubType.UpdateTagQueryStatus)]
+        [VersionedRoute(KnownRoutes.UpdateExtendedQueryTagQueryStatusRoute, Name = KnownRouteNames.VersionedUpdateExtendedQueryTagQueryStatus)]
+        [Route(KnownRoutes.UpdateExtendedQueryTagQueryStatusRoute, Name = KnownRouteNames.UpdateExtendedQueryTagQueryStatus)]
+        [AuditEventType(AuditEventSubType.UpdateExtendedQueryTagQueryStatus)]
         public async Task<IActionResult> UpdateTagQueryStatusAsync([FromRoute] string tagPath, QueryStatus queryStatus)
         {
             _logger.LogInformation("DICOM Web Update Extended Query Tag Query Status request received for extended query tag {tagPath} and query status {queryStatus}", tagPath, queryStatus);
 
             EnsureFeatureIsEnabled();
-            var response = await _mediator.UpdateTagQueryStatusAsync(tagPath, queryStatus, HttpContext.RequestAborted);
+            var response = await _mediator.UpdateExtendedQueryTagQueryStatusAsync(tagPath, queryStatus, HttpContext.RequestAborted);
 
             return StatusCode((int)HttpStatusCode.OK, response.TagEntry);
         }
--- a/src/Microsoft.Health.Dicom.Api/Controllers/ExtendedQueryTagController.cs
+++ b/src/Microsoft.Health.Dicom.Api/Controllers/ExtendedQueryTagController.cs
@@ -181,18 +181,19 @@ namespace Microsoft.Health.Dicom.Api.Controllers
 
         [HttpPatch]
         [Produces(KnownContentTypes.ApplicationJson)]
+        [BodyModelStateValidator]
         [ProducesResponseType(typeof(IEnumerable<ExtendedQueryTagError>), (int)HttpStatusCode.OK)]
         [ProducesResponseType(typeof(string), (int)HttpStatusCode.BadRequest)]
         [ProducesResponseType(typeof(string), (int)HttpStatusCode.NotFound)]
         [VersionedRoute(KnownRoutes.UpdateExtendedQueryTagQueryStatusRoute, Name = KnownRouteNames.VersionedUpdateExtendedQueryTagQueryStatus)]
         [Route(KnownRoutes.UpdateExtendedQueryTagQueryStatusRoute, Name = KnownRouteNames.UpdateExtendedQueryTagQueryStatus)]
         [AuditEventType(AuditEventSubType.UpdateExtendedQueryTagQueryStatus)]
-        public async Task<IActionResult> UpdateTagQueryStatusAsync([FromRoute] string tagPath, [Required][FromBody] UpdateExtendedQueryTagEntry updateEntry)
+        public async Task<IActionResult> UpdateTagAsync([FromRoute] string tagPath, [Required][FromBody] UpdateExtendedQueryTagEntry newValue)
         {
-            _logger.LogInformation("DICOM Web Update Extended Query Tag Query Status request received for extended query tag {tagPath} and new value {updateEntry}", tagPath, updateEntry);
+            _logger.LogInformation("DICOM Web Update Extended Query Tag Query Status request received for extended query tag {tagPath} and new value {newValue}", tagPath, newValue);
 
             EnsureFeatureIsEnabled();
-            var response = await _mediator.UpdateExtendedQueryTagAsync(tagPath, updateEntry, HttpContext.RequestAborted);
+            var response = await _mediator.UpdateExtendedQueryTagAsync(tagPath, newValue, HttpContext.RequestAborted);
 
             return StatusCode((int)HttpStatusCode.OK, response.TagEntry);
         }
--- a/src/Microsoft.Health.Dicom.Api/Controllers/ExtendedQueryTagController.cs
+++ b/src/Microsoft.Health.Dicom.Api/Controllers/ExtendedQueryTagController.cs
@@ -188,7 +188,7 @@ namespace Microsoft.Health.Dicom.Api.Controllers
         [VersionedRoute(KnownRoutes.UpdateExtendedQueryTagQueryStatusRoute, Name = KnownRouteNames.VersionedUpdateExtendedQueryTagQueryStatus)]
         [Route(KnownRoutes.UpdateExtendedQueryTagQueryStatusRoute, Name = KnownRouteNames.UpdateExtendedQueryTagQueryStatus)]
         [AuditEventType(AuditEventSubType.UpdateExtendedQueryTagQueryStatus)]
-        public async Task<IActionResult> UpdateTagAsync([FromRoute] string tagPath, [Required][FromBody] UpdateExtendedQueryTagEntry newValue)
+        public async Task<IActionResult> UpdateTagAsync([FromRoute] string tagPath, [FromBody] UpdateExtendedQueryTagEntry newValue)
         {
             _logger.LogInformation("DICOM Web Update Extended Query Tag Query Status request received for extended query tag {tagPath} and new value {newValue}", tagPath, newValue);
 
--- a/src/Microsoft.Health.Dicom.Api/Features/Filters/BodyModelStateValidatorAttribute.cs
+++ b/src/Microsoft.Health.Dicom.Api/Features/Filters/BodyModelStateValidatorAttribute.cs
@@ -18,7 +18,11 @@ namespace Microsoft.Health.Dicom.Api.Features.Filters
             EnsureArg.IsNotNull(context, nameof(context));
             if (!context.ModelState.IsValid)
             {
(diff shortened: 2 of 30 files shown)


src/Microsoft.Health.Dicom.Api/Features/Filters/BodyModelStateValidatorAttribute.cs
```diff
@@ -18,7 +18,11 @@ public override void OnActionExecuting(ActionExecutingContext context)
             EnsureArg.IsNotNull(context, nameof(context));
             if (!context.ModelState.IsValid)
             {
-                throw new InvalidRequestBodyException(Environment.NewLine + string.Join(Environment.NewLine, context.ModelState.SelectMany(x => x.Value.Errors).Select(x => x.ErrorMessage)));
+                var message = string.Join(Environment.NewLine,
+                    context.ModelState.Where(x => x.Value.ValidationState == AspNetCore.Mvc.ModelBinding.ModelValidationState.Invalid)
+                     .Select(x => $"\t{x.Key} - {x.Value.Errors.First().ErrorMessage}"));
+
+                throw new InvalidRequestBodyException(Environment.NewLine + message);
```
wsugarman, line 25: This was the existing behavior, but it's kind of weird that that a newline is passed as the first element. If this is internally concatenating strings, perhaps the `InvalidRequestBodyException` should take care of inserting the newline.

src/Microsoft.Health.Dicom.Api/Features/Filters/BodyModelStateValidatorAttribute.cs
```diff
@@ -18,7 +18,11 @@ public override void OnActionExecuting(ActionExecutingContext context)
             EnsureArg.IsNotNull(context, nameof(context));
             if (!context.ModelState.IsValid)
             {
-                throw new InvalidRequestBodyException(Environment.NewLine + string.Join(Environment.NewLine, context.ModelState.SelectMany(x => x.Value.Errors).Select(x => x.ErrorMessage)));
+                var message = string.Join(Environment.NewLine,
```
wsugarman, line 21: Would it be easier to pass key and error message separately to the exception?

src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
```diff
@@ -95,7 +95,10 @@ public static IDicomServerBuilder AddBackgroundWorkers(this IDicomServerBuilder
                     options.RespectBrowserAcceptHeader = true;
                     options.OutputFormatters.Insert(0, new DicomJsonOutputFormatter());
                 })
-                .AddJsonSerializerOptions(o => o.Converters.Add(new JsonStringEnumConverter()));
+                //
+                //  JsonStringEnumConverter doesn't provide good enough error message, make custom one to fit our requirements.
+                //  CustomJsonStringEnumConverter is used for extended query tag Apis.
+                .AddJsonSerializerOptions(o => o.Converters.Add(new CustomJsonStringEnumConverter()));
```
wsugarman, line 101: Is the default error message really that bad? For parsing enums, it appears to be like `"The value '<value>' is not valid for <Type>."` when I test it locally. I'd definitely prefer not to rewrite the enum parsing simply because of a less-than-ideal error message. If we want to change the error message, I think we can configure the `ModelBindingMessageProvider` when configuring `AddMvc`

src/Microsoft.Health.Dicom.Client/Models/UpdateExtendedQueryTagEntry.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    /// <summary>
+    /// Encapsulate parameters for updating extended query tag.
+    /// </summary>
+    public class UpdateExtendedQueryTagEntry
+    {
+        /// <summary>
+        /// Gets or sets query status.
+        /// </summary>
+        public QueryStatus QueryStatus { get; set; }
```
wsugarman, line 16: Do we want to be strict when parsing these values? For examples, should we throw an exception if a user thinks they can update other fields in the tag by including things like:
```json
{
  "TagPath": "foo",
  "QueryStatus": "Enabled"
}
```

I can see for `System.Text.Json` we can add a property with `[JsonExtensionData]` and perhaps check its count before continuing processing

src/Microsoft.Health.Dicom.Client/DicomWebClient.ExtendedQueryTag.cs
```diff
@@ -91,5 +91,25 @@ public async Task<DicomWebResponse<IEnumerable<ExtendedQueryTagError>>> GetExten
...
+        {
+            EnsureArg.IsNotNullOrWhiteSpace(tagPath, nameof(tagPath));
+            EnsureArg.IsNotNull(newValue, nameof(newValue));
+            EnsureArg.EnumIsDefined(newValue.QueryStatus, nameof(newValue.QueryStatus));
+            string jsonString = JsonConvert.SerializeObject(newValue);
+            var uri = new Uri($"/{_apiVersion}{DicomWebConstants.BaseExtendedQueryTagUri}", UriKind.Relative);
+
+            using var request = new HttpRequestMessage(HttpMethod.Patch, uri);
+            {
+                request.Content = new StringContent(jsonString);
```
wsugarman, line 105: I think you can use `JsonContent` instead

src/Microsoft.Health.Dicom.Client/DicomWebClient.ExtendedQueryTag.cs
```diff
@@ -91,5 +91,25 @@ public async Task<DicomWebResponse<IEnumerable<ExtendedQueryTagError>>> GetExten
...
+                request.Content.Headers.ContentType = new System.Net.Http.Headers.MediaTypeHeaderValue(DicomWebConstants.ApplicationJsonMediaType);
+            }
+
+            HttpResponseMessage response = await HttpClient.SendAsync(request, cancellationToken)
+                .ConfigureAwait(false);
```
wsugarman, line 110: nit: this can probably got on the above line

src/Microsoft.Health.Dicom.Client/DicomWebClient.ExtendedQueryTag.cs
```diff
@@ -91,5 +91,25 @@ public async Task<DicomWebResponse<IEnumerable<ExtendedQueryTagError>>> GetExten
             await EnsureSuccessStatusCodeAsync(response).ConfigureAwait(false);
             return new DicomWebResponse<IEnumerable<ExtendedQueryTagError>>(response, ValueFactory<IEnumerable<ExtendedQueryTagError>>);
         }
+
+        public async Task<DicomWebResponse<GetExtendedQueryTagEntry>> UpdateExtendedQueryTagAsync(string tagPath, UpdateExtendedQueryTagEntry newValue, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNullOrWhiteSpace(tagPath, nameof(tagPath));
+            EnsureArg.IsNotNull(newValue, nameof(newValue));
+            EnsureArg.EnumIsDefined(newValue.QueryStatus, nameof(newValue.QueryStatus));
```
wsugarman, line 99: nit: name here probably should be `newValue` to properly format the error: `EnsureArg.EnumIsDefined(newValue.QueryStatus, nameof(newValue));`

wsugarman (changes requested):

src/Microsoft.Health.Dicom.Api/Features/Filters/BodyModelStateValidatorAttribute.cs
pengchen0692: I found first error message should be good enough. will simplify

src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
pengchen0692: * There are 2 reason the `JsonStringEnumConverter` not fit well in our case. 

1.  Error message is good for developer, but not for end user.   Check the error message below, we know what is `[Microsoft.Health.Dicom.Core.Features.ExtendedQueryTag.QueryStatus]`, but custom don't know this, and no idea how to fix it -- we need to give valid value sets.


**Vs**. improved error message:


2.  `JsonStringEnumConverter` treat "1" as string, so accept as valid enum value, which I'm trying to block -- the number value is what we use internally, don't want to expose and confuse customer (even setting `allowIntegerValues` to `false` cannot block it)

* I explored `ModelBindingMessageProvider`, but seems not work in our case, best guess is the error comes from `JsonDeserializer` where model binding doesn't have enough insight --- I override accessor for every message in `ModelBindingMessageProvider` like below, but they don't show in error message.

pengchen0692: Maybe I should not block number  string (e.g"1"), that's default behavior of `Enum.TryParse`

src/Microsoft.Health.Dicom.Client/Models/UpdateExtendedQueryTagEntry.cs
pengchen0692: Did that, please take a look. :)

src/Microsoft.Health.Dicom.Client/DicomWebClient.ExtendedQueryTag.cs
pengchen0692: `JsonContent` use `System.Text.Json` for serialization,  as we already use `Newtonsoft.Json`,  better to use same.

src/Microsoft.Health.Dicom.Api/Features/Converter/CustomJsonStringEnumConverter.cs
```diff
@@ -0,0 +1,33 @@
...
+using System;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using EnsureThat;
+
+namespace Microsoft.Health.Dicom.Api.Features.Converter
+{
+    internal class CustomJsonStringEnumConverter : JsonConverterFactory
+    {
+        public CustomJsonStringEnumConverter() { }
```
wsugarman, line 17: nit: put the `{ }` on its own line

src/Microsoft.Health.Dicom.Api/Features/Converter/CustomJsonStringEnumConverterImp.cs
```diff
@@ -0,0 +1,40 @@
...
+
+using System;
+using System.Globalization;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using EnsureThat;
+
+namespace Microsoft.Health.Dicom.Api.Features.Converter
+{
+    internal class CustomJsonStringEnumConverterImp<T> : JsonConverter<T>
```
wsugarman, line 14: nit: perhaps names like `EnumNameJsonConverter` for the factory and `EnumNameConverter` for the actual converter?

src/Microsoft.Health.Dicom.Api/Features/Converter/CustomJsonStringEnumConverter.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        public override bool CanConvert(Type typeToConvert)
+        {
+            EnsureArg.IsNotNull(typeToConvert, nameof(typeToConvert));
```
wsugarman, line 21: nit: you can do this in one line like
```suggestion
            => EnsureArg.IsNotNull(typeToConvert, nameof(typeToConvert)).IsEnum;
```

src/Microsoft.Health.Dicom.Api/Features/Converter/CustomJsonStringEnumConverter.cs
```diff
@@ -0,0 +1,33 @@
...
+            return typeToConvert.IsEnum;
+        }
+
+        public override JsonConverter CreateConverter(Type typeToConvert, JsonSerializerOptions options)
+        {
+            EnsureArg.IsNotNull(typeToConvert, nameof(typeToConvert));
+            var baseType = typeof(CustomJsonStringEnumConverterImp<>);
+            var genericType = baseType.MakeGenericType(typeToConvert);
+            return (JsonConverter)Activator.CreateInstance(genericType);
```
wsugarman, line 30: nit: do all of this within `Activator.CreateInstance`. No need to create these variables3

src/Microsoft.Health.Dicom.Api/Features/Converter/CustomJsonStringEnumConverter.cs
```diff
@@ -0,0 +1,33 @@
...
+
+
+using System;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using EnsureThat;
+
+namespace Microsoft.Health.Dicom.Api.Features.Converter
+{
+    internal class CustomJsonStringEnumConverter : JsonConverterFactory
```
wsugarman, line 15: nit: let's make this `sealed`

src/Microsoft.Health.Dicom.Api/Features/Converter/CustomJsonStringEnumConverterImp.cs
wsugarman, line 14: nit: let's make this `sealed`

src/Microsoft.Health.Dicom.Api/Features/Converter/CustomJsonStringEnumConverterImp.cs
```diff
@@ -0,0 +1,40 @@
...
+            {
+                string content = reader.GetString();
+                if (Enum.TryParse(typeToConvert, content, true, out object result))
+                {
+                    return (T)result;
+                }
+            }
+
+            throw new JsonException(
+                string.Format(CultureInfo.InvariantCulture, DicomApiResource.InvalidEnumValue, string.Join("\",\"", Enum.GetNames(typeToConvert))));
```
wsugarman, line 31: While helpful, I am unsure we always want to print all of the possible values in the error. When things like `Enum.Parse` fail, they simply say they cannot parse it. Perhaps that is sufficient, especially if we ever have very prolific enums. Users can always refer to the swagger

src/Microsoft.Health.Dicom.Api/Features/Converter/CustomJsonStringEnumConverterImp.cs
```diff
@@ -0,0 +1,40 @@
...
+        where T : Enum
+    {
+        public override T Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            EnsureArg.IsNotNull(typeToConvert, nameof(typeToConvert));
+
+            if (reader.TokenType == JsonTokenType.String)
+            {
+                string content = reader.GetString();
+                if (Enum.TryParse(typeToConvert, content, true, out object result))
```
wsugarman, line 24:
```suggestion
                if (Enum.TryParse<T>(content, true, out T result))
```

Unless there is some screwiness between the `enum` type constraint and `Enum.TryParse<T>` which is totally possible.

src/Microsoft.Health.Dicom.Api/Features/Converter/CustomJsonStringEnumConverter.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        public CustomJsonStringEnumConverter() { }
+
+        public override bool CanConvert(Type typeToConvert)
+        {
+            EnsureArg.IsNotNull(typeToConvert, nameof(typeToConvert));
+            return typeToConvert.IsEnum;
```
wsugarman, line 22: Will we allow `Nullable<T>` where `T` is an `Enum`?

src/Microsoft.Health.Dicom.Api/Controllers/ExtendedQueryTagController.cs
```diff
@@ -183,6 +183,26 @@ public async Task<IActionResult> GetTagAsync(string tagPath)
             return StatusCode((int)HttpStatusCode.OK, response.ExtendedQueryTagErrors);
         }
 
+        [HttpPatch]
+        [Produces(KnownContentTypes.ApplicationJson)]
+        [BodyModelStateValidator]
+        [ProducesResponseType(typeof(IEnumerable<ExtendedQueryTagError>), (int)HttpStatusCode.OK)]
```
wsugarman, line 189: I don't think this returns errors 😛

src/Microsoft.Health.Dicom.Api/Features/Converter/CustomJsonStringEnumConverterImp.cs
wsugarman, line 24: Unfortunately, `TryParse` accepts both names and literals, right? We should have a test for that, but I'm afraid you'll need to probably cache a list of all of the names in this type and check against it.

src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
wsugarman: I think it's a bit strange. I am not wholly opposed to leaving that functionality in for now and we can remove it from our API later on. We just won't document it that behavior.

wsugarman: If we decide to leave it in for now, I'd say remove the custom converter until we remove the functionality.

wsugarman (changes requested):

src/Microsoft.Health.Dicom.Core/Features/ExtendedQueryTag/UpdateExtendedQueryTagService.cs
```diff
@@ -0,0 +1,47 @@
...
+            _dicomTagParser = EnsureArg.IsNotNull(dicomTagParser, nameof(dicomTagParser));
+            _urlResolver = EnsureArg.IsNotNull(urlResolver, nameof(urlResolver));
+        }
+
+        public async Task<GetExtendedQueryTagEntry> UpdateExtendedQueryTagAsync(string tagPath, UpdateExtendedQueryTagEntry newValue, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNull(tagPath, nameof(tagPath));
+            EnsureArg.IsNotNull(newValue?.QueryStatus, nameof(newValue.QueryStatus));
+            EnsureArg.EnumIsDefined(newValue.QueryStatus, nameof(UpdateExtendedQueryTagEntry.QueryStatus));
+            DicomTag[] tags;
```
wsugarman, line 36: nit: can we declare this in-line?

src/Microsoft.Health.Dicom.Core/Features/ExtendedQueryTag/UpdateExtendedQueryTagService.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            _extendedQueryTagStore = EnsureArg.IsNotNull(extendedQueryTagStore, nameof(extendedQueryTagStore));
+            _dicomTagParser = EnsureArg.IsNotNull(dicomTagParser, nameof(dicomTagParser));
+            _urlResolver = EnsureArg.IsNotNull(urlResolver, nameof(urlResolver));
+        }
+
+        public async Task<GetExtendedQueryTagEntry> UpdateExtendedQueryTagAsync(string tagPath, UpdateExtendedQueryTagEntry newValue, CancellationToken cancellationToken)
+        {
+            EnsureArg.IsNotNull(tagPath, nameof(tagPath));
+            EnsureArg.IsNotNull(newValue?.QueryStatus, nameof(newValue.QueryStatus));
```
wsugarman, line 34: nit: the 2nd parameter should probably still be `nameof(newValue)` as it refers to the argument name with the problem

src/Microsoft.Health.Dicom.Api/Features/Converter/CustomJsonStringEnumConverterImp.cs
pengchen0692: Another approach is to create custom jsonconverter for these enums specifically.

src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
pengchen0692: As discussed, will only use custom converter for our own Enum `QueryStatus`

src/Microsoft.Health.Dicom.Api/Features/Converter/CustomJsonStringEnumConverter.cs
pengchen0692: This class is removed.

src/Microsoft.Health.Dicom.Api/Features/Converter/CustomJsonStringEnumConverterImp.cs
pengchen0692: Get this failure if I do that:
image

src/Microsoft.Health.Dicom.Api/Features/Converter/CustomJsonStringEnumConverter.cs
pengchen0692: the class is deleted.

src/Microsoft.Health.Dicom.Api/Features/Converter/CustomJsonStringEnumConverterImp.cs
pengchen0692: I'm thinking just using the default behavior which accept both name and integer string is a good or as least acceptable idea.

src/Microsoft.Health.Dicom.Api/Features/Converter/EnumNameJsonConverter.cs
```diff
@@ -0,0 +1,44 @@
...
+        where T : Enum
+    {
+        public override T Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            EnsureArg.IsNotNull(typeToConvert, nameof(typeToConvert));
+
+            if (reader.TokenType == JsonTokenType.String)
+            {
+                string content = reader.GetString();
+                if (Enum.TryParse(typeToConvert, content, true, out object result))
```
wsugarman, line 28: After playing around, it appears that the `Enum` type constraint doesn't imply `struct`. So if you simply change your type constraint at the top to `where T : struct, Enum` then you should be able to use `Enum.TryParse<T>(content, true, out T result)`

src/Microsoft.Health.Dicom.Api/Features/Converter/EnumNameJsonConverter.cs
```diff
@@ -0,0 +1,44 @@
...
+            }
+
+            throw new JsonException(
+                string.Format(CultureInfo.InvariantCulture, DicomApiResource.InvalidEnumValue, string.Join("\",\"", Enum.GetNames(typeToConvert))));
+        }
+
+        public override void Write(Utf8JsonWriter writer, T value, JsonSerializerOptions options)
+        {
+            EnsureArg.IsNotNull(writer);
+            writer.WriteStringValue(value?.ToString());
```
wsugarman, line 41: If we support `null` writes, then we should support `null` reads.

wsugarman (changes requested):

PR #1066, opened 2021-09-28
SQL level changes for supporting DT and TM in Extended Query Tags.

alishahahmed:
## Description
SQL level changes made for supporting DT and TM in Extended Query Tags.
Mainly, following changes have been made:
- Added a column called TagValueUTC in ExtendedQueryTagDateTime table. It's been kept as nullable for backward compatibility.
- Updated respective index for this table to include TagValueUTC
- Updated table valued parameter InsertDateTimeExtendedQueryTagTableType_1 to expect TagValueUTC.
- Updated stored procedures AddInstance and IndexInstance to also insert TagValueUTC in ExtendedQueryTagDateTime table.

## Related issues
[AB#85475](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/85475)

## Testing
All existing tests are passing.


src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.diff.sql
```diff
@@ -0,0 +1,668 @@
...
+BEGIN
+    ALTER TABLE dbo.ExtendedQueryTagDateTime
+    ADD TagValueUTC DATETIME2(7) NULL
+END
+
+CREATE UNIQUE CLUSTERED INDEX IXC_ExtendedQueryTagDateTime ON dbo.ExtendedQueryTagDateTime
+(
+    TagKey,
+    TagValue,
+    TagValueUTC,
```
smithago, line 24: Can you give me 2 example of how will the QIDO generated SQL looks like for using this new column? Based on that con you confirm if this is the right index?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.diff.sql
```diff
@@ -0,0 +1,668 @@
...
+    StudyKey,
+    SeriesKey,
+    InstanceKey
+) WITH (DROP_EXISTING = ON)
+
+/*************************************************************
+    Table valued parameter to insert into Extended Query Tag table for data type Date Time.
+    V2 contains the TagValueUTC which separates it from V1.
+*************************************************************/
+IF TYPE_ID(N'InsertDateTimeExtendedQueryTagTableType_2') IS NULL
```
smithago, line 34: Since all of this is running in a trasaction, do you need this IF?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.diff.sql
```diff
@@ -0,0 +1,668 @@
...
+--     @doubleExtendedQueryTags
+--         * Double extended query tag data
+--     @dateTimeExtendedQueryTags
+--         * DateTime extended query tag data
+--     @personNameExtendedQueryTags
+--         * PersonName extended query tag data
+-- RETURN VALUE
+--     None
+/***************************************************************************************/
+CREATE OR ALTER PROCEDURE dbo.EndAddInstance
```
smithago, line 380: This is not used. Can we not make this change. Will is deleting this I think

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/Error/SqlExtendedQueryTagErrorStoreV4.cs
```diff
@@ -81,20 +81,20 @@ public override async Task<IReadOnlyList<ExtendedQueryTagError>> GetExtendedQuer
...
                 while (await reader.ReadAsync(cancellationToken))
                 {
                     (int tagkey, short errorCode, DateTime createdTime, string studyInstanceUid, string seriesInstanceUid, string sopInstanceUid) = reader.ReadRow(
-                        VLatest.ExtendedQueryTagError.TagKey,
-                        VLatest.ExtendedQueryTagError.ErrorCode,
-                        VLatest.ExtendedQueryTagError.CreatedTime,
-                        VLatest.Instance.StudyInstanceUid,
-                        VLatest.Instance.SeriesInstanceUid,
-                        VLatest.Instance.SopInstanceUid);
+                        V4.ExtendedQueryTagError.TagKey,
```
smithago, line 92: Why did ExtendedQueryTagErrorStore change?

smithago (changes requested):
Please review the comments for the details

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.diff.sql
alishahahmed: So for now, we don't need TagValueUTC. I placed it considering that we will eventually support offset in QIDO. Once that is implemented, TagValueUTC will help in the indexing. I can remove it for now and add it once we implement offset support in QIDO.

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/Error/SqlExtendedQueryTagErrorStoreV4.cs
alishahahmed: Thanks for pointing out. There is no need to make these changes. I made them initially just to test things out and forgot to revert. I will update this and other similar changes in the next iteration.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.diff.sql
alishahahmed: Since Will is not in this week, is it ok if I delete this stored procedure in my review?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.diff.sql
alishahahmed: Theoretically this if check can be removed. However, there is a test that tries to apply the same diff twice to check the idempotency. This would fail claiming that the type "InsertDateTimeExtendedQueryTagTableType_2" already exists.

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/Error/SqlExtendedQueryTagErrorStoreV4.cs
alishahahmed: Updated code in the latest iteration.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.diff.sql
alishahahmed: I have updated the code not to make changes to this stored procedure.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.diff.sql
smithago: Sounds good. Just a FYI, the way you had placed the TagValueUTC will probably not work. Please read how composite indexes work in SQL

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/ExtendedQueryTagDataRowsBuilderV2.cs
```diff
@@ -0,0 +1,123 @@
...
+using Microsoft.Health.Dicom.Core.Extensions;
+using Microsoft.Health.Dicom.Core.Features.ExtendedQueryTag;
+using Microsoft.Health.Dicom.SqlServer.Features.Schema.Model;
+
+namespace Microsoft.Health.Dicom.SqlServer.Features.ExtendedQueryTag
+{
+    /// <summary>
+    /// Class that build ExtendedQueryTagRows.
+    /// </summary>
+    internal static class ExtendedQueryTagDataRowsBuilderV2
```
jovinson-ms, line 21: While we are following the pattern of versioning the various stores to maintain backward compatibility, I don't think we need to version the builder and the data transfer object. We should be able to simply add the static methods to the builder and properties to the rows DTO.

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/ExtendedQueryTagDataRowsV2.cs
```diff
@@ -0,0 +1,29 @@
...
+
+using System.Collections.Generic;
+using Microsoft.Health.Dicom.SqlServer.Features.Schema.Model;
+
+namespace Microsoft.Health.Dicom.SqlServer.Features.ExtendedQueryTag
+{
+    /// <summary>
+    /// Represent rows for extended query tag index data.
+    /// </summary>
+    internal class ExtendedQueryTagDataRowsV2
```
jovinson-ms, line 14: See comment on builder.

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStoreV5.cs
```diff
@@ -0,0 +1,142 @@
...
+                        SqlErrorCodes.NotFound => new InstanceNotFoundException(),
+                        SqlErrorCodes.Conflict => new PendingInstanceException(),
+                        _ => new DataStoreException(ex),
+                    };
+
+                }
+            }
+        }
+
+        private static byte[] UlongToRowVersion(ulong? value)
```
jovinson-ms, line 130: Don't need to replicate this method.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.diff.sql
```diff
@@ -0,0 +1,567 @@
...
+    @studyDescription                   NVARCHAR(64) = NULL,
+    @accessionNumber                    NVARCHAR(64) = NULL,
+    @modality                           NVARCHAR(16) = NULL,
+    @performedProcedureStepStartDate    DATE = NULL,
+    @patientBirthDate                   DATE = NULL,
+    @manufacturerModelName              NVARCHAR(64) = NULL,
+    @stringExtendedQueryTags dbo.InsertStringExtendedQueryTagTableType_1 READONLY,
+    @longExtendedQueryTags dbo.InsertLongExtendedQueryTagTableType_1 READONLY,
+    @doubleExtendedQueryTags dbo.InsertDoubleExtendedQueryTagTableType_1 READONLY,
+    @dateTimeExtendedQueryTags dbo.InsertDateTimeExtendedQueryTagTableType_2 READONLY,
```
smithago, line 99: Can you check with Brandon if this works. Your code changes are deployed, SQl schema version is 5, but the code still has not gotten the latest version info, it still things it is 4. When it send AddInstance from 4 Store version, won't this fail?

alishahahmed: That is indeed a very good point. I will sync up with him to see what can be done to avoid this issue.

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStoreV5.cs
alishahahmed: Good point. I will update it.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.diff.sql
alishahahmed: I checked with Brandon. I went to him with one idea and his recommendation was aligned with my idea. We need to have a new version of the stored procedure as well. Ultimately, once we are past supporting the minimum version for backward compatibility, we can safely remove the older version of the stored procedure. This change will be part of the next iteration which I will push shortly.

alishahahmed: New iteration has been updated which should fix this issue.

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStoreV5.cs
alishahahmed: Updated.

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/ExtendedQueryTagDataRowsV2.cs
alishahahmed: Done.

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/ExtendedQueryTagDataRowsBuilderV2.cs
alishahahmed: Fixed this. I have used a new schema version constant and used that in the builder for considering the correct field.

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/ExtendedQueryTagDataRowsBuilder.cs
```diff
@@ -36,6 +39,7 @@ internal static class ExtendedQueryTagDataRowsBuilder
             var longRows = new List<InsertLongExtendedQueryTagTableTypeV1Row>();
             var doubleRows = new List<InsertDoubleExtendedQueryTagTableTypeV1Row>();
             var dateTimeRows = new List<InsertDateTimeExtendedQueryTagTableTypeV1Row>();
+            var dateTimeWithUTCRows = new List<InsertDateTimeExtendedQueryTagTableTypeV2Row>();
             var personNamRows = new List<InsertPersonNameExtendedQueryTagTableTypeV1Row>();
```
jovinson-ms, line 43: nit: `personName`

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/ExtendedQueryTagDataRowsBuilder.cs
```diff
@@ -91,6 +105,18 @@ private static void AddDateTimeRow(DicomDataset instance, List<InsertDateTimeExt
             }
         }
 
+        private static void AddDateTimeRow(DicomDataset instance, List<InsertDateTimeExtendedQueryTagTableTypeV2Row> dateTimeRows, QueryTag queryTag)
```
jovinson-ms, line 108: were you going to add a qualifier here, like `...WithUtc`?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.diff.sql
```diff
@@ -0,0 +1,567 @@
...
+IF NOT EXISTS 
+(
+    SELECT *
+    FROM    sys.columns
+    WHERE   NAME = 'TagValueUTC'
+            AND object_id = OBJECT_ID('dbo.ExtendedQueryTagDateTime')
+)
+BEGIN
+    ALTER TABLE dbo.ExtendedQueryTagDateTime
+    ADD TagValueUTC DATETIME2(7) NULL
```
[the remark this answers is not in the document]
alishahahmed: thanks for pointing this out. I will make the change.

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/ExtendedQueryTagDataRowsBuilder.cs
alishahahmed: I was debating, but left it as an overload of AddDateTimeRow. If there is a preference, I don't mind making the change.

src/Microsoft.Health.Dicom.SqlServer.UnitTests/Features/ExtendedQueryTag/AddInstanceTableValuedParametersBuilderTests.cs
```diff
@@ -23,7 +24,7 @@ public void GivenSupportedDicomElement_WhenRead_ThenShouldReturnExpectedValue(Di
             DicomDataset dataset = new DicomDataset();
             dataset.Add(element);
             QueryTag tag = new QueryTag(element.Tag.BuildExtendedQueryTagStoreEntry(vr: element.ValueRepresentation.Code));
-            var parameters = ExtendedQueryTagDataRowsBuilder.Build(dataset, new QueryTag[] { tag });
+            var parameters = ExtendedQueryTagDataRowsBuilder.Build(dataset, new QueryTag[] { tag }, SchemaVersion.V5);
```
jovinson-ms, line 27: Since there's logic based this value, how about including schema version as an argument to this test method and specifying it for each data member?

alishahahmed: Yep, that would be valuable! I will make the necessary changes here.

alishahahmed: Latest iteration has the updated tests that run for all the schema versions between min and max.

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/ExtendedQueryTagDataRowsBuilder.cs
alishahahmed: Updated in the latest iteration.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/5.diff.sql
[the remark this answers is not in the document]
alishahahmed: Done.

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/ExtendedQueryTagDataRowsBuilder.cs
```diff
@@ -91,6 +105,18 @@ private static void AddDateTimeRow(DicomDataset instance, List<InsertDateTimeExt
...
 
+        private static void AddDateTimeWithUtcRow(DicomDataset instance, List<InsertDateTimeExtendedQueryTagTableTypeV2Row> dateTimeRows, QueryTag queryTag)
+        {
+            DateTime? dateVal = DataTimeReaders.TryGetValue(
+                             queryTag.VR,
+                             out Func<DicomDataset, DicomTag, DicomVR, DateTime?> reader) ? reader.Invoke(instance, queryTag.Tag, queryTag.VR) : null;
+
+            if (dateVal.HasValue)
+            {
+                dateTimeRows.Add(new InsertDateTimeExtendedQueryTagTableTypeV2Row(queryTag.ExtendedQueryTagStoreEntry.Key, dateVal.Value, null, (byte)queryTag.Level));
```
pengchen0692, line 116: As we discussed, whether or not tagvalue contains timezone offset, the `TagValueUtc` should be set to some value, but here is `null`

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/ExtendedQueryTagDataRows.cs
```diff
@@ -23,6 +24,8 @@ internal class ExtendedQueryTagDataRows
 
         public IEnumerable<InsertDateTimeExtendedQueryTagTableTypeV1Row> DateTimeRows { get; set; }
 
+        public IEnumerable<InsertDateTimeExtendedQueryTagTableTypeV2Row> DateTimeWithUTCRows { get; set; }
```
pengchen0692, line 27: nit: UTC -> Utc

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/ExtendedQueryTagDataRowsBuilder.cs
alishahahmed: This is actually just part of the changes, kind of a place holder. Once this PR gets checked in, my following PR will contain backend changes which will update this code.

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/Error/SqlExtendedQueryTagErrorStoreV5.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    internal class SqlExtendedQueryTagErrorStoreV5 : SqlExtendedQueryTagErrorStoreV4
+    {
+        public SqlExtendedQueryTagErrorStoreV5(
+           SqlConnectionWrapperFactory sqlConnectionWrapperFactory,
+           ILogger<SqlExtendedQueryTagErrorStoreV5> logger) : base(sqlConnectionWrapperFactory, logger)
+        {
+        }
+
+        public override SchemaVersion Version => SchemaVersion.V5;
```
smithago, line 20: Maybe for the future, we should look at making this better. I feels boiler plate code to create a class to just inherit, maybe we can make Version property a collection and multiple versions per class.

jovinson-ms (approved):

pengchen0692 (approved):

smithago (approved):

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/Error/SqlExtendedQueryTagErrorStoreV5.cs
alishahahmed: I agree! This is something that should be fixed sooner than later. As we have increasing number of schemas, this will get harder to manage as well.

PR #1072, opened 2021-10-01
Personal/alishah/support dt tm extended query tags

alishahahmed:
## Description
Supporting DT and TM as Extended Query Tags.

- DT and TM are now accepted as Extended Query Tags.
- On STOW, newly created tags are indexed properly.
- Single Value Match working for DT and TM.
- Range Value Match working for DT and TM.
- Deletion: when an instance is deleted, it's respective entries are deleted from respective extended query tags tables.
- Deletion: when DT or TM extended query tag is deleted, respective entries are also deleted from extended query tags tables.

## Related issues
AB#84581
AB#84582
AB#84583
AB#84585
AB#84643
AB#84851
AB#84858
AB#84859

## Testing

- Added unit tests for DT and TM, and range query matching.
- All existing test cases are passing.


src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.TagValueParser.cs
```diff
@@ -42,6 +42,64 @@ private static QueryFilterCondition ParseDateTagValue(QueryTag queryTag, string
             return new DateSingleValueMatchCondition(queryTag, parsedDate);
         }
 
+        private static QueryFilterCondition ParseDateTimeTagValue(QueryTag queryTag, string value)
+        {
+            if (QueryLimit.IsValidRangeQueryTag(queryTag))
```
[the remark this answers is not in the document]
alishahahmed: I believe so. It's a preventive check to validate if the VR is supported for range queries.

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.TagValueParser.cs
```diff
@@ -76,5 +134,25 @@ private static DateTime ParseDate(string date, string tagName)
 
             return parsedDate;
         }
+
+        private static DateTime ParseDateTime(string dateTime, string tagName)
```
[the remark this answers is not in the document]
alishahahmed: That's a very good suggestion. I will make the necessary changes. Thank you!

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.TagValueParser.cs
```diff
@@ -76,5 +134,25 @@ private static DateTime ParseDate(string date, string tagName)
...
+        {
+            if (!DateTimeOffset.TryParseExact(dateTime, DateTimeTagValueFormats, null, System.Globalization.DateTimeStyles.None, out DateTimeOffset parsedDateTimeOffset))
```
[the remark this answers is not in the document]
alishahahmed: Since that are specific formats that we accept (DateTimeTagValueFormats), we don't necessarily need the CultureInfo.InvariantCulture. This is also the case with the already existing ParseDate method.
If you feel like it should be changed to CultureInfo.InvariantCulture, please feel free to respond or reach out offline.

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.cs
```diff
@@ -50,6 +52,32 @@ public partial class QueryParser : IQueryParser
 
         public const string DateTagValueFormat = "yyyyMMdd";
 
+        public static readonly string[] DateTimeTagValueFormats =
+        {
+            "yyyyMMddHHmmss.FFFFFFzzz",
```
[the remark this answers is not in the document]
alishahahmed: Good call! I will make this change along with the above mentioned change where, if offset is passed in the query attribute, will result in an exception.

src/Microsoft.Health.Dicom.Core/DicomCoreResource.resx
```diff
@@ -418,4 +418,28 @@ The first part date {1} should be lesser than or equal to the second part date {
     <value>Query is disabled on specified attribute '{0}'.</value>
     <comment>'{0}' attribute in QIDO query. </comment>
   </data>
+  <data name="InvalidDateTimeRangeValue" xml:space="preserve">
+    <value>Invalid QIDO-RS query. Specified date range '{0}' is invalid.
+The first part DateTime {1} should be lesser than or equal to the second part DateTime {2}.</value>
+    <comment>{0} DateTime range with &lt;DateTime1&gt;-&lt;DateTime2&gt; format. {1} is DateTime1. {2} is DateTime2.</comment>
+  </data>
+  <data name="InvalidDateTimeValue" xml:space="preserve">
+    <value>Invalid QIDO-RS query. Specified DateTime value '{0}' is invalid for parameter '{1}'. DateTime should be valid and formatted as yyyyMMddHHmmss.FFFFFFzzz where yyyy is mandatory.</value>
```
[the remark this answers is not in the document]
alishahahmed: Interesting. Parameter was already used for InvalidDateValue. I will make this change for both.
Also, I will update the exception to not include timezone when I fix the offset exception issue as discussed earlier.

src/Microsoft.Health.Dicom.Core/Features/Validation/ValidationErrorCode.cs
```diff
@@ -78,6 +78,16 @@ public enum ValidationErrorCode : short
         /// </summary>
         DateIsInvalid = 1100,
 
+        /// <summary>
+        /// DateTime element has invalid value.
+        /// </summary>
+        DateTimeIsInvalid = 1101,
```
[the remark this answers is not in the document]
alishahahmed: Ohhh I didn't notice that. Thank you! I will update this.

src/Microsoft.Health.Dicom.Core/Features/Validation/DateTimeValidation.cs
```diff
@@ -0,0 +1,51 @@
...
+using Dicom;
+using Microsoft.Health.Dicom.Core.Exceptions;
+using Microsoft.Health.Dicom.Core.Extensions;
+
+namespace Microsoft.Health.Dicom.Core.Features.Validation
+{
+    internal class DateTimeValidation : ElementValidation
+    {
+        private static readonly string[] DateTimeFormatsDT =
+        {
```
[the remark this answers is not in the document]
alishahahmed: Makes sense. I will keep the one in Extensions and use the same here.

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.TagValueParser.cs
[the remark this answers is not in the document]
alishahahmed: Done.

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.cs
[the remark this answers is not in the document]
alishahahmed: Fixed.

src/Microsoft.Health.Dicom.Core/Features/Validation/TimeValidation.cs
```diff
@@ -0,0 +1,41 @@
...
+            string value = dicomElement.Get<string>();
+            string name = dicomElement.Tag.GetFriendlyName();
+            if (string.IsNullOrEmpty(value))
+            {
+                return;
+            }
+
+            if (!DateTime.TryParseExact(value, TimeFormatsTM, CultureInfo.InvariantCulture, DateTimeStyles.NoCurrentDateDefault, out _))
+            {
+                throw ElementValidationExceptionFactory.CreateDateIsInvalidException(name, value);
```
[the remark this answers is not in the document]
alishahahmed: Updated. Good catch!

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -62,13 +88,43 @@ public static T GetSingleValueOrDefault<T>(this DicomDataset dicomDataset, Dicom
         /// <param name="dicomDataset">The dataset to get the VR value from.</param>
         /// <param name="dicomTag">The DICOM tag.</param>
         /// <param name="expectedVR">Expected VR of the element.</param>
-        /// /// <remarks>If expectedVR is provided, and not match, will return null.</remarks>
-        /// <returns>An instance of <see cref="DateTime"/> if the value exists and comforms to the DA format; otherwise <c>null</c>.</returns>
-        public static DateTime? GetStringDateAsDate(this DicomDataset dicomDataset, DicomTag dicomTag, DicomVR expectedVR = null)
+        /// <remarks>If expectedVR is provided, and not match, will return null.</remarks>
+        /// <returns>An instance of <see cref="DateTimeOffset"/> if the value exists and comforms to the DA format; otherwise <c>null</c>.</returns>
+        public static DateTimeOffset? GetStringDateAsDate(this DicomDataset dicomDataset, DicomTag dicomTag, DicomVR expectedVR = null)
```
smithago, line 93: Why change this to DateTimeOffset?

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -62,13 +88,43 @@ public static T GetSingleValueOrDefault<T>(this DicomDataset dicomDataset, Dicom
...
         {
             EnsureArg.IsNotNull(dicomDataset, nameof(dicomDataset));
             string stringDate = dicomDataset.GetSingleValueOrDefault<string>(dicomTag, expectedVR: expectedVR);
-            return DateTime.TryParseExact(stringDate, DateFormatDA, CultureInfo.InvariantCulture, DateTimeStyles.None, out DateTime result) ? result : null;
+            return DateTimeOffset.TryParseExact(stringDate, DateFormatDA, CultureInfo.InvariantCulture, DateTimeStyles.None, out DateTimeOffset result) ? result : null;
+        }
+
+        /// <summary>
+        /// Gets the DT VR value as <see cref="DateTime"/>.
```
smithago, line 101: comment is wrong this is returning Offset

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -62,13 +88,43 @@ public static T GetSingleValueOrDefault<T>(this DicomDataset dicomDataset, Dicom
...
+        /// <returns>An instance of <see cref="DateTime"/> if the value exists and conforms to the DT format; othewise <c>null</c>.</returns>
+        public static DateTimeOffset? GetStringDateTimeAsDateTimeOffset(this DicomDataset dicomDataset, DicomTag dicomTag, DicomVR expectedVR = null)
+        {
+            EnsureArg.IsNotNull(dicomDataset, nameof(dicomDataset));
+            string stringDateTime = dicomDataset.GetSingleValueOrDefault<string>(dicomTag, expectedVR: expectedVR);
+            return DateTimeOffset.TryParseExact(stringDateTime, DateTimeFormatsDT, CultureInfo.InvariantCulture, DateTimeStyles.None, out DateTimeOffset result) ? result : null;
+        }
+
+        /// <summary>
+        /// Gets the TM VR value as <see cref="DateTime"/>.
```
smithago, line 116: same here

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.TagValueParser.cs
```diff
@@ -42,6 +43,64 @@ private static QueryFilterCondition ParseDateTagValue(QueryTag queryTag, string
             return new DateSingleValueMatchCondition(queryTag, parsedDate);
         }
 
+        private static QueryFilterCondition ParseDateTimeTagValue(QueryTag queryTag, string value)
+        {
```
smithago, line 47: Can we consolidate these methods, they all look the same, except the parser is different

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.cs
```diff
@@ -50,6 +52,36 @@ public partial class QueryParser : IQueryParser
 
         public const string DateTagValueFormat = "yyyyMMdd";
 
+        public static readonly string[] DateTimeTagValueFormats =
```
smithago, line 55: They seem to be duplicated also in DicomDatasetExtensions. Can we consolidate?

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
alishahahmed: This was done to make DT part of the range queries. DA was a DateTimeReaders supporting range queries. Since, the same could be used for DT, I ended up updating DA as offset.
For reference, please check ExtendedQueryTagDataRowsBuilder.cs.
As an alternative, a new reader can be created for DT avoiding changes for DA, but I felt like this way was cleaner. Hence, proceeded with it.

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.cs
alishahahmed: In the updated iteration, these are different. In DicomDatasetExtensions, we also accept the offset. However, while parsing query, we have two separate sets for date times with offsets and without offsets such that appropriate exception can be displayed to the user.

src/Microsoft.Health.Dicom.Core/Features/Validation/DateTimeValidation.cs
```diff
@@ -0,0 +1,33 @@
...
+            base.Validate(dicomElement);
+
+            string value = dicomElement.Get<string>();
+            string name = dicomElement.Tag.GetFriendlyName();
+            if (string.IsNullOrEmpty(value))
+            {
+                return;
+            }
+
+            if (!DateTimeOffset.TryParseExact(value, Extensions.DicomDatasetExtensions.DateTimeFormatsDT, CultureInfo.InvariantCulture, DateTimeStyles.None, out _))
```
smithago, line 27: Did you look at https://github.com/fo-dicom/fo-dicom/blob/development/DICOM/DicomValidation.cs?

Is this something we can use?

src/Microsoft.Health.Dicom.Core/Features/Validation/TimeValidation.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Globalization;
+using Dicom;
+using Microsoft.Health.Dicom.Core.Exceptions;
+using Microsoft.Health.Dicom.Core.Extensions;
+
+namespace Microsoft.Health.Dicom.Core.Features.Validation
+{
+    internal class TimeValidation : ElementValidation
+    {
+        private static readonly string[] TimeFormatsTM =
```
smithago, line 16: same here

src/Microsoft.Health.Dicom.SqlServer/Features/Query/SqlQueryGenerator.cs
```diff
@@ -331,6 +331,24 @@ public override void Visit(DoubleSingleValueMatchCondition doubleSingleValueMatc
                 .AppendLine();
         }
 
+        public override void Visit(LongRangeValueMatchCondition rangeValueMatchCondition)
```
smithago, line 334: I am using you are not checking on UTC value and only supporting literal value

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -62,13 +88,43 @@ public static T GetSingleValueOrDefault<T>(this DicomDataset dicomDataset, Dicom
...
+        /// <param name="dicomDataset">The dataset to get the VR value from.</param>
+        /// <param name="dicomTag">The DICOM tag.</param>
+        /// <param name="expectedVR">Expected VR of the element.</param>
+        /// <remarks>If expectedVR is provided, and not match, will return null.</remarks>
+        /// <returns>An instance of <see cref="DateTime"/> if the value exists and conforms to the DT format; othewise <c>null</c>.</returns>
+        public static DateTimeOffset? GetStringDateTimeAsDateTimeOffset(this DicomDataset dicomDataset, DicomTag dicomTag, DicomVR expectedVR = null)
+        {
+            EnsureArg.IsNotNull(dicomDataset, nameof(dicomDataset));
+            string stringDateTime = dicomDataset.GetSingleValueOrDefault<string>(dicomTag, expectedVR: expectedVR);
+            return DateTimeOffset.TryParseExact(stringDateTime, DateTimeFormatsDT, CultureInfo.InvariantCulture, DateTimeStyles.None, out DateTimeOffset result) ? result : null;
```
smithago, line 112: Is this current? Should you be looking at the http://dicom.nema.org/medical/Dicom/2016e/output/chtml/part03/sect_C.12.5.html that tag to confirm the current culture?

You need to parse 2 DTs, one as is, put in TagValue and another in UTC needs you to config the Offset tag and if specified Get the UTC DT and use that in STOW

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/ExtendedQueryTagDataRowsBuilder.cs
```diff
@@ -95,25 +101,25 @@ private static void AddPersonNameRow(DicomDataset instance, List<InsertPersonNam
...
-            DateTime? dateVal = DataTimeReaders.TryGetValue(
+            DateTimeOffset? dateVal = DataTimeReaders.TryGetValue(
                              queryTag.VR,
-                             out Func<DicomDataset, DicomTag, DicomVR, DateTime?> reader) ? reader.Invoke(instance, queryTag.Tag, queryTag.VR) : null;
+                             out Func<DicomDataset, DicomTag, DicomVR, DateTimeOffset?> reader) ? reader.Invoke(instance, queryTag.Tag, queryTag.VR) : null;
 
             if (dateVal.HasValue)
             {
-                dateTimeRows.Add(new InsertDateTimeExtendedQueryTagTableTypeV2Row(queryTag.ExtendedQueryTagStoreEntry.Key, dateVal.Value, null, (byte)queryTag.Level));
+                dateTimeRows.Add(new InsertDateTimeExtendedQueryTagTableTypeV2Row(queryTag.ExtendedQueryTagStoreEntry.Key, dateVal.Value.DateTime, queryTag.VR == DicomVR.DT ? dateVal.Value.UtcDateTime : null, (byte)queryTag.Level));
```
smithago, line 122: dateVal.Value.UtcDateTime does not seem correct. It will use the server timezone and convert to UTC.

src/Microsoft.Health.Dicom.Core/Features/Query/QueryParser.TagValueParser.cs
alishahahmed: This is doable but would make things a bit complicated for the fact that each of these methods has following differences:
- Parser
- Exception
- Single value match condition
- Range value match condition

I will proceed by combining Date and DateTime methods as they are fairly similar, however, will leave Time method as is as it uses long instead of DateTime which would make things messier.

src/Microsoft.Health.Dicom.Core/Features/Validation/DateTimeValidation.cs
alishahahmed: thanks for referring to this. I will check it out and use this existing validation.

src/Microsoft.Health.Dicom.SqlServer/Features/Query/SqlQueryGenerator.cs
alishahahmed: For now, we are not supporting offsets in QIDO. Once we work on implementing this feature, that's when UTCValue will come in handy. For now, we are simply storing it such that we have enough data to work on whenever we support this in future.

src/Microsoft.Health.Dicom.Core/Features/Validation/DateTimeValidation.cs
pengchen0692: Maybe we should utilize this validation method for all other VRs. I remember there was a bug where it fails to write (or maybe read) dicom data into json file, this change probably can solve that issue.

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
alishahahmed: I have changed everything back to DateTime. Now that I am using two separate requests for calculating local time and UTC time, DateTimeOffset is not required.

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
alishahahmed: Updated. Now we make two calls, one for getting the local time and one for UTC time.

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/ExtendedQueryTagDataRowsBuilder.cs
alishahahmed: Updated.

Issue #1087, opened 2021-10-05
Support more date formats in QIDO-RS range matches


smithago:
**User story**
Support more dates in QIDO-RS range matches

We attempted to query the studies prior 2021-01-15 with the following:
'/studies?limit=200&fuzzymatching=true&00080020=-20210115'

However it was rejected with the following error:
'Invalid QIDO-RS query. Specified Date value '' is invalid for parameter 'StudyDate'. Date should be valid and formatted as yyyyMMdd.'

However the notation -<date> is actually valid: DICOM standard on QIDO matching (part 18 section 6.7.1..2.1:
The matching semantics for each attribute are determined by the types of matching allowed by C-FIND (see Section C.2.2.2 in PS3.4 ).

 And part 4, section C.2.2.2.4 (Range Matching) : 

In the absence of extended negotiation, if the Attribute is a date, then: 

•	A string of the form "<date1> - <date2>", where <date1> is less or equal to <date2>, shall match all occurrences of dates that fall between <date1> and <date2> inclusive
•	A string of the form "- <date1>" shall match all occurrences of dates prior to and including <date1>
•	A string of the form "<date1> -" shall match all occurrences of <date1> and subsequent dates

**Acceptance criteria**
Support below scenarios
•	A string of the form "- <date1>" shall match all occurrences of dates prior to and including <date1>
•	A string of the form "<date1> -" shall match all occurrences of <date1> and subsequent dates

PR #1109, opened 2021-10-12
Support more date formats in QIDO-RS range matches

alishahahmed:
## Description
**Reference:** [Support more date formats in QIDO-RS range matches](https://github.com/microsoft/dicom-server/issues/1087)

This pull request addresses the following issue:
Currently, DICOM OSS project supports range matching on DA, DT and TM VR types. For all these three types, following pattern needs to be followed: _TagName=Value1-Value2_. If, user does not provide _Value1_ or _Value2_, the query fails. However, according to the DICOM standard, it is possible to have one of these two values as empty. Following is the description of such cases.

**TagName=-Value2**
In such case, we need to match all the values that occur prior to Value2 where Value2 is also inclusive.

**TagName=Value1-**
In such case, we need to match all the values that occur post Value1 where Value1 is also inclusive.

## Related issues
[AB#85848](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/85848).

## Testing
Tests are also added for positive and negative cases.
All existing tests are also passing

### DIFF
--- a/docs/how-to-guides/extended-query-tags.md
+++ b/docs/how-to-guides/extended-query-tags.md
@@ -209,7 +209,7 @@ The matching types stated below are valid for extended query tags.
 
 | Search Type | Supported VR    | Example                                                      |
 | :---------- | :-------------- | :----------------------------------------------------------- |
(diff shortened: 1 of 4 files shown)


alishahahmed:
> can you update the docs along with the PR. I do not see md file changes

I have updated the documentation in the latest iteration. Thanks for pointing out.

PR #1147, opened 2021-10-28
Data Partitions - Changefeed client updates

bcarthic:
## Description
This PR updates Change feed client to return partition name.

## Related issues
Addresses [[AB#84991](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/84991)]

## Testing
We updated existing and new E2E test to validate basic functionality. We also tested against DICOM Cast running locally and changes were properly propagated to FHIR. Since the partition is not used in the cast, the same dataset will only be represented once in FHIR when stored in multiple partitions.


test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/DataPartitionEnabledTests.cs
```diff
@@ -130,6 +132,48 @@ public async Task GivenDatasetInstancesWithDifferentPartitions_WhenDeleted_OneDe
...
+            {
+                // Data partition feature flag only enabled locally. For Remote servers, feature flag is by default disabled
+                return;
+            }
+
+            var newPartition = TestUidGenerator.Generate();
+            string studyInstanceUID = TestUidGenerator.Generate();
+            string seriesInstanceUID = TestUidGenerator.Generate();
+            string sopInstanceUID = TestUidGenerator.Generate();
+
```
wsugarman, line 148: nit: extra new line

test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/DataPartitionEnabledTests.cs
```diff
@@ -130,6 +132,48 @@ public async Task GivenDatasetInstancesWithDifferentPartitions_WhenDeleted_OneDe
...
+
+            DicomFile dicomFile = Samples.CreateRandomDicomFile(studyInstanceUID, seriesInstanceUID, sopInstanceUID);
+            using DicomWebResponse<DicomDataset> response1 = await _client.StoreAsync(new[] { dicomFile }, partitionName: newPartition);
```
wsugarman, line 151: Do we need to assert this was successful?

src/Microsoft.Health.Dicom.Client/Models/ChangeFeedEntry.cs
```diff
@@ -36,10 +37,13 @@ public class ChangeFeedEntry
             Action = action;
             Timestamp = timestamp;
             State = state;
+            PartitionName = partitionName;
```
wsugarman, line 40: Are there any rules we need to enforce about partition names? Like can't be white space?

wsugarman (approved):

test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/DataPartitionEnabledTests.cs
bcarthic: Ideally yes :). Fixed

src/Microsoft.Health.Dicom.Client/Models/ChangeFeedEntry.cs
bcarthic: It is enforced at the service level when the user issue a request. This will be anyway part of the response and not the request.

PR #1148, opened 2021-10-29
Data partitions - Ensure feature cannot be disabled when data is partitioned

bcarthic:
## Description
This PR includes changes to ensure user cannot disable DataPartitions feature flag if they already created partitioned data other than the default partition.

## Related issues
Addresses [[AB#85406](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/85406)].

### DIFF
--- a/src/Microsoft.Health.Dicom.Api/Features/Partition/DataPartitionFeatureValidatorService.cs
+++ b/src/Microsoft.Health.Dicom.Api/Features/Partition/DataPartitionFeatureValidatorService.cs
@@ -1,0 +1,60 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using EnsureThat;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Options;
+using Microsoft.Health.Dicom.Core.Configs;
+using Microsoft.Health.Dicom.Core.Exceptions;
+using Microsoft.Health.Dicom.Core.Features.Partition;
+
+namespace Microsoft.Health.Dicom.Api.Features.Partition
+{
+    /// <summary>
+    /// This hosted service performs the check at startup to ensure user cannot disable DataPartitions feature flag
+    /// if they already created partitioned data other than the default partition.
+    /// If a user created partitioned data other than the default partition, then an exception is thrown to block startup.
+    /// We will allow users to enable DataPartitions feature even if they have already created data,
+    /// it will be accessible by specifying Microsoft.Default partition name in the request.
+    /// </summary>
+    public class DataPartitionFeatureValidatorService : IHostedService
+    {
+        private readonly IServiceProvider _serviceProvider;
+        private readonly bool _isPartitionEnabled;
+
+        public DataPartitionFeatureValidatorService(
+            IServiceProvider serviceProvider,
+            IOptions<FeatureConfiguration> featureConfiguration)
+        {
+            _serviceProvider = serviceProvider;
+            EnsureArg.IsNotNull(featureConfiguration, nameof(featureConfiguration));
+            _isPartitionEnabled = featureConfiguration.Value.EnableDataPartitions;
+        }
+
+        public async Task StartAsync(CancellationToken cancellationToken)
+        {
+            if (!_isPartitionEnabled)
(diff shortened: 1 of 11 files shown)


src/Microsoft.Health.Dicom.Api/Features/Partition/DataPartitionFeatureValidatorService.cs
```diff
@@ -0,0 +1,60 @@
...
+            _serviceProvider = serviceProvider;
+            EnsureArg.IsNotNull(featureConfiguration, nameof(featureConfiguration));
+            _isPartitionEnabled = featureConfiguration.Value.EnableDataPartitions;
+        }
+
+        public async Task StartAsync(CancellationToken cancellationToken)
+        {
+            if (!_isPartitionEnabled)
+            {
+                using (var scope = _serviceProvider.CreateScope())
```
wsugarman, line 44: Do we need the new service scope for `IPartitionService`?

src/Microsoft.Health.Dicom.Api/Features/Partition/DataPartitionFeatureValidatorService.cs
```diff
@@ -0,0 +1,60 @@
...
+    {
+        private readonly IServiceProvider _serviceProvider;
+        private readonly bool _isPartitionEnabled;
+
+        public DataPartitionFeatureValidatorService(
+            IServiceProvider serviceProvider,
+            IOptions<FeatureConfiguration> featureConfiguration)
+        {
+            _serviceProvider = serviceProvider;
+            EnsureArg.IsNotNull(featureConfiguration, nameof(featureConfiguration));
```
wsugarman, line 36: Can you double-check, but I don't think `IOptions<T>` guarantees that `Value` is non-`null`, so you should probably `null`-check it. Although that would be pretty rare it is actually `null`.

src/Microsoft.Health.Dicom.Core/Exceptions/DataPartitionsFeatureCannotBeDisabledException.cs
```diff
@@ -0,0 +1,20 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System.Globalization;
+
+namespace Microsoft.Health.Dicom.Core.Exceptions
+{
+    /// <summary>
+    /// Exception that is thrown when data partitions feature is disabled.
+    /// </summary>
+    public class DataPartitionsFeatureCannotBeDisabledException : BadRequestException
```
wsugarman, line 13: Is this really a `BadRequest`? It seems more general purpose as this is thrown at the server start-up, like an `InvalidOperationException`?

wsugarman (approved):

src/Microsoft.Health.Dicom.Core/DicomCoreResource.resx
```diff
@@ -465,4 +465,7 @@ For details on valid range queries, please refer to Search Matching section in C
     <value>Specified PartitionName '{0}' does not exist.</value>
     <comment>{0} PartitionName value.</comment>
   </data>
+  <data name="DataPartitionFeatureCannotBeDisabled" xml:space="preserve">
+    <value>Data partitions feature cannot be disabled when the service has existing data.</value>
```
wsugarman, line 469: nit:
```suggestion
    <value>Data partitions cannot be disabled while existing data has already been partitioned.</value>
```

Is there also any way for users to disable it? If they deleted all of their data / partitions, then could they disable the feature? If so, perhaps the error should indicate this.

bcarthic: Currently there is no way to delete all the partitions unless they have access to the DB. But in future may be.

src/Microsoft.Health.Dicom.Api/Features/Partition/DataPartitionFeatureValidatorService.cs
bcarthic: Yes we need because we cant inject a scoped service in a hosted services. The service needs to be singleton.

src/Microsoft.Health.Dicom.Api/Features/Partition/DataPartitionFeatureValidatorService.cs
bcarthic: Yeah it can be null, fixed it

src/Microsoft.Health.Dicom.Core/Exceptions/DataPartitionsFeatureCannotBeDisabledException.cs
bcarthic: Yes makes sense.

PR #1152, opened 2021-10-29
Data Partitions - New partition controller to get all partitions

bcarthic:
## Description
This PR updates Partition functionality from the persistence layer to the API surface. We have updated the client to allow get all partitions when the feature flag is enabled.

## Related issues
Addresses [[AB#86314](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/86314)]

Testing
We added a new E2E test to validate basic functionality.


src/Microsoft.Health.Dicom.Api/Controllers/PartitionController.cs
```diff
@@ -0,0 +1,67 @@
...
+using DicomAudit = Microsoft.Health.Dicom.Api.Features.Audit;
+
+namespace Microsoft.Health.Dicom.Api.Controllers
+{
+    [ApiVersion("1.0-prerelease")]
+    [ServiceFilter(typeof(DicomAudit.AuditLoggingFilterAttribute))]
+    public class PartitionController : Controller
+    {
+        private readonly IMediator _mediator;
+        private readonly ILogger<QueryController> _logger;
```
jovinson-ms, line 31: Should be `PartitionController`

jovinson-ms (approved):

src/Microsoft.Health.Dicom.Api/Controllers/PartitionController.cs
bcarthic: Fixed

PR #1158, opened 2021-11-04
Parallel tests

jovinson-ms:
## Description
This change cleans up a few artifacts related to tests and parallelizes the integration tests. The biggest gain here is seen by splitting schema upgrade tests into separate classes - see my comment. 

As a result, the integration test run time is cut from ~20 minutes to ~10 minutes:

### Previous run
image

### Current run
image

## Related issues
Addresses [AB#83141](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/83141)

## Testing
[PR test run](https://microsofthealthoss.visualstudio.com/DicomServer/_build/results?buildId=15868&view=results)


src/Microsoft.Health.Dicom.Tests.Common/Microsoft.Health.Dicom.Tests.Common.csproj
```diff
@@ -8,6 +8,7 @@
   <ItemGroup>
     <PackageReference Include="fo-dicom" Version="$(FoDicomVersion)" />
     <PackageReference Include="NSubstitute" Version="4.2.2" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.11.0" />
```
jovinson-ms, line 11: Resolves errors in test discovery.

src/Microsoft.Health.Dicom.Web/Microsoft.Health.Dicom.Web.csproj
```diff
@@ -32,10 +32,4 @@
     <Folder Include="wwwroot\" />
   </ItemGroup>
 
-  <ProjectExtensions>
```
jovinson-ms, line 35: Removing developer-specific file.

test/Microsoft.Health.Dicom.Tests.Integration/Persistence/SqlServerSchemaUpgradeTests.cs
```diff
@@ -15,7 +15,7 @@
 
 namespace Microsoft.Health.Dicom.Tests.Integration.Persistence
 {
-    public class SqlServerSchemaUpgradeTests
+    public class SqlServerSchemaUpgradeTests1
```
jovinson-ms, line 18: Tests within a collection (here created implicitly by a class) are always run sequentially - there's no built-in way to override this behavior. These tests take several minutes each on the build agent, so splitting them into separate classes is the easiest way to achieve parallelization. There is a workaround we could try, or just wait for xUnit 3: https://github.com/xunit/xunit/issues/1986

test/Microsoft.Health.Dicom.Tests.Integration/Persistence/SqlServerSchemaUpgradeTests.cs
```diff
@@ -108,6 +119,14 @@ public async Task GivenASchemaVersion_WhenApplyingDiffTwice_ShouldSucceed(int sc
             await snapshotFixture.DisposeAsync();
         }
 
+        public static IEnumerable<object[]> SchemaDiffVersions = Enumerable
+            .Range(start: SchemaVersionConstants.Min + 1, count: SchemaVersionConstants.Max - SchemaVersionConstants.Min)
+            .Select(x => new object[] { x })
+            .ToList();
```
bcarthic, line 125: Should we put this into a separate helper class

bcarthic (approved):

PR #1191, opened 2021-11-12
Add additional indexes for extended query tag and improve QIDO indexes

bcarthic:
## Description
This PR includes 
1. Introducing partitionKey to Extended query tag tables
1. Few small improvements to existing indexes on Instance table. Currently the cross-apply  queries are sorted by watermark column, but the watermark columns are part of include which results in index scan instead of an index seek. Moving the watermark to primary column will results in index seek.

Below are the query plan comparison

**Without  index change**
WithOutIndex

**With index change **
WithIndex

**Query**
```sql
SELECT  x.StudyInstanceUid, x.SeriesInstanceUid, x.SopInstanceUid, x.Watermark
FROM
( 
  SELECT st.StudyKey, se.SeriesKey
  FROM dbo.Study st
  INNER JOIN dbo.Series se
  ON se.StudyKey = st.StudyKey
  WHERE 1 = 1
  AND se.SeriesInstanceUid=@p0
  ORDER BY se.SeriesKey DESC
  OFFSET 0 ROWS
  FETCH NEXT 100 ROWS ONLY
) f
CROSS APPLY
(
    SELECT TOP 1 a.StudyInstanceUid, a.SeriesInstanceUid, a.SopInstanceUid,a.Watermark
    FROM dbo.Instance a
    WHERE 1 = 1
    AND a.StudyKey = f.StudyKey
    AND a.SeriesKey = f.SeriesKey
    AND a.Status = 1
    ORDER BY a.Watermark DESC
) x
OPTION(RECOMPILE)
```
## Related issues
Addresses
[[AB#86552](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/86552)].
[[AB#86014](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/86014)]

## Testing
Ran all the tests locally


jovinson-ms (approved):

src/Microsoft.Health.Dicom.SqlServer/Microsoft.Health.Dicom.SqlServer.csproj
```diff
@@ -72,6 +72,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <None Remove="Features\Schema\Migrations\7.diff.sql" />
```
smithago, line 75: What are these None remove tags helping with?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/AddInstanceV7.sql
```diff
@@ -0,0 +1,181 @@
...
+--     @doubleExtendedQueryTags
+--         * Double extended query tag data
+--     @dateTimeExtendedQueryTags
+--         * DateTime extended query tag data
+--     @personNameExtendedQueryTags
+--         * PersonName extended query tag data
+-- RETURN VALUE
+--     The watermark (version).
+------------------------------------------------------------------------
+CREATE OR ALTER PROCEDURE dbo.AddInstanceV7
```
smithago, line 52: Why do we need a new version? Since the signature is not changing? Basically tables and sprocs change together and client does not need to know anything the new prc will start pumping partitionKey.

Same may apply to other procedures.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/7.diff.sql
```diff
@@ -0,0 +1,1123 @@
...
+
+        END CATCH
+
+    COMMIT TRANSACTION
+END
+GO
+
+COMMIT TRANSACTION
+
+BEGIN TRANSACTION
```
smithago, line 842: Why are some indexes created online and some not? Why not create all indexes online without a transaction?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/AddInstanceV7.sql
bcarthic: No, this procedure in turn calls `IIndexInstanceCoreV7` sproc, which uses new added partitionKey column .

src/Microsoft.Health.Dicom.SqlServer/Microsoft.Health.Dicom.SqlServer.csproj
bcarthic: This is not needed, can be removed. Added by vs automatically

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/7.diff.sql
bcarthic: We include everything in a transaction to make it re-runnable. Since we are not checking IF EXISTS and IF NOT EXISTS for each one, we rely on the transaction to rollback

bcarthic: Existing indexes are dropped and created ONLINE and the new indexes are just created OFFLINE. I see we follow this pattern before.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/AddInstanceV7.sql
bcarthic: as discussed moved to older versions

smithago (approved):

PR #1200, opened 2021-11-16
AB#72602 - Reject implicit value representations

shivakumar-ms:
## Description
Added Transfer Syntax validation to reject if it is not Explicit VR.

## Related issues
Addresses [issue [AB#72602](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/72602)].

## Testing
Added test coverage to validate Explicit VR and Implicit VR transfer syntax.


src/Microsoft.Health.Dicom.Tests.Common/Samples.cs
```diff
@@ -215,5 +216,58 @@ private static IByteBuffer CreateRandomPixelData(int pixelDataSize)
 
             return new MemoryByteBuffer(result);
         }
+
+        public static IEnumerable<DicomTransferSyntax> GetAllDicomeTransferSyntax()
```
jovinson-ms, line 220: nit: extra `e` after `Dicom`

shivakumar-ms: Sorry about that.  Fixed.

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/StoreDatasetValidatorTests.cs
```diff
@@ -67,6 +67,20 @@ public async Task GivenAValidDicomDataset_WhenValidated_ThenItShouldSucceed()
...
 
+        [Theory]
+        [MemberData(nameof(GetNonExplicitVRTransferSyntax))]
+        public async Task GivenAValidDicomDatasetWithImplicitVR_WhenValidated_ThenItShouldThrowNotAcceptableException(DicomTransferSyntax transferSyntax)
+        {
+            var dicomDataset = Samples
+                .CreateRandomInstanceDataset(dicomTransferSyntax: transferSyntax)
+                .NotValidated();
+
+            var exception = await Record.ExceptionAsync(() => _dicomDatasetValidator.ValidateAsync(dicomDataset, requiredStudyInstanceUid: null));
```
jovinson-ms, line 78: Style question - why prefer this approach over `Assert.ThrowsAsync`?

jovinson-ms (approved):

jovinson-ms:
One other note - can you add a line to the [conformance statement](https://github.com/microsoft/dicom-server/blob/main/docs/resources/conformance-statement.md) stating that only explicit transfer syntaxes are accepted for STOW requests under the STOW section?

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/StoreDatasetValidatorTests.cs
shivakumar-ms: Good question. I guess we could change this to ThrowsAsync. Initially I was thinking about adding Assert to ex.Message to verify. Then I punted on it, as this check is not this tests responsibility.

shivakumar-ms:
> One other note - can you add a line to the [conformance statement](https://github.com/microsoft/dicom-server/blob/main/docs/resources/conformance-statement.md) stating that only explicit transfer syntaxes are accepted for STOW requests under the STOW section?

Done.

src/Microsoft.Health.Dicom.Core/Features/Validation/ValidationErrorCode.cs
```diff
@@ -54,6 +54,11 @@ public enum ValidationErrorCode : short
         /// </summary>
         UnexpectedVR = 0005,
 
+        /// <summary>
+        /// Implicit VR Transfer Syntax is not allowed.
+        /// </summary>
+        ImplicitVRNotAllowed = 0011,
```
wsugarman, line 60: Why not 6 instead of 11? This doesn't look like binary

bcarthic (approved):

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/StoreDatasetValidatorTests.cs
wsugarman: Yeah, I wouldn't worry too much about asserting the correct message. Asserting the type is sufficient. Otherwise, I think you open an unnecessary can of worms if you have multiple languages for your error messages.

wsugarman:
Would you also mind including the line `+semver: breaking` at the end of your commit message when merging in these changes? I forgot to annotate my `net6.0` as breaking for the next package release...!

wsugarman (approved):

PR #1228, opened 2021-12-08
Update DICOMCast table storage provider with managed identity support

brandonpollett:
## Description
This PR Updates the DICOMCast Table Storage provider to have managed identity support. 

## Related issues
Addresses [issue #].

## Testing
Manual testing with local emulator, remote connection string, remote azure credential.

### DIFF
--- a/converter/dicom-cast/src/Microsoft.Health.DicomCast.TableStorage/Features/Storage/TableClientReadWriteTestProivder.cs
+++ b/converter/dicom-cast/src/Microsoft.Health.DicomCast.TableStorage/Features/Storage/TableClientReadWriteTestProivder.cs
@@ -5,8 +5,8 @@
 
 using System.Threading;
 using System.Threading.Tasks;
+using Azure.Data.Tables;
 using EnsureThat;
-using Microsoft.Azure.Cosmos.Table;
 using Microsoft.Health.DicomCast.TableStorage.Configs;
 using Microsoft.Health.DicomCast.TableStorage.Features.Storage.Entities;
 
@@ -20,27 +20,23 @@ namespace Microsoft.Health.DicomCast.TableStorage.Features.Storage
         private const string TestTable = "testTable";
 
         /// <inheritdoc/>
-        public async Task PerformTestAsync(CloudTableClient client, TableDataStoreConfiguration configuration, CancellationToken cancellationToken = default)
+        public async Task PerformTestAsync(TableServiceClient testServiceClient, TableDataStoreConfiguration configuration, CancellationToken cancellationToken = default)
         {
-            EnsureArg.IsNotNull(client, nameof(client));
+            EnsureArg.IsNotNull(testServiceClient, nameof(testServiceClient));
             EnsureArg.IsNotNull(configuration, nameof(configuration));
 
-            CloudTable table = client.GetTableReference(TestTable);
-            await table.CreateIfNotExistsAsync(cancellationToken);
+            await testServiceClient.CreateTableIfNotExistsAsync(TestTable, cancellationToken: cancellationToken);
 
+            var tableClient = testServiceClient.GetTableClient(TestTable);
             var entity = new HealthEntity(TestPartitionKey, TestRowKey) { Data = TestData };
 
-            var insertOrMergeOperation = TableOperation.InsertOrMerge(entity);
+            await tableClient.UpsertEntityAsync(entity, cancellationToken: cancellationToken);
 
-            await table.ExecuteAsync(insertOrMergeOperation, cancellationToken);
+            await tableClient.GetEntityAsync<HealthEntity>(TestPartitionKey, TestRowKey, cancellationToken: cancellationToken);
 
-            var retrieveOperation = TableOperation.Retrieve<HealthEntity>(TestPartitionKey, TestRowKey);
-            await table.ExecuteAsync(retrieveOperation, cancellationToken);
+            await tableClient.DeleteEntityAsync(TestPartitionKey, TestRowKey, cancellationToken: cancellationToken);
 
-            var deleteOperation = TableOperation.Delete(entity);
-            await table.ExecuteAsync(deleteOperation, cancellationToken);
-
-            await table.DeleteAsync(cancellationToken);
+            await testServiceClient.DeleteTableAsync(TestTable, cancellationToken);
         }
     }
 }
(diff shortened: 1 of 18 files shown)


converter/dicom-cast/src/Microsoft.Health.DicomCast.TableStorage/Features/Storage/TableClientReadWriteTestProivder.cs
```diff
@@ -5,42 +5,41 @@
...
+using Azure.Data.Tables;
 using EnsureThat;
-using Microsoft.Azure.Cosmos.Table;
-using Microsoft.Health.DicomCast.TableStorage.Configs;
 using Microsoft.Health.DicomCast.TableStorage.Features.Storage.Entities;
 
 namespace Microsoft.Health.DicomCast.TableStorage.Features.Storage
 {
-    public class TableClientReadWriteTestProivder : ITableClientTestProvider
+    public class TableClientReadWriteTestProvider : ITableClientTestProvider
```
wsugarman, line 14: nit: file is still misspelled

converter/dicom-cast/src/Microsoft.Health.DicomCast.TableStorage/Features/Storage/TableClientReadWriteTestProivder.cs
```diff
@@ -5,42 +5,41 @@
...
-            var retrieveOperation = TableOperation.Retrieve<HealthEntity>(TestPartitionKey, TestRowKey);
-            await table.ExecuteAsync(retrieveOperation, cancellationToken);
+            await tableClient.GetEntityAsync<HealthEntity>(TestPartitionKey, TestRowKey, cancellationToken: cancellationToken);
 
-            var deleteOperation = TableOperation.Delete(entity);
-            await table.ExecuteAsync(deleteOperation, cancellationToken);
+            await tableClient.DeleteEntityAsync(TestPartitionKey, TestRowKey, cancellationToken: cancellationToken);
 
-            await table.DeleteAsync(cancellationToken);
+            await _testServiceClient.DeleteTableAsync(TestTable, cancellationToken);
```
wsugarman, line 42: Should we be wrapping the deletes in finally blocks? Or is that not important since we're upserting?

converter/dicom-cast/src/Microsoft.Health.DicomCast.TableStorage/Features/Storage/TableExceptionStore.cs
```diff
@@ -20,28 +20,25 @@ namespace Microsoft.Health.DicomCast.TableStorage.Features.Storage
...
 
         /// <inheritdoc/>
         public async Task WriteExceptionAsync(ChangeFeedEntry changeFeedEntry, Exception exceptionToStore, ErrorType errorType, CancellationToken cancellationToken)
         {
-            CloudTable table;
-            TableEntity entity;
-            string tableName;
-
             EnsureArg.IsNotNull(changeFeedEntry, nameof(changeFeedEntry));
+            string tableName;
```
wsugarman, line 41: nit: I think you could make this a switch expression if you want

converter/dicom-cast/src/Microsoft.Health.DicomCast.TableStorage/Features/Storage/TableServiceClientInitializer.cs
```diff
@@ -28,32 +29,29 @@ public class TableClientInitializer : ITableClientInitializer
...
-
-            CloudTableClient tableClient = storageAccount.CreateCloudTableClient(new TableClientConfiguration());
+            // TODO: Add retry policy for accessing the table storage
+            if (!string.IsNullOrWhiteSpace(_configuration.ConnectionString))
+            {
+                return new TableServiceClient(_configuration.ConnectionString);
+            }
 
-            return tableClient;
+            var credential = new DefaultAzureCredential(new DefaultAzureCredentialOptions { ManagedIdentityClientId = _configuration.ManagedIdentityClientId });
```
wsugarman, line 40: You can also let the Azure APIs handle this for you through the use of the package [`Microsoft.Extensions.Azure`](https://www.nuget.org/packages/Microsoft.Extensions.Azure/). They have a type `AzureComponentFactory` that has a method `CreateTokenCredential`. You give it the `IConfigurationSection` for the credentials and it creates the correct `TokenCredential` option (or the `DefaultAzureCredential` if none are specified). However, if we need logic like "use this managed identity if found, otherwise use defaults," then we may not be able to use it.

The configuration section looks for names like `Credential` = `"managedidentity"` and `ClientId`. I had to look up these values in their source code [here](https://github.com/Azure/azure-sdk-for-net/blob/e1ec26c5401e836fe3ddee778423b1a628949599/sdk/extensions/Microsoft.Extensions.Azure/src/Internal/ClientFactory.cs#L86).

wsugarman: If you went this route, you could have `Connection` either be a connection string or a section containing more configuration info about the connection.

converter/dicom-cast/src/Microsoft.Health.DicomCast.TableStorage/Registration/DicomCastTableRegistrationExtension.cs
```diff
@@ -61,19 +61,19 @@ private static IServiceCollection AddTableDataStore(this IServiceCollection serv
...
                 .AsSelf()
                 .AsService<IHostedService>()
                 .AsService<IRequireInitializationOnFirstRequest>();
 
-            serviceCollection.Add(sp => sp.GetService<TableClientProvider>().CreateTableClient())
+            serviceCollection.Add(sp => sp.GetService<TableServiceClientProvider>().CreateTableServiceClient())
                 .Singleton()
                 .AsSelf();
 
-            serviceCollection.Add<TableClientInitializer>()
```
wsugarman, line 74: You may want to use the registration methods that I think are from `Azure.Core`. That way the clients leverage the loggers from our dependency injection container. You can use the extension method `AddAzureClients` on an `IServiceCollection` and add your table client within the configure delegate. I did something similar for blob clients in healthcare shared components [here](https://github.com/microsoft/healthcare-shared-components/blob/713e0357a06ac697eda616cd9d0486089bb1efbd/src/Microsoft.Health.Blob/Registration/BlobClientRegistrationExtensions.cs#L124).

If you don't need to special-case the token credential and can let the built-in factories create it, you can probably invoke something like `AddTableServiceClient` with an `IConfiguration` containing both the authentication and the table info.

wsugarman (approved):

converter/dicom-cast/src/Microsoft.Health.DicomCast.TableStorage/Features/Storage/TableClientReadWriteTestProivder.cs
brandonpollett: I don't think it actually matters since we're doing both `CreateTableIfNotExistsAsync` and `UpsertEntityAsync`. This should mean that the method will complete regardless of the existing state of the table and entity.

PR #1251, opened 2021-12-22
Users/v shivakumar/ab#87263 refactor

shivakumar-ms:
## Description
Removed Metadata project. Merged Blob and Metadata store registrations.

## Related issues
Addresses [issue [AB#87263](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/87263)].

## Testing
Updated and executed unit tests locally.


src/Microsoft.Health.Dicom.Blob/Registration/DicomServerBuilderBlobRegistrationExtensions.cs
```diff
@@ -17,59 +17,113 @@ namespace Microsoft.Extensions.DependencyInjection
...
         /// <summary>
-        /// Adds the blob data store for the DICOM server.
+        /// Adds the data store for the DICOM server.
         /// </summary>
-        /// <param name="serverBuilder">The DICOM server builder instance.</param>
+        /// <param name="serverBuilder">The DICOM server builder.</param>
         /// <param name="configuration">The configuration for the server.</param>
-        /// <returns>The server builder.</returns>
-        public static IDicomServerBuilder AddBlobStorageDataStore(this IDicomServerBuilder serverBuilder, IConfiguration configuration)
+        /// <returns>The service collection.</returns>
```
jovinson-ms, line 25: Looks like we are still returning a serverBuilder, not service collection

src/Microsoft.Health.Dicom.Blob/Registration/DicomServerBuilderBlobRegistrationExtensions.cs
```diff
@@ -17,59 +17,113 @@ namespace Microsoft.Extensions.DependencyInjection
...
+            var blobConfig = configuration.GetSection(BlobServiceClientOptions.DefaultSectionName);
+            functionsBuilder.Services
+                .AddBlobServiceClient(blobConfig)
+                .AddPersistence<IMetadataStore, BlobMetadataStore, LoggingMetadataStore>()
+                .Configure<BlobContainerConfiguration>(MetadataContainerConfigurationAware.ConfigurationSectionName, c => c.ContainerName = containerName);
+
+            return functionsBuilder;
+        }
+
+        private static IServiceCollection AddPersistence(this IServiceCollection services)
```
brandonpollett, line 91: Is there much benefit from having this method vs replacing the one call to it with 
```
.AddPersistence<IFileStore, BlobFileStore, LoggingFileStore>()
.AddPersistence<IMetadataStore, BlobMetadataStore, LoggingMetadataStore>()
```

src/Microsoft.Health.Dicom.Blob/Utilities/MetadataContainerDetailAware.cs
```diff
@@ -0,0 +1,21 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Dicom.Blob.Utilities
+{
+    /// <summary>
+    ///
```
brandonpollett, line 9: No summary

src/Microsoft.Health.Dicom.Blob/Utilities/StoreConfigurationAware.cs
```diff
@@ -0,0 +1,23 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Dicom.Blob.Utilities
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    internal abstract class StoreConfigurationAware : IStoreConfigurationAware
```
brandonpollett, line 11: What about just having a constants file with the 4 things we need in it:
```
        internal const string MetadataConfigurationSectionName = "DicomWeb:MetadataStore";
        internal const string MetadataContainerName = "dicomMetadata";
        internal const string BlobConfigurationSectionName = "DicomWeb:DicomStore";
        internal const string BlobContainerName = "dicomBlob";
```

I think the only place this becomes a bit trickier is when registering the health checks, but I believe that can be fixed by adding a name as a parameter and using `AddTypeActivatedCheck` instead of `AddCheck`

brandonpollett, line 11: Also, I find this naming a bit weird. Seems like it should be something like `BlobConfigurationSection`

jovinson-ms: We could do that - one thing to note is that this is a preparation for adding a new blob type for workitems that will be in a separate container.

shivakumar-ms: > What about just having a constants file with the 4 things we need in it:
> 
> ```
>         internal const string MetadataConfigurationSectionName = "DicomWeb:MetadataStore";
>         internal const string MetadataContainerName = "dicomMetadata";
>         internal const string BlobConfigurationSectionName = "DicomWeb:DicomStore";
>         internal const string BlobContainerName = "dicomBlob";
> ```
> 
> I think the only place this becomes a bit trickier is when registering the health checks, but I believe that can be fixed by adding a name as a parameter and using `AddTypeActivatedCheck` instead of `AddCheck`

"AddTypeActivatedCheck" does take params. But it requires us passing in the resolved instances for other parameters injected in that consumer.

shivakumar-ms: Sure will change it. @josiahvinson @bcarthic do you have any preference or thoughts on the naming? Otherwise will go with Brandon's suggestion.

src/Microsoft.Health.Dicom.Blob/Registration/DicomServerBuilderBlobRegistrationExtensions.cs
shivakumar-ms: No other benefit other than increasing the lines of code in a single method, as we continue to add Workitem related changes. I can move it into the main method.

src/Microsoft.Health.Dicom.Blob/Registration/DicomServerBuilderBlobRegistrationExtensions.cs
```diff
@@ -17,59 +17,113 @@ namespace Microsoft.Extensions.DependencyInjection
...
+        }
+
+        /// <summary>
+        /// Adds the metadata store for the DICOM functions.
+        /// </summary>
+        /// <param name="functionsBuilder">The DICOM functions builder instance.</param>
+        /// <param name="containerName">The name of the metadata container.</param>
+        /// <param name="configuration">The configuration for the function.</param>
+        /// <returns>The functions builder.</returns>
+        public static IDicomFunctionsBuilder AddMetadataStorageDataStore(
```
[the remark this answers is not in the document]
shivakumar-ms: I just moved these from a project that we removed. I'll separate this into it's own registration extension.

src/Microsoft.Health.Dicom.Blob/Registration/DicomServerBuilderBlobRegistrationExtensions.cs
```diff
@@ -17,59 +17,113 @@ namespace Microsoft.Extensions.DependencyInjection
...
-        /// Adds the blob data store for the DICOM server.
+        /// Adds the data store for the DICOM server.
         /// </summary>
-        /// <param name="serverBuilder">The DICOM server builder instance.</param>
+        /// <param name="serverBuilder">The DICOM server builder.</param>
         /// <param name="configuration">The configuration for the server.</param>
-        /// <returns>The server builder.</returns>
-        public static IDicomServerBuilder AddBlobStorageDataStore(this IDicomServerBuilder serverBuilder, IConfiguration configuration)
+        /// <returns>The service collection.</returns>
+        public static IDicomServerBuilder AddDataStore(this IDicomServerBuilder serverBuilder, IConfiguration configuration)
```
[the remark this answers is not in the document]
shivakumar-ms: I sort of agree. But just wondering whether changing this method name to "BlobDataStore" could cause confusion. As this method takes care of registering both Blob Store as well as Metadata store. and we are going to be adding Workitem store in this.

src/Microsoft.Health.Dicom.Blob/Registration/DicomServerBuilderBlobRegistrationExtensions.cs
shivakumar-ms: Done.

src/Microsoft.Health.Dicom.Blob/Registration/DicomServerBuilderBlobRegistrationExtensions.cs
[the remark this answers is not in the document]
shivakumar-ms: Done

src/Microsoft.Health.Dicom.Blob/Registration/DicomServerBuilderBlobRegistrationExtensions.cs
shivakumar-ms: Done.

src/Microsoft.Health.Dicom.Blob/Registration/DicomServerBuilderBlobRegistrationExtensions.cs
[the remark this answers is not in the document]
bcarthic: We can separate methods for adding different stores.

src/Microsoft.Health.Dicom.Blob/Utilities/StoreConfigurationAware.cs
bcarthic: i am fine with BlobConfigurationSection

src/Microsoft.Health.Dicom.Blob/Registration/DicomServerBuilderBlobRegistrationExtensions.cs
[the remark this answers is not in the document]
jovinson-ms: I like `BlobDataStores` since the method will add multiple stores: `File`, `Metadata`, and `Workitem`.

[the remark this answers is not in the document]
shivakumar-ms: Done

src/Microsoft.Health.Dicom.Blob/Utilities/StoreConfigurationAware.cs
shivakumar-ms: Done

src/Microsoft.Health.Dicom.Blob/Registration/DicomServerBuilderBlobRegistrationExtensions.cs
```diff
@@ -17,58 +18,83 @@ namespace Microsoft.Extensions.DependencyInjection
...
-
         /// <summary>
-        /// Adds the blob data store for the DICOM server.
+        /// Adds the blob data stores for the DICOM server.
         /// </summary>
         /// <param name="serverBuilder">The DICOM server builder instance.</param>
         /// <param name="configuration">The configuration for the server.</param>
         /// <returns>The server builder.</returns>
-        public static IDicomServerBuilder AddBlobStorageDataStore(this IDicomServerBuilder serverBuilder, IConfiguration configuration)
+        public static IDicomServerBuilder AddDataStores(this IDicomServerBuilder serverBuilder, IConfiguration configuration)
```
jovinson-ms, line 27: Can we change this to `AddBlobDataStores`? I think the important thing here is that we are accessing blob storage specifically, and I'd like that to be clear when viewing the registration methods.

jovinson-ms (approved):

PR #1258, opened 2022-01-05
cleanup dicom cast logging to match dicom service

abdullah248:
## Description
Change Dicom Cast logging for uids to match that of dicom service for log redaction.

## Related issues
[AB#87220](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/87220)

## Testing

### DIFF
--- a/converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/ChangeFeedProcessor.cs
+++ b/converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/ChangeFeedProcessor.cs
@@ -92,9 +92,9 @@ namespace Microsoft.Health.DicomCast.Core.Features.Worker
                     {
                         if (ex is FhirNonRetryableException || ex is DicomTagException || ex is TimeoutRejectedException)
                         {
-                            string studyUid = changeFeedEntry.StudyInstanceUid;
-                            string seriesUid = changeFeedEntry.SeriesInstanceUid;
-                            string instanceUid = changeFeedEntry.SopInstanceUid;
+                            string studyInstanceUid = changeFeedEntry.StudyInstanceUid;
+                            string seriesInstanceUid = changeFeedEntry.SeriesInstanceUid;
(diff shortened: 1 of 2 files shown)


alishahahmed (approved):

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/ChangeFeedProcessor.cs
```diff
@@ -114,7 +114,7 @@ public async Task ProcessAsync(TimeSpan pollIntervalDuringCatchup, CancellationT
                                 errorType,
                                 cancellationToken);
 
-                            _logger.LogError("Failed to process DICOM event with SequenceID: {SequenceId}, StudyUid: {StudyUid}, SeriesUid: {SeriesUid}, instanceUid: {InstanceUid}  and will not be retried further. Continuing to next event.", changeFeedEntry.Sequence, studyUid, seriesUid, instanceUid);
+                            _logger.LogError("Failed to process DICOM event with SequenceID: {SequenceId}, StudyUid: {StudyInstanceUid}, SeriesUid: {SeriesInstanceUid}, instanceUid: {SopInstanceUid}  and will not be retried further. Continuing to next event.", changeFeedEntry.Sequence, studyInstanceUid, seriesInstanceUid, sopInstanceUid);
```
alishahahmed, line 117: I would just confirm once if the property name is being matched with IgnoreCase. If not, just double check if the property name should be StudyInstanceUid or studyInstanceUid.

alishahahmed: Same goes for other property names as well.

abdullah248: As we talked offline, looks like it is not ignore case, but it needs to be uppercase here so will handle on redaction side.

PR #1266, opened 2022-01-09
Preserve Function Configuration as Activity Input for Re-playability

wsugarman:
## Description
Some activities rely on configuration values, like the size of batches when re-indexing. However, if this value were to change it would change the output of the method. So, this must be preserved in the input so it can be replayed properly.

- Create new "V2" versions of the `GetInstanceBatchesAsync` and `ReindexBatchAsync` activities
  - Forward calls from the old versions to the new ones by populating the input from the config
  - Mark the old versions obsolete
  - Create new classes to encapsulate the input
- Update packages

## Related issues
N/A

## Testing
Test pipeline


src/Microsoft.Health.Dicom.Operations/Indexing/ReindexDurableFunction.Activity.cs
```diff
@@ -92,27 +92,42 @@ public partial class ReindexDurableFunction
         /// A task representing the asynchronous get operation. The value of its <see cref="Task{TResult}.Result"/>
         /// property contains a list of batches as defined by their smallest and largest watermark.
         /// </returns>
+        [Obsolete("Please use GetInstanceBatchesAsyncV2 instead.")]
```
wsugarman, line 95: whoops -- name is wrong

PR #1270, opened 2022-01-12
Updates to DICOM Cast Cert Authentication

brandonpollett:
## Description
This PR introduces some changes to the registration of the authentication providers for DICOM Cast such that they can be take more advantage of the options pattern.

## Related issues
Addresses [AB#87283](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/87283)

## Testing
Manually tested downstream.


converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Modules/DicomModule.cs
```diff
@@ -31,15 +32,18 @@ public void Load(IServiceCollection services)
...
+            var dicomWebConfiguration = new DicomWebConfiguration();
+            IConfigurationSection dicomWebConfigurationSection = _configuration.GetSection(DicomWebConfigurationSectionName);
+            dicomWebConfigurationSection.Bind(dicomWebConfiguration);
+
+            services.AddSingleton(Options.Create(dicomWebConfiguration));
 
             services.AddHttpClient<IDicomWebClient, DicomWebClient>(sp =>
                 {
                     sp.BaseAddress = dicomWebConfiguration.Endpoint;
+
```
jovinson-ms, line 44: nit: extra newline

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Modules/ServiceCollectionExtensions.cs
```diff
@@ -15,68 +16,78 @@ namespace Microsoft.Health.DicomCast.Core.Modules
...
             EnsureArg.IsNotNull(serviceCollection, nameof(serviceCollection));
             EnsureArg.IsNotNull(managedIdentityCredentialConfiguration, nameof(managedIdentityCredentialConfiguration));
             EnsureArg.IsNotNullOrWhiteSpace(name, nameof(name));
 
+            serviceCollection.Configure<ManagedIdentityCredentialConfiguration>(name, managedIdentityCredentialConfiguration);
+
             serviceCollection.Add(provider =>
                 {
-                    var options = Options.Create(managedIdentityCredentialConfiguration);
+                    IOptionsMonitor<ManagedIdentityCredentialConfiguration> options = provider.GetService<IOptionsMonitor<ManagedIdentityCredentialConfiguration>>();
```
jovinson-ms, line 29: `Monitor` will change values mid-request, is that preferable to `Snapshot`'s behavior? I'm not familiar enough with the network patterns here to have strong opinion.

jovinson-ms (approved):

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Extensions/IHttpClientBuilderRegistrationExtensions.cs
```diff
@@ -13,31 +14,33 @@ namespace Microsoft.Health.DicomCast.Core.Extensions
...
-            if (!authenticationConfiguration.Enabled)
+            var auth = new AuthenticationConfiguration();
+            authenticationConfigurationSection.Bind(auth);
+            if (!auth.Enabled)
             {
                 return;
             }
 
-            switch (authenticationConfiguration.AuthenticationType)
+            switch (auth.AuthenticationType)
```
wsugarman, line 31: As talked about offline, I think we can push this all off into a separate class loaded via DI such that the configuration resolution binds later, but it's not necessary. It would be sort of how like `ILogger` providers are loaded, or even `IOptions` itself.

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Modules/ServiceCollectionExtensions.cs
wsugarman: It's definitely a bit odd because we're creating a `Singleton` service that relies on options where we have a monitor but are removing it. So if there are changes, this service won't see them. But I think Brandon is going to use `IOptionsMonitor` for these providers in a future PR. Then this custom service becomes moot.

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Extensions/IHttpClientBuilderRegistrationExtensions.cs
wsugarman: In theory, you could even create services for every auth scenario, but only read the correct service based on the config when creating an instance of `NamedCredentialProvider`

wsugarman (approved):

PR #1274, opened 2022-01-13
Fix Delete Extended Query Tag Sproc

wsugarman:
## Description
If there is a problem, like a timeout when acquiring the lock, between the two transactions in the `DeleteExtendedQueryTag` sproc, then the status will be changed without any tags being deleted. Furthermore, subsequent calls will be unable to delete the tag.

Changes:
- Add new V8 schema with a new version of `DeleteExtendedQueryTag` that ignores the current state of the tag, as the operation is synchronous
- Add missing database settings from V1 into the script generation

## Related issues
[AB#87827](https://microsofthealth.visualstudio.com/Health/_workitems/edit/87827)

## Testing
Pass new test that validates bug fix


src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/SqlExtendedQueryTagStoreV2.cs
```diff
@@ -134,29 +134,63 @@ internal static AddExtendedQueryTagsInputTableTypeV1Row ToAddExtendedQueryTagsIn
...
+            using SqlConnectionWrapper sqlConnectionWrapper = await ConnectionWrapperFactory.ObtainSqlConnectionWrapperAsync(cancellationToken);
+            using SqlCommandWrapper sqlCommandWrapper = sqlConnectionWrapper.CreateSqlCommand();
 
-                try
-                {
-                    await sqlCommandWrapper.ExecuteNonQueryAsync(cancellationToken);
-                }
-                catch (SqlException ex)
+            sqlCommandWrapper.CommandText = @"
+SET NOCOUNT     ON
```
alishahahmed, line 141: Would it be better to create a fallback stored procedure and call that in case of an error?

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/SqlExtendedQueryTagStoreV2.cs
```diff
@@ -134,29 +134,63 @@ internal static AddExtendedQueryTagsInputTableTypeV1Row ToAddExtendedQueryTagsIn
...
+
+    SELECT @tagKey = TagKey
+    FROM dbo.ExtendedQueryTag
+    WHERE dbo.ExtendedQueryTag.TagPath = @tagPath
+
+    -- Check existence
+    IF @@ROWCOUNT = 0
+        THROW 50404, 'extended query tag not found', 1
+
+    -- Delete index data
```
alishahahmed, line 156: I am not sure what other statuses are there, but wouldn't it be safer to only execute this if the state is "Deleting"?

wsugarman: Not exactly. You should be able to delete in the middle of re-indexing (so you can cancel) and after re-indexing. In fact, it was a bit premature to add the `Deleting` status as the deletion operation is synchronous (users in theory should never see `Deleting` for now).

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/SqlExtendedQueryTagStoreV2.cs
wsugarman: What do you mean? As opposed to executing this query?

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/SqlExtendedQueryTagStoreV2.cs
```diff
@@ -134,29 +134,63 @@ internal static AddExtendedQueryTagsInputTableTypeV1Row ToAddExtendedQueryTagsIn
 
         public override async Task DeleteExtendedQueryTagAsync(string tagPath, string vr, CancellationToken cancellationToken = default)
```
pengchen0692, line 135: Why not rollout a new sql schema to fix the problem as we normally do?

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/SqlExtendedQueryTagStoreV2.cs
```diff
@@ -134,29 +134,63 @@ internal static AddExtendedQueryTagsInputTableTypeV1Row ToAddExtendedQueryTagsIn
...
-                try
-                {
-                    await sqlCommandWrapper.ExecuteNonQueryAsync(cancellationToken);
-                }
-                catch (SqlException ex)
+            sqlCommandWrapper.CommandText = @"
+SET NOCOUNT     ON
+SET XACT_ABORT  ON
+
+BEGIN TRANSACTION
```
smithago, line 144: I like this is more clean. What will happen if the Tag was still being indexed by the Async job. Does it know that it needs to cancel after this foes through?

Also, do we know why the second transaction timed out, is there a overlocking issue we need to fix?

wsugarman: I've reverted back to 2 transaction. The lock + status should prevent the re-indexing activity from attempting to re-index additional rows.
But yes, there may be an issue with too many locks. Tracking the bug here: https://microsofthealth.visualstudio.com/Health/_workitems/edit/87860

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/8.sql
```diff
@@ -0,0 +1,2441 @@
+
+/*************************************************************************************************
+    Auto-Generated from Sql build task. Do not manually edit it.
```
pengchen0692: 8.sql doesn't have the `Configure database` section

pengchen0692 (approved):

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/8.sql
wsugarman: it just got reformatted

alishahahmed (approved):

PR #1275, opened 2022-01-13
Fix Dependabot Alert

alishahahmed:
## Description
As per the [alert] (https://github.com/microsoft/dicom-server/security/dependabot/tools/dicom-web-electron/package-lock.json/follow-redirects/open), follows-redirect version has been updated to 1.14.7.


wsugarman:
Can you add our electron app to the dependabot config?

alishahahmed:
> Can you add our electron app to the dependabot config?

Updated.

wsugarman (approved):

PR #1283, opened 2022-01-16
Centralize and Expand Json Serializer Options

wsugarman:
## Description
- Expand the set of specified JSON serializer options for stricter API contracts
- Centralize the options for consistent rules across services

## Related issues
N/A

## Testing
Pipeline + New tests

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Extensions/JsonSerializerOptionsExtensions.cs
+++ b/src/Microsoft.Health.Dicom.Core/Extensions/JsonSerializerOptionsExtensions.cs
@@ -1,0 +1,49 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using EnsureThat;
+using FellowOakDicom.Serialization;
+using Microsoft.Health.Dicom.Core.Serialization;
+
+namespace Microsoft.Health.Dicom.Core.Extensions
+{
+    /// <summary>
+    /// Provides <see langword="static"/> methods configuring <see cref="JsonSerializerOptions"/>.
+    /// </summary>
+    public static class JsonSerializerOptionsExtensions
+    {
+        /// <summary>
+        /// Configures an instance of <see cref="JsonSerializerOptions"/> for usage within
+        /// the DICOM server and related services.
+        /// </summary>
+        /// <param name="options">A set of existing options.</param>
+        /// <exception cref="ArgumentNullException"><paramref name="options"/> is <see langword="null"/>.</exception>
+        public static void ConfigureDefaultDicomSettings(this JsonSerializerOptions options)
+        {
+            EnsureArg.IsNotNull(options, nameof(options));
+
+            options.Converters.Clear();
+            options.Converters.Add(new StrictStringEnumConverterFactory());
+            options.Converters.Add(new DicomJsonConverter(writeTagsAsKeywords: false));
+
+            options.AllowTrailingCommas = false;
+            options.DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull;
+            options.DictionaryKeyPolicy = JsonNamingPolicy.CamelCase;
+            options.Encoder = null;
+            options.IgnoreReadOnlyFields = false;
+            options.IgnoreReadOnlyProperties = false;
+            options.IncludeFields = false;
+            options.MaxDepth = 0;
+            options.NumberHandling = JsonNumberHandling.Strict;
+            options.PropertyNameCaseInsensitive = true;
+            options.PropertyNamingPolicy = JsonNamingPolicy.CamelCase;
+            options.ReadCommentHandling = JsonCommentHandling.Disallow;
+            options.WriteIndented = false;
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


src/Microsoft.Health.Dicom.Core/Extensions/JsonSerializerOptionsExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// </summary>
+    public static class JsonSerializerOptionsExtensions
+    {
+        /// <summary>
+        /// Configures an instance of <see cref="JsonSerializerOptions"/> for usage within
+        /// the DICOM server and related services.
+        /// </summary>
+        /// <param name="options">A set of existing options.</param>
+        /// <exception cref="ArgumentNullException"><paramref name="options"/> is <see langword="null"/>.</exception>
+        public static void ConfigureDefaultDicomSettings(this JsonSerializerOptions options)
```
jovinson-ms, line 26: General note - why specify default settings here? Do we want to pin to this version of defaults?

src/Microsoft.Health.Dicom.Core/Extensions/JsonSerializerOptionsExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            EnsureArg.IsNotNull(options, nameof(options));
+
+            options.Converters.Clear();
+            options.Converters.Add(new StrictStringEnumConverterFactory());
+            options.Converters.Add(new DicomJsonConverter(writeTagsAsKeywords: false));
+
+            options.AllowTrailingCommas = false;
```
jovinson-ms, line 34: We could make a case to allow as part of being a lenient consumer.

src/Microsoft.Health.Dicom.Core/Extensions/JsonSerializerOptionsExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+            options.DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull;
+            options.DictionaryKeyPolicy = JsonNamingPolicy.CamelCase;
+            options.Encoder = null;
+            options.IgnoreReadOnlyFields = false;
```
jovinson-ms, line 38: If fields are not included, readonly fields will be ignored, right?

src/Microsoft.Health.Dicom.Core/Extensions/JsonSerializerOptionsExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+            options.IgnoreReadOnlyProperties = false;
+            options.IncludeFields = false;
+            options.MaxDepth = 0;
+            options.NumberHandling = JsonNumberHandling.Strict;
+            options.PropertyNameCaseInsensitive = true;
+            options.PropertyNamingPolicy = JsonNamingPolicy.CamelCase;
+            options.ReadCommentHandling = JsonCommentHandling.Disallow;
```
jovinson-ms, line 45: Same above, I would suggest `Skip`

src/Microsoft.Health.Dicom.Core/Extensions/JsonSerializerOptionsExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+
+            options.AllowTrailingCommas = false;
+            options.DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull;
+            options.DictionaryKeyPolicy = JsonNamingPolicy.CamelCase;
+            options.Encoder = null;
+            options.IgnoreReadOnlyFields = false;
+            options.IgnoreReadOnlyProperties = false;
+            options.IncludeFields = false;
+            options.MaxDepth = 0;
+            options.NumberHandling = JsonNumberHandling.Strict;
```
jovinson-ms, line 42: Should we use `AllowReadingFromString`? There are Dicom VRs like DecimalString.

src/Microsoft.Health.Dicom.Core/Extensions/JsonSerializerOptionsExtensions.cs
wsugarman: Well, it is really in case of breaking changes in the underlying json serializer libs. If those defaults change in like v7, our web API should change its behavior (unless we explicitly want it to)

src/Microsoft.Health.Dicom.Core/Extensions/JsonSerializerOptionsExtensions.cs
wsugarman: Yeah, that's correct.

src/Microsoft.Health.Dicom.Core/Extensions/JsonSerializerOptionsExtensions.cs
wsugarman: This is actually ok. I just tested this myself, and I can add a test, but the Dicom JSON converter must recognize both formats as acceptable regardless of token kind.

jovinson-ms (approved):

src/Microsoft.Health.Dicom.Core/Extensions/JsonSerializerOptionsExtensions.cs
wsugarman: As agreed offline, we'll keep these defaults

src/Microsoft.Health.Dicom.Core/Extensions/JsonSerializerOptionsExtensions.cs
wsugarman: I confirmed this behavior offline

src/Microsoft.Health.Dicom.Core/Extensions/JsonSerializerOptionsExtensions.cs
wsugarman: Done

src/Microsoft.Health.Dicom.Core/Extensions/JsonSerializerOptionsExtensions.cs
wsugarman: I have added a test to illustrate the dicom tag parsing in unaffected

PR #1291, opened 2022-01-18
Sequence element parsing

bcarthic:
## Description
This PR changes includes DICOM Sequence element parsing.

Currently we are only supporting 2 levels. Nested sequence elements are not supported.

## Related issues
Addresses [AB#87263](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/87263)

## Testing
Added new unit tests to test parsing and matching sequence elements


src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/9.diff.sql
```diff
@@ -1666,31 +1666,31 @@ BEGIN
 
         -- ReferencedRequestSequence.Accesionnumber
         INSERT INTO dbo.WorkitemQueryTag (TagKey, TagPath, TagVR)
-        VALUES (NEXT VALUE FOR TagKeySequence, '0040A370.00080050', 'SH')
+        VALUES (NEXT VALUE FOR TagKeySequence, '0040A370.00080050', 'SQ')
 
         -- ReferencedRequestSequence.Requested​Procedure​ID
         INSERT INTO dbo.WorkitemQueryTag (TagKey, TagPath, TagVR)
-        VALUES (NEXT VALUE FOR TagKeySequence, '0040A370.00401001', 'SH')
+        VALUES (NEXT VALUE FOR TagKeySequence, '0040A370.00401001', 'SQ')
```
pengchen0692, line 1673: do we want to get rid of `.` to save a bit space, since tagpath always be 8 letter?

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
```diff
@@ -27,24 +28,79 @@ public bool TryParse(string dicomTagPath, out DicomTag[] dicomTags, bool support
...
+
+
+        private static DicomTag[] ParseMultipleStandardTag(string dicomTagPath)
+        {
+            var paths = dicomTagPath.Split(".");
+            var dicomTags = new List<DicomTag>();
+
+            foreach (var path in paths)
+            {
+                dicomTags.Add(GetTagFromKeywordOrNumber(path));
```
pengchen0692, line 80: Do we want to handle case where `GetTagFromKeywordOrNumber` returns `null`? which indicates the input is invalid.

bcarthic: It is handled in the parent method, where we check for if it is not null. If any one of them is null we return false for parsing the elements. A unit test also covers that scenario

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/9.diff.sql
bcarthic: Yeah we thought about that, but keeping it for now. May be if we support real sequence parsing then we can get rid of all these. We need a major refactoring

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -329,5 +329,35 @@ public static void ValidateQueryTag(this DicomDataset dataset, QueryTag queryTag
...
+            var foundDatasets = new List<DicomDataset>();
+
+            var foundSequence = dataset.GetDicomItem<DicomSequence>(searchSequence.Tag);
+            if (foundSequence == null) return foundDatasets;
+
+            var searchDataset = searchSequence.Items.FirstOrDefault();
+
+            foreach (var childDataset in foundSequence.Items)
+            {
+                var elements = childDataset.Intersect(searchDataset, new DicomTagComparer());
```
smithago, line 353: What does this return?

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -329,5 +329,35 @@ public static void ValidateQueryTag(this DicomDataset dataset, QueryTag queryTag
...
+        /// <param name="searchSequence">The Dicom sequence modelling the path.</param>
+        /// <returns>Lists of DicomDataset that matches the search sequence.</returns>
+        public static IEnumerable<DicomDataset> SearchSequencePath(this DicomDataset dataset, DicomSequence searchSequence)
+        {
+            EnsureArg.IsNotNull(dataset, nameof(dataset));
+            EnsureArg.IsNotNull(searchSequence, nameof(searchSequence));
+
+            var foundDatasets = new List<DicomDataset>();
+
+            var foundSequence = dataset.GetDicomItem<DicomSequence>(searchSequence.Tag);
```
smithago, line 346: GetSequence?

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -329,5 +329,35 @@ public static void ValidateQueryTag(this DicomDataset dataset, QueryTag queryTag
...
             }
         }
+
+        /// <summary>
+        /// Gets DicomDatasets that matches a DicomSequence.
+        /// </summary>
+        /// <param name="dataset">The DicomDataset to be traversed.</param>
+        /// <param name="searchSequence">The Dicom sequence modelling the path.</param>
+        /// <returns>Lists of DicomDataset that matches the search sequence.</returns>
+        public static IEnumerable<DicomDataset> SearchSequencePath(this DicomDataset dataset, DicomSequence searchSequence)
```
smithago, line 339: I am trying to understand this API. Are you trying to find an array of values that match the TagPath?

If that is the case, is there a opportunity to simplify the input and return type?

Input can be TagPath and output can be List of values based on the VR type

smithago: Also to be consistent call the Method GetSequencePathValues<T>()

src/Microsoft.Health.Dicom.Core/Features/Common/DicomValuelessItem.cs
```diff
@@ -0,0 +1,27 @@
...
+
+using System.Linq;
+using FellowOakDicom;
+
+namespace Microsoft.Health.Dicom.Core.Features.Common
+{
+    /// <summary>
+    /// A DicomItem with a tag but no value - used to model either concrete DicomElements or DicomSequences.
+    /// </summary>
+    public class DicomValuelessItem : DicomItem
```
smithago, line 14: What is DicomValueLessItem adding to DicomItem? Do we need this?

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
```diff
@@ -27,24 +28,79 @@ public bool TryParse(string dicomTagPath, out DicomTag[] dicomTags, bool support
                 return false;
             }
 
+            DicomTag dicomTag = GetTagFromKeywordOrNumber(dicomTagPath);
 
+            dicomTags = new DicomTag[] { dicomTag };
+            return dicomTag != null;
+        }
+
+        public bool TryParseToDicomItem(string dicomTagPath, out DicomItem dicomItem)
```
smithago, line 37: What value are you getting from returing the DicomItem? Is'nt it implicit that the more than one tag returned is a SQ?

Can we change the above TryParse(supportMultiple = true)? for WorkItem? and then later for DicomInstance?

You have one method that handles all cases

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
```diff
@@ -27,24 +28,79 @@ public bool TryParse(string dicomTagPath, out DicomTag[] dicomTags, bool support
...
+            if (dicomTags.Length == 1)
+            {
+                var tag = GetTagFromKeywordOrNumber(dicomTagPath);
+                if (tag != null) dicomItem = new DicomValuelessItem(tag);
+                return tag != null;
+            }
+
+            if (dicomTags.Length > 2)
+            {
+                throw new NotImplementedException(DicomCoreResource.NestedSequencesNotSupported);
```
smithago, line 57: Is this tracked for 500 return code? We general wrap all exceptions with DicomValidationExceptions

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
```diff
@@ -27,24 +28,79 @@ public bool TryParse(string dicomTagPath, out DicomTag[] dicomTags, bool support
...
+            return dicomTags.ToArray();
+        }
+
+        private static DicomTag GetTagFromKeywordOrNumber(string dicomTagPath)
+        {
             DicomTag dicomTag = ParseStardandDicomTagKeyword(dicomTagPath);
 
             if (dicomTag == null)
             {
                 dicomTag = ParseDicomTagNumber(dicomTagPath);
```
smithago, line 92: If this returns null? We don't care?

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
```diff
@@ -27,24 +28,79 @@ public bool TryParse(string dicomTagPath, out DicomTag[] dicomTags, bool support
...
+
+            foreach (var path in paths)
+            {
+                dicomTags.Add(GetTagFromKeywordOrNumber(path));
+            }
+
+            return dicomTags.ToArray();
+        }
+
+        private static DicomTag GetTagFromKeywordOrNumber(string dicomTagPath)
```
smithago, line 86: Can we keep the names consistent Parse*

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
```diff
@@ -27,24 +28,79 @@ public bool TryParse(string dicomTagPath, out DicomTag[] dicomTags, bool support
...
+
+                dicomItem = sequenceItem;
+            }
+
+
+            return dicomTags[0] != null && dicomTags[1] != null;
+        }
+
+
+        private static DicomTag[] ParseMultipleStandardTag(string dicomTagPath)
```
smithago, line 73: ParseSequenceStandardTag?

src/Microsoft.Health.Dicom.Core/Features/ExtendedQueryTag/QueryTag.cs
```diff
@@ -45,6 +46,20 @@ public QueryTag(ExtendedQueryTagStoreEntry entry)
             ExtendedQueryTagStoreEntry = entry;
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="QueryTag"/> class.
+        /// </summary>
+        /// <remarks>Used for constructing from <see cref="WorkitemQueryTagStoreEntry"/> (to model sequences).</remarks>
+        /// <param name="entry">The WorkitemQueryTagStore entry.</param>
+        public QueryTag(WorkitemQueryTagStoreEntry entry)
```
smithago, line 54: Consider separating out to different classes with a base abstract class

src/Microsoft.Health.Dicom.Core/Features/Workitem/WorkitemQueryTagService.cs
```diff
@@ -53,9 +58,19 @@ public async Task<IReadOnlyCollection<WorkitemQueryTagStoreEntry>> GetQueryTagsA
...
         {
-            return await _indexWorkitemStore.GetWorkitemQueryTagsAsync(cancellationToken);
+            var workitemQueryTags = await _indexWorkitemStore.GetWorkitemQueryTagsAsync(cancellationToken);
+
+            foreach (var tag in workitemQueryTags)
+            {
+                if (_dicomTagParser.TryParseToDicomItem(tag.Path, out var dicomItem))
+                {
+                    tag.Item = dicomItem;
+                }
```
smithago, line 70: Log if something falses to parse

src/Microsoft.Health.Dicom.Core/Features/Workitem/WorkitemQueryTagService.cs
```diff
@@ -5,23 +5,28 @@
...
+using Microsoft.Health.Dicom.Core.Features.ExtendedQueryTag;
 using Microsoft.Health.Dicom.Features.Common;
 
 namespace Microsoft.Health.Dicom.Core.Features.Workitem
 {
     public class WorkitemQueryTagService : IWorkitemQueryTagService, IDisposable
     {
         private readonly IIndexWorkitemStore _indexWorkitemStore;
-        private readonly AsyncCache<IReadOnlyCollection<WorkitemQueryTagStoreEntry>> _queryTagCache;
+        private readonly AsyncCache<IReadOnlyCollection<QueryTag>> _queryTagCache;
```
smithago, line 21: What is AsyncCache?  Why not use MemoryCache?

src/Microsoft.Health.Dicom.Core/Features/Workitem/WorkitemQueryTagStoreEntry.cs
```diff
@@ -24,5 +25,11 @@ public WorkitemQueryTagStoreEntry(int key, string path, string vr)
         /// Key of this extended query tag entry.
         /// </summary>
         public int Key { get; }
+
+        /// <summary>
+        /// Get the DicomItem that is the representation of the path for this tag
+        /// This is populated while fetching query tag from the db
+        /// </summary>
+        public DicomItem Item { get; set; }
```
smithago, line 33: Is'nt DicomItem a Tag and VR, how can this represent the TagPath?

wsugarman:
Unrelated to this specific PR, but related to the feature, should we re-name all of the SQL tables and data type that house the actual extended query tag index (eg. `dbo.ExtendedQueryTagString`, dbo.ExtendedQueryTagDateTime`, etc)? I think that adds quite a bit of confusion to the data model. We can still keep the `dbo.ExtendedQueryTag` table, as there is a corresponding work item table. Perhaps we can call it `dbo.Index[Type]Data`? Or something to allude that it is a generic index? `TagKey` could become like `ResourceKey` to match `ResourceType`?

src/Microsoft.Health.Dicom.Core/Features/Workitem/WorkitemQueryTagService.cs
wsugarman: `AsyncCache<T>` is a type I wrote for resolving schema versions. It's meant to be like an `async` version of `Lazy<T>`

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -329,5 +329,35 @@ public static void ValidateQueryTag(this DicomDataset dataset, QueryTag queryTag
...
+        {
+            EnsureArg.IsNotNull(dataset, nameof(dataset));
+            EnsureArg.IsNotNull(searchSequence, nameof(searchSequence));
+
+            var foundDatasets = new List<DicomDataset>();
+
+            var foundSequence = dataset.GetDicomItem<DicomSequence>(searchSequence.Tag);
+            if (foundSequence == null) return foundDatasets;
```
wsugarman, line 347: nit: we don't typically write if statements in a single line.

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
wsugarman, line 353: nit: should there be a singleton property for an instance of `DicomTagComparer`, like `DicomTagComparer.Instance`? Similar to how there is `Comparer<T>.Default`

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -329,5 +329,35 @@ public static void ValidateQueryTag(this DicomDataset dataset, QueryTag queryTag
...
+
+            var searchDataset = searchSequence.Items.FirstOrDefault();
+
+            foreach (var childDataset in foundSequence.Items)
+            {
+                var elements = childDataset.Intersect(searchDataset, new DicomTagComparer());
+                if (elements.Any())
```
wsugarman, line 354: Is there not a `Count` property?

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -329,5 +329,35 @@ public static void ValidateQueryTag(this DicomDataset dataset, QueryTag queryTag
...
+        {
+            EnsureArg.IsNotNull(dataset, nameof(dataset));
+            EnsureArg.IsNotNull(searchSequence, nameof(searchSequence));
+
+            var foundDatasets = new List<DicomDataset>();
+
+            var foundSequence = dataset.GetDicomItem<DicomSequence>(searchSequence.Tag);
+            if (foundSequence == null) return foundDatasets;
+
+            var searchDataset = searchSequence.Items.FirstOrDefault();
```
wsugarman, line 349: Should we check if `searchDataset` has `Count == 0`?

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
```diff
@@ -27,24 +28,79 @@ public bool TryParse(string dicomTagPath, out DicomTag[] dicomTags, bool support
...
+            {
+                return false;
+            }
+
+            DicomTag[] dicomTags = ParseMultipleStandardTag(dicomTagPath);
+
+            if (dicomTags.Length == 1)
+            {
+                var tag = GetTagFromKeywordOrNumber(dicomTagPath);
+                if (tag != null) dicomItem = new DicomValuelessItem(tag);
```
wsugarman, line 51: nit:
```suggestion
if (tag != null)
    dicomItem = new DicomValuelessItem(tag);
    
return dicomItem == null;
```

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
bcarthic: It returns the matching DICOM elements within a DICOM data set.

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
bcarthic: This method returns the matching elements inside a dicom sequence elements, the original dataset can contain all the items but the search sequence will contain only the indexed element that we need. And the returned values can be list of elements but for our logic further we need to know the dataset so we can use the existing code.

src/Microsoft.Health.Dicom.Core/Features/Common/DicomValuelessItem.cs
bcarthic: Yes, we need this, Dicom Sequence is a DICOM item and it is abstract. unlike our core tag which can be represented as DICOM tag. We need a way to store it in an object model from the DB.

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
bcarthic: We are getting a DICOM sequence out, which is a dicom item. Dicom sequence is not just list of tags but list of datasets

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
bcarthic: Updated to GetSequencePathValues and instead of sequence we now pass list of dicom tags.

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
bcarthic: Done

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
bcarthic: Done

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
bcarthic: Removed

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
bcarthic: Removed

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
bcarthic: Removed

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
bcarthic: Changed as per our discussion, now it returns multiple tags for sequence tag path. We removed supportMultiple flag to keep it simple and it wasnt necessary now

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
bcarthic: Changed the logic, not needed now

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
bcarthic: Since we are parsing all tag types, we named it as ParseTag

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
bcarthic: Done

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
bcarthic: Yes, we dont care. When we return back we check if all the results are not null.

src/Microsoft.Health.Dicom.Core/Features/Common/DicomValuelessItem.cs
bcarthic: Removing this file

src/Microsoft.Health.Dicom.Core/Features/Workitem/WorkitemQueryTagStoreEntry.cs
bcarthic: Changed to list of dicomtags

src/Microsoft.Health.Dicom.Core/Features/ExtendedQueryTag/QueryTag.cs
bcarthic: Yeah agree, but as we discussed we will refactor to better approach later and keep this design for now.

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
bcarthic: Yes this return 500. Changed to DicomValidationException

src/Microsoft.Health.Dicom.Core/Features/Workitem/WorkitemQueryTagService.cs
bcarthic: Logging and throwing an exception, as this should never happen. It can only happen if some one manually change the db

src/Microsoft.Health.Dicom.Core/Features/Query/QueryLimit.cs
```diff
@@ -84,6 +90,10 @@ public static QueryTagLevel GetQueryTagLevel(DicomTag coreTag)
             {
                 return QueryTagLevel.Instance;
             }
+            if (CoreWorkitemTags.Contains(coreTag))
```
wsugarman, line 93: Is it possible for extended query tags to overlap with the work item query tags? Eg. A user adds a work item query tag as an extended item query tag? Should we explicitly prevent this scenario in the extended query tag API?

src/Microsoft.Health.Dicom.Core/Features/Workitem/WorkitemQueryTagService.cs
```diff
@@ -5,23 +5,33 @@
...
+        public WorkitemQueryTagService(IIndexWorkitemStore indexWorkitemStore, IDicomTagParser dicomTagParser, ILogger<WorkitemQueryTagService> logger)
         {
             _indexWorkitemStore = EnsureArg.IsNotNull(indexWorkitemStore, nameof(indexWorkitemStore));
-            _queryTagCache = new AsyncCache<IReadOnlyCollection<WorkitemQueryTagStoreEntry>>(ResolveQueryTagsAsync);
+            _queryTagCache = new AsyncCache<IReadOnlyCollection<QueryTag>>(ResolveQueryTagsAsync);
+            _dicomTagParser = EnsureArg.IsNotNull(dicomTagParser, nameof(dicomTagParser));
+            _logger = EnsureArg.IsNotNull(logger, nameof(logger));
         }
 
         public void Dispose()
```
wsugarman, line 37: nit: in general, if this class is `sealed` because we don't foresee derived implementations, we don't need to implement the entire disposal pattern and instead can have a simply `Dispose` method, like:
```csharp
public void Dispose()
{
    _queryTagCache.Dispose();
    GC.SuppressFinalize(this);
}
```

wsugarman: `AsyncCache<T>` will also throw an `ObjectDisposedException`, so you don't need to maintain your own `_disposed` flag

src/Microsoft.Health.Dicom.Core/Features/Workitem/WorkitemQueryTagService.cs
```diff
@@ -53,9 +63,24 @@ public async Task<IReadOnlyCollection<WorkitemQueryTagStoreEntry>> GetQueryTagsA
...
+        private async Task<IReadOnlyCollection<QueryTag>> ResolveQueryTagsAsync(CancellationToken cancellationToken)
         {
-            return await _indexWorkitemStore.GetWorkitemQueryTagsAsync(cancellationToken);
+            var workitemQueryTags = await _indexWorkitemStore.GetWorkitemQueryTagsAsync(cancellationToken);
+
+            foreach (var tag in workitemQueryTags)
+            {
+                if (_dicomTagParser.TryParse(tag.Path, out DicomTag[] dicomTags))
+                {
+                    tag.PathTags = Array.AsReadOnly(dicomTags);
```
wsugarman, line 74: Is this necessary? Surely no one is casting the `IReadOnlyCollection<T>` to a mutable type internally

jovinson-ms (approved):

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/ExtendedQueryTagDataRowsBuilder.cs
```diff
@@ -52,26 +52,46 @@ internal static class ExtendedQueryTagDataRowsBuilder
             var dateTimeWithUtcRows = new List<InsertDateTimeExtendedQueryTagTableTypeV2Row>();
             var personNameRows = new List<InsertPersonNameExtendedQueryTagTableTypeV1Row>();
 
-            foreach (QueryTag queryTag in queryTags.Where(x => x.IsExtendedQueryTag))
+            foreach (QueryTag queryTag in queryTags.Where(x => x.IsExtendedQueryTag || x.WorkitemQueryTagStoreEntry != null))
+            {
+                if (queryTag.VR == DicomVR.SQ)
+                {
+                    var dicomDatasets = instance.GetSequencePathValues(queryTag.WorkitemQueryTagStoreEntry.PathTags);
+                    foreach (var dataset in dicomDatasets)
```
wsugarman, line 60: nit:
```suggestion
                    foreach (var dicomItem in dicomDatasets.SelectMany(x => x))
```

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/ExtendedQueryTagDataRowsBuilder.cs
```diff
@@ -182,17 +202,22 @@ private static void AddLongRow(DicomDataset instance, List<InsertLongExtendedQue
...
             if (stringVal != null)
             {
-                stringRows.Add(new InsertStringExtendedQueryTagTableTypeV1Row(queryTag.ExtendedQueryTagStoreEntry.Key, stringVal, (byte)queryTag.Level));
+                stringRows.Add(new InsertStringExtendedQueryTagTableTypeV1Row(tagKey, stringVal, (byte)queryTag.Level));
             }
         }
+
+        private static int GetKeyFromQueryTag(QueryTag queryTag)
+        {
+            return queryTag.WorkitemQueryTagStoreEntry != null ? queryTag.WorkitemQueryTagStoreEntry.Key : queryTag.ExtendedQueryTagStoreEntry.Key;
```
wsugarman, line 220: Should this just be a property on the `QueryTag` class?

src/Microsoft.Health.Dicom.Tests.Common/Extensions/DicomTagExtensions.cs
```diff
@@ -25,5 +28,17 @@ public static ExtendedQueryTagStoreEntry BuildExtendedQueryTagStoreEntry(this Di
...
             return new ExtendedQueryTagStoreEntry(key: key, path: tag.GetPath(), vr: vr ?? tag.GetDefaultVR().Code, privateCreator: privateCreator, level: level, status: status, queryStatus: QueryStatus.Enabled, errorCount: 0);
         }
+
+        public static WorkitemQueryTagStoreEntry BuildWorkitemQueryTagStoreEntry(string path, int key, string vr)
+        {
+            var dicomTagParser = new DicomTagParser();
+            var entry = new WorkitemQueryTagStoreEntry(key, path, vr);
+            if (dicomTagParser.TryParse(path, out var dicomTags))
+            {
+                entry.PathTags = Array.AsReadOnly(dicomTags);
```
wsugarman, line 38: Is this necessary?

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -329,5 +330,42 @@ public static void ValidateQueryTag(this DicomDataset dataset, QueryTag queryTag
...
+        /// <param name="dicomTags">The Dicom tags modelling the path.</param>
+        /// <returns>Lists of DicomDataset that matches the list of tags.</returns>
+        public static IEnumerable<DicomDataset> GetSequencePathValues(this DicomDataset dataset, ReadOnlyCollection<DicomTag> dicomTags)
+        {
+            EnsureArg.IsNotNull(dataset, nameof(dataset));
+            EnsureArg.IsNotNull(dicomTags, nameof(dicomTags));
+
+            if (dicomTags.Count != 2)
+            {
+                throw new DicomValidationException(string.Join(",", dicomTags.Select(x => x.GetPath())), DicomVR.SQ, DicomCoreResource.NestedSequencesNotSupported);
```
wsugarman, line 347: nit: `", "` (with a space)

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -329,5 +330,42 @@ public static void ValidateQueryTag(this DicomDataset dataset, QueryTag queryTag
...
+            }
+
+            var foundDatasets = new List<DicomDataset>();
```
wsugarman, line 350: Instead of populating a list, should we just use `yield` here and below?

wsugarman (approved):

src/Microsoft.Health.Dicom.Core/Features/Query/QueryLimit.cs
bcarthic: Yes it is possible for overlap. And we shouldn't restrict this scenario in ExtendedQueryTag since they index any standard tags they want. For workitem, we dont use level, when we refactor we need to change this.

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/ExtendedQueryTagDataRowsBuilder.cs
bcarthic: I dont want to do this since we need the dataset to get the value for that item

src/Microsoft.Health.Dicom.Tests.Common/Extensions/DicomTagExtensions.cs
bcarthic: Yes it is necessary to populate the pathTags, since it is a readOnly collection, this is the way we found to populate

src/Microsoft.Health.Dicom.Core/Features/Workitem/WorkitemQueryTagService.cs
bcarthic: We couldnt assign dicomTags directly to PathTags, so we have create a readOnly list

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
```diff
@@ -14,37 +16,58 @@ namespace Microsoft.Health.Dicom.Core.Features.Common
...
+            {
+                dicomTags.Add(ParseTagFromKeywordOrNumber(path));
             }
 
+            return dicomTags.ToArray();
+        }
 
+        private static DicomTag ParseTagFromKeywordOrNumber(string dicomTagPath)
+        {
             DicomTag dicomTag = ParseStardandDicomTagKeyword(dicomTagPath);
```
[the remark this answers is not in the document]
bcarthic: Done

src/Microsoft.Health.Dicom.Core/Features/Common/IDicomTagParser.cs
```diff
@@ -13,12 +13,11 @@ namespace Microsoft.Health.Dicom.Core.Features.Common
     public interface IDicomTagParser
     {
         /// <summary>
-        /// Parse dicom tag path.
+        /// Parse dicom tag path. Can support multiple dicom tags like '00101002.00100024'.
```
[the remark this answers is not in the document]
bcarthic: Done

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
```diff
@@ -14,37 +16,58 @@ namespace Microsoft.Health.Dicom.Core.Features.Common
...
+
+            var result = tags.All(x => x != null);
+            dicomTags = result ? tags : null;
+
+            return result;
+        }
+
+        private static DicomTag[] ParseTag(string dicomTagPath)
+        {
+            var paths = dicomTagPath.Split(".");
```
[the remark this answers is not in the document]
bcarthic: Done

src/Microsoft.Health.Dicom.Core/Features/Workitem/WorkitemQueryTagService.cs
bcarthic: Done

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
bcarthic: Done

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
bcarthic: Done

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
```diff
@@ -14,37 +15,41 @@ namespace Microsoft.Health.Dicom.Core.Features.Common
...
 
+            DicomTag[] tags = dicomTagPath
+                     .Split('.')
+                     .Select(ParseTagFromKeywordOrNumber)
+                     .ToArray();
 
-            DicomTag dicomTag = ParseStardandDicomTagKeyword(dicomTagPath);
-
-            if (dicomTag == null)
+            if (tags.Length > 2)
```
smithago, line 31: Should this be only true for workItem tags or will be get this for Dicom instance tags for free?

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/ExtendedQueryTagDataRowsBuilder.cs
bcarthic: Yes that is correct. we thought about that. It needs some refactoring and change existing code.

src/Microsoft.Health.Dicom.Core/Features/Workitem/WorkitemQueryTagService.cs
```diff
@@ -5,57 +5,63 @@
...
-                }
-
-                _disposed = true;
-            }
+            return await _queryTagCache.GetAsync(cancellationToken: cancellationToken);
         }
 
-        public async Task<IReadOnlyCollection<WorkitemQueryTagStoreEntry>> GetQueryTagsAsync(CancellationToken cancellationToken = default)
+        private async Task<IReadOnlyCollection<QueryTag>> ResolveQueryTagsAsync(CancellationToken cancellationToken)
         {
```
smithago, line 48: See if there logs, both info or perf once you want to track to help debug easily.

smithago (approved):

src/Microsoft.Health.Dicom.Core/Features/Common/DicomTagParser.cs
bcarthic: we get this for Dicom instance tags for free. but we can query for multi level tags in QIDO but it wont return any results with current implementation,

bcarthic: You are right, we should block for other than workitem tags. Throwing exception now

src/Microsoft.Health.Dicom.Core/Features/Workitem/WorkitemQueryTagService.cs
bcarthic: Sure

PR #1305, opened 2022-01-27
Remove old SQL client(system.Data.SqlClient) package reference

smithago:
## Description
It is replaced already with the new Microsoft.Data.SqlClient

## Related issues
[AB#88198](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/88198)

## Testing
Existing tests pass


jovinson-ms (approved):

wsugarman (approved):

wsugarman:
Thank you for making this change. I think this was a strange dependency brought in by the `fo-dicom.json` (if I recall correctly) package, and now that we no longer use it we can get rid of these package specifications.

PR #1326, opened 2022-02-03
Two step adding workitem

bcarthic:
## Description
This PR includes changes that will save workitem in two step. First it will create the workitem record in the DB with Creating status and after the blob upload is successful it will update the status to Created.

For Retrieve and search transaction only Created status workitem are returned back.

## Related issues
Addresses [[AB#87263](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/87263)].

## Testing
Integration tests added to ensure workitem is added.

### DIFF
--- a/src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/UpdateWorkitemStatus.sql
+++ b/src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/UpdateWorkitemStatus.sql
@@ -1,0 +1,44 @@
+﻿/*************************************************************
+    Stored procedure for updating a workitem status.
+**************************************************************/
+--
+-- STORED PROCEDURE
+--     UpdateWorkitemStatus
+--
+-- DESCRIPTION
+--     Updates a workitem status.
+--
+-- PARAMETERS
+--     @partitionKey
+--         * The system identifier of the data partition.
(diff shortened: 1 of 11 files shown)


src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/UpdateWorkitemStatus.sql
```diff
@@ -0,0 +1,44 @@
...
+-- DESCRIPTION
+--     Updates a workitem status.
+--
+-- PARAMETERS
+--     @partitionKey
+--         * The system identifier of the data partition.
+--     @workitemKey
+--         * The workitem key.
+--     @status
+--         * Initial status of the workitem status, Either 0(Creating) or 1(Created)
```
jovinson-ms, line 17: This should say `New status of the workitem, ...`

jovinson-ms (approved):

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/UpdateWorkitemStatus.sql
bcarthic: Added this in the next PR

PR #1328, opened 2022-02-03
Support DicomCast name as a prefix to Dicomcast tables

poadhika:
## Description
This Pr allows the paas service to pass the dicomcast name to oss such that the flow will store the exception to correct dicomcast specific table.

## Related issues
Addresses [AB#88226](https://microsofthealth.visualstudio.com/Health/_workitems/edit/88226)

## Testing
Added unit tests.


jnlycklama:
AllTables in Constants does not have the dicomCastName added, and is used here - https://github.com/microsoft/dicom-server/blob/a7251027997c56584495c5a8157f9ded3b105838/converter/dicom-cast/src/Microsoft.Health.DicomCast.TableStorage/Features/Storage/TableServiceClientInitializer.cs#L32

jnlycklama:
Should this one be updated too? 
https://github.com/microsoft/dicom-server/blob/a7251027997c56584495c5a8157f9ded3b105838/converter/dicom-cast/src/Microsoft.Health.DicomCast.TableStorage/Features/Storage/TableSyncStateStore.cs#L44

Would it make sense to update constants so they are strings like "{dicomCastName}FhirFailToStoreExceptionTable" so it is clear what the full table names are?

converter/dicom-cast/src/Microsoft.Health.DicomCast.TableStorage/Features/Storage/TableServiceClientProvider.cs
```diff
@@ -19,19 +22,29 @@ public class TableServiceClientProvider : IHostedService, IRequireInitialization
     {
         private readonly TableServiceClient _tableServiceClient;
         private readonly RetryableInitializationOperation _initializationOperation;
+        private readonly TableDataStoreConfiguration _tableDataStoreConfiguration;
+
+        public Dictionary<string, string> TableList { get; }
```
jnlycklama, line 27: Can you add a comment here to describe what the key and value should be? I think it is key = constant, value = prefix+constant but it's not clear

converter/dicom-cast/src/Microsoft.Health.DicomCast.TableStorage/Features/Storage/TableServiceClientInitializer.cs
```diff
@@ -22,16 +23,18 @@ public class TableServiceClientInitializer : ITableServiceClientInitializer
...
         {
             EnsureArg.IsNotNull(tableServiceClient, nameof(tableServiceClient));
+            EnsureArg.IsNotNull(tableList, nameof(tableList));
 
             try
             {
                 _logger.LogInformation("Initializing Table Storage and tables");
-                foreach (string tableName in Constants.AllTables)
+
+                foreach (string tableName in tableList.Keys)
```
jnlycklama, line 35: Should this be tableList.Values?

jnlycklama (approved):

PR #1364, opened 2022-02-17
Support lazy streaming of dicom instances in WADO-RS to reduce memory pressure

smithago:
## Description
Fixes several bugs in WADO-RS resource 

1. Transcoding and Frames needs the file to be fully read in Memory from fo-dicom. To protect the service, this is limited to single file of size <= 100MB
2. GetInstance with accept multipart will now return multipart response.
3. GetStudy/GetSeries with multiple instances was parallelly initializing all the instance streams which could result in OOM exception. We now have `LazyMultipartReadonlyStream` that takes AsyncEnumerable of instance streams and JIT pulls down and returns the content. Time to first byte is faster now, compared to before.

## Related issues
[AB#88299](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/88299)
[AB#86306](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/86306)
[AB#86308](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/86308)
[AB#88758](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/88758)

## Testing
- Existing tests passing
- New tests for restricts
- New tests for LazyMultipartReadonlyStream
- Validate reduced memory usage when multiple instances are requested. There is a constant 8MB used now.


src/Microsoft.Health.Dicom.Core/Features/Common/FileProperties.cs
```diff
@@ -0,0 +1,13 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Dicom.Core.Features.Common
+{
+    public class FileProperties
+    {
+        public long ContentLength { get; init; }
+
```
bcarthic, line 11: nit: new line

src/Microsoft.Health.Dicom.Api/Features/Responses/ResourceResult.cs
```diff
@@ -0,0 +1,107 @@
...
+{
+    internal class ResourceResult : IActionResult
+    {
+        private readonly RetrieveResourceResponse _response;
+        private readonly RetrieveConfiguration _retrieveConfiguration;
+
+        internal ResourceResult(RetrieveResourceResponse response, RetrieveConfiguration retrieveConfiguration)
+        {
+            _response = EnsureArg.IsNotNull(response);
+            _retrieveConfiguration = EnsureArg.IsNotNull(retrieveConfiguration);
```
brandonpollett, line 31: Would it make sense to only pass the property we need off of here?

src/Microsoft.Health.Dicom.Api/Features/Responses/ResourceResult.cs
```diff
@@ -0,0 +1,107 @@
...
+                objectResult = await GetSingelPartResult(context.HttpContext, context.HttpContext.RequestAborted);
+            }
+            else
+            {
+                objectResult = GetMultiPartResult(context.HttpContext, context.HttpContext.RequestAborted);
+            }
+            await objectResult.ExecuteResultAsync(context);
+        }
+
+        private async Task<ObjectResult> GetSingelPartResult(HttpContext context, CancellationToken cancellationToken)
```
brandonpollett, line 48: Singel => Single

src/Microsoft.Health.Dicom.Api/Features/Responses/ResourceResult.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            var enumerator = _response.GetResponseInstancesEnumerator(cancellationToken);
+            await enumerator.MoveNextAsync();
```
brandonpollett, line 51: Should we debug assert that there is nothing left in this enumerator?

src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
```diff
@@ -63,47 +70,85 @@ public async Task<RetrieveResourceResponse> GetInstanceResourceAsync(RetrieveRes
...
+                        {
+                            _dicomRequestContextAccessor.RequestContext.IsTranscodeRequested = true;
+                            _dicomRequestContextAccessor.RequestContext.BytesTranscoded = fileProperties.ContentLength;
+                        }
+
+                        Stream stream = await _blobDataStore.GetFileAsync(instance.VersionedInstanceIdentifier, cancellationToken);
+
+                        if (message.ResourceType == ResourceType.Frames)
+                        {
+                            // egarly doing getFrames to validate frame numbers are valid before returning a response
```
brandonpollett, line 115: egarly => agerly

src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
```diff
@@ -63,47 +70,85 @@ public async Task<RetrieveResourceResponse> GetInstanceResourceAsync(RetrieveRes
...
+                        if (fileProperties.ContentLength > _retrieveConfiguration.MaxDicomFileSize)
+                        {
+                            throw new NotAcceptableException(string.Format(DicomCoreResource.RetrieveServiceFileTooBig, _retrieveConfiguration.MaxDicomFileSize));
+                        }
+
+                        // set properties used for billing
+                        if (needsTranscoding)
+                        {
+                            _dicomRequestContextAccessor.RequestContext.IsTranscodeRequested = true;
+                            _dicomRequestContextAccessor.RequestContext.BytesTranscoded = fileProperties.ContentLength;
```
brandonpollett, line 108: Does this mean on a frames request that needs transcoding, we count the entire file as transcoded?

src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
```diff
@@ -133,7 +177,60 @@ private static string GetResponseTransferSyntax(string requestedTransferSyntax,
...
+        private static async IAsyncEnumerable<RetrieveResourceInstance> GetAsyncEnumerableFrameStreams(
+            IEnumerable<Stream> frameStreams,
+            InstanceMetadata instanceMetadata,
+            bool isOriginalTransferSyntaxRequested,
+            string requestedTransferSyntax)
+        {
+            // fake await to return AsyncEnumerable and keep the response consistent
+            await Task.Run(() => 1);
+            // responseTransferSyntax is same for all frames in a instance
+            var responseTranferSyntax = GetResponseTransferSyntax(isOriginalTransferSyntaxRequested, requestedTransferSyntax, instanceMetadata);
```
brandonpollett, line 216: responseTranferSyntax => responseTransferSyntax

src/Microsoft.Health.Dicom.Api/Features/Responses/ResourceResult.cs
smithago: I wanted the resultObject not to know about specific bufferSize, so kept it generic.

src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
smithago: good catch, that is a bug, fixed it.

src/Microsoft.Health.Dicom.Core/Configs/RetrieveConfiguration.cs
```diff
@@ -0,0 +1,21 @@
...
+        /// <summary>
+        /// Maximum dicom file supported for getting frames and transcoding
+        /// </summary>
+        public long MaxDicomFileSize { get; } = 1024 * 1024 * 100; //100 MB
+
+        /// <summary>
+        /// Response uses lazy streams that copy data from storage JIT into a buffer and then to the output stream
+        /// This is the size of the buffer
+        /// </summary>
+        public int LazyResponseStreamBufferSize { get; } = 1024 * 1024 * 4; //4 MB
```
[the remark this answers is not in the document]
smithago: +1 on the article. I did a quick digging, since Azure Blob SDK uses the same size.

Here is the source code of arrayPool https://github.com/dotnet/corert/blob/master/src/System.Private.CoreLib/shared/System/Buffers/TlsOverPerCoreLockedStacksArrayPool.cs#L281

Looks like there are 16 buckets of varying sizes (16 bytes to 1MB) and each bucket has 8 buffers.

I also did a quick perfview check and Array pool created reason was PoolExhausted and not overMaximumSize. 

image

src/Microsoft.Health.Dicom.Api/Web/LazyMultipartReadOnlyStream.cs
```diff
@@ -0,0 +1,228 @@
...
+            {
+                ArrayPool<byte>.Shared.Return(_buffer, clearArray: true);
+                _buffer = null;
+            }
+        }
+
+        #region MultiPart boundary write
+        private string GetHeaderBoundary(bool isStart)
+        {
+            var multiPartStringBuilder = new StringBuilder();
```
brandonpollett, line 182: Would it make sense to set the capacity since we're essentially building the same string builder multiple times? Just wondering on allocations since the initial size seems to be 16 characters and we'll always overrun that and this could be called many times on a large request.

brandonpollett: Also, do we have to have a different start boundary from an intermediate boundary or can they both just be the same as the intermediate boundary?

src/Microsoft.Health.Dicom.Api/Web/LazyMultipartReadOnlyStream.cs
```diff
@@ -0,0 +1,228 @@
...
+        }
+
+        private string GetTerminatingBoundary()
+        {
+            return Crlf + "--" + _boundary + "--" + Crlf;
+        }
+
+        private static string GetContentHeader(IEnumerable<KeyValuePair<string, IEnumerable<string>>> headers)
+        {
+            StringBuilder builder = new StringBuilder();
```
brandonpollett, line 218: Should we use a new StringBuilder here or just use the one in the calling method?

src/Microsoft.Health.Dicom.Api/Features/Responses/ResourceResult.cs
```diff
@@ -0,0 +1,111 @@
...
+            return objectResult;
+        }
+
+        private ObjectResult GetMultiPartResult(HttpContext context, CancellationToken cancellationToken)
+        {
+            string boundary = Guid.NewGuid().ToString();
+            var mediaType = new MediaTypeHeaderValue(KnownContentTypes.MultipartRelated);
+            mediaType.Parameters.Add(new NameValueHeaderValue(KnownContentTypes.Boundary, boundary));
+#pragma warning disable CA2000 // Dispose objects before losing scope, registered for dispose in response
+            LazyMultipartReadOnlyStream lazyStream = new LazyMultipartReadOnlyStream(
```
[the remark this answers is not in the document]
smithago: Yes, I missed it.

src/Microsoft.Health.Dicom.Api/Web/LazyMultipartReadOnlyStream.cs
smithago: Start and intermediate have different formats

smithago: I am not sure if we need to optimize StringBuilder, it is not immutable like string, so no copying around, and creates additional char[]s to allows easy updates.

src/Microsoft.Health.Dicom.Core/Configs/RetrieveConfiguration.cs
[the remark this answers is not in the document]
jovinson-ms: @smitha the buffer sizes in the logs above are all less than 1MB so they aren't testing the 4MB behavior. My concern is that the behavior of ArrayPool in the file you linked will not use ArrayPool at all - it will [check to find a bucket index](https://github.com/dotnet/corert/blob/c6af4cfc8b625851b91823d9be746c4f7abdc667/src/System.Private.CoreLib/shared/System/Buffers/TlsOverPerCoreLockedStacksArrayPool.cs#L100) that fits 4MB.

Since the value returned (18) is not less than the number of buckets (17), [a new array will be allocated](https://github.com/dotnet/corert/blob/c6af4cfc8b625851b91823d9be746c4f7abdc667/src/System.Private.CoreLib/shared/System/Buffers/TlsOverPerCoreLockedStacksArrayPool.cs#L143) and **not** returned to the pool.

src/Microsoft.Health.Dicom.Api/Web/LazyMultipartReadOnlyStream.cs
```diff
@@ -0,0 +1,227 @@
...
+            CancellationToken cancellation)
+            : base()
+        {
+            EnsureArg.IsNotNull(enumerableStreams, nameof(enumerableStreams));
+            EnsureArg.IsNotEmptyOrWhiteSpace(boundary, nameof(boundary));
+            EnsureArg.IsGt(bufferSize, KB, nameof(bufferSize));
+
+            _asyncEnumerator = enumerableStreams.GetAsyncEnumerator(cancellation);
+            _bufferSize = bufferSize;
+            _buffer = ArrayPool<byte>.Shared.Rent(bufferSize);
```
jovinson-ms, line 53: Would it make sense to defer renting until a call to `ReadAsync`?

smithago: Right pattern, but there is not much value add in our scenario

src/Microsoft.Health.Dicom.Core/Configs/RetrieveConfiguration.cs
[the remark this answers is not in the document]
smithago: Here is another snapshot of 4MB, with the same poolExhausted. I have also asked the experts. We can always change it later if they come back with the better suggestion

image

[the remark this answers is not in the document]
jovinson-ms: Just figured it out - the article I linked and the code you linked is from the `dotnet/corert` repo, which is now archived. That repo did have the maximum array size of 1MB, but there's a completely new repo for these libraries at `dotnet/runtime`. There they did increase the [maximum array size](https://github.com/dotnet/runtime/blob/2a9745176ac0cea4a86b37c52e9863d7d13f24fc/src/libraries/System.Private.CoreLib/src/System/Buffers/TlsOverPerCoreLockedStacksArrayPool.cs#L24) to **1GB**. I ran the same perfview and was also seeing that 4MB was not over the maximum. Sorry about the confusion! 🙂

jovinson-ms (approved):

brandonpollett (approved):

PR #1402, opened 2022-03-07
Fix Deadlock in AddExtendedQueryTag and AddExtendedQueryTagErorr

pengchen0692:
This PR fix 2 deadlocks.
# AddExtendedQueryTagError Deadlock
### Context
https://microsofthealth.visualstudio.com/Health/_workitems/edit/89214. We see deadlocks in Dicom-Paas-CI, need investigation and fix it

### RootCause
The problem is with [AddExtendedQueryTagError](https://github.com/microsoft/dicom-server/blob/main/src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/AddExtendedQueryTagError.sql)

Let's say 2 process want to add error for `tagA` but different Dicom instance at same time:
Process1 hold S lock on `tagA` in `ExtendedQueryTag` table,  then request range lock on same tag in `ExtendedQueryTagError` table, while Process2 completed updated `ExtendedQueryTagError` table, so hold X lock on the tag in `ExtendedQueryTagError`  table, and need to update the tag in `ExtendedQueryTag` table,  so Process1 and Process2 are deadlock

### Solution
Change **HOLDLOCK**
```sql
IF NOT EXISTS (SELECT * FROM dbo.ExtendedQueryTag WITH (HOLDLOCK) WHERE TagKey = @tagKey AND TagStatus = 0)
```
to **UPDLOCK**
```sql
IF NOT EXISTS (SELECT * FROM dbo.ExtendedQueryTag WITH (UPDLOCK) WHERE TagKey = @tagKey AND TagStatus = 0)
```
so that allow only 1 operation on same tag.

## Related issues
Addresses [AB#89214](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/89214)

# AddExtendedQueryTag deadlock
### Simplified Repro:
1. Clear dbo.ExtendedQueryTag
2. Run queries below simultaneously 

- Query1:

```sql
begin tran
 select COUNT(*) from dbo.ExtendedQueryTag with (HOLDLOCK)
 waitfor delay '00:00:10'
 insert into dbo.ExtendedQueryTag values(4,'tag4','vr', NULL,0,0,0,0)
commit tran
```

- Query2:
```sql
begin tran
 select COUNT(*) from dbo.ExtendedQueryTag with (HOLDLOCK)
 waitfor delay '00:00:10'
 insert into dbo.ExtendedQueryTag values(3,'tag3','vr', NULL,0,0,0,0)
 commit tran
```
One of the 2 queries will fail for deadlock

### Root cause
The problem is with [AddExtendedQueryTags](https://github.com/microsoft/dicom-server/blob/main/src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/AddExtendedQueryTags.sql)
- Query1:
_Select COUNT(*) with (HOLDLOCK)_ add rangeS-S  lock which blocks any modification into the table -> try to insert tag1 which pending on Query2 to release rangeS-S lock
- Query2:
_Select COUNT(*) with (HOLDLOCK)_ add rangeS-S  lock which blocks any modification into the table -> try to insert tag2 which pending on Query1 to release rangeS-S lock 
The deadlock surface.

### Solution
Change **HOLDLOCK** to **UPDLOCK,HOLDLOCK**, so that prevent another process enters.
```SQL
IF (SELECT COUNT(*)
            FROM dbo.ExtendedQueryTag AS XQT WITH(HOLDLOCK)
            FULL OUTER JOIN @extendedQueryTags AS input 
            ON XQT.TagPath = input.TagPath) > @maxAllowedCount
            THROW 50409, 'extended query tags exceed max allowed count', 1
```
==>
```SQL
IF (SELECT COUNT(*)
            FROM dbo.ExtendedQueryTag AS XQT WITH(UPDLOCK,HOLDLOCK)
            FULL OUTER JOIN @extendedQueryTags AS input 
            ON XQT.TagPath = input.TagPath) > @maxAllowedCount
            THROW 50409, 'extended query tags exceed max allowed count', 1
```

## Related issues
Addresses [AB#88732](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/88732)


src/Microsoft.Health.Dicom.SqlServer/Microsoft.Health.Dicom.SqlServer.csproj
```diff
@@ -69,6 +69,7 @@
     <TSqlScript Include="Features\Schema\Sql\Scripts\Sequences.sql" />
     <TSqlScript Include="Features\Schema\Sql\Tables\*.sql" />
     <TSqlScript Include="Features\Schema\Sql\Types\*.sql" />
+    <None Remove="Features\Schema\Migrations\14.diff.sql" />
```
pengchen0692, line 72: remove this line

test/Microsoft.Health.Dicom.Tests.Integration/Persistence/SqlDataStoreTestsFixture.cs
```diff
@@ -139,7 +139,7 @@ internal SqlDataStoreTestsFixture(string databaseName, SchemaInformation schemaI
                {
                     new SqlExtendedQueryTagErrorStoreV1(),
                     new SqlExtendedQueryTagErrorStoreV4(SqlConnectionWrapperFactory, NullLogger<SqlExtendedQueryTagErrorStoreV4>.Instance),
-                    new SqlExtendedQueryTagErrorStoreV6(SqlConnectionWrapperFactory, NullLogger<SqlExtendedQueryTagErrorStoreV6>.Instance),
+                    new SqlExtendedQueryTagErrorStoreV6(SqlConnectionWrapperFactory, NullLogger<SqlExtendedQueryTagErrorStoreV6>.Instance)
```
pengchen0692, line 142: remove change in this file

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/AddExtendedQueryTagError.sql
```diff
@@ -32,7 +32,7 @@ AS
             THROW 50404, 'Instance does not exist or has not been created.', 1;
 
         --Check if tag exists and in Adding status.
-        IF NOT EXISTS (SELECT * FROM dbo.ExtendedQueryTag WITH (HOLDLOCK) WHERE TagKey = @tagKey AND TagStatus = 0)
+        IF NOT EXISTS (SELECT * FROM dbo.ExtendedQueryTag WITH (UPDLOCK) WHERE TagKey = @tagKey AND TagStatus = 0)
```
wsugarman, line 35: This makes sense. We can allow INSERTs to happen concurrently, as if the entry were added in the middle of our transaction we would fail anyway here without it.

wsugarman:
Unfortunately, I think you'll have to migrate this to v15 as I think Shiva's change is 14.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/AddExtendedQueryTagError.sql
pengchen0692: If I understand correctly, what you mean is that `UPDLOCK` doesn't block concurrent `INSERT`, but this procedure will fail here as no qualified tag, so it should be fine?

pengchen0692:
> 

Thanks Will, I'll check with Shiva.

wsugarman (approved):
Approved for v15 after 14 is checked in

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/14.diff.sql
```diff
@@ -0,0 +1,74 @@
...
+            VALUES (@tagKey, @errorCode, @watermark, @currentDate)
+        OUTPUT INSERTED.TagKey;
+
+        -- Disable query on the tag and update error count
+        UPDATE dbo.ExtendedQueryTag
+        SET QueryStatus = 0, ErrorCount = ErrorCount + @addedCount
+        WHERE TagKey = @tagKey
+
+    COMMIT TRANSACTION
+
```
pengchen0692, line 73: Missed `GO` here

wsugarman (approved):

PR #1425, opened 2022-03-09
Fix deadlock in AddExtendedQueryTag

pengchen0692:
### Simplified Repro:
1. Clear dbo.ExtendedQueryTag
2. Run queries below simultaneously 

- Query1:

```sql
begin tran
 select COUNT(*) from dbo.ExtendedQueryTag with (HOLDLOCK)
 waitfor delay '00:00:10'
 insert into dbo.ExtendedQueryTag values(4,'tag4','vr', NULL,0,0,0,0)
commit tran
```

- Query2:
```sql
begin tran
 select COUNT(*) from dbo.ExtendedQueryTag with (HOLDLOCK)
 waitfor delay '00:00:10'
 insert into dbo.ExtendedQueryTag values(3,'tag3','vr', NULL,0,0,0,0)
 commit tran
```
One of the 2 queries will fail for deadlock

### Root cause
The problem is with [AddExtendedQueryTags](https://github.com/microsoft/dicom-server/blob/main/src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/AddExtendedQueryTags.sql)
- Query1:
_Select COUNT(*) with (HOLDLOCK)_ add rangeS-S  lock which blocks any modification into the table -> try to insert tag1 which pending on Query2 to release rangeS-S lock
- Query2:
_Select COUNT(*) with (HOLDLOCK)_ add rangeS-S  lock which blocks any modification into the table -> try to insert tag2 which pending on Query1 to release rangeS-S lock 
The deadlock surface.

### Solution
Change **HOLDLOCK** to **UPDLOCK,HOLDLOCK**, so that prevent another process enters.
```SQL
IF (SELECT COUNT(*)
            FROM dbo.ExtendedQueryTag AS XQT WITH(HOLDLOCK)
            FULL OUTER JOIN @extendedQueryTags AS input 
            ON XQT.TagPath = input.TagPath) > @maxAllowedCount
            THROW 50409, 'extended query tags exceed max allowed count', 1
```
==>
```SQL
IF (SELECT COUNT(*)
            FROM dbo.ExtendedQueryTag AS XQT WITH(UPDLOCK,HOLDLOCK)
            FULL OUTER JOIN @extendedQueryTags AS input 
            ON XQT.TagPath = input.TagPath) > @maxAllowedCount
            THROW 50409, 'extended query tags exceed max allowed count', 1
```

## Related issues
Addresses [AB#88732](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/88732)


src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/14.diff.sql
```diff
@@ -0,0 +1,97 @@
...
+    @maxAllowedCount INT = 128, -- Default value for backwards compatibility
+    @ready BIT = 0
+AS
+    SET NOCOUNT     ON
+    SET XACT_ABORT  ON
+BEGIN
+    BEGIN TRANSACTION
+
+        -- Check if total count exceed @maxCount
+        -- HOLDLOCK to prevent adding queryTags from other transactions at same time.
```
pengchen0692, line 45: update comments

PR #1461, opened 2022-03-28
Eagerly Fail XQT Tests if Deletion Fails due to Unknown Reason

wsugarman:
## Description
- Attempt to retrieve more diagnostic information by increasing strictness for tag deletion before tests. If the deletion fails for a reason other than it's not found, throw the exception

## Related issues
[AB#88316](https://microsofthealth.visualstudio.com/Health/_workitems/edit/88316/)

## Testing
Pipeline

### DIFF
--- a/test/Microsoft.Health.Dicom.Web.Tests.E2E/Common/DicomTagsManager.cs
+++ b/test/Microsoft.Health.Dicom.Web.Tests.E2E/Common/DicomTagsManager.cs
@@ -36,14 +36,7 @@ internal class DicomTagsManager : IAsyncDisposable
     {
         foreach (var tag in _tags)
         {
-            try
-            {
-                await _dicomWebClient.DeleteExtendedQueryTagAsync(tag);
-            }
-            catch (DicomWebException)
-            {
-
-            }
+            await _dicomWebClient.DeleteExtendedQueryTagAsync(tag);
         }
     }
 
@@ -114,8 +107,16 @@ internal class DicomTagsManager : IAsyncDisposable
     public async Task<bool> DeleteExtendedQueryTagAsync(string tagPath, CancellationToken cancellationToken = default)
     {
         EnsureArg.IsNotNull(tagPath, nameof(tagPath));
+        _tags.Remove(tagPath);
 
-        var response = await _dicomWebClient.DeleteExtendedQueryTagAsync(tagPath, cancellationToken);
-        return response.StatusCode == System.Net.HttpStatusCode.NoContent;
+        try
+        {
+            var response = await _dicomWebClient.DeleteExtendedQueryTagAsync(tagPath, cancellationToken);
+            return response.StatusCode == System.Net.HttpStatusCode.NoContent;
+        }
+        catch (DicomWebException dwe) when (dwe.StatusCode == System.Net.HttpStatusCode.NotFound)
+        {
+            return false;
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


alishahahmed (approved):

wsugarman:
I think I found the issue and will follow up in this same PR with the fix

test/Microsoft.Health.Dicom.Web.Tests.E2E/Common/DicomTagsManager.cs
```diff
@@ -114,8 +107,16 @@ public async Task<GetExtendedQueryTagEntry> UpdateExtendedQueryTagAsync(string t
...
-        var response = await _dicomWebClient.DeleteExtendedQueryTagAsync(tagPath, cancellationToken);
-        return response.StatusCode == System.Net.HttpStatusCode.NoContent;
+        try
+        {
+            var response = await _dicomWebClient.DeleteExtendedQueryTagAsync(tagPath, cancellationToken);
+            return response.StatusCode == System.Net.HttpStatusCode.NoContent;
+        }
+        catch (DicomWebException dwe) when (dwe.StatusCode == System.Net.HttpStatusCode.NotFound)
+        {
+            return false;
```
jovinson-ms, line 119: Why not return `true` here? This method is only being called as a cleanup utility, not to validate the deletion response.

pengchen0692 (approved):

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/18.diff.sql
```diff
@@ -0,0 +1,374 @@
...
+                @watermark,
+                S.TagValueUtc,
+                @resourceType
+            );
+    END
+
+    -- PersonName Key tags
+    IF EXISTS (SELECT 1 FROM @personNameExtendedQueryTags)
+    BEGIN
+        MERGE INTO dbo.ExtendedQueryTagPersonNam AS T
```
wsugarman, line 338: Oops! Typo and I should actually use this new version by bumping the max version

test/Microsoft.Health.Dicom.Web.Tests.E2E/Common/DicomTagsManager.cs
wsugarman: Delete will throw an exception if the status code is not success, so it should only return true when it really deleted an entry. But I think it could be useful for the return value to indicate whether deleted succeeded, even if it's not used in this particular scenario.

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/18.diff.sql
```diff
@@ -335,7 +327,7 @@ BEGIN
     -- PersonName Key tags
     IF EXISTS (SELECT 1 FROM @personNameExtendedQueryTags)
     BEGIN
-        MERGE INTO dbo.ExtendedQueryTagPersonNam AS T
```
pengchen0692: wow, we have this bug!

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/18.diff.sql
```diff
@@ -200,9 +204,6 @@ BEGIN
             AND ISNULL(T.SopInstanceKey2, @seriesKey) = @seriesKey
             -- Null InstanceKey indicates a Study/Series level tag, no to compare InstanceKey
             AND ISNULL(T.SopInstanceKey3, @instanceKey) = @instanceKey
-        WHEN MATCHED AND @watermark > T.Watermark THEN
```
pengchen0692: Add an integration test to cover this?

Issue #1396, opened 2022-03-04
Error deleting a dicom study or retrieving a study using REST API


mhassaninmsft:
**Describe the bug**
The application does not respect the HTTP 'Accept' headers in the request. When we provide -H 'accept:multipart/related; type=application/dicom'  The application only receives. `application/dicom` and not the multipart/related part of the header which causes some checks later in the request to fail (the checks enforce both application/dicom as the only supported media type and also expect the multipart/related section.
* The headers are conformat to the spec as specified [here](https://dicom.nema.org/medical/dicom/2019a/output/chtml/part18/sect_6.5.html)
* After debugging, it seems there is a bug with ASP.NET Typed headers as used in `\dicom-server\src\Microsoft.Health.Dicom.Api\Extensions\HttpRequestExtensions.cs`  The call to the method `httpRequest.GetTypedHeaders().Accept` returns a list of `MediaTypeHeaderValue` which seems strange, As per the class [doc](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.headers.mediatypeheadervalue?view=net-6.0) on Microsoft, the class is used for content type and not accept headers. The call to `httpRequest.GetTypedHeaders().Accept` only returns `application/dicom` and not the `multipart/related` section. However, a call to ` httpRequest.Headers.GetCommaSeparatedValues(HeaderNames.Accept)` returns the correct headers. 
* Ideally there should be a bug request for ASP.NET to fix how accept headers are handled. or in the meantime, the dicom code can change from the `MediaTypeHeaderValue` for accept header to a custom implemnted header handler.
* I tried forcing the `MediaTypeHeaderValue` to accept the correct headler by calling `new MediaTypeHeaderValue(new StringSegment(x, 0, x.Length))` where `x = "multipart/related; type=application/dicom"`, but that threw a formatting exception confirming it is not the correct abstraction.

**To Reproduce**
1.  Enable the openAPI Spec (optional).
2. Follow the instructions in the README to upload a dicom file using fiddler or postman
3. Try to delete or get that study using the swagger button which correpsonds to 
 curl -X 'GET' 'https://localhost:63838/v1/studies/1.2.826.0.1.3680043.8.498.13230779778012324449356534479549187420' -H 'accept:multipart/related; type=application/dicom'  -H 'content-type: application/dicom' substitue the study instace id with the apporpiate study instance if of the file you uploaded in step 2.
You will get an exception:
"Headers are not supported"

**Expected behavior**
[...]

PR #1462, opened 2022-03-28
Bug #89350: media type validation

mitchell-fox:
## Description
[PS3.18 (nema.org)](https://dicom.nema.org/medical/dicom/current/output/html/part18.html#sect_8.7):
"An mtp-value can be transmitted either as a token or quoted-string. The quoted and unquoted values are equivalent."

Currently, WADO with `Accept: multipart/related; type=application/dicom` will throw an error. It will accept `Accept: multipart/related; type="application/dicom"`, but they should be equivalent.

##Changes
Altered AcceptContentFilterAttribute.cs to account for media type values with missing quotes such as "multipart/related; type=application/dicom" instead of the usual "multipart/related; type='application/dicom'".

## Addresses
#1396 and [AB#89350](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/89350)

## Testing
Added new tests to AcceptContentFilterAttributeTests.cs.


src/Microsoft.Health.Dicom.Api/Features/Filters/AcceptContentFilterAttribute.cs
```diff
@@ -83,18 +105,12 @@ public override void OnActionExecuting(ActionExecutingContext context)
                     var stringHeaders = _mediaTypes.Select(x => x.ToString()).ToList();
                     if (split.Any(x => stringHeaders.Contains(x, StringComparer.InvariantCultureIgnoreCase)))
                     {
-                        acceptable = true;
-                        break;
+                        return true;
```
jovinson-ms, line 108: This whole block can be simplified to: 
```dotnet
if (_mediaTypes.Any(x => x.MatchesMediaType(acceptHeader.MediaType))) 
{
    return true;
}
```

src/Microsoft.Health.Dicom.Api.UnitTests/Features/Filters/AcceptContentFilterAttributeTests.cs
```diff
@@ -130,7 +130,11 @@ public void GivenARequestWithNoAcceptHeader_WhenValidatingTheContentTypeAgainstS
     [InlineData("multipart/related; type=\"application/octet_stream\"", (int)HttpStatusCode.NotAcceptable)]
     [InlineData("multipart/related; type=\"application/octetStream\"", (int)HttpStatusCode.NotAcceptable)]
     [InlineData("multipart/related; type=\"application/dicom\"", null)]
+    [InlineData("multipart/related; type='application/dicom'", (int)HttpStatusCode.NotAcceptable)]
+    [InlineData("multipart/related; type=application/dicom", null)]
     [InlineData("multipart/relATed; type=\"applICation/DICom\"", null)]
+    [InlineData("multipart/relATed; type='applICation/DICom'", (int)HttpStatusCode.NotAcceptable)]
+    [InlineData("multipart/relATed; type=applICation/DICom", null)]
```
jovinson-ms, line 137: We should add a test case with two parameters, like: `multipart/related; type=application/dicom; transfer-syntax="*"`

jovinson-ms (dismissed):

jovinson-ms:
As discussed, the Dicom standard requires the `type` parameter of the `Accept` header to be quoted, since it can include a `/` special character. This is already reflected in our conformance statement.

shivakumar-ms (dismissed):

mitchell-fox:
Committed code for updated PR scope. New code simplifies AcceptContentFilterAttribute so it no longer handles multipart Accept headers as it was never used in the codebase.

bcarthic:
I am surprised none of our E2E tests are broken. Should we add a test?

mitchell-fox:
That's a good idea. I can put them under Microsoft.Health.Dicom.Web.Tests.E2E -> Rest -> AcceptHeaderTests.cs. Should only need a couple E2E tests to validate everything past the updated unit tests.

src/Microsoft.Health.Dicom.Api/Features/Filters/AcceptContentFilterAttribute.cs
```diff
@@ -13,22 +13,17 @@
...
     private const int NotAcceptableResponseCode = (int)HttpStatusCode.NotAcceptable;
-    private const string TypeParameter = "type";
-
-    private readonly bool _allowSingle;
-    private readonly bool _allowMultiple;
 
     private readonly HashSet<MediaTypeHeaderValue> _mediaTypes;
 
-    public AcceptContentFilterAttribute(string[] mediaTypes, bool allowSingle, bool allowMultiple)
+    public AcceptContentFilterAttribute(string[] mediaTypes)
```
smithago, line 26: Can we confirm this change also works down stream? There are places like dicom-server\src\Microsoft.Health.Dicom.Api\Extensions\MediaTypeHeaderValueExtensions.cs, where the acceptHeader gets converted to internal objects and passed into service, is the type with and without quotes passed in as same value downstream?

jovinson-ms: The service currently uses two code paths to handle media type content negotiation - `AcceptContentFilterAttribute` is used only for single part media types like `application/dicom`, and the `GetAcceptHeaders` extension method is only used on controller actions that use the `Produces` attribute for `multipart/related` media types, and then parses to find the media type from the `type` parameter. 

Once we realized that the `AcceptContentFilterAttribute` doesn't handle `multipart/related`, we decided to explicitly handle only single part media types there, and consider a future refactor to handle multipart media type content negotiation at the API layer as well.

src/Microsoft.Health.Dicom.Api/Features/Filters/AcceptContentFilterAttribute.cs
```diff
@@ -46,55 +41,46 @@ public AcceptContentFilterAttribute(string[] mediaTypes, bool allowSingle, bool
...
-                        break;
-                    }
-                }
+    private bool AcceptHeadersContainKnownTypes(HttpRequest request)
+    {
+        IList<MediaTypeHeaderValue> acceptHeaders = ParseAcceptHeaders(request.Headers.Accept);
+
+        foreach (MediaTypeHeaderValue acceptHeader in acceptHeaders)
+        {
+            if (_mediaTypes.Any(x => x.MatchesMediaType(acceptHeader.MediaType)))
```
jovinson-ms, line 66: As discussed, we will flip this check - the defined media types for this action should be a subset of the Accept header media range, e.g. `application/dicom` is a subset of `application/*` and so will match.

jovinson-ms (approved):

PR #1472, opened 2022-04-04
Refactor Indexing Input

wsugarman:
## Description
- Update shared operations package to latest
- Separate input from checkpoint for re-indexing
- Move batch options to orchestration input, as it determines the number of generated activities in addition to its output
- Refactor client options to allow function names to be specified

## Related issues
N/A

## Testing
Pipeline


src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Orchestration.cs
```diff
@@ -35,13 +36,18 @@ public partial class ReindexDurableFunction
...
         logger = context.CreateReplaySafeLogger(logger);
-        ReindexInput input = context.GetInput<ReindexInput>();
+        ReindexCheckpoint input = context.GetInput<ReindexCheckpoint>();
 
-        // The ID should be a GUID as generated by the trigger, but we'll assert here just to make sure!
-        context.ThrowIfInvalidOperationId();
+        // Backfill batching options
+        input.Batching ??= new BatchingOptions
+        {
+            MaxParallelCount = _options.MaxParallelBatches,
```
wsugarman, line 48: should probably deprecate the function options and instead rely on input

pengchen0692: Why we prefer moving this setting to input?

The benefit I could think about is to better optimize based on input? 

The drawback I could think about is we could not globally change these value by deployment -- like last time we reduced parallel level to reduce CPU usage.

src/Microsoft.Health.Dicom.Core/Models/Indexing/ReindexCheckpoint.cs
```diff
@@ -0,0 +1,38 @@
...
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using Microsoft.Health.Dicom.Core.Features.Model;
+using Microsoft.Health.Operations;
+
+namespace Microsoft.Health.Dicom.Core.Models.Indexing;
+
+internal class ReindexCheckpoint : ReindexInput, IOperationCheckpoint
```
pengchen0692, line 15: we should figure out a way to test such change to avoid breaking existing operations.

src/Microsoft.Health.Dicom.Web/appsettings.Development.json
```diff
@@ -3,8 +3,15 @@
     "ConnectionString": "UseDevelopmentStorage=true"
   },
   "DicomFunctions": {
-    "ConnectionName": "DicomFunctions:ConnectionString",
-    "ConnectionString": "UseDevelopmentStorage=true"
+    "DurableTask": {
+      "ConnectionName": "DicomFunctions:DurableTask:ConnectionString",
+      "ConnectionString": "UseDevelopmentStorage=true"
+    },
+    "Indexing": {
```
pengchen0692, line 10: reason for adding this? for easy testing?

src/Microsoft.Health.Dicom.Functions/Indexing/ReindexDurableFunction.Orchestration.cs
wsugarman: We have to set this information in the input, as it impacts the number of generated activities. However, the number of threads does not need to be based as input

pengchen0692: so  for different reindex job, we will have different parallel level in the future?

pengchen0692 (dismissed):

pengchen0692 (approved):

PR #1475, opened 2022-04-04
When there are multiple values, index the first one instead of failing validation.

pengchen0692:
## Description
When there are multiple values, index the first one instead of failing validation
* Add Warning header when there are multiple values for indexed dicom tags
* Change `GetSingleValueOrDefault` to `GetFirstValueOrDefault`
* Update all validations to verify only the first value.

## Related issues
Addresses [AB#89792](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/89792)

## Testing
All tests passed


jovinson-ms:
Looks good - can you add some language to the QIDO portion of the conformance statement and maybe the extended query tag doc as well about only indexing the first value of a single valued data element that incorrectly has multiple values?

docs/resources/conformance-statement.md
```diff
@@ -435,6 +434,7 @@ The query API will return one of the following status codes in the response:
 - Paged results are optimized to return matched *newest* instance first, this may result in duplicate records in subsequent pages if newer data matching the query was added.
 - Matching is case in-sensitive and accent in-sensitive for PN VR types.
 - Matching is case in-sensitive and accent sensitive for other string VR types.
+- Only index the first value of a single valued data element that incorrectly has multiple values.
```
jovinson-ms, line 437: I would reword this as "Only the first value will be indexed of a ..."

docs/how-to-guides/extended-query-tags.md
```diff
@@ -71,7 +71,7 @@ The following VR types are supported:
 
 > You can add up to 128 extended query tags.
 
-> If multiple values are specified for DICOM tag, only first one is indexed.
+> Only index the first value of a single valued data element that incorrectly has multiple values.
```
jovinson-ms, line 74: I would reword this as "Only the first value will be indexed of a ..."

jovinson-ms (approved):

PR #1477, opened 2022-04-05
Track number of parts in multipart request

jnlycklama:
## Description
Track number of parts in multipart request. RequestParts will be used to determine the # of ApiRequests to bill

We are billing for each part sent to us in STOW (aka - each instance) & each part returned in the response for WADO requests. For most requests this is the # of instances returned, except for /frames requests, where we bill for the number of frames returned.

## Related issues
Addresses #89153.

## Testing
Generate local nuget packages, deploy to test dicom service and try sending multipart and normal requests


src/Microsoft.Health.Dicom.Api.UnitTests/Web/AspNetCoreMultipartReaderTests.cs
```diff
@@ -174,6 +180,7 @@ public async Task GivenMissingMultipartBodyPartException_WhenReading_NoMoreSecti
             MultipartBodyPart result = await aspNetCoreMultipartReader.ReadNextBodyPartAsync(cancellationToken: default);
 
             Assert.Null(result);
+            Assert.True(_dicomRequestContextAccessor.RequestContext.RequestParts == 0);
```
wsugarman, line 183: nit: use `Assert.Equals`

src/Microsoft.Health.Dicom.Api.UnitTests/Web/AspNetCoreMultipartReaderTests.cs
```diff
@@ -280,17 +293,24 @@ private async Task ExecuteAndValidateAsync(string body, string requestContentTyp
...
+                if (result != null)
+                {
+                    numberOfParts++;
+                }
             }
 
             result = await aspNetCoreMultipartReader.ReadNextBodyPartAsync(cancellationToken: default);
 
             Assert.Null(result);
+            Assert.True(_dicomRequestContextAccessor.RequestContext.RequestParts == numberOfParts);
```
wsugarman, line 313: nit: use `Assert.Equals`

src/Microsoft.Health.Dicom.Core/Features/Context/IDicomRequestContext.cs
```diff
@@ -20,5 +20,7 @@ public interface IDicomRequestContext : IRequestContext
 
     long BytesTranscoded { get; set; }
 
+    int RequestParts { get; set; }
```
wsugarman, line 23: How exactly will this be used? Somewhere we'll wait for each of the parts to be read and then pass this information as a billing metric?

jnlycklama: In `BillingMiddleware.cs` in workspace-platform repo. It already reads some other fields from the request context (like BytesTranscoded), so now it will also read RequestParts and use that to bill for ApiRequests

jovinson-ms (dismissed):

src/Microsoft.Health.Dicom.Api/Web/AspNetCoreMultipartReader.cs
```diff
@@ -125,9 +129,13 @@ public async Task<MultipartBodyPart> ReadNextBodyPartAsync(CancellationToken can
             // The stream must be consumed before the next ReadNextSectionAsync is called.
             // Also, the stream returned by the MultipartReader is not seekable. We need to make
             // it seekable so that we can process the stream multiple times.
-            return new MultipartBodyPart(
+            MultipartBodyPart part = new MultipartBodyPart(
                 contentType,
                 await _seekableStreamConverter.ConvertAsync(section.Body, cancellationToken));
+
+            _dicomRequestContextAccessor.RequestContext.RequestParts++;
```
smithago, line 136: Can we track the billing informations in service layer? StoreService/StoreOrchestrator can give you the same

src/Microsoft.Health.Dicom.Core/Features/Context/DicomRequestContext.cs
```diff
@@ -75,6 +75,8 @@ public class DicomRequestContext : IDicomRequestContext
 
     public long ResponseSize { get; set; }
 
+    public int RequestParts { get; set; }
```
smithago, line 78: How about mulitpart response? WADO can get multiple instances in a study?

Should we have more generic InstanceCount or something like that?

smithago (changes requested):
How about multi-part response

src/Microsoft.Health.Dicom.Api/Web/AspNetCoreMultipartReader.cs
jnlycklama: Yeah that's a good idea - Moved to StoreService

src/Microsoft.Health.Dicom.Core/Features/Context/DicomRequestContext.cs
jnlycklama: Just updated to "PartCount", since we need to count the number of frame for frame WADO requests (as discussed), so "InstanceCount" doesn't make sense anymore

jnlycklama:
> How about multi-part response

Updated to bill for multipart WADO responses

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Retrieve/RetrieveMetadataServiceTests.cs
```diff
@@ -137,6 +137,7 @@ public async Task GivenRetrieveInstanceMetadataRequestForStudy_WhenIsSuccessful_
         RetrieveMetadataResponse response = await _retrieveMetadataService.RetrieveStudyInstanceMetadataAsync(_studyInstanceUid, ifNoneMatch, DefaultCancellationToken);
 
         Assert.Equal(response.ResponseMetadata.Count(), versionedInstanceIdentifiers.Count);
+        Assert.Equal(response.ResponseMetadata.Count(), _dicomRequestContextAccessor.RequestContext.PartCount);
```
wsugarman, line 140: nit: I think this equality check is backwards (at least for the purpose of the possible error message if it fails). I think it should be `Assert.Fails(<expected>, <actual>)`

src/Microsoft.Health.Dicom.Core.UnitTests/Microsoft.Health.Dicom.Core.UnitTests.csproj
```diff
@@ -24,6 +24,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\Microsoft.Health.Dicom.Api.UnitTests\Microsoft.Health.Dicom.Api.UnitTests.csproj" />
```
wsugarman, line 27: We should avoid this sort of dependency. Can we use `NSubstitute` instead?

src/Microsoft.Health.Dicom.Core/Features/Store/StoreService.cs
```diff
@@ -80,6 +84,7 @@ public class StoreService : IStoreService
     {
         if (instanceEntries != null)
         {
+            _dicomRequestContextAccessor.RequestContext.PartCount = instanceEntries.Count;
```
wsugarman, line 87: I think this is fine for now, but seeing this mysterious type collect assignments around the code seems weird. Like, is there a place where this sort of computation/interjection makes sense. This is the current pattern though, so I think it's ok for now lol.

src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
```diff
@@ -119,6 +121,7 @@ public async Task<RetrieveResourceResponse> GetInstanceResourceAsync(RetrieveRes
                             isOriginalTransferSyntaxRequested,
                             requestedTransferSyntax);
 
+                        _dicomRequestContextAccessor.RequestContext.PartCount = message.Frames.Count();
```
wsugarman, line 124: nit: can the `Frames` property be some sort of read-only collection instead of only `IEnumerable<T>`?

wsugarman (changes requested):

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Retrieve/RetrieveMetadataServiceTests.cs
jnlycklama: I want to assert that the PartCount has been updated to the response count - so in this case wouldn't the "actual" be PartCount and the "expected" be response count?

src/Microsoft.Health.Dicom.Core/Features/Store/StoreService.cs
jnlycklama: Yeah, I agree it's a bit odd but I can't think of a singular place that makes sense. I assume you are thinking of all the other fields on the request context used to capture billing as well (BytesTranscoded, IsTranscodeRequested), so it would need to have access to all of the internal data here.

Maybe in the future we could have something after the service layer (StoreService, RetrieveResourceService, RetrieveMetadataService), but before it returns a response object. It could take the stream data & then grab the necessary data to add to the request context. But I haven't fully thought this out so I don't know if it would make it better or worse haha

smithago: Plus 1, someway to separate billing concern

smithago (approved):

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Retrieve/RetrieveMetadataServiceTests.cs
wsugarman: Ah yes, I see. You're correct

wsugarman (approved):

PR #1491, opened 2022-04-07
Handle race conditions during parallel instance inserts in the same study

smithago:
## Description
We are holding a UPDLOCK on study which still allows inserts. Parallel new instance inserts in a study can result in dupe key on studyInstanceUID. Holding a SERIALIZATION lock is correct but results in slow instance inserts and deadLocks on rangelock.

## Related issues
[AB#90184](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/90184)

## Testing
Added new integration test.
Manually tests


src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/AddInstanceV6.sql
```diff
@@ -112,14 +112,24 @@ BEGIN
         AND StudyInstanceUid = @studyInstanceUid
 
     IF @@ROWCOUNT = 0
-    BEGIN
+    BEGIN TRY
```
wsugarman, line 115: Do we need to worry about this happening when inserting into the `dbo.Series` table  too in case the `dbo.Study` row already existed but the `dbo.Series` row did not?

test/Microsoft.Health.Dicom.Tests.Integration/Persistence/IndexDataStoreTests.cs
```diff
@@ -557,15 +557,29 @@ public async Task GivenMaxTagKeyNotMatch_WhenCreateIndex_ThenShouldThrowExceptio
...
+        DicomDataset instance1 = CreateTestDicomDataset(studyInstanceUid, seriesInstanceUid);
+        DicomDataset instance2 = CreateTestDicomDataset(studyInstanceUid, seriesInstanceUid);
+
+        Task.WaitAll(
+                _indexDataStore.BeginCreateInstanceIndexAsync(1, instance1),
+                _indexDataStore.BeginCreateInstanceIndexAsync(1, instance2));
+    }
+
     private static void ValidateStudyMetadata(
-        string expectedStudyInstanceUid,
```
wsugarman, line 561: nit: I think these parameters need to be indented one level

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/AddInstanceV6.sql
wsugarman: Oh, never mind. The `UPDATE` on `dbo.Study` will force the transactions to execute serially so that there is never a problem adding the row into `dbo.Series`

wsugarman (approved):

PR #1519, opened 2022-04-21
Fix warning header: include agent as host

pengchen0692:
## Description
Fix warning header: include agent as host

## Related issues
Addresses [AB#91168](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/91168)


src/Microsoft.Health.Dicom.Api/DicomApiResource.resx
```diff
@@ -173,7 +173,7 @@
     <comment>{0} is field.</comment>
   </data>
   <data name="WarningHeader" xml:space="preserve">
-    <value>299: {0}</value>
-    <comment>{0} is the warning message</comment>
+    <value>299 {0}: "{1}"</value>
+    <comment>{0} is host name, {1} is the warning message</comment>
```
wsugarman, line 177: This probably shouldn't be in here, as this `string` is not culture-specific.

src/Microsoft.Health.Dicom.Api/Extensions/HttpResponseExtensions.cs
```diff
@@ -25,6 +25,19 @@ public static void AddLocationHeader(this HttpResponse response, Uri locationUrl
...
+    /// Set Response Warning header.
+    /// </summary>
+    /// <param name="response">The httpResponse.</param>
+    /// <param name="host">Host name.</param>
+    /// <param name="message">The warning message.</param>
+    public static void SetWarning(this HttpResponse response, string host, string message)
+    {
+        EnsureArg.IsNotNull(response, nameof(response));
+        EnsureArg.IsNotEmptyOrWhiteSpace(message, nameof(message));
+        response.Headers.Warning = string.Format(DicomApiResource.WarningHeader, host ?? string.Empty, message);
```
wsugarman, line 38: If `host` is `null` or empty I think it needs to be `'-'` instead

pengchen0692: yeah, that's right!

src/Microsoft.Health.Dicom.Api/DicomApiResource.resx
wsugarman: While no longer in use, this change still needs to be deleted

jovinson-ms (approved):

src/Microsoft.Health.Dicom.Core/Models/HttpWarningCode.cs
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Dicom.Core.Models;
+
+/// <summary>
+/// Represents Http Warning code.  https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Warning#warning_codes
+/// </summary>
+[System.Diagnostics.CodeAnalysis.SuppressMessage("Design", "CA1008:Enums should have zero value", Justification = "0 is not valid HttpWarningCode")]
+public enum HttpWarningCode
```
bcarthic, line 12: Do we need this class? Why can't we direcly set the value from the controller

src/Microsoft.Health.Dicom.Api/Controllers/WorkitemController.Cancel.cs
```diff
@@ -63,9 +63,16 @@ public async Task<IActionResult> CancelAsync(string workitemInstanceUid)
         if (response.Status is Core.Messages.Workitem.WorkitemResponseStatus.Conflict
             && !string.IsNullOrEmpty(response.Message))
         {
-            Response.Headers.Warning = string.Format(DicomApiResource.WarningHeader, response.Message);
+            Response.SetWarning(Core.Models.HttpWarningCode.MiscPersistentWarning, GetHost(), response.Message);
```
bcarthic, line 66: Why can't we directly set the code, why do we need the enum?

src/Microsoft.Health.Dicom.Api/DicomApiResource.resx
pengchen0692: yeah, forgot to do that...

src/Microsoft.Health.Dicom.Api/Controllers/WorkitemController.Cancel.cs
```diff
@@ -63,9 +63,16 @@ public async Task<IActionResult> CancelAsync(string workitemInstanceUid)
...
         }
 
         return StatusCode((int)response.Status.CancelResponseToHttpStatusCode(), response.Message);
     }
+
+    private string GetHost()
+    {
+        string host = Request.Host.Host;
+        // As Dicom standard, should append colon after service. https://dicom.nema.org/medical/dicom/current/output/chtml/part18/sect_11.7.3.2.html
+        return string.IsNullOrWhiteSpace(host) ? host : host + ":";
```
wsugarman, line 76: Is the colon still present if the host is missing and replaced with `'-'`?

wsugarman (approved):

src/Microsoft.Health.Dicom.Api/Controllers/WorkitemController.Cancel.cs
pengchen0692: yeah, that's the trickly part.  As http standard, it should be `'_'`.

PR #1520, opened 2022-04-21
Add KeyVault client to store and retrieve secrets

smithago:
## Description
- SecretStore implementing Azure KeyVault
-  For running in docker, since there is no KeyVault we need probably a local SecretManager?

## Related issues
[AB#91155](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/91155)

## Testing
Will be tested with Export


src/Microsoft.Health.Dicom.Azure/Registration/KeyVaultClientRegistrationExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+using EnsureThat;
+using Microsoft.Extensions.Azure;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Health.Dicom.Azure.Config;
+using Microsoft.Health.Dicom.AzureKeyVault;
+using Microsoft.Health.Dicom.Core.Features.Common;
+using Microsoft.Health.Dicom.Core.Registration;
+
+namespace Microsoft.Health.Dicom.Azure.Registration;
```
wsugarman, line 16: nit: missing new line

src/Microsoft.Health.Dicom.Azure/Registration/KeyVaultClientRegistrationExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        var config = new KeyVaultConfiguration();
+
+        configuration.GetSection(KeyVaultConfiguration.SectionName).Bind(config);
+        configureAction?.Invoke(config);
+
+        // conditional KeyVault registration to support running on local docker
+        if (!string.IsNullOrWhiteSpace(config.Endpoint))
+        {
+
```
wsugarman, line 36: nit: extra new line

src/Microsoft.Health.Dicom.Azure/Registration/KeyVaultClientRegistrationExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+            dicomServerBuilder.Services.AddAzureClients(
+                    builder =>
+                    {
+                        builder.AddSecretClient(new Uri(config.Endpoint))
+                        .WithCredential(new DefaultAzureCredential());
+                    });
+
+            dicomServerBuilder.Services.AddScoped<ISecretStore, KeyVaultSecretStore>();
```
wsugarman, line 44: Should there be a default "passthrough" secret store if its not enabled? Although perhaps that'd be configured by the core dicom lib instead of this one?

src/Microsoft.Health.Dicom.Azure/KeyVault/KeyVaultSecretStore.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace Microsoft.Health.Dicom.AzureKeyVault;
+
+public class KeyVaultSecretStore : ISecretStore
+{
+    private readonly SecretClient _secretClient;
+
+    public KeyVaultSecretStore(SecretClient secretClient)
+    {
+        _secretClient = secretClient;
```
wsugarman, line 17: nit: `null`-check?

src/Microsoft.Health.Dicom.Azure/KeyVault/KeyVaultSecretStore.cs
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public async Task SetSecretAsync(string secretName, string secretValue, CancellationToken cancellationToken)
+    {
+        await _secretClient.SetSecretAsync(secretName, secretValue, cancellationToken);
```
wsugarman, line 22: nit: `await` is unnecessary

src/Microsoft.Health.Dicom.Core/Features/Common/ISecretStore.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Health.Dicom.Core.Features.Common;
+
+public interface ISecretStore
+{
+    Task DeleteSecretAsync(string secretName, CancellationToken cancellationToken = default);
+
+    Task<string> GetSecretAsync(string secretName, CancellationToken cancellationToken = default);
```
wsugarman, line 15: So are secrets created per export operation? I was under the impression we were going to encrypt SAS tokens instead or storing them as secrets.

src/Microsoft.Health.Dicom.Azure/Microsoft.Health.Dicom.Azure.csproj
```diff
@@ -0,0 +1,22 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>Implementation of Azure dependencies used in DICOM core library</Description>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
```
wsugarman, line 6: nit: I don't think we want implicit usings. But we could try `nullable` here lol

src/Microsoft.Health.Dicom.Azure/Registration/KeyVaultClientRegistrationExtensions.cs
smithago: Correct, we can create dummy one in another library if we want it

src/Microsoft.Health.Dicom.Azure/Microsoft.Health.Dicom.Azure.csproj
smithago: Nullable is painful. I can remove implicit

src/Microsoft.Health.Dicom.Core/Features/Common/ISecretStore.cs
smithago: We can do that too, but we will have to rotate the encryption key and all I think to meet the requirements.

smithago: Mohammad's code was just storing and retrieving the secret has here

wsugarman: I am just worried that if we store each SAS key individually then we'll have to manage their lifecycle (like their eventual removal). And then if the export were to fail for some reason, we'll need to remove old SAS keys... which means we'll need to probably have some timer task that runs periodically to clean up.

And at the point, perhaps we should just have a task whose job it is to rotate the encryption key every so often? Maybe there is a third approach?

smithago: Rotation also adds the need of overlapping keys to make sure we can gracefully handle existing running ones.

smithago: And I am not sure if it still passes/recommended in the security requirement. Meaning you can encrypt and store the secret in a normal datastore vs using a secret store

src/Microsoft.Health.Dicom.Azure/Registration/KeyVaultClientRegistrationExtensions.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        // conditional KeyVault registration to support running on local docker
+        if (!string.IsNullOrWhiteSpace(config.Endpoint))
+        {
+
+            dicomServerBuilder.Services.AddAzureClients(
+                    builder =>
+                    {
+                        builder.AddSecretClient(new Uri(config.Endpoint))
+                        .WithCredential(new DefaultAzureCredential());
```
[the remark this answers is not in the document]
smithago: This is using DefaultAzureCredential() which will use MI to connect

src/Microsoft.Health.Dicom.Azure/Registration/KeyVaultClientRegistrationExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+        configureAction?.Invoke(config);
+
+        // conditional KeyVault registration to support running on local docker
+        if (!string.IsNullOrWhiteSpace(config.Endpoint))
+        {
+            dicomServerBuilder.Services.AddAzureClients(
+                    builder =>
+                    {
+                        builder.AddSecretClient(new Uri(config.Endpoint))
+                        .WithCredential(new DefaultAzureCredential());
```
wsugarman, line 42: Should we support a non-identity based connection to key vault, if such a thing exists? For example, we support using connection strings and secrets to authenticate to SQL and azure storage today. Is there a similar mechanism necessary for azure key vault?

wsugarman (approved):

samples/templates/default-azuredeploy.json
```diff
@@ -193,9 +193,10 @@
             "DicomServer__Features__EnableExtendedQueryTags": "true",
             "DicomServer__Security__Enabled": "[variables('securityAuthenticationEnabled')]",
             "DicomServer__Security__Authentication__Authority": "[parameters('securityAuthenticationAuthority')]",
-            "DicomServer__Security__Authentication__Audience": "[parameters('securityAuthenticationAudience')]"
+            "DicomServer__Security__Authentication__Audience": "[parameters('securityAuthenticationAudience')]",
+            "KeyVault__Endpoint": "[variables('keyVaultEndpoint')]"
```
wsugarman, line 197: I think we can directly use the `reference` function with the [`vaultUri`](https://docs.microsoft.com/en-us/azure/templates/microsoft.keyvault/vaults?tabs=bicep) property. Although we may need to wait until it has been created?

```suggestion
            "KeyVault__Endpoint": "[reference(resourceId('Microsoft.KeyVault/vaults', variables('serviceName'))).vaultUri]"
```

Or later we have to dynamically add the JSON like: `[reference(variables('keyVaultResourceId')).vaultUri]`

wsugarman (approved):

PR #1529, opened 2022-04-22
Fix CI Deployment

wsugarman:
## Description
- Replace deprecated parameters in deploy script

## Related issues
N/A

## Testing
Look very closely

### DIFF
--- a/release/templates/featuresenabled-azuredeploy.json
+++ b/release/templates/featuresenabled-azuredeploy.json
@@ -10,6 +10,13 @@
                 "description": "Name of the DICOM Service Web App."
             }
         },
+        "appServicePlanResourceGroup": {
+            "type": "string",
+            "defaultValue": "",
+            "metadata": {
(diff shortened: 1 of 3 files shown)


release/templates/featuresenabled-azuredeploy.json
```diff
@@ -62,7 +55,7 @@
     "variables": {
         "loweredServiceName": "[toLower(parameters('serviceName'))]",
         "featuresEnabledServiceName": "[concat(variables('loweredServiceName'),'-featuresenabled')]",
-        "appServicePlanResourceId": "[resourceId(parameters('appServicePlanResourceGroup'), 'Microsoft.Web/serverfarms/', parameters('appServicePlanName'))]",
```
brandonpollett, line 65: I suppose this means we're always targeting an app service plan that's  in the same resource group as everything else now?

brandonpollett (approved):

PR #1545, opened 2022-04-27
Add Export Operation

wsugarman:
## Description
- Create source and sink APIs for copying data
- Add new durable function for export

## Related issues
Addresses [AB#90232](https://microsofthealth.visualstudio.com/Health/_workitems/edit/90232/)

## Testing
Pipeline


src/Microsoft.Health.Dicom.Functions/Export/ExportDurableFunction.Activity.cs
```diff
@@ -0,0 +1,96 @@
...
+    {
+        EnsureArg.IsNotNull(context, nameof(context));
+        EnsureArg.IsNotNull(logger, nameof(logger));
+
+        ExportBatchArguments args = context.GetInput<ExportBatchArguments>();
+        await using IExportSource source = _sourceFactory.CreateSource(args.Source);
+        await using IExportSink sink = _sinkFactory.CreateSink(args.Destination, context.GetOperationId());
+
+        // Export
+        source.ReadFailure += (source, e) => logger.LogError(e.Exception, "Cannot read desired DICOM file(s)");
```
pengchen0692, line 49: Are these failures for internal tracking or write into the log file shown to customer?

pengchen0692 (dismissed):

src/Microsoft.Health.Dicom.Functions/Export/ExportDurableFunction.Activity.cs
wsugarman: Both. The event and logging is for internal use, but the sink should also write the error to the error file

pengchen0692 (approved):

PR #1557, opened 2022-05-04
Implement Azure Blob Export Sink

wsugarman:
## Description
- Implement `AzureBlobExportSink`, and its corresponding providers
- Add `ExportService`
- Add new registration methods
- Refactor operations client:
  - Accept operation Id as input instead of generating it. This way callers can perform actions based on the ID of the job they're about to submit (e.g. for export, the store uploads a secret with the job ID as the key)
  - Standard APIs to return an `OperationReference` for the operation they submitted
- Remove `ISensitive` as it's no longer needed

## Related issues
Addresses [AB#90232](https://microsofthealth.visualstudio.com/Health/_workitems/edit/90232)
Addresses [AB#91160](https://microsofthealth.visualstudio.com/Health/_workitems/edit/91160)

## Testing
New Tests


src/Microsoft.Health.Dicom.Core/Features/Common/GuidFactory.cs
```diff
@@ -5,7 +5,7 @@
 
 using System;
 
-namespace Microsoft.Health.Dicom.Functions.Client.DurableTask;
+namespace Microsoft.Health.Dicom.Core.Features.Common;
```
smithago, line 8: I am curious, why do we need a factory to get the functionality of Guid.NewGuid()?

wsugarman: Mostly so I can inject a deterministic `Guid` value during testing, like `Clock` for `DateTime.Now`

src/Microsoft.Health.Dicom.Blob/Features/Export/ExportPatternPlaceholder.cs
```diff
@@ -0,0 +1,17 @@
+﻿// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+
+[Flags]
+internal enum ExportPatternPlaceholders : sbyte
```
smithago, line 9: Do not understand. Do we need this? Should we need to parameterize this?

src/Microsoft.Health.Dicom.Blob/Features/Export/AzureBlobExportSink.cs
```diff
@@ -0,0 +1,166 @@
...
+using Microsoft.Health.Dicom.Core.Features.Common;
+using Microsoft.Health.Dicom.Core.Features.Export;
+using Microsoft.Health.Dicom.Core.Models.Common;
+using Microsoft.Health.Dicom.Core.Models.Export;
+
+namespace Microsoft.Health.Dicom.Blob.Features.Export;
+
+internal sealed class AzureBlobExportSink : IExportSink
+{
+    public Uri ErrorHref
```
smithago, line 31: nit: should this go after the constructor?

src/Microsoft.Health.Dicom.Blob/Features/Export/AzureBlobExportSink.cs
```diff
@@ -0,0 +1,166 @@
...
+    private readonly AzureBlobExportFormatOptions _output;
+    private readonly BlobOperationOptions _blobOptions;
+    private readonly JsonSerializerOptions _jsonOptions;
+
+    private const int BlockSize = 2 * 1024 * 1024;
+
+    public AzureBlobExportSink(
+        IFileStore source,
+        BlobContainerClient dest,
+        IOptions<AzureBlobExportFormatOptions> outputOptions,
```
smithago, line 58: If this is not user controlled, should this be passed in or implicit?

src/Microsoft.Health.Dicom.Blob/Features/Export/ExportFilePattern.cs
```diff
@@ -0,0 +1,72 @@
...
+using System.Text;
+using EnsureThat;
+using Microsoft.Health.Dicom.Core.Features.Model;
+using Microsoft.Health.Operations;
+
+namespace Microsoft.Health.Dicom.Blob.Features.Export;
+
+internal static class ExportFilePattern
+{
+    public static string Parse(string pattern, ExportPatternPlaceholders placeholders = ExportPatternPlaceholders.All)
```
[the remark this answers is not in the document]
wsugarman: Yeah, I can add some comments. Here I am trying to parse from some format of like `/%Study%/%Series%/%SopInstance%.dcm` to a .NET format string like `{0}/{1}/{2}.dcm`. The idea was to provide users with a convenient way of specifying their custom paths without having to use our own format strings with numbers. However, I needed some translation layer in our code

src/Microsoft.Health.Dicom.Core/Features/Export/ExportService.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        // Validate the input and update the specification
+        specification = new ExportSpecification
+        {
+            Source = await _sourceFactory.ValidateAsync(specification.Source, cancellationToken),
+            Destination = await _sinkFactory.ValidateAsync(specification.Destination, operationId, cancellationToken),
+        };
+
+        // Start the operation
+        return await _client.StartExportAsync(operationId, specification, _partition, cancellationToken);
```
smithago, line 55: I was expecting SAS token to be put in the KV here? Did I miss that code? I also did see it being read in Export Sink

src/Microsoft.Health.Dicom.Blob/Features/Export/ExportFilePattern.cs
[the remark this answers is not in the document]
smithago: Lets add it when we expose that feature.

src/Microsoft.Health.Dicom.Core/Features/Export/ExportService.cs
wsugarman: As discussed offline, I have separated the logic outside of `ValidateAsync`

src/Microsoft.Health.Dicom.Blob/Features/Export/AzureBlobExportSink.cs
wsugarman: I have made the output options `internal`, but we can still pass it in as an options as if it's customizable, should we ever need to change this.

src/Microsoft.Health.Dicom.Blob/Features/Export/AzureBlobExportSink.cs
wsugarman: Good catch

src/Microsoft.Health.Dicom.Blob/Features/Export/ExportPatternPlaceholder.cs
wsugarman: As discussed offline, I went all in on the path specification lol. I have made this `internal` for now

src/Microsoft.Health.Dicom.Blob/Features/Export/ExportFilePattern.cs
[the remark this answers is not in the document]
wsugarman: As discussed, I have made this `internal` for now

src/Microsoft.Health.Dicom.Blob/Features/Export/AzureBlobExportFormatOptions.cs
```diff
@@ -0,0 +1,39 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Text;
+using EnsureThat;
+using Microsoft.Health.Dicom.Core.Features.Model;
+
+namespace Microsoft.Health.Dicom.Blob.Features.Export;
+
+internal sealed class AzureBlobExportFormatOptions
```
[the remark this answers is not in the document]
wsugarman: For now it's an option, but the user can't control it. It's more about for the future if we ever need to change things

src/Microsoft.Health.Dicom.Core/Features/Common/GuidFactory.cs
wsugarman: Will leave a comment

smithago (approved):

PR #1562, opened 2022-05-04
Fix Deploy Button in Quick Start

wsugarman:
## Description
- Fix button in quick start

## Related issues
N/A

## Testing
Clicked it

### DIFF
--- a/docs/quickstarts/deploy-via-azure.md
+++ b/docs/quickstarts/deploy-via-azure.md
@@ -4,11 +4,8 @@ In this quickstart, you'll learn how to deploy the Medical Imaging Server for DI
 
 If you do not have an Azure subscription, create a [free account](https://azure.microsoft.com/free) before you begin.
 
-Once you have your subscription, click the link below to begin deployment.
-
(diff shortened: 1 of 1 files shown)


docs/quickstarts/deploy-via-azure.md
```diff
@@ -4,11 +4,8 @@ In this quickstart, you'll learn how to deploy the Medical Imaging Server for DI
 
 If you do not have an Azure subscription, create a [free account](https://azure.microsoft.com/free) before you begin.
 
-Once you have your subscription, click the link below to begin deployment.
-
-<a href="https://portal.azure.com/#create/Microsoft.Template/uri/https%3A%2F%2Fdcmcistorage.blob.core.windows.net%2Fcibuild%2Fdefault-azuredeploy.json" target="_blank">
```
brandonpollett, line 9: Look like we're still publishing to this storage account, is there a reason it no longer works?

brandonpollett (approved):

docs/quickstarts/deploy-via-azure.md
wsugarman: I am unsure at the moment. I'm digging into it, but the samples on our README work. Furthermore, the images are indeed in our blob storage and the container is publicly accessible.

PR #1563, opened 2022-05-05
Add InMemorySecretStore and Minor Refactors

wsugarman:
## Description
- Add new `InMemorySecretStore`
- Update `ISecretStore` API to use version in addition to secret name
- Add new `ISensitive` interface for storing sensitive information
- Add new unit tests for the Azure project

## Related issues
Addresses [AB#90232](https://microsofthealth.visualstudio.com/Health/_workitems/edit/90232)
Addresses [AB#91160](https://microsofthealth.visualstudio.com/Health/_workitems/edit/91160)

## Testing
New unit tests


src/Microsoft.Health.Dicom.Azure/KeyVault/InMemorySecretStore.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        int nextVersion;
+        ConcurrentDictionary<int, string> versions;
+        do
+        {
+            versions = _secrets.GetOrAdd(name, _ => new ConcurrentDictionary<int, string>());
+            nextVersion = versions
+                .OrderByDescending(x => x.Key)
+                .FirstOrDefault(KeyValuePair.Create(0, string.Empty)).Key + 1;
+        } while (!versions.TryAdd(nextVersion, value)
```
wsugarman, line 58: I need to make sure someone didn't remove the key and that other callers will see any modifications

src/Microsoft.Health.Dicom.Core/Features/Common/ISecretStore.cs
```diff
@@ -3,16 +3,71 @@
...
+using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 
 namespace Microsoft.Health.Dicom.Core.Features.Common;
 
+/// <summary>
+/// Represents a data store that securely stores named secrets.
+/// </summary>
 public interface ISecretStore
```
pengchen0692, line 16: Where `ISecretStore` will be used? I guess it's part of Export, and the code using it will be checked in later?

src/Microsoft.Health.Dicom.Core/Features/Common/ISensitive.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Health.Dicom.Core.Features.Common;
+
+/// <summary>
+/// Represents something that may have sensitive information that has been secured
+/// by an instance of the <see cref="ISecretStore"/> class.
+/// </summary>
+public interface ISensitive
```
pengchen0692, line 16: Implementation will be checked in later?

src/Microsoft.Health.Dicom.Web/appsettings.json
```diff
@@ -101,6 +101,9 @@
       "ContainerName": "workitemcontainer"
     }
   },
+  "KeyVault": {
+    "Enabled":  false
```
pengchen0692, line 105: Where `Endpoint` will come from?

src/Microsoft.Health.Dicom.Core/Features/Common/ISecretStore.cs
wsugarman: The options for the blob store sink will be "sensitive," as it may containing a SasToken. So when validating the options, the sink provider will use the secret store to "classify" (ie. store) the sensitive info in the secret store.

pengchen0692 (dismissed):

src/Microsoft.Health.Dicom.Web/appsettings.json
wsugarman: This is found via reflection inside of the Azure libs. I think it will look for a property called `vaultUri` in the config

pengchen0692 (approved):

src/Microsoft.Health.Dicom.Web/appsettings.json
smithago: Will this won't work in PaaS and OSS, because we set Endpoint property. You will have to change that also

wsugarman: But we don't use KeyVault today for DICOM directly, right? Like in OSS we use it to store secrets, but ultimately we use environment variables whose values are from key vault:
image

So we could simply add new properties for the key vault connection too. Similarly, we don't use key vault for the DICOM service itself today in PAAS (just indirectly)

wsugarman: Oh, I see it in DICOMcast: https://github.com/microsoft/dicom-server/blob/5cbc30ca3004f10b56a6d10c3d3cb044aade824d/converter/dicom-cast/src/Microsoft.Health.DicomCast.Hosting/Program.cs#L28

But I think the pattern will still work on its own -- it just uses a different API that should contradict the othe one. And I think it's probably best we align with the new SDK instead of our own setting names, especially seeing that the APIs used in DICOMcast are deprecated.

PR #1575, opened 2022-05-09
Remove Combine method

pengchen0692:
## Description
As watermarkRange list comes from database, there could be gap between, so Combine won't work

## Related issues
Addresses [AB#91432](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/91432)


src/Microsoft.Health.Dicom.Functions/Copy/CopyDurableFunction.Orchestration.cs
```diff
@@ -61,8 +61,8 @@ public partial class CopyDurableFunction
             logger.LogInformation("Completed copying the range {Range}. Continuing with new execution...", batchRange);
 
             WatermarkRange completed = input.Completed.HasValue
-                ? input.Completed.Value.Combine(batchRange)
-                : batchRange;
+                    ? new WatermarkRange(batchRange.Start, input.Completed.Value.End)
```
wsugarman, line 64: nit: this alignment is off

wsugarman (approved):

wsugarman (approved):

PR #1581, opened 2022-05-10
Create Identifier-Based Export Source

wsugarman:
## Description
- Create new export source: `IdentifierExportSource`
- Add partitioning to export API
- Refactor default partition key
- Refactor export abstractions
- Update `IInstanceStore` API to return `IReadOnlyList<T>`

## Related issues
Addresses [#AB91160](https://microsofthealth.visualstudio.com/Health/_workitems/edit/91160)
Addresses [#AB90232](https://microsofthealth.visualstudio.com/Health/_workitems/edit/90232)

## Testing
New tests


src/Microsoft.Health.Dicom.Core/Features/Export/IdentifierExportOptions.cs
```diff
@@ -0,0 +1,45 @@
...
+        else
+        {
+            try
+            {
+                foreach (string value in Values)
+                {
+                    if (value == null)
+                        throw new ValidationException(string.Format(CultureInfo.CurrentCulture, DicomCoreResource.InvalidDicomIdentifier, value));
+
+                    DicomIdentifier.Parse(value);
```
smithago, line 34: nit: generaly not a fan of catching and wrapping exception. They should be used to bubble up problems. Does that mean we need a version of this that returns parse status with a message?

src/Microsoft.Health.Dicom.Core/Features/Export/IdentifierExportSource.cs
```diff
@@ -0,0 +1,120 @@
...
+
+            // Attempt to read the data
+            IReadOnlyList<VersionedInstanceIdentifier> instances = identifier.Type switch
+            {
+                ResourceType.Study => await _instanceStore.GetInstanceIdentifiersInStudyAsync(_partition.PartitionKey, identifier.StudyInstanceUid, cancellationToken),
+                ResourceType.Series => await _instanceStore.GetInstanceIdentifiersInSeriesAsync(_partition.PartitionKey, identifier.StudyInstanceUid, identifier.SeriesInstanceUid, cancellationToken),
+                _ => await _instanceStore.GetInstanceIdentifierAsync(_partition.PartitionKey, identifier.StudyInstanceUid, identifier.SeriesInstanceUid, identifier.SopInstanceUid, cancellationToken),
+            };
+
+            if (instances.Count > 0)
```
smithago, line 64: nit: generally like the flow like, instead of if and else
if (invalid case)
{
    throw;
}
valid case

src/Microsoft.Health.Dicom.Core/Features/Export/IdentifierExportSource.cs
```diff
@@ -0,0 +1,120 @@
...
+            ? Math.Min(maxSize.GetValueOrDefault(), _options.Values.Count - _startIndex)
+            : _options.Values.Count - _startIndex;
+
+        if (count == 0)
+            return null;
+
+        // Create a configuration that describes the source over this subset
+        var source = new TypedConfiguration<ExportSourceType>
+        {
+            Configuration = new ConfigurationBuilder().AddInMemoryCollection().Build(),
```
smithago, line 107: For my understanding, What is this doing? Feels too much for every batch get.

src/Microsoft.Health.Dicom.Core/Features/Export/IdentifierExportSourceProvider.cs
```diff
@@ -0,0 +1,55 @@
...
+        config.Bind(options);
+
+        return Task.FromResult<IExportSource>(
+            new IdentifierExportSource(
+                provider.GetRequiredService<IInstanceStore>(),
+                partition,
+                Options.Create(options)));
+    }
+
+    public Task<IConfiguration> ValidateAsync(IConfiguration config, CancellationToken cancellationToken = default)
```
smithago, line 42: nit: ValidateSourceAsync or CreateAsync to stay consistent on naming

src/Microsoft.Health.Dicom.Core/Features/Export/IdentifierExportOptions.cs
wsugarman: I tried to look at `Parse` in isolation and thought that a failed parse would return `FormatException` like `DateTime.Parse`. And then when we call it here, we wrap that exception in our own `ValidationException` used by this API. I think this is generally a good idea, but here in this case I could easily write a `TryParse` method instead and use its result to throw

src/Microsoft.Health.Dicom.Core/Features/Export/ExportHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+
+    public ExportHandler(IAuthorizationService<DataActions> authorizationService, IExportService exportService)
+        : base(authorizationService)
+        => _service = EnsureArg.IsNotNull(exportService, nameof(exportService));
+
+    public async Task<ExportResponse> Handle(ExportRequest request, CancellationToken cancellationToken = default)
+    {
+        EnsureArg.IsNotNull(request, nameof(request));
+
+        if (await AuthorizationService.CheckAccess(DataActions.Read, cancellationToken) != DataActions.Read)
```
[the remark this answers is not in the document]
wsugarman: I actually had an `Export` action previously. I'll add it back as I was going back-and-forth.

src/Microsoft.Health.Dicom.Core/Features/Export/IdentifierExportSource.cs
```diff
@@ -0,0 +1,120 @@
...
+            if (instances.Count > 0)
+            {
+                foreach (VersionedInstanceIdentifier read in instances)
+                    yield return ReadResult.ForIdentifier(read);
+            }
+            else
+            {
+                var args = new ReadFailureEventArgs(
+                    identifier,
+                    new FileNotFoundException(
```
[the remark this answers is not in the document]
wsugarman: Yeah, that makes a lot of sense

src/Microsoft.Health.Dicom.Core/Features/Export/IdentifierExportSource.cs
wsugarman: Yeah, it might be a bit much. If I recall the type names correctly, I think it's creating a `InMemoryConfigurationProvider` which has an `InMemoryConfigurationSource` which ultimately uses a `Dictionary<string, string> for storing the data lol.

I think I can avoid some of this overhead by changing the API

src/Microsoft.Health.Dicom.Core/Features/Export/IdentifierExportSourceProvider.cs
wsugarman: Oh, good catch! I changed this at the end, but I think `ValidateAsync` and `CreateAsync` are more succinct

src/Microsoft.Health.Dicom.Core/Features/Export/IdentifierExportSource.cs
wsugarman: Here we actual preserve the error to send to the error log, but I can put the error first as that is more typical

src/Microsoft.Health.Dicom.Api/Controllers/ExportController.cs
```diff
@@ -37,24 +38,28 @@ namespace Microsoft.Health.Dicom.Api.Controllers;
...
     /// <param name="mediator">An <see cref="IMediator"/> used to send requests.</param>
+    /// <param name="context">Dicom-specific request context.</param>
     /// <param name="options">Options concerning which features are enabled.</param>
     /// <param name="logger">A diagnostic logger.</param>
     /// <exception cref="ArgumentNullException">
     /// <paramref name="mediator"/>, <paramref name="options"/>, or <paramref name="logger"/> is <see langword="null"/>.
     /// </exception>
     public ExportController(
         IMediator mediator,
+        IDicomRequestContext context,
```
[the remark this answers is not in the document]
wsugarman: Yeah, you're right. I was hoping to bring up the context as I am a bit hesitant about its current usage pattern. However, I think I brought it up too high, and it seems natural a DICOM service/client would be scoped to a partition, so I have pushed this back down.

smithago (approved):

PR #1582, opened 2022-05-11
Enforce API Version

alishahahmed:
## Description
- Coding related changes have been made to enforce versioning.
- Changes to documentation have also been made.
- No changes to the postman collection are required as they already make use of the versioning.
- Same goes for electron tool.

## Related issues
[AB#91611](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/91611).
[AB#91673](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/91673)
[AB#91674](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/91674)
[AB#91675](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/91675)


src/Microsoft.Health.Dicom.Api/Features/Filters/PopulateDataPartitionFilterAttribute.cs
```diff
@@ -30,11 +30,8 @@ public sealed class PopulateDataPartitionFilterAttribute : ActionFilterAttribute
     private readonly HashSet<string> _partitionCreationSupportedRouteNames = new HashSet<string>
     {
         KnownRouteNames.PartitionStoreInstance,
-        KnownRouteNames.VersionedPartitionStoreInstance,
         KnownRouteNames.PartitionStoreInstancesInStudy,
```
bcarthic, line 33: Shouldn't we keep the versioned one and remove the non-versioned route?

alishahahmed: These are just the route names. They shouldn't have an impact on the actual route (please correct me if I am wrong). Given that now all routes are supposed to be versioned, I didn't like the idea of calling every one of them Versioned.

bcarthic: As long as it is same in the controller it should work fine. If you want to remove version, do you want to rename the VersionedRoute to Route and VersionedPartitionedRoute to PartitionedRoute

jovinson-ms: Here's my suggestion - we can use the convention approach to create a convention that will add the API version template at the beginning of all routes at the controller level. Then we can simply convert all `VersionedRoute` attributes to `Route`, and `VersionedPartitionedRoute` to `PartitionedRoute`.

This blog post shows how to do it: https://tpodolak.com/blog/2017/05/04/asp-net-core-api-versioning-convention, although they also do implicit API versioning which I don't think we want to do.

jovinson-ms (changes requested):
My main comment is that we should register a new convention for API version - replied to the existing comment thread.

We also need to update the failing E2E test to specify the version, and we should remove [this line](https://github.com/microsoft/dicom-server/blob/da745e14141f0cd2cc17897fc324285c95aaf880/src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs#L117).

src/Microsoft.Health.Dicom.Api/Features/Filters/PopulateDataPartitionFilterAttribute.cs
alishahahmed: I have updated the code. Now we are using conventions.

jovinson-ms (changes requested):
I just realized that the `_schema` endpoints get affected by this convention. Unless there's a clean way to avoid applying the version template to those endpoints we should probably go back to the previous solution until we resolve what to do with `_schema`. Sorry for the confusion!

jovinson-ms (approved):

PR #1583, opened 2022-05-11
Handle InvalidSchemaVersionException

brandonpollett:
## Description
This PR changes how we handle `InvalidSchemaVersionException` in two different cases. Each one is a case where the service doesn't yet know what the current schema version is on startup and we should be resilient to this.

- `DeletedInstanceCleanupWorker` - If this is the case, we should ignore this exception and continue to process on our normal cadence
- `BackgroundServiceHealthCheck` - Return an unhealthy result since this exception would constitute that we're unhealthy.

## Related issues
Addresses [AB#90262](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/90262).

## Testing
Unit test added for cleanup worker.

### DIFF
--- a/src/Microsoft.Health.Dicom.Api.UnitTests/Features/BackgroundServices/DeletedInstanceCleanupWorkerTests.cs
+++ b/src/Microsoft.Health.Dicom.Api.UnitTests/Features/BackgroundServices/DeletedInstanceCleanupWorkerTests.cs
@@ -10,6 +10,7 @@ using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.Extensions.Options;
 using Microsoft.Health.Dicom.Api.Features.BackgroundServices;
 using Microsoft.Health.Dicom.Core.Configs;
+using Microsoft.Health.Dicom.Core.Exceptions;
 using Microsoft.Health.Dicom.Core.Features.Delete;
 using NSubstitute;
 using Xunit;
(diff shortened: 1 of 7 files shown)


src/Microsoft.Health.Dicom.Api.UnitTests/Features/BackgroundServices/DeletedInstanceCleanupWorkerTests.cs
```diff
@@ -65,4 +66,28 @@ public async Task GivenANumberOfDeletedEntriesAndBatchSize_WhenCallingExecute_Th
         await _deletedInstanceCleanupWorker.ExecuteAsync(_cancellationTokenSource.Token);
         await _deleteService.ReceivedWithAnyArgs(expectedDeleteCount).CleanupDeletedInstancesAsync();
     }
+
+    [Fact]
+    public async Task GivenANotReadyDataStore_WhenCallingExecute_ThenNothingShouldHappen()
+    {
+        int iterations = 3;
+        int count = 0;
+        (bool, int) GenerateCleanupDeletedInstancesAsyncResponse()
```
wsugarman, line 75: nit: I think local functions are typically declared at the bottom

src/Microsoft.Health.Dicom.SqlServer/Exceptions/InvalidSchemaVersionException.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace Microsoft.Health.Dicom.SqlServer.Exceptions;
 
-internal class InvalidSchemaVersionException : DicomServerException
+internal class InvalidSchemaVersionException : DataStoreNotReadyException
```
wsugarman, line 10: There are really two cases this happens:
(1) we hecked up and there isn't a data store that supports this version of the schema OR
(2) the DB hasn't been initialized

Should we change the schema version cache to throw `InvalidSchemaVersionException` (or perhaps directly throw a `DataStoreNotReadyException` in case (2) and a regular ol' `KeyNotFoundException` for case (1)?

wsugarman (approved):

PR #1616, opened 2022-05-17
Fix versioning to allow _schema endpoint

bcarthic:
## Description
Fix versioning to allow _schema endpoint

## Related issues
Addresses [issue #].

## Testing

### DIFF
--- a/src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
+++ b/src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
@@ -113,6 +113,7 @@ public static class DicomServerServiceCollectionExtensions
         services.AddApiVersioning(c =>
         {
             c.ApiVersionReader = new UrlSegmentApiVersionReader();
+            c.AssumeDefaultVersionWhenUnspecified = true;
             c.ReportApiVersions = true;
             c.UseApiBehavior = false;
         });


src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
```diff
@@ -113,6 +113,7 @@ public static IDicomServerBuilder AddHostedServices(this IDicomServerBuilder ser
         services.AddApiVersioning(c =>
         {
             c.ApiVersionReader = new UrlSegmentApiVersionReader();
+            c.AssumeDefaultVersionWhenUnspecified = true;
```
abdullah248, line 116: What is the default version that it points to?

abdullah248: because this will affect all endpoints, not only _schema

jovinson-ms: @abdullah248 it's a confusing option and the author of the library has said so. Here's some discussion of it: https://github.com/dotnet/aspnet-api-versioning/issues/351#issuecomment-425106940] - basically this option will apply a null api version selector to the route of the _schema controller, but will not affect controllers that include the version template. Here's an example running with the change: 

```
info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/2 GET https://localhost:63838/studies - -
dbug: Microsoft.AspNetCore.Builder.RouterMiddleware[1]
      Request did not match any routes
dbug: Microsoft.AspNetCore.Builder.RouterMiddleware[1]
      Request did not match any routes
info: Microsoft.AspNetCore.Hosting.Diagnostics[2]
      Request finished HTTP/2 GET https://localhost:63838/studies - - - 404 0 - 11.9198ms
info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/2 GET https://localhost:63838/v1/studies - -
dbug: Microsoft.AspNetCore.Routing.Tree.TreeRouter[1]
      Request successfully matched the route with name '(null)' and template 'v{version:apiVersion}/studies'
dbug: Microsoft.AspNetCore.Mvc.Versioning.ApiVersionActionSelector[2]
```

alishahahmed:
We should update developer's guide for API pointing this out.

PR #1622, opened 2022-05-19
Prevent Concurrent Re-Indexing

wsugarman:
## Description
- Add new API for searching for operations based on some criteria to the `IDicomOperationsClient`
- Update extended query tag API to return a `409` (Conflict) response if there is a re-index operation already running when a new set of tags is requested
    - This API returns the existing operation in the header and its ID in the message

## Related issues
N/A

## Testing
New tests


src/Microsoft.Health.Dicom.Core/Models/Operations/OperationQueryCondition.cs
```diff
@@ -0,0 +1,43 @@
...
+
+namespace Microsoft.Health.Dicom.Core.Models.Operations;
+
+/// <summary>
+/// Represents a set of search criteria when querying for potentially multiple operations.
+/// </summary>
+public sealed class OperationQueryCondition<T>
+{
+    /// <summary>
+    /// Gets the optional collection of operation types to include in the search result.s
```
bcarthic, line 18: nit: results

src/Microsoft.Health.Dicom.Functions.Client/DicomAzureFunctionsClient.cs
```diff
@@ -66,63 +67,71 @@ internal class DicomAzureFunctionsClient : IDicomOperationsClient
...
+        EnsureArg.IsNotNull(query, nameof(query));
 
-        if (state == null)
+        var operations = new HashSet<DicomOperation>(query.Operations ?? Array.Empty<DicomOperation>());
+        var result = new OrchestrationStatusQueryResult();
+        do
         {
-            return null;
-        }
+            result = await _durableClient.ListInstancesAsync(
```
bcarthic, line 98: Not sure about the performance of this query, if the operation status is not given, should we default to running status?

src/Microsoft.Health.Dicom.Core/Exceptions/OngoingReindexException.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Globalization;
+using EnsureThat;
+using Microsoft.Health.Operations;
+
+namespace Microsoft.Health.Dicom.Core.Exceptions;
+
+/// <summary>
+/// The exception that is thrown when a re-index request is submitted while one is already active.
+/// </summary>
+public sealed class ExistingReindexException : Exception
```
bcarthic, line 16: Class name not matching the file name

src/Microsoft.Health.Dicom.Api/Controllers/ExtendedQueryTagController.cs
```diff
@@ -65,10 +66,24 @@ public async Task<IActionResult> PostAsync([Required][FromBody] IReadOnlyCollect
         _logger.LogInformation("DICOM Web Add Extended Query Tag request received, with extendedQueryTags {ExtendedQueryTags}.", extendedQueryTags);
 
         EnsureFeatureIsEnabled();
```
bcarthic, line 68: just checking, do we still need this flag. This is On everywhere right?

wsugarman: Yeah, it is enable everywhere today. I'll create a separate PR to remove this flag

src/Microsoft.Health.Dicom.Core/Exceptions/OngoingReindexException.cs
wsugarman: Good catch!

src/Microsoft.Health.Dicom.Functions.Client/DicomAzureFunctionsClient.cs
wsugarman: Well, it's ultimately a query on the table in Azure Storage. We don't expose this API publicly, but it could be good to provide a more granular default

wsugarman: I'll leave it without a default for now, but if we ever expose this publicly we'll need to consider the defaults

bcarthic (dismissed):

bcarthic (approved):

PR #1627, opened 2022-05-21
Add E2E Test for Export

wsugarman:
## Description
- Add E2E test
- Initialize sink before submitting operation
- Delete secret after successful operation

## Related issues
Resolves [AB#90232](https://microsofthealth.visualstudio.com/Health/_workitems/edit/90232)

## Testing
New tests


wsugarman:
For the CI test, here is an example entry that is stored in key vault before deletion:
image

build/ci/variables.yml
```diff
@@ -4,6 +4,7 @@ variables:
   testFeaturesEnabledEnvironmentUrl: 'https://$(deploymentName)-featuresenabled.azurewebsites.net/'
   testApplicationScope: 'https://$(deploymentName).resoluteopensource.onmicrosoft.com/.default'
   testApplicationResource: 'https://$(deploymentName).resoluteopensource.onmicrosoft.com'
+  azureStorageAccountName: 'dcmcipermanpmlxszw4sayty'
```
smithago, line 7: Why is the account name hardcoded here?

src/Microsoft.Health.Dicom.Azure/KeyVaultSecretClientOptions.cs
```diff
@@ -11,5 +12,19 @@ internal sealed class KeyVaultSecretClientOptions : SecretClientOptions
 {
     public const string SectionName = "KeyVault";
 
-    public bool Enabled { get; set; }
+    // This value is based on the SecretClient's parameter name and is read automatically from the configuration
+    public Uri VaultUri { get; set; }
+
+    [Obsolete]
```
smithago, line 18: Do we need to mark it obsolete? Should we just add comments for a future PR to cleanup as needed?

smithago (approved):

src/Microsoft.Health.Dicom.Azure/KeyVaultSecretClientOptions.cs
wsugarman: I think I'll add a comment in addition to the `[Obsolete]` attribute. I like the attribute as it will prevent further dependency on it

build/ci/variables.yml
wsugarman: Here is a snippet of the pipeline resolving it now:
image

bcarthic (approved):

PR #1639, opened 2022-05-27
Refactor Hashing Prefix to use Hex Values

wsugarman:
## Description
- Convert watermarks to `int` and transform the resulting hash into a 3 character hex string
- Rename types to be consistent with interface
- Include work item names in file name interface
- Use pointers for faster hashing

## Related issues
N/A

## Testing
Pipeline and new tests


bcarthic:
These changes are awesome. But unfortunately it will break WADO.

PR #1674, opened 2022-06-09
Updated Conformance-as-Postman.postman_collection.json

v-pbartley:
Added a "POST AuthorizeGetToken" call for obtaining an access token

Renamed "POST Store-single-instance (red-triangle.dcm)"

Reverted single instance paths back to original

Restored 'raw' option for dicom files in single instance calls


mmitrik (approved):
Tried it out, new auth request worked great 😄

PR #1682, opened 2022-06-13
Remove remaining UIDs and partition names from logs

alishahahmed:
## Description
Remove remaining UIDs and partition names from logs which will not be redacted through PaaS solution.

## Related issues
AB#[92521](https://microsofthealth.visualstudio.com/Health/_workitems/edit/92521).

## Testing
All the tests are passing.

Query that can later be used to verify if this pull request solved the problem:
```
LinuxTraceTelemetry
| where TIMESTAMP > ago(3d)
|where cloud_RoleName == "dicom"
|where message !contains "404"
|where message !contains "403"
| distinct message
```

### DIFF
--- a/src/Microsoft.Health.Dicom.Api/Features/Filters/PopulateDataPartitionFilterAttribute.cs
+++ b/src/Microsoft.Health.Dicom.Api/Features/Filters/PopulateDataPartitionFilterAttribute.cs
@@ -80,7 +80,7 @@ public sealed class PopulateDataPartitionFilterAttribute : ActionFilterAttribute
             }
             else
             {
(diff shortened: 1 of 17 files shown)


mitchell-fox (approved):
Looks good to me!

pengchen0692 (commented):
A general question: why we choose removing these information instead of redact?

alishahahmed:
> A general question: why we choose removing these information instead of redact?

Good question. These strings are formatted through resource file because of which they are not treated as properties which can be then redacted in the PaaS solution. That's why had to remove any sensitive information that was passed through the resource file.

wsugarman (approved):

PR #1686, opened 2022-06-14
Add Export Documentation

wsugarman:
## Description
- Add new markdown file describing export API

## Related issues
N/A

## Testing
N/A

### DIFF
--- a/docs/how-to-guides/export-data.md
+++ b/docs/how-to-guides/export-data.md
@@ -1,0 +1,119 @@
+# Export DICOM Files
+
+The Medical Imaging Server for DICOM supports the bulk export of data to an [Azure Blob Storage account](https://azure.microsoft.com/en-us/services/storage/blobs/). Before starting, be sure the feature is enabled in your configuration by setting `DicomServer:Features:EnableExtendedQueryTags` to `true`.
+
+The export API is available at `POST /export`. Given a *source*, the set of data to be exported, and a *destination*, the location to which data will be exported, the export endpoint returns a reference the long-running export operation. The duration of this operation depends on the volume of data to be exported.
+
+## Request
+
+The request body consists of the export source and destination.
+
+```json
+{
+    "source": {
+        "type": "<source type>",
+        "settings": {
+            // Source settings
+        }
+    },
+    "destination": {
+        "type": "<destination type>",
+        "settings": {
+            // Destination settings
+        }
+    }
+}
+```
+
+## Response
+
+Upon successfully starting the export operation, the export API returns a `202` response. The body of the response contains a reference to the newly start operation, while the value `Location` header is the URL of the export operation's status.
+
+```json
+{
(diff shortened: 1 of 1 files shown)


docs/how-to-guides/export-data.md
```diff
@@ -0,0 +1,124 @@
...
+
+**Type**: `"azureblob"`
+
+The Azure Blob storage destination can be used to export DICOM files to an existing Azure Blob container.
+
+#### Settings
+
+The connection to the Azure Blob storage account can be specified with either a `ConnectionString` and `BlobContainerName` or a `BlobContainerUri`. One of these settings is required!
+
+If the storage account requires authentication, a [SAS token](https://docs.microsoft.com/en-us/azure/storage/common/storage-sas-overview) can be included in either the `ConnectionString` or `BlobContaienrUri`. A managed identity can also be used to access the storage account with the `UseManagedIdentity` option. Note that the identity used by both the DICOM server and the functions must have access to the container.
```
mmitrik, line 101: > Note that the identity used by both the DICOM server and the functions must have access to the container.

What access do they need?  Storage Blob Data Reader?  Storage Blob Data Owner?  I ran into a similar problem in the OHIF viewer docs, so I think being specific can help the user a lot 😄

docs/how-to-guides/export-data.md
```diff
@@ -0,0 +1,124 @@
...
+
+The only setting is the list of identifiers to export.
+
+| Property | Required | Default | Description |
+| :------- | :------- | :------ | :---------- |
+| `Values` | Yes      |         | A list of one or more DICOM studies, series, and/or SOP instances identifiers in the format of `"<study instance UID>[/<series instance UID>[/<SOP instance UID>]]"` |
+
+#### Example
+
+```json
```
mmitrik, line 74: This example was very helpful ✨

docs/how-to-guides/export-data.md
```diff
@@ -0,0 +1,124 @@
...
+The error log can be found at `<export blob container uri>/<operation ID>/Errors.log`.
+
+### Example
+
+```json
+{
+    "type": "azureblob",
+    "settings": {
+        "blobContainerName": "export",
+        "connectionString": "UseDevelopmentStorage=true"
```
mmitrik, line 121: Is this correct?  It says "connectionString" but looks like a setting 🤔

mmitrik (approved):
Overall, pretty good.  My main feedback is that I think an example request up front would add a lot of value.  Also, the docs can probably be cleaned up by focusing on what the API does today, not what it might do in the future.

docs/how-to-guides/export-data.md
wsugarman: This is indeed correct lol. The connection strings for Azure Storage are a series key-value pairs. For the emulator, it just needs this value

mmitrik (approved):
Looks great 😄

PR #1725, opened 2022-06-27
Use blob DownloadContentAsync with Stream

bcarthic:
## Description
DICOM metadata file was saved using UTF-8 encoding with BOM, when using JsonSerializer, by design UTF-8 BOM should be ignored
Else, the following exception will occur `System.Text.Json.JsonReaderException: ''0xEF' is an invalid start of a value. LineNumber: 0 | BytePositionInLine: 0." at the line Reader.Read.`
So, reading the content into stream which will remove the UTF-8 BOM. https://github.com/dotnet/runtime/issues/29838

This PR related to #1720 
## Related issues
Addresses [issue AB#92911].

## Testing

### DIFF
--- a/src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobMetadataStore.cs
+++ b/src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobMetadataStore.cs
@@ -33,7 +33,6 @@ namespace Microsoft.Health.Dicom.Blob.Features.Storage;
 /// </summary>
 public class BlobMetadataStore : IMetadataStore
 {
-    private const string GetInstanceMetadataStreamTagName = nameof(BlobMetadataStore) + "." + nameof(GetInstanceMetadataAsync);
     private const string StoreInstanceMetadataStreamTagName = nameof(BlobMetadataStore) + "." + nameof(StoreInstanceMetadataAsync);
     private const string StoreInstanceFramesRangeTagName = nameof(BlobMetadataStore) + "." + nameof(StoreInstanceFramesRangeAsync);
     private readonly BlobContainerClient _container;
@@ -130,12 +129,14 @@ public class BlobMetadataStore : IMetadataStore
 
         return ExecuteAsync(async t =>
         {
-            await using (Stream stream = _recyclableMemoryStreamManager.GetStream(GetInstanceMetadataStreamTagName))
+            BlobDownloadResult result = await blobClient.DownloadContentAsync(t);
+
+            // DICOM metadata file was saved using UTF-8 encoding with BOM, When using JsonSerializer, by design UTF-8 BOM should be ignored
+            // Else, the following exception will occur "System.Text.Json.JsonReaderException: ''0xEF' is an invalid start of a value. LineNumber: 0 | BytePositionInLine: 0." at the line Reader.Read.
+            // So reading the content into stream which will remove the UTF-8 BOM. https://github.com/dotnet/runtime/issues/29838
+            using (Stream stream = _recyclableMemoryStreamManager.GetStream(result.Content))
             {
-                await blobClient.DownloadToAsync(stream, cancellationToken);
-
                 stream.Seek(0, SeekOrigin.Begin);
-
                 return await JsonSerializer.DeserializeAsync<DicomDataset>(stream, _jsonSerializerOptions, t);
             }
         }, cancellationToken);


src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobMetadataStore.cs
```diff
@@ -130,12 +129,14 @@ public Task<DicomDataset> GetInstanceMetadataAsync(VersionedInstanceIdentifier v
...
+            BlobDownloadResult result = await blobClient.DownloadContentAsync(t);
 
+            // DICOM metadata file was saved using UTF-8 encoding with BOM, When using JsonSerializer, by design UTF-8 BOM should be ignored
+            // Else, the following exception will occur "System.Text.Json.JsonReaderException: ''0xEF' is an invalid start of a value. LineNumber: 0 | BytePositionInLine: 0." at the line Reader.Read.
+            // So reading the content into stream which will remove the UTF-8 BOM. https://github.com/dotnet/runtime/issues/29838
+            using (Stream stream = _recyclableMemoryStreamManager.GetStream(result.Content))
+            {
                 stream.Seek(0, SeekOrigin.Begin);
-
                 return await JsonSerializer.DeserializeAsync<DicomDataset>(stream, _jsonSerializerOptions, t);
```
wsugarman, line 140: Shouldn't the `Stream`-based APIs automatically handle the BOM? The linked issue in dotnet/runtime seems to indicate this too

wsugarman: I just loaded a UTF8-BOM file locally, and can see the `JsonSerializer` method parses just fine:
```csharp
using Stream file = File.OpenRead(@"...");
JsonElement element = JsonSerializer.Deserialize<JsonElement>(file);
```

wsugarman: Oh, I see! You are changing it back. The reason it didn't work is indeed because of the BOM, but it is because of the API we used: [`ToObjectFromJson`](https://github.com/Azure/azure-sdk-for-net/blob/d46247724a5467a71269f2bbb80811f7c1838178/sdk/core/System.Memory.Data/src/BinaryData.cs#L257). Looking at the implementation, it attempts to deserialize the bytes directly which does not process the BOM unlike the `Stream` overload.

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobMetadataStore.cs
```diff
@@ -130,12 +129,14 @@ public Task<DicomDataset> GetInstanceMetadataAsync(VersionedInstanceIdentifier v
...
         {
-            await using (Stream stream = _recyclableMemoryStreamManager.GetStream(GetInstanceMetadataStreamTagName))
-            {
-                await blobClient.DownloadToAsync(stream, cancellationToken);
+            BlobDownloadResult result = await blobClient.DownloadContentAsync(t);
 
+            // DICOM metadata file was saved using UTF-8 encoding with BOM, When using JsonSerializer, by design UTF-8 BOM should be ignored
+            // Else, the following exception will occur "System.Text.Json.JsonReaderException: ''0xEF' is an invalid start of a value. LineNumber: 0 | BytePositionInLine: 0." at the line Reader.Read.
+            // So reading the content into stream which will remove the UTF-8 BOM. https://github.com/dotnet/runtime/issues/29838
+            using (Stream stream = _recyclableMemoryStreamManager.GetStream(result.Content))
```
wsugarman, line 137: We can skip the use of this `MemoryStream`. Instead just pass `result.ToStream()` to the `DeserializeAsync` method. No need to seek it back to the beginning either. This is not a network stream; it is the entire buffered file.

In some future when we don't need the serializer to seek, we can simply use `DownloadStreaming` instead

wsugarman (changes requested):

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobMetadataStore.cs
bcarthic: Yeah, it wasn't because of the API. It was deserializing

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobMetadataStore.cs
bcarthic: Yeah, the deserialization was the issue. Passing stream directly

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobMetadataStore.cs
```diff
@@ -130,14 +129,12 @@ public Task<DicomDataset> GetInstanceMetadataAsync(VersionedInstanceIdentifier v
...
-            await using (Stream stream = _recyclableMemoryStreamManager.GetStream(GetInstanceMetadataStreamTagName))
-            {
-                await blobClient.DownloadToAsync(stream, cancellationToken);
-
-                stream.Seek(0, SeekOrigin.Begin);
+            BlobDownloadResult result = await blobClient.DownloadContentAsync(t);
 
-                return await JsonSerializer.DeserializeAsync<DicomDataset>(stream, _jsonSerializerOptions, t);
-            }
+            // DICOM metadata file was saved using UTF-8 encoding with BOM, When using JsonSerializer, by design UTF-8 BOM should be ignored
```
wsugarman, line 134: I would just say that the DICOM metadata file includes the UTF-8 BOM and that there is a bug with the `BinaryData.ToObjectFromJson` method as seen in this issue: https://github.com/dotnet/runtime/issues/71447. The other linked issue doesn't seem like a real issue lol

bcarthic: Updated comment

wsugarman (approved):

PR #1729, opened 2022-06-28
Update Directory.Packages.props

brandonpollett:
## Description

## Related issues
Addresses [issue #].

## Testing


jovinson-ms (approved):

jovinson-ms:
@brandonpollett 👋

brandonpollett:
I think we may need to look at this a bit deeper. There was an internal conversation about if `tools\dicom-auth-test\Microsoft.Health.Web.Dicom.Tool\Microsoft.Health.Web.Dicom.Tool.sln` really belongs in this repo since it is really a sample application. If we wanted, in the meantime we could include a `tools` directory in the main solution and update the dependency to be a project reference to the client. I'll update this PR to be that at least so we can get out of the circular dependency update loop this currently causes.

PR #1730, opened 2022-06-28
Dicom Modules Extension

emilylo2:
## Description
Created extension for DicomModules

## Related issues
Part of User Story 92759 - [DICOM Events] Local Implementation

PR #1731, opened 2022-06-29
Refactored Dicom Modules

emilylo2:
## Description
Refactored Dicom Modules

## Related issues
Resolved comment from Dicom Events Library PR


converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/DicomWeb/Service/ChangeFeedRetrieveService.cs
```diff
@@ -39,11 +39,6 @@ public async Task<IReadOnlyList<ChangeFeedEntry>> RetrieveChangeFeedAsync(long o
 
         ChangeFeedEntry[] changeFeedEntries = await result.ToArrayAsync(cancellationToken);
 
-        if (changeFeedEntries?.Any() != null)
-        {
-            return changeFeedEntries;
-        }
-
-        return Array.Empty<ChangeFeedEntry>();
+        return await result.ToArrayAsync(cancellationToken) ?? Array.Empty<ChangeFeedEntry>();
```
wsugarman, line 42: My comment doesn't concern your change, but I have wondered if the client API shouldn't handle pagination on the users behalf. When you're using the `ChangeFeedRetrieveService`, are you fetching paginated results until the count is 0? If so, perhaps this method (or better yet the underlying client) should return an `IAsyncEnumerable<T>` that automatically handles pagination

jnlycklama:
havent we already done these changes? In https://github.com/microsoft/dicom-server/pull/1730/files

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/DicomWeb/Service/ChangeFeedRetrieveService.cs
jnlycklama: The way Dicom Cast currently uses this kind of relies on only getting one batch of changes at a time. We process those changes, send them to fhir, and then call ChangeFeedRetrieveService again to see if there's anything else. I'm not sure how getting a larger list of results, or having to wait for multiple pages, could affect the service.

We could change this to handle pagination in theory, and dicom cast might work fine, but I don't want to inflate the work of this too much so that we need to test out changes to dicom cast as well as functions. So I'd prefer to leave it how it is for now

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/DicomWeb/Service/ChangeFeedRetrieveService.cs
```diff
@@ -39,11 +39,6 @@ public async Task<IReadOnlyList<ChangeFeedEntry>> RetrieveChangeFeedAsync(long o
 
         ChangeFeedEntry[] changeFeedEntries = await result.ToArrayAsync(cancellationToken);
```
jnlycklama, line 40: looks like you don't need this line anymore since we aren't using changefeedentries

emilylo2: Removed unnecessary line

jnlycklama (approved):

PR #1739, opened 2022-06-30
Add Azure Function Health Checks

wsugarman:
## Description
- Add new Azure Function with an HTTP trigger for return a health check status
- Add new health check for Key Vault secrets
- Add health check to docker-compose for functions
- Fix SQL server docker-compose health check
- Integrate health check into PR pipeline

## Related issues
N/A

## Testing
PR Build


wsugarman:
TODO: More tests

PR #1746, opened 2022-07-02
Continue Updating Code Style

wsugarman:
## Description
- Use new [`.globalconfig`](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/configuration-files#global-analyzerconfig) file which can allow for per-project rules
- Enforce more code style changes. In particular, we'll begin enforcing:
    - Culture info for methods like `ToString()` and `Format`
        - Generally, errors use current culture while culture-agnostic strings use invariant
    - Checks on `IDisposable` ownership
    - Consistent parameter names in interface implementations
- We'll also encourage the use of brace-less if-statements if they're only 1 line (probably the most controversial)
- Fix `Directory.Packages.props`
- Minor refactors
- Update DICOM server and function dockerfiles to enforce en_US locale

## Related issues
N/A

## Testing
Test pipeline


converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/FhirTransaction/Observation/ObservationParser.cs
```diff
@@ -6,59 +6,48 @@
...
 
-        var observations = new List<Observation>();
-        ParseDicomDataset(dataset, patientReference, imagingStudyReference, observations, identifier);
-
-        return observations;
+        return ParseDicomDataset(dataset, patientReference, imagingStudyReference, identifier).ToList();
     }
 
-    private void ParseDicomDataset(DicomDataset dataset, ResourceReference patientReference, ResourceReference imagingStudyReference, List<Observation> observations, Identifier identifier)
+    private static IEnumerable<Observation> ParseDicomDataset(DicomDataset dataset, ResourceReference patientReference, ResourceReference imagingStudyReference, Identifier identifier)
```
wsugarman, line 29: @brandonpollett / @KnicKnic - I did a more significant refactor here, as I tried to remove the exception-based flow control and avoid passing in the list to mutate it. An analyzer flagged the list parameter

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/FhirTransaction/MissingRequiredDicomTagException.cs
```diff
@@ -22,6 +23,6 @@ private static string FormatMessage(string dicomTagName)
     {
         EnsureArg.IsNotNullOrWhiteSpace(dicomTagName, nameof(dicomTagName));
 
-        return string.Format(DicomCastCoreResource.MissingRequiredDicomTag, dicomTagName);
+        return string.Format(CultureInfo.CurrentCulture, DicomCastCoreResource.MissingRequiredDicomTag, dicomTagName);
```
jovinson-ms, line 26: I don't think we should localize these exception messages: the [Microsoft Rest API guidelines](https://github.com/Microsoft/api-guidelines/blob/vNext/Guidelines.md#7102-error-condition-responses) say "Services SHOULD NOT localize "message" for the end user, because doing so might make the value unreadable to the app developer who may be logging the value, as well as make the value less searchable on the Internet."

wsugarman: That makes sense. Although our usage of a resource file is a bit misleading, as presumably we'll never add other languages. Ideally we'd always set the culture to invariant with `DOTNET_SYSTEM_GLOBALIZATION_INVARIANT` but we have to keep it disabled (`false`) for the SQL Client apparently. So instead I suppose we can just ensure we don't add any other languages to our resources and use `InvariantCulture` for formatting?

wsugarman: Alternatively, we could keep the `CurrentCulture` and `.resx` for our exceptions, but explicitly set the local in our `dockerfile` like:
```dockerfile
ENV LC_ALL=en_US.UTF-8
ENV LANG=en_US.UTF-8
```

jovinson-ms: I'd argue that we should not be applying localization anywhere - logs, error messages, responses. The Dicom standard specifies literal English strings in many cases, and all of our strings are for internal use (wouldn't be directly exposed to an end user).

wsugarman: We can certainly do that for our code base, but our dependencies may include different localization. I think we even depend on certain error messages in a few cases! So I think it's best to enforce a certain locale within our application such that all dependencies are aligned with our own assemblies.

test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/RetrieveTransactionResourceTests.Frame.cs
```diff
@@ -63,18 +63,6 @@ await foreach (Stream item in response)
         }
     }
 
-    [Theory]
-    [MemberData(nameof(GetUnsupportedAcceptHeadersForFrames))]
-    public async Task GivenUnsupportedAcceptHeaders_WhenRetrieveFrame_ThenServerShouldReturnNotAcceptable(bool singlePart, string mediaType, string transferSyntax)
-    {
```
jovinson-ms: I think we need this

jovinson-ms (changes requested):

test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/RetrieveTransactionResourceTests.Frame.cs
wsugarman: Oops!

wsugarman: @jovinson-ms Actually, it looks like this was removed in main too: https://github.com/microsoft/dicom-server/commit/6d397cc4d20363e6bc26aea1f99f99360ee4e625

jovinson-ms (approved):

PR #1751, opened 2022-07-05
Update DICOM Schema Manager to run with empty DB

brandonpollett:
## Description
Previously there was a dependency on a copy of the DICOM service to be running when the Schema Manager was executing. This PR updates this to capture specific exceptions that are thrown when there is no instance running when the schema manager is running. The assumption is that if there is no instance registered with the DB then it is safe to pursue the upgrade to the specified version.

## Related issues
Addresses [AB#93113](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/93113).

## Testing
Manual testing with local instances and also downstream testing.


jovinson-ms (dismissed):

mitchell-fox (approved):
Looks good!

PR #1777, opened 2022-07-12
Delete blob orchestration

bcarthic:
## Description
This PR includes changes to delete blob after we switched to 'new' format.

## Related issues
Addresses [AB# 93319].

## Testing
Added unit tests and updated existing ones


src/Microsoft.Health.Dicom.Api/Features/BackgroundServices/StartBlobDeleteMigrationService.cs
```diff
@@ -0,0 +1,109 @@
...
+        IOptions<BlobMigrationConfiguration> blobMigrationFormatConfiguration,
+        ILogger<StartBlobDeleteMigrationService> logger)
+    {
+        EnsureArg.IsNotNull(serviceProvider, nameof(serviceProvider));
+        EnsureArg.IsNotNull(blobMigrationFormatConfiguration, nameof(blobMigrationFormatConfiguration));
+        EnsureArg.IsNotNull(logger, nameof(logger));
+
+        _serviceProvider = serviceProvider;
+        _blobMigrationFormatType = blobMigrationFormatConfiguration.Value.FormatType;
+        _startBlobDelete = blobMigrationFormatConfiguration.Value.StartCopy;
```
wsugarman, line 42: Perhaps we should just save the options as a member if we're saving 4 values from it?

src/Microsoft.Health.Dicom.Api/Features/BackgroundServices/StartBlobDeleteMigrationService.cs
```diff
@@ -0,0 +1,109 @@
...
+        _deleteOperationId = blobMigrationFormatConfiguration.Value.DeleteOperationId;
+        _copyOperationId = blobMigrationFormatConfiguration.Value.CopyOperationId;
+        _logger = logger;
+    }
+
+    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    {
+        try
+        {
+            using (var scope = _serviceProvider.CreateScope())
```
wsugarman, line 52: nit: change to a `using` declaration

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -183,6 +184,27 @@ public async Task CopyFileAsync(VersionedInstanceIdentifier versionedInstanceIde
...
+    {
+        EnsureArg.IsNotNull(versionedInstanceIdentifier, nameof(versionedInstanceIdentifier));
+
+        var blobClient = GetInstanceBlockBlobClient(versionedInstanceIdentifier, BlobMigrationFormatType.Old);
+        var newBlobClient = GetInstanceBlockBlobClient(versionedInstanceIdentifier, BlobMigrationFormatType.New);
+
+        // Delete the old file only if the new file exists
+        if (await newBlobClient.ExistsAsync(cancellationToken))
+        {
+            await ExecuteAsync(() => blobClient.DeleteIfExistsAsync(DeleteSnapshotsOption.IncludeSnapshots, conditions: null, cancellationToken));
```
wsugarman, line 200: Is it necessary to call `ExistsAsync` if we're going to call `DeleteIfExistsAsync`? Presumably `DeleteIfExistsAsync` returns a result based on whether the blob exists, and we could use that for the log below?

src/Microsoft.Health.Dicom.Core/Features/Common/IFileStore.cs
```diff
@@ -70,4 +70,14 @@ public interface IFileStore
...
         FrameRange range,
         CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Asynchronously deletes a old DICOM file from the file store if the file exists.
+    /// </summary>
+    /// <param name="versionedInstanceIdentifier">The DICOM identifier.</param>
+    /// <param name="cancellationToken">The cancellation token.</param>
+    /// <returns>A task that represents the asynchronous delete operation.</returns>
+    Task DeleteOldFileIfExistsAsync(
```
wsugarman, line 80: Don't we have some class that represents the format of the remote data? If that's the case, should we not just have `DeleteIfExists` methods that optionally accept these formatters?

wsugarman, line 80: Should this method return a flag indicate whether it was deleted?

src/Microsoft.Health.Dicom.Core/Features/Common/IFileStore.cs
```diff
@@ -70,4 +70,14 @@ public interface IFileStore
...
+        VersionedInstanceIdentifier versionedInstanceIdentifier,
+        CancellationToken cancellationToken);
```
wsugarman, line 82: nit: there should be a default value for the token

src/Microsoft.Health.Dicom.Core/Features/Common/IMetadataStore.cs
```diff
@@ -87,4 +87,14 @@ public interface IMetadataStore
...
         VersionedInstanceIdentifier versionedInstanceIdentifier,
         CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Asynchronously deletes a old DICOM instance metadata.
+    /// </summary>
+    /// <param name="versionedInstanceIdentifier">The DICOM instance identifier.</param>
+    /// <param name="cancellationToken">The cancellation token.</param>
+    /// <returns>A task that represents the asynchronous delete operation.</returns>
+    Task DeleteOldInstanceMetadataIfExistsAsync
```
wsugarman, line 97: Same comments as the `IFileStore`

src/Microsoft.Health.Dicom.Core/Features/Operations/IDicomOperationsClient.cs
```diff
@@ -111,4 +111,17 @@ public interface IDicomOperationsClient
...
+
+    /// <summary>
+    /// Asynchronously begins the old blob delete.
+    /// </summary>
+    /// <param name="operationId">The desired ID for the delete operation.</param>
+    /// <param name="previousCheckpoint">Optional checkpoint from a previous execution.</param>
+    /// <param name="cancellationToken">The token to monitor for cancellation requests. The default value is <see cref="CancellationToken.None"/>.</param>
+    /// <returns>
+    /// A task representing the <see cref="StartBlobDeleteAsync"/> operation.
+    /// The value of its <see cref="Task{TResult}.Result"/> that is performing the asynchronous copy.
```
wsugarman, line 123: I don't think this sentence makes sense lol. Do we need it?

src/Microsoft.Health.Dicom.Core/Models/Operations/DicomOperation.cs
```diff
@@ -29,4 +29,9 @@ public enum DicomOperation
     /// Specifies an copy operation that copies DICOM instance and metadata file in the same container.
     /// </summary>
     Copy,
+
+    /// <summary>
+    /// Specifies an delete operation that deletes DICOM instance and metadata file in the same container.
+    /// </summary>
+    Delete,
```
wsugarman, line 36: nit: I suspect we'll have an async delete at some point. Perhaps this is more like `MigrationDeletion` or `MigrationEnd`?

src/Microsoft.Health.Dicom.Functions/BlobMigration/CopyDurableFunction.Orchestration.cs
```diff
@@ -33,7 +33,7 @@ public partial class CopyDurableFunction
     {
         // The ID should be a GUID as generated by the trigger, but we'll assert here just to make sure!
         EnsureArg.IsNotNull(context, nameof(context)).ThrowIfInvalidOperationId();
-        CopyCheckpoint input = context.GetInput<CopyCheckpoint>();
+        BlobMigrationCheckpoint input = context.GetInput<BlobMigrationCheckpoint>();
```
wsugarman, line 36: Just as a reminder: this is only valid because we explicitly do not serialize our JSON input with type information (i.e. `"$type": "foo"`). If we did that, which is the default behavior, this change would not be backwards compatible and would require a new function name

src/Microsoft.Health.Dicom.Functions/BlobMigration/DeleteDurableFunction.Orchestration.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        if (batches.Count > 0)
+        {
+            // Note that batches are in reverse order because we start from the highest watermark
+            // Example: [8,11][4,7][1,3]
+            var batchRange = new WatermarkRange(batches[^1].Start, batches[0].End);
+
+            logger.LogInformation("Beginning to delete the range {Range}.", batchRange);
+            await Task.WhenAll(batches
+                .Select(x => context.CallActivityWithRetryAsync(
```
wsugarman, line 55: nit: this can probably all go on one line

src/Microsoft.Health.Dicom.Functions/BlobMigration/DeleteDurableFunction.Orchestration.cs
```diff
@@ -0,0 +1,82 @@
...
+{
+    /// <summary>
+    /// Asynchronously copy DICOM instances.
+    /// It goes through DICOM instances in the past, copy to files as new blob format.
+    /// </summary>
+    /// <param name="context">The context for the orchestration instance.</param>
+    /// <param name="logger">A diagnostic logger.</param>
+    /// <returns>A task representing the <see cref="DeleteFilesAsync"/> operation.</returns>
+    [FunctionName(nameof(DeleteFilesAsync))]
+    public async Task DeleteFilesAsync(
```
wsugarman, line 30: Are the name for the orchestration and activity too generic? Should they indicate they are part of the the migration, as officially the name of the functions are like `DeleteFileAsync` and I suspect we'll have an async delete in the near-ish future.

src/Microsoft.Health.Dicom.Functions/BlobMigration/DeleteDurableFunction.cs
```diff
@@ -0,0 +1,32 @@
...
+
+using EnsureThat;
+using Microsoft.Extensions.Options;
+using Microsoft.Health.Dicom.Core.Features.BlobMigration;
+using Microsoft.Health.Dicom.Core.Features.Retrieve;
+
+namespace Microsoft.Health.Dicom.Functions.BlobMigration;
+
+/// <summary>
+/// Represents the Azure Durable Functions that copy existing DICOM instances to new format.
```
wsugarman, line 14: nit: this XML doc is incorrect

src/Microsoft.Health.Dicom.Functions/Microsoft.Health.Dicom.Functions.csproj
```diff
@@ -41,6 +41,9 @@
     <None Include="Indexing\ReindexDurableFunction.*.cs">
       <DependentUpon>ReindexDurableFunction.cs</DependentUpon>
     </None>
+    <None Include="BlobMigration\DeleteDurableFunction.*.cs">
```
wsugarman, line 44: nit: alphabetize

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
bcarthic: Yes, we need to make sure new file exists before deleting the old one.

wsugarman: Oh, I misread this

src/Microsoft.Health.Dicom.Core/Features/Common/IFileStore.cs
bcarthic: Done

src/Microsoft.Health.Dicom.Api/Features/BackgroundServices/StartBlobDeleteMigrationService.cs
bcarthic: Done

src/Microsoft.Health.Dicom.Core/Features/Common/IMetadataStore.cs
bcarthic: Done.

src/Microsoft.Health.Dicom.Core/Features/Operations/IDicomOperationsClient.cs
bcarthic: Not needed. Deleted it

src/Microsoft.Health.Dicom.Core/Models/Operations/DicomOperation.cs
bcarthic: Done

src/Microsoft.Health.Dicom.Functions/BlobMigration/DeleteDurableFunction.Orchestration.cs
bcarthic: Done

src/Microsoft.Health.Dicom.Functions/BlobMigration/DeleteDurableFunction.Orchestration.cs
bcarthic: Yeah make sense, Changing to MigrateDeleteFilesAsync

src/Microsoft.Health.Dicom.Core/Features/Common/IFileStore.cs
bcarthic: What do you mean by optionally accepting these formatters. The logic here is unique, it is not just deleting if it exists. It needs to ensure it has copied the new format before deleting the old one

bcarthic: Successful running of this task means it is deleted. Why do you think we need a flag?

wsugarman: Typically "if exists" methods return some indication of what actually happened. Did this thing actually exist and we deleted it, or was it skipped? For example, the Azure Storage SDKs include methods like this that have different results depending on whether the element was present. I think their pattern is typically like `Task<Response> DeleteIfExists(CancellationToken)` where the response object is `null` if the value didn't exist.

src/Microsoft.Health.Dicom.Api/Features/BackgroundServices/StartBlobDeleteMigrationService.cs
bcarthic: Done

src/Microsoft.Health.Dicom.Core/Features/Common/IFileStore.cs
bcarthic: We have a method `DeleteFileIfExistsAsync` which doesn't return any flag. Do you want two different implementations in the same file?

src/Microsoft.Health.Dicom.Api/Features/BackgroundServices/StartBlobDeleteMigrationService.cs
```diff
@@ -0,0 +1,109 @@
...
+            {
+                // Start the background service only when the flag is turned on and the format type is not new service.
```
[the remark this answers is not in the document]
bcarthic: Yes, corrected

src/Microsoft.Health.Dicom.Api/Features/BackgroundServices/StartBlobDeleteMigrationService.cs
```diff
@@ -0,0 +1,109 @@
...
+using Microsoft.Extensions.Options;
+using Microsoft.Health.Dicom.Core.Configs;
+using Microsoft.Health.Dicom.Core.Features.Operations;
+using Microsoft.Health.Dicom.Core.Models.BlobMigration;
+using Microsoft.Health.Dicom.Core.Models.Operations;
+using Microsoft.Health.Operations;
+
+namespace Microsoft.Health.Dicom.Api.Features.BackgroundServices;
+
+public class StartBlobDeleteMigrationService : BackgroundService
```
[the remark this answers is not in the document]
bcarthic: Added some

src/Microsoft.Health.Dicom.Core/Features/Common/IFileStore.cs
wsugarman: Ah that's unfortunate. I suppose for symmetry's sake we should keep as-is, but we should be more intentional going forward

src/Microsoft.Health.Dicom.Api/Features/BackgroundServices/StartBlobDeleteMigrationService.cs
```diff
@@ -0,0 +1,106 @@
...
+                if (existingInstance == null)
+                {
+                    _logger.LogDebug("No existing delete operation.");
+                }
+                else
+                {
+                    _logger.LogDebug("Existing delete operation is in status: '{Status}'", existingInstance.Status);
+                }
+
+                OperationCheckpointState<DicomOperation> copyOperation = await operationsClient.GetLastCheckpointAsync(_blobMigrationFormatConfiguration.CopyOperationId, stoppingToken);
```
wsugarman, line 68: Shouldn't this line move inside of the first `if` block?

src/Microsoft.Health.Dicom.Functions.Client/DicomAzureFunctionsClient.cs
```diff
@@ -251,8 +267,9 @@ public async Task StartBlobCopyAsync(Guid operationId, WatermarkRange? previousC
     private static IOperationCheckpoint ParseCheckpoint(DicomOperation type, DurableOrchestrationStatus status)
         => type switch
         {
-            DicomOperation.Copy => status.Input?.ToObject<CopyCheckpoint>() ?? new CopyCheckpoint(),
+            DicomOperation.Copy => status.Input?.ToObject<BlobMigrationCheckpoint>() ?? new BlobMigrationCheckpoint(),
```
wsugarman, line 270: nit: I think you can use `or`:
```csharp
DicomOperation.Copy or DicomOperation.MigrationDeletion => status.Input?.ToObject<BlobMigrationCheckpoint>() ?? new BlobMigrationCheckpoint(),
```

wsugarman (dismissed):

src/Microsoft.Health.Dicom.Api/Features/BackgroundServices/StartBlobDeleteMigrationService.cs
bcarthic: Yeah, I think it was changed after the merge conflict

wsugarman (approved):

PR #1794, opened 2022-07-19
Enable multiple authentication schemes

jnlycklama:
## Description
The goal of this change is to allow our internal managed identities to be used to authenticate to the dicom server. These managed identities will not match the audience/authority we are currently using for authentication, so this change will allow us to specify multiple schemes

## Related issues
Addresses [issue #].

## Testing
Generated nuget packages and tested with dicom server deployed in k8s


src/Microsoft.Health.Dicom.Api/Features/Audit/AuditLogger.cs
```diff
@@ -78,8 +78,8 @@ public class AuditLogger : IAuditLogger
             AuditMessageFormat,
             auditAction,
             AuditEventType,
-            _securityConfiguration.Authentication?.Audience,
-            _securityConfiguration.Authentication?.Authority,
+            _securityConfiguration.AuthenticationSchemes.FirstOrDefault()?.Audience,
```
jnlycklama, line 81: Still need to figure out exactly how to get the correct audience & authority here

jnlycklama:
FYI - I spoke to Joyce about potential changes needed to Audit logs because of this. She said there is no requirement she knows of to specify if a call is internal or external in the audit logs.
Confirmed with Nate that there were no specific changes made to FHIR audit logs for IOT call using an internal identity. He said they rely on the identity objectId being included in the logs, and if the customer ever asks they can clarify that it is our internal identity

src/Microsoft.Health.Dicom.Api/Modules/SecurityModule.cs
```diff
@@ -41,26 +47,60 @@ public void Load(IServiceCollection services)
...
+                    {
+                        // Find the first authentication header with a JWT Bearer token whose issuer
+                        // contains one of the scheme names and return the found scheme name.
+                        StringValues authHeaders = default(StringValues);
+
+                        if (!StringValues.IsNullOrEmpty(context.Request.Headers.Authorization))
+                        {
+                            authHeaders = context.Request.Headers.Authorization;
+                        }
+                        else if (!StringValues.IsNullOrEmpty(context.Request.Headers.WWWAuthenticate))
```
brandonpollett, line 74: I believe that `WWW-Authenticate` is a response header. I don't know of a case where we should do something with it on a request - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate

src/Microsoft.Health.Dicom.Core/Configs/AuthenticationConfiguration.cs
```diff
@@ -9,6 +9,8 @@ namespace Microsoft.Health.Dicom.Core.Configs;
 
 public class AuthenticationConfiguration
 {
+    public string Name { get; set; }
```
brandonpollett, line 12: Could we just make use of the `authority` in this case instead of specifying a `name`?

src/Microsoft.Health.Dicom.Api/Modules/SecurityModule.cs
```diff
@@ -25,11 +28,14 @@ namespace Microsoft.Health.Dicom.Api.Modules;
...
 {
     private readonly SecurityConfiguration _securityConfiguration;
+    private readonly Dictionary<string, string> _audienceToSchemeMapper;
+    private string _defaultSchemeName;
 
     public SecurityModule(DicomServerConfiguration dicomServerConfiguration)
     {
         EnsureArg.IsNotNull(dicomServerConfiguration, nameof(dicomServerConfiguration));
         _securityConfiguration = dicomServerConfiguration.Security;
+        _audienceToSchemeMapper = new Dictionary<string, string>();
```
brandonpollett, line 38: Would authority to schema mapper make more sense? In the case of AAD, you could have a multi-tenant application that represents your service in multiple tenants, but they would have different authorities.

brandonpollett: Also, I believe that there's on a single `iss` claim in a `jwt` so that may also simplify other logic below.

src/Microsoft.Health.Dicom.Core/Configs/AuthenticationConfiguration.cs
jnlycklama: Make sense. Name is only to identify which scheme should be used, so authority works here too

src/Microsoft.Health.Dicom.Api/Modules/SecurityModule.cs
jnlycklama: I was initially thinking of doing this but noticed that the authority and issuer were a bit different - https://login.microsoftonline.com/72f988bf-86f1-41af-91ab-2d7cd011db47/ vs https://sts.windows.net/72f988bf-86f1-41af-91ab-2d7cd011db47/
But I guess it could make sense to get the tenant Id out of these claims and match on that. I've updated to do that

src/Microsoft.Health.Dicom.Api/Modules/SecurityModule.cs
jnlycklama: The example I found online was also checking that header so I decided to mimic that - https://stackoverflow.com/questions/52952600/multiple-jwt-authorities-issuers-in-asp-net-core

But after a quick search it looks like you are right, its mainly a response header. I will remove this check

PR #1806, opened 2022-07-22
DowloadStreaming API for export instead of OpenReadAsync that buffers

smithago:
## Description
Testing perf improvements

## Related issues
Addresses [issue #].

## Testing


src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -142,6 +142,26 @@ public class BlobFileStore : IFileStore
...
+        EnsureArg.IsNotNull(versionedInstanceIdentifier, nameof(versionedInstanceIdentifier));
+
+        BlockBlobClient blobClient = GetInstanceBlockBlobClient(versionedInstanceIdentifier, _blobMigrationFormatType);
+
+        Stream stream = null;
+        var blobOpenReadOptions = new BlobOpenReadOptions(allowModifications: false);
+
+        await ExecuteAsync(async () =>
+        {
+            Response<BlobDownloadStreamingResult> result = await blobClient.DownloadStreamingAsync(range: default, conditions: null, rangeGetContentHash: false, cancellationToken);
```
wsugarman, line 158: I know we had a number of conversations during the hackathon surrounding the various blob APIs we could use during export.  Has something changed, or could we always simply stream the blob to the destination without buffering?

src/Microsoft.Health.Dicom.Core/Features/Common/IFileStore.cs
```diff
@@ -80,4 +80,14 @@ public interface IFileStore
...
         VersionedInstanceIdentifier versionedInstanceIdentifier,
         CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Asynchronously gets a streaming file from the file store.
+    /// </summary>
+    /// <param name="versionedInstanceIdentifier">The DICOM identifier.</param>
+    /// <param name="cancellationToken">The cancellation token.</param>
+    /// <returns>A task that represents the asynchronous get operation.</returns>
+    Task<Stream> GetStreamingFileAsync(
```
wsugarman, line 90: Should we double-down on the "file" aspect and make this a method like `OpenReadAsync`?

wsugarman (approved):

src/Microsoft.Health.Dicom.Core/Features/Common/IFileStore.cs
smithago: Probably. Right now kept it in the same pattern as the rest of the method names

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
smithago: Not yet, we need to follow up with the Blob team to see, if the new SDK is available

PR #1817, opened 2022-07-27
Adding a new prop to track the Application Insights package version

jovinson-ms:
…shared across multiple libraries. This version of Application Insights marks instrumentation key as obsolete - we will build connection strings with the instrumentation key in Dicom and DicomCast: https://github.com/microsoft/ApplicationInsights-dotnet/issues/2560

Tracked via [AB#93836](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/93836)

Tested by sending telemetry to Application Insights via the new configuration.

### DIFF
--- a/converter/dicom-cast/src/Microsoft.Health.DicomCast.Hosting/Program.cs
+++ b/converter/dicom-cast/src/Microsoft.Health.DicomCast.Hosting/Program.cs
@@ -7,6 +7,7 @@ using System;
 using Azure.Extensions.AspNetCore.Configuration.Secrets;
 using Azure.Identity;
 using Azure.Security.KeyVault.Secrets;
+using Microsoft.ApplicationInsights.WorkerService;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
@@ -53,8 +54,7 @@ public static class Program
     }
 
     /// <summary>
-    /// Adds ApplicationInsights for telemetry and logging. We need to migrate to Application Insights
-    /// connection strings: https://github.com/microsoft/ApplicationInsights-dotnet/issues/2560
+    /// Adds ApplicationInsights for telemetry and logging.
     /// </summary>
     private static void AddApplicationInsightsTelemetry(IServiceCollection services, IConfiguration configuration)
     {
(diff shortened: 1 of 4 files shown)


src/Microsoft.Health.Dicom.Web/Startup.cs
```diff
@@ -52,15 +52,18 @@ public virtual void Configure(IApplicationBuilder app)
...
+    /// connection strings: https://github.com/microsoft/ApplicationInsights-dotnet/issues/2560
     /// </summary>
     private void AddApplicationInsightsTelemetry(IServiceCollection services)
     {
         string instrumentationKey = Configuration["ApplicationInsights:InstrumentationKey"];
 
         if (!string.IsNullOrWhiteSpace(instrumentationKey))
         {
+#pragma warning disable CS0618 // Type or member is obsolete
             services.AddApplicationInsightsTelemetry(instrumentationKey);
```
brandonpollett, line 65: Instead of marking this with a pragma, can we just adopt the suggested change from https://github.com/microsoft/ApplicationInsights-dotnet/issues/2560?

```
var applicationInsightsServiceOptions = new ApplicationInsightsServiceOptions
{
    ConnectionString = $"InstrumentationKey={instrumentationKey}",
};
services.AddApplicationInsightsTelemetry(applicationInsightsServiceOptions);
```

converter/dicom-cast/src/Microsoft.Health.DicomCast.Hosting/Program.cs
```diff
@@ -53,16 +53,19 @@ public static void Main(string[] args)
...
+    /// Adds ApplicationInsights for telemetry and logging. We need to migrate to Application Insights
+    /// connection strings: https://github.com/microsoft/ApplicationInsights-dotnet/issues/2560
     /// </summary>
     private static void AddApplicationInsightsTelemetry(IServiceCollection services, IConfiguration configuration)
     {
         string instrumentationKey = configuration["ApplicationInsights:InstrumentationKey"];
 
         if (!string.IsNullOrWhiteSpace(instrumentationKey))
         {
+#pragma warning disable CS0618 // Type or member is obsolete
```
brandonpollett, line 65: Same here?

src/Microsoft.Health.Dicom.Web/Startup.cs
jovinson-ms: I only read through the issue and two of the links - should have read the third 😄 

I have updated the PR to build connection strings in both startups, and tested with AI to validate telemetry is flowing.

converter/dicom-cast/src/Microsoft.Health.DicomCast.Hosting/Program.cs
jovinson-ms: fixed

brandonpollett (approved):

PR #1834, opened 2022-08-01
Even More Coding Style Changes

wsugarman:
## Description
- Enforce new coding styles:
  - Enforce `static`
  - Discourage use of `catch` blocks that capture every `Exception`
      - Annotated existing usages

## Related issues
N/A

## Testing
Existing tests

### DIFF
--- a/converter/dicom-cast/src/Microsoft.Health.DicomCast.TableStorage/Features/Health/TableHealthCheck.cs
+++ b/converter/dicom-cast/src/Microsoft.Health.DicomCast.TableStorage/Features/Health/TableHealthCheck.cs
@@ -40,8 +40,7 @@ public class TableHealthCheck : IHealthCheck
         catch (Exception ex)
         {
             _logger.LogWarning(ex, "Failed to connect to the table data store.");
-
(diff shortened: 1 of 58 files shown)


converter/dicom-cast/src/Microsoft.Health.DicomCast.TableStorage/Features/Health/TableHealthCheck.cs
```diff
@@ -40,8 +40,7 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
         catch (Exception ex)
         {
             _logger.LogWarning(ex, "Failed to connect to the table data store.");
-
-            return HealthCheckResult.Unhealthy("Failed to connect to the table data store.");
+            throw;
```
wsugarman, line 43: No need to explicitly return unhealthy result. An exception will automatically generate a result

jnlycklama (approved):

jnlycklama (dismissed):

jnlycklama (approved):

PR #1842, opened 2022-08-03
Handle workitem request body deserialization as part of the request pipeline.

shivakumar-ms:
## Description
Removed passing request.body to handlers. And let the workitem datasets be deserialized as part of the request pipeline, and as a [FromBody] parameter in the controller/action-method.

## Related issues
Addresses [AB#93067].

## Testing
Updated Unit and Integration tests.

### DIFF
--- a/src/Microsoft.Health.Dicom.Api.UnitTests/Controllers/WorkitemController.AddTests.cs
+++ b/src/Microsoft.Health.Dicom.Api.UnitTests/Controllers/WorkitemController.AddTests.cs
@@ -1,4 +1,4 @@
-﻿// -------------------------------------------------------------------------------------------------
+// -------------------------------------------------------------------------------------------------
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
@@ -56,7 +56,7 @@ public sealed class WorkitemControllerAddTests
                 Arg.Is(_controller.HttpContext.RequestAborted))
             .Returns(new AddWorkitemResponse(WorkitemResponseStatus.Failure, new Uri("https://www.microsoft.com")));
 
-        ObjectResult result = await _controller.AddAsync() as ObjectResult;
+        ObjectResult result = await _controller.AddAsync(null) as ObjectResult;
 
         Assert.IsType<ObjectResult>(result);
         Assert.Equal(HttpStatusCode.BadRequest, (HttpStatusCode)result.StatusCode);
@@ -72,7 +72,7 @@ public sealed class WorkitemControllerAddTests
                 Arg.Is(_controller.HttpContext.RequestAborted))
             .Returns(new AddWorkitemResponse(WorkitemResponseStatus.Conflict, new Uri("https://www.microsoft.com")));
 
-        ObjectResult result = await _controller.AddAsync() as ObjectResult;
+        ObjectResult result = await _controller.AddAsync(null) as ObjectResult;
 
         Assert.IsType<ObjectResult>(result);
         Assert.Equal(HttpStatusCode.Conflict, (HttpStatusCode)result.StatusCode);
@@ -90,7 +90,7 @@ public sealed class WorkitemControllerAddTests
                 Arg.Is(_controller.HttpContext.RequestAborted))
             .Returns(new AddWorkitemResponse(WorkitemResponseStatus.Success, new Uri(url)));
 
-        ObjectResult result = await _controller.AddAsync() as ObjectResult;
+        ObjectResult result = await _controller.AddAsync(null) as ObjectResult;
 
         Assert.IsType<ObjectResult>(result);
         Assert.Equal(HttpStatusCode.Created, (HttpStatusCode)result.StatusCode);
--- a/src/Microsoft.Health.Dicom.Api/Controllers/WorkitemController.Add.cs
+++ b/src/Microsoft.Health.Dicom.Api/Controllers/WorkitemController.Add.cs
@@ -1,11 +1,13 @@
-﻿// -------------------------------------------------------------------------------------------------
+// -------------------------------------------------------------------------------------------------
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using System.Threading.Tasks;
+using FellowOakDicom;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Logging;
 using Microsoft.Health.Api.Features.Audit;
@@ -46,23 +48,23 @@ public partial class WorkitemController
     [VersionedPartitionRoute(KnownRoutes.AddWorkitemInstancesRoute, Name = KnownRouteNames.PartitionedAddWorkitemInstance)]
     [VersionedRoute(KnownRoutes.AddWorkitemInstancesRoute, Name = KnownRouteNames.AddWorkitemInstance)]
     [AuditEventType(AuditEventSubType.AddWorkitem)]
-    public async Task<IActionResult> AddAsync()
+    public async Task<IActionResult> AddAsync([FromBody] IReadOnlyCollection<DicomDataset> dicomDatasets)
     {
         // The Workitem UID is passed as the name of the first query parameter 
         string workitemUid = HttpContext.Request.Query.Keys.FirstOrDefault();
 
-        return await PostAddAsync(workitemUid);
+        return await PostAddAsync(workitemUid, dicomDatasets);
     }
 
-    private async Task<IActionResult> PostAddAsync(string workitemInstanceUid)
+    private async Task<IActionResult> PostAddAsync(string workitemInstanceUid, IReadOnlyCollection<DicomDataset> dicomDatasets)
     {
-        long fileSize = Request.ContentLength ?? 0;
-        _logger.LogInformation("DICOM Web Add Workitem Transaction request received, with Workitem instance UID {WorkitemInstanceUid}, and file size of {FileSize} bytes",
+        long numberOfDatasets = dicomDatasets?.Count ?? 0;
+        _logger.LogInformation("DICOM Web Add Workitem Transaction request received, with Workitem instance UID {WorkitemInstanceUid}, and {NumberOfDatasets} DICOM dataset",
             workitemInstanceUid ?? string.Empty,
-            fileSize);
+            numberOfDatasets);
 
         AddWorkitemResponse response = await _mediator.AddWorkitemAsync(
-            Request.Body,
+            dicomDatasets?.FirstOrDefault(),
             Request.ContentType,
             workitemInstanceUid,
             HttpContext.RequestAborted);
--- a/src/Microsoft.Health.Dicom.Api/Controllers/WorkitemController.Cancel.cs
+++ b/src/Microsoft.Health.Dicom.Api/Controllers/WorkitemController.Cancel.cs
@@ -1,10 +1,13 @@
-﻿// -------------------------------------------------------------------------------------------------
+// -------------------------------------------------------------------------------------------------
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System.Collections.Generic;
+using System.Linq;
 using System.Net;
 using System.Threading.Tasks;
+using FellowOakDicom;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Logging;
 using Microsoft.Health.Api.Features.Audit;
@@ -35,6 +38,7 @@ public partial class WorkitemController
     /// 
     /// </remarks>
     /// <param name="workitemInstanceUid">The workitem Uid</param>
+    /// <param name="dicomDatasets">The DICOM dataset payload in the body.</param>
     /// <returns>Returns a string status report.</returns>
     [HttpPost]
     [Produces(KnownContentTypes.ApplicationDicomJson)]
@@ -47,13 +51,13 @@ public partial class WorkitemController
     [VersionedPartitionRoute(KnownRoutes.CancelWorkitemInstancesRoute, Name = KnownRouteNames.PartitionedCancelWorkitemInstance)]
     [VersionedRoute(KnownRoutes.CancelWorkitemInstancesRoute, Name = KnownRouteNames.CancelWorkitemInstance)]
     [AuditEventType(AuditEventSubType.CancelWorkitem)]
-    public async Task<IActionResult> CancelAsync(string workitemInstanceUid)
+    public async Task<IActionResult> CancelAsync(string workitemInstanceUid, [FromBody] IReadOnlyCollection<DicomDataset> dicomDatasets)
     {
         _logger.LogInformation("DICOM Web Cancel Workitem Transaction request received, with Workitem instance UID {WorkitemInstanceUid}",
             workitemInstanceUid ?? string.Empty);
 
         var response = await _mediator.CancelWorkitemAsync(
-                Request.Body,
+                dicomDatasets?.FirstOrDefault(),
                 Request.ContentType,
                 workitemInstanceUid,
                 HttpContext.RequestAborted)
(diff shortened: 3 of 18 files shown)


src/Microsoft.Health.Dicom.Api/Controllers/WorkitemController.Add.cs
```diff
@@ -46,23 +48,23 @@ public partial class WorkitemController
...
-        return await PostAddAsync(workitemUid);
+        return await PostAddAsync(workitemUid, dicomDatasets);
     }
 
-    private async Task<IActionResult> PostAddAsync(string workitemInstanceUid)
+    private async Task<IActionResult> PostAddAsync(string workitemInstanceUid, IReadOnlyCollection<DicomDataset> dicomDatasets)
     {
-        long fileSize = Request.ContentLength ?? 0;
-        _logger.LogInformation("DICOM Web Add Workitem Transaction request received, with Workitem instance UID {WorkitemInstanceUid}, and file size of {FileSize} bytes",
+        long numberOfDatasets = dicomDatasets?.Count ?? 0;
```
wsugarman, line 61: We can log both the number and the size

src/Microsoft.Health.Dicom.Api/Controllers/WorkitemController.Add.cs
```diff
@@ -46,23 +48,23 @@ public partial class WorkitemController
     [VersionedPartitionRoute(KnownRoutes.AddWorkitemInstancesRoute, Name = KnownRouteNames.PartitionedAddWorkitemInstance)]
     [VersionedRoute(KnownRoutes.AddWorkitemInstancesRoute, Name = KnownRouteNames.AddWorkitemInstance)]
     [AuditEventType(AuditEventSubType.AddWorkitem)]
-    public async Task<IActionResult> AddAsync()
+    public async Task<IActionResult> AddAsync([FromBody] IReadOnlyCollection<DicomDataset> dicomDatasets)
```
wsugarman, line 51: We can add other annotations here, like `[Required]`, `[MinLength]`, and `[MaxLength]`. I think. Test 'em out. That way we don't have to validate further down the stack

src/Microsoft.Health.Dicom.Api/Controllers/WorkitemController.Add.cs
```diff
@@ -46,23 +48,23 @@ public partial class WorkitemController
...
     }
 
-    private async Task<IActionResult> PostAddAsync(string workitemInstanceUid)
+    private async Task<IActionResult> PostAddAsync(string workitemInstanceUid, IReadOnlyCollection<DicomDataset> dicomDatasets)
     {
-        long fileSize = Request.ContentLength ?? 0;
-        _logger.LogInformation("DICOM Web Add Workitem Transaction request received, with Workitem instance UID {WorkitemInstanceUid}, and file size of {FileSize} bytes",
+        long numberOfDatasets = dicomDatasets?.Count ?? 0;
+        _logger.LogInformation("DICOM Web Add Workitem Transaction request received, with Workitem instance UID {WorkitemInstanceUid}, and {NumberOfDatasets} DICOM dataset",
             workitemInstanceUid ?? string.Empty,
```
wsugarman, line 63: No need to null-coalesce here. `null` is valid in format strings

src/Microsoft.Health.Dicom.Core/Extensions/DicomMediatorExtensions.cs
```diff
@@ -281,13 +282,17 @@ public static class DicomMediatorExtensions
...
-        this IMediator mediator, Stream requestBody, string requestContentType, string workitemInstanceUid, string transactionUid = default, CancellationToken cancellationToken = default)
+        this IMediator mediator,
+        DicomDataset dicomDataset,
+        string requestContentType,
+        string workitemInstanceUid,
+        string transactionUid = default,
+        CancellationToken cancellationToken = default)
     {
         EnsureArg.IsNotNull(mediator, nameof(mediator));
-        EnsureArg.IsNotNull(requestBody, nameof(requestBody));
```
wsugarman, line 287: Don't we still care about `null`-checking?

src/Microsoft.Health.Dicom.Core/Features/Workitem/AddWorkitemRequestHandler.cs
```diff
@@ -47,12 +44,8 @@ public class AddWorkitemRequestHandler : BaseHandler, IRequestHandler<AddWorkite
 
         request.Validate();
 
-        var workitems = await _workitemSerializer
```
wsugarman, line 50: Do we still need the `IWorkitemSerializer` interface?

wsugarman (changes requested):

src/Microsoft.Health.Dicom.Api/Controllers/WorkitemController.Add.cs
shivakumar-ms: Required, Min/Max length are FromRoute specific. I have added FromBody behavior, I hope this is acceptable.

src/Microsoft.Health.Dicom.Api/Controllers/WorkitemController.Add.cs
shivakumar-ms: Done

src/Microsoft.Health.Dicom.Api/Controllers/WorkitemController.Add.cs
shivakumar-ms: Done

src/Microsoft.Health.Dicom.Core/Extensions/DicomMediatorExtensions.cs
shivakumar-ms: It is validated in the API pipeline, as well as in the Handler/Validator.

src/Microsoft.Health.Dicom.Core/Features/Workitem/AddWorkitemRequestHandler.cs
shivakumar-ms: Done

src/Microsoft.Health.Dicom.Api/Controllers/WorkitemController.Add.cs
```diff
@@ -46,23 +49,23 @@ public partial class WorkitemController
...
     }
 
-    private async Task<IActionResult> PostAddAsync(string workitemInstanceUid)
+    private async Task<IActionResult> PostAddAsync(string workitemInstanceUid, IReadOnlyCollection<DicomDataset> dicomDatasets)
     {
-        long fileSize = Request.ContentLength ?? 0;
-        _logger.LogInformation("DICOM Web Add Workitem Transaction request received, with Workitem instance UID {WorkitemInstanceUid}, and file size of {FileSize} bytes",
-            workitemInstanceUid ?? string.Empty,
-            fileSize);
+        long numberOfDatasets = dicomDatasets?.Count ?? 0;
```
wsugarman, line 62: The length is guaranteed to be 1 and there's no need to `null` check either, right? We are validating this now

src/Microsoft.Health.Dicom.Api/Controllers/WorkitemController.Add.cs
```diff
@@ -46,23 +49,23 @@ public partial class WorkitemController
...
+        _logger.LogInformation("DICOM Web Add Workitem Transaction request received, with Workitem instance UID {WorkitemInstanceUid}, and {NumberOfDatasets} DICOM dataset",
+            workitemInstanceUid,
+            numberOfDatasets);
 
         AddWorkitemResponse response = await _mediator.AddWorkitemAsync(
-            Request.Body,
+            dicomDatasets?.FirstOrDefault(),
```
wsugarman, line 68: This is now unnecessary here and across the controllers

src/Microsoft.Health.Dicom.Core/Features/Workitem/AddWorkitemRequestHandler.cs
wsugarman: I still don't see this removal

src/Microsoft.Health.Dicom.Api.UnitTests/Controllers/WorkitemController.AddTests.cs
```diff
@@ -72,7 +72,7 @@ public async Task GivenWorkitemInstanceUid_WhenItAlreadyExists_ThenReturnConflic
                 Arg.Is(_controller.HttpContext.RequestAborted))
             .Returns(new AddWorkitemResponse(WorkitemResponseStatus.Conflict, new Uri("https://www.microsoft.com")));
 
-        ObjectResult result = await _controller.AddAsync() as ObjectResult;
+        ObjectResult result = await _controller.AddAsync(null) as ObjectResult;
```
wsugarman, line 75: Given that we cannot actually pass `null` values into the controller, this should be a non-`null` value and in the other tests

wsugarman: However, we'll still need to `null` check that first value down the call stack (that we presumably already do)

src/Microsoft.Health.Dicom.Api/Controllers/WorkitemController.Add.cs
```diff
@@ -46,23 +49,23 @@ public partial class WorkitemController
...
     {
         // The Workitem UID is passed as the name of the first query parameter 
         string workitemUid = HttpContext.Request.Query.Keys.FirstOrDefault();
 
-        return await PostAddAsync(workitemUid);
+        return await PostAddAsync(workitemUid, dicomDatasets);
     }
 
-    private async Task<IActionResult> PostAddAsync(string workitemInstanceUid)
+    private async Task<IActionResult> PostAddAsync(string workitemInstanceUid, IReadOnlyCollection<DicomDataset> dicomDatasets)
```
wsugarman, line 60: Don't we need to validate this argument too?

src/Microsoft.Health.Dicom.Api/Controllers/WorkitemController.Update.cs
```diff
@@ -39,21 +42,23 @@ public partial class WorkitemController
...
     {
         // The Transaction UID is passed as the first query parameter 
         string transactionUid = HttpContext.Request.Query.Keys.FirstOrDefault();
 
-        return await PostUpdateAsync(workitemInstanceUid, transactionUid);
+        return await PostUpdateAsync(workitemInstanceUid, transactionUid, dicomDatasets);
     }
 
-    private async Task<IActionResult> PostUpdateAsync(string workitemInstanceUid, string transactionUid)
+    private async Task<IActionResult> PostUpdateAsync(string workitemInstanceUid, string transactionUid, IReadOnlyCollection<DicomDataset> dicomDatasets)
```
wsugarman, line 53: Validate argument?

src/Microsoft.Health.Dicom.Api/Controllers/WorkitemController.Add.cs
```diff
@@ -46,23 +49,23 @@ public partial class WorkitemController
     [VersionedPartitionRoute(KnownRoutes.AddWorkitemInstancesRoute, Name = KnownRouteNames.PartitionedAddWorkitemInstance)]
     [VersionedRoute(KnownRoutes.AddWorkitemInstancesRoute, Name = KnownRouteNames.AddWorkitemInstance)]
     [AuditEventType(AuditEventSubType.AddWorkitem)]
-    public async Task<IActionResult> AddAsync()
+    public async Task<IActionResult> AddAsync([FromBody][Required][MinLength(1)][MaxLength(1)] IReadOnlyList<DicomDataset> dicomDatasets)
```
wsugarman, line 52: I think the only thing we want to check before merging this is how we expose these errors as per the UPS-RS standard

src/Microsoft.Health.Dicom.Api/Controllers/WorkitemController.Add.cs
```diff
@@ -46,23 +49,23 @@ public partial class WorkitemController
...
     }
 
-    private async Task<IActionResult> PostAddAsync(string workitemInstanceUid)
+    private async Task<IActionResult> PostAddAsync(string workitemInstanceUid, IReadOnlyList<DicomDataset> dicomDatasets)
     {
-        long fileSize = Request.ContentLength ?? 0;
-        _logger.LogInformation("DICOM Web Add Workitem Transaction request received, with Workitem instance UID {WorkitemInstanceUid}, and file size of {FileSize} bytes",
-            workitemInstanceUid ?? string.Empty,
-            fileSize);
+        long numberOfDatasets = dicomDatasets.Count;
```
wsugarman, line 62: Isn't this always `1`? Do we need to log it?

wsugarman (commented):
I'm not seeing any deleted files. Shouldn't there be some?

src/Microsoft.Health.Dicom.Api/Controllers/WorkitemController.Add.cs
shivakumar-ms: QueryStringValidatorMiddleware kicks in and throws 400 error with message "The query parameter '' is invalid. A non-empty request body is required.".

image

jovinson-ms: Why does it talk about query parameter and also body?

shivakumar-ms:
@wsugarman based on the feedback from @jovinson-ms I have removed the max-length attribute for the datasets parameter. Also, we are going ahead with the Bad Request response/message from the middleware as is, for this iteration. @jovinson-ms please chime in.

src/Microsoft.Health.Dicom.Api/Controllers/WorkitemController.Add.cs
shivakumar-ms: It is due to [Required] and [MinLength] attributes.

build/common/versioning.yml
```diff
@@ -32,11 +32,11 @@ jobs:
...
 
-  - task: PowerShell@2
-    displayName: 'Check for breaking API / Swagger changes'
-    inputs:
-      pwsh: true
-      filepath: './build/common/scripts/CheckForBreakingAPISwaggerChanges.ps1'
-      arguments: >
-        -SwaggerDir 'swagger'
-        -Version 'v1-prerelease','v1'
+  # - task: PowerShell@2
```
wsugarman, line 35: We probably need a better solution for this going forward lol

src/Microsoft.Health.Dicom.Api/Controllers/WorkitemController.Add.cs
```diff
@@ -46,23 +49,20 @@ public partial class WorkitemController
...
     }
 
-    private async Task<IActionResult> PostAddAsync(string workitemInstanceUid)
+    private async Task<IActionResult> PostAddAsync(string workitemInstanceUid, IReadOnlyList<DicomDataset> dicomDatasets)
     {
-        long fileSize = Request.ContentLength ?? 0;
-        _logger.LogInformation("DICOM Web Add Workitem Transaction request received, with Workitem instance UID {WorkitemInstanceUid}, and file size of {FileSize} bytes",
-            workitemInstanceUid ?? string.Empty,
-            fileSize);
+        _logger.LogInformation("DICOM Web Add Workitem Transaction request received with file size of {FileSize} bytes.", Request.ContentLength);
```
wsugarman, line 62: Do we no longer want to log the instance UID?

src/Microsoft.Health.Dicom.Api/Controllers/WorkitemController.Cancel.cs
```diff
@@ -47,13 +51,12 @@ public partial class WorkitemController
     [VersionedPartitionRoute(KnownRoutes.CancelWorkitemInstancesRoute, Name = KnownRouteNames.PartitionedCancelWorkitemInstance)]
     [VersionedRoute(KnownRoutes.CancelWorkitemInstancesRoute, Name = KnownRouteNames.CancelWorkitemInstance)]
     [AuditEventType(AuditEventSubType.CancelWorkitem)]
-    public async Task<IActionResult> CancelAsync(string workitemInstanceUid)
+    public async Task<IActionResult> CancelAsync(string workitemInstanceUid, [FromBody][Required][MinLength(1)] IReadOnlyList<DicomDataset> dicomDatasets)
     {
-        _logger.LogInformation("DICOM Web Cancel Workitem Transaction request received, with Workitem instance UID {WorkitemInstanceUid}",
-            workitemInstanceUid ?? string.Empty);
+        _logger.LogInformation("DICOM Web Cancel Workitem Transaction request received with file size of {FileSize} bytes.", Request.ContentLength);
```
wsugarman, line 56: Same comment about the UID

wsugarman (approved):

PR #1870, opened 2022-08-15
Fix - Impacting SLA: System.ArgumentNullException

alishahahmed:
## Description
Fixed SLA impacting issue where customer's request without content-type header somehow made it to the service. This was resulting in a 5xx error because of the EnsureArg check in place.

I have removed this check and also updated the Handler and Reader tests such that we validate the possible input values for content-type header.

After this change, instead of 5xx status code, user should receive a 415 result code indicating that the media type is unsupported.

## Related issues
AB#94228.

## Testing
- Updated handler and receiver tests.
- All existing tests are passing.

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Features/Store/Entries/DicomInstanceEntryReaderManager.cs
+++ b/src/Microsoft.Health.Dicom.Core/Features/Store/Entries/DicomInstanceEntryReaderManager.cs
@@ -1,4 +1,4 @@
-﻿// -------------------------------------------------------------------------------------------------
+// -------------------------------------------------------------------------------------------------
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
@@ -26,8 +26,6 @@ public class DicomInstanceEntryReaderManager : IDicomInstanceEntryReaderManager
     /// <inheritdoc />
     public IDicomInstanceEntryReader FindReader(string contentType)
     {
-        EnsureArg.IsNotNull(contentType, nameof(contentType));
-
         return _dicomInstanceEntryReaders.FirstOrDefault(reader => reader.CanRead(contentType));
     }
 }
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.Dicom.Core/Features/Store/Entries/DicomInstanceEntryReaderManager.cs
```diff
@@ -26,8 +26,6 @@ public DicomInstanceEntryReaderManager(IEnumerable<IDicomInstanceEntryReader> di
     /// <inheritdoc />
     public IDicomInstanceEntryReader FindReader(string contentType)
     {
-        EnsureArg.IsNotNull(contentType, nameof(contentType));
```
jovinson-ms, line 29: We should probably keep this check in place while we implement the error handling.

jovinson-ms (commented):
Are we able to reproduce the conditions that would lead to this happening?

src/Microsoft.Health.Dicom.Core/Features/Store/Entries/DicomInstanceEntryReaderManager.cs
wsugarman, line 29: Is a `ContentType` required by the spec or is there some default value we can assume? If it's the latter, then we should populate the default.

alishahahmed: ContentType is required by spec. No default is specified for this. In addition to the standard, we have also provided a list of accepted ContentType values in the conformance statement.

alishahahmed:
> Are we able to reproduce the conditions that would lead to this happening?

No, we were not able to reproduce it. The controller has a Consumes attribute filter for the ContentType header. This should not have happened in the first place. I also synced with Smitha on this, and removing the EnsureArg check seemed to be the proper solution for this issue.

src/Microsoft.Health.Dicom.Core/Features/Store/Entries/DicomInstanceEntryReaderManager.cs
alishahahmed: We don't necessarily need this. Given that the reader returns null if `ContentType` is null, we are good. The caller is responsible for making sure that the reader is not null. In our case, we always check the reader and return `UnsupportedMediaType` status code if the reader is not found.
Please do let me know if I misunderstood the comment.

alishahahmed: I have also updated the `StoreHandler` tests to verify that we return `UnsupportedMediaType` exception when `ContentType` is null.

jovinson-ms (approved):

PR #1874, opened 2022-08-18
Flush metrics

poadhika:
## Description
Flush the metrics when appliation crashes

## Related issues
Addresses [issue #93947](https://microsofthealth.visualstudio.com/Health/_backlogs/backlog/Medical%20Imaging/Epics/?workitem=93947)

## Testing
Tested in test environment.

### DIFF
--- a/converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/DicomCastWorker.cs
+++ b/converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/DicomCastWorker.cs
@@ -132,5 +132,9 @@ public class DicomCastWorker : IDicomCastWorker
             // Please refer to .net core issue on github for more details: "Exceptions in BackgroundService ExecuteAsync are (sometimes) hidden" https://github.com/dotnet/extensions/issues/2363
             _hostApplicationLifetime.StopApplication();
         }
+        finally
+        {
+            _telemetryClient.Flush();
+        }
     }
 }


converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/DicomCastWorker.cs
```diff
@@ -132,5 +132,9 @@ public async Task ExecuteAsync(CancellationToken cancellationToken)
             // Please refer to .net core issue on github for more details: "Exceptions in BackgroundService ExecuteAsync are (sometimes) hidden" https://github.com/dotnet/extensions/issues/2363
             _hostApplicationLifetime.StopApplication();
         }
+        finally
+        {
+            _telemetryClient.Flush();
```
wsugarman, line 137: How does this rectify the auth failure? Presumably the host is crashing upon being misconfigured? Are we hoping logs are flushed? Metrics? Will the Telemetry Client affect logs, metrics, or both?

poadhika: Right now the approach is to identify and categorize different type of faiures. That's where the metrics will come handy. Different types of metrics will be emitted depending upon the reason it is crashing. If the failure is related to to rbac rules that client has misconfigured let the customer know and supress the alert. If the metrics indicate other reasons then go and investigate.

jnlycklama (approved):

PR #1897, opened 2022-08-29
Refactor DICOM Server Registrations

brandonpollett:
## Description
This PR refactors some of the DICOM Server Registrations to make it easier to integrate the core libraries and capabilities into a non REST-fronted service. While most of the changes should be transparent to the end-user, the registration of the web API has changed. There is now a registration extension `AddWebApi` which must be called after `AddDicomServer` to enable the REST API to the services.

**Previous**
```csharp
services.AddDicomServer(Configuration)
            .AddBlobDataStores(Configuration)
            .AddSqlServer(Configuration)
```

**New**
```csharp
services.AddDicomCore(Configuration)
            .AddWebApi(Configuration)
            .AddBlobDataStores(Configuration)
            .AddSqlServer(Configuration)
```

## Related issues
Addresses AB#94685

## Testing
Existing tests continue to pass and tested use without the `AddWebApi` manually downstream.

+semver: breaking


src/Microsoft.Health.Dicom.Core/Configs/CoreConfiguration.cs
```diff
@@ -3,24 +3,17 @@
...
-using Microsoft.Health.Api.Configuration;
-using Microsoft.Health.Api.Features.Cors;
 using Microsoft.Health.Core.Configs;
-using Microsoft.Health.Dicom.Core.Configs;
 
-namespace Microsoft.Health.Dicom.Api.Configs;
+namespace Microsoft.Health.Dicom.Core.Configs;
 
-public class DicomServerConfiguration : IApiConfiguration
+public class CoreConfiguration
```
brandonpollett, line 10: Totally open to another name here, landed on `CoreConfiguration` to map to `Microsoft.Health.Dicom.Core` much like `ApiConfiguration` now maps to `Microsoft.Health.Dicom.Api`

src/Microsoft.Health.Dicom.Core/Registration/DicomServerApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,66 @@
...
+using Microsoft.Health.Extensions.DependencyInjection;
+using Microsoft.IO;
+
+namespace Microsoft.Health.Dicom.Core.Registration;
+
+public static class DicomServerApplicationBuilderExtensions
+{
+    private const string DicomServerConfigurationSectionName = "DicomServer";
+
+    public static IDicomServerBuilder AddDicomServer(this IServiceCollection services,
```
brandonpollett, line 23: We could potentially rename this as something like `AddDicomCore` too if we wanted.

src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
```diff
@@ -67,96 +65,73 @@ public static IDicomServerBuilder AddHostedServices(this IDicomServerBuilder ser
...
     /// </summary>
-    /// <param name="services">The services collection.</param>
+    /// <param name="dicomServerBuilder">The services collection.</param>
     /// <param name="configurationRoot">An optional configuration root object. This method uses the "DicomServer" section.</param>
-    /// <param name="configureAction">An optional delegate to set <see cref="DicomServerConfiguration"/> properties after values have been loaded from configuration.</param>
+    /// <param name="configureAction">An optional delegate to set <see cref="ApiConfiguration"/> properties after values have been loaded from configuration.</param>
     /// <returns>A <see cref="IDicomServerBuilder"/> object.</returns>
-    public static IDicomServerBuilder AddDicomServer(
-        this IServiceCollection services,
+    public static IDicomServerBuilder AddWebApi(
```
wsugarman, line 74: I wonder if the builder should be closer to something like a state machine, where the API guides callers to add the various components necessary for the server to run, such as the metadata store, file store, etc?

src/Microsoft.Health.Dicom.Core/Configs/CoreConfiguration.cs
```diff
@@ -3,24 +3,17 @@
...
 {
+    public AuditConfiguration Audit { get; } = new AuditConfiguration("X-MS-AZUREDICOM-AUDIT-");
```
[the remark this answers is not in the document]
brandonpollett: I 100% agree. I think this is something that we'll need to address once we're ready to refactor the auditing.

src/Microsoft.Health.Dicom.Core/Registration/DicomServerApplicationBuilderExtensions.cs
jovinson-ms: That makes sense to me.

src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
brandonpollett: Hmm...  I could see where this would be useful. I think the hard part of that today is that we are targeting the implementations vs the types of stores. e.g. `AddSqlServer` and `AddBlobDataStores` vs `AddFileStore`.

jovinson-ms (approved):

PR #1908, opened 2022-09-01
Fix for null reference issue while dicomcast processes erred DELETE entries

arunmk-ms:
## Description
Pods were restarting while dicomcast processes change feed entries for Delete resources that has multiple resource matches in Fhir. The issue was due to the metadata field being blank in Delete changefeed entries. 

## Related issues
Addresses [issue #331589309](https://portal.microsofticm.com/imp/v3/incidents/details/331589309/home).

## Testing

- Run the fhir, dicom and dicomcast apps in local. 
- Delete an image in dicom using the DELETE study endpoint that has multiple entries in Fhir. 
- Make sure there are no null reference exceptions while dicomcast processes the change feed entry


abdullah248 (approved):
looks good

poadhika (approved):

arunmk-ms:
> Is it expected for metadata to be blank for change feed entries on delete?

Tested a couple of delete images and noticed that metadata is blank for all of them.

arunmk-ms:
> I would rename this PR as there are no pods in OSS. The hosting is an implementation detail

Done

PR #1929, opened 2022-09-09
[AB#93875] Forked fo-dicom JsonDicomConverter locally to fix deserialization issue

shivakumar-ms:
## Description
Forked JsonDicomConverter to fix Deserialization issue with DS, IS, UV, and SV VR types with non-numeric string values

## Related issues
Addresses [AB#93875].

## Testing
Unit Tests in Fo-Dicom repository


forks/.editorconfig
```diff
@@ -0,0 +1,18 @@
+# Analyzers
+[*.cs]
```
wsugarman, line 2: We've migrated to using `.globalconfig` for analyzer rules. Please use that instead with a higher `global_level` number like we do in healthcare-shared-components.

Eg.
- [root config](https://github.com/microsoft/healthcare-shared-components/blob/main/.globalconfig)
- [per-project config](https://github.com/microsoft/healthcare-shared-components/blob/main/src/Microsoft.Health.Operations.Functions/.globalconfig)

forks/Microsoft.Health.FellowOakDicom/Microsoft.Health.FellowOakDicom.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>library</OutputType>
+    <TargetFramework>net6.0</TargetFramework>
+    <Nullable>disable</Nullable>
+    <AnalysisLevel></AnalysisLevel>
```
wsugarman, line 7: What happens if you leave out `AnalysisLevel`, `AnalysisMode`, and `EnforceCodeStyleInBuild`? Shouldn't the rules specified in the .editorconfig/.globalconfig account for these?

forks/Microsoft.Health.FellowOakDicom/Serialization/Utf8JsonReaderExtensions.cs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2012-2021 fo-dicom contributors.
+// Licensed under the Microsoft Public License (MS-PL).
+
+using System.Text.Json;
+
+namespace Microsoft.Health.FellowOakDicom.Serialization
+{
+    internal static class Utf8JsonReaderExtensions
+    {
+        public static void Assume(this ref Utf8JsonReader reader, JsonTokenType tokenType)
```
wsugarman, line 10: Do you need this file? Is this from Fo.Dicom too?

shivakumar-ms: Yes. Otherwise the converter code doesn't compile. Not sure how this is not in the published fo-dicom 5.0.3 version.

forks/Microsoft.Health.FellowOakDicom/Microsoft.Health.FellowOakDicom.csproj
shivakumar-ms: Removed the AnalysisMode. But AnalysisLevel and EnforceCodeStyleInBuild are being used; removing them will cause the code compilation failures (CLSCompliant and code style errors).

wsugarman: Are there too many to include in project-specific global config?

forks/.editorconfig
shivakumar-ms: Done.

wsugarman (approved):

forks/Microsoft.Health.FellowOakDicom/Microsoft.Health.FellowOakDicom.csproj
shivakumar-ms: I added the globalconfig for "Forks". So that in-case if we fork something else, as long as that code lives inside "forks" this globalconfig would apply for that too.

PR #1942, opened 2022-09-13
Fix the json de-serialization bug with private creator data element

smithago:
## Description
Private creator data element is in wrong format compared to the standard.
Standard expects it be LO and single value. We had some customers with MV vr 

## Related issues
AB#95242

## Testing
UT added

## Pending
Need to port this changes to fo-dicom


forks/Microsoft.Health.FellowOakDicom/Serialization/JsonDicomConverter.cs
```diff
@@ -233,7 +233,15 @@ private DicomDataset ReadJsonDataset(ref Utf8JsonReader reader)
 
                     if (dataset.Contains(privateCreatorTag))
                     {
-                        item.Tag.PrivateCreator = new DicomPrivateCreator(dataset.GetSingleValue<string>(privateCreatorTag));
+                        var privateCreatorItem = dataset.GetDicomItem<DicomItem>(privateCreatorTag);
+                        // Based on standard private creator data item should be VM = 1, but there are buggy dcms
+                        // To allow serialization to continue to work instead of throwing checking valid single value before calling GetSingleValue
+                        if (privateCreatorItem is DicomElement element
```
jovinson-ms, line 239: We can simplify this to `privateCreatorItem is DicomLongString element` and leave out the VR check.

jovinson-ms (dismissed):

jovinson-ms (approved):

PR #1943, opened 2022-09-14
[AB#93875] Fixed deserialization issue with DS, IS, SV and UV VR types with non-numeric string values.

shivakumar-ms:
## Description
Updated forked JsonDicomConverter ReadJson behavior to support non-numeric string values for DS, IS, SV and UV vr types.

Corresponding PR with unit tests is in progress in fo-dicom repository: https://github.com/fo-dicom/fo-dicom/pull/1441

Json string similar to the following one, would throw a Json DICOM exception without this fix.

> var json = @"
>             {
>                 ""00101030"": {
>                     ""vr"":""SV"",
>                     ""Value"":[84]
>                 },
>                 ""00101020"": {
>                     ""vr"":""SV"",
>                     ""Value"":[""asd""]
>                 }}";

## Related issues
Addresses [AB#93875].

## Testing
Unit/Integration/e2e tests.


forks/Microsoft.Health.FellowOakDicom/Core/DicomElement.cs
```diff
@@ -0,0 +1,185 @@
...
+
+    #region Public Properties
+
+    public override DicomVR ValueRepresentation => DicomVR.SV;
+
+    #endregion
+
+    #region Public Members
+
+    public override T Get<T>(int item = -1)
```
smithago, line 47: I am not sure why we are overriding the whole methos here, why can't we just use the second constructor which takes byeBuffer

forks/Microsoft.Health.FellowOakDicom/Serialization/JsonDicomConverter.cs
```diff
@@ -886,6 +892,11 @@ private static T[] ReadJsonMultiNumberOrStringValue<T>(ref Utf8JsonReader reader
...
                     childValues.Add(parsed);
                 }
+                else if (reader.TokenType == JsonTokenType.String && (_numberSerializationMode == NumberSerializationMode.AsString || _numberSerializationMode == NumberSerializationMode.PreferablyAsNumber))
+                {
+                    hasNonNumericString = true;
+                    childValues.Add(reader.GetString());
+                }
                 else
                 {
                     throw new JsonException("Malformed DICOM json, number expected");
```
smithago, line 902: Should'nt we be fixing this? else just return the bytearray?

forks/Microsoft.Health.FellowOakDicom/Core/DicomElement.cs
shivakumar-ms: We are using ByteBuffer constructor. But when we call "GetString" it would call the Get (virtual) method, and the base class implementation does not handle a case, which requires a override in the derived class. (I modeled this similar to how DS and IS VR types are being handled already)

smithago (approved):

forks/Microsoft.Health.FellowOakDicom/Serialization/JsonDicomConverter.cs
```diff
@@ -862,7 +867,7 @@ private object ReadJsonMultiNumberOrString<T>(ref Utf8JsonReader reader, GetValu
             }
         }
 
-        private static T[] ReadJsonMultiNumberOrStringValue<T>(ref Utf8JsonReader reader, GetValue<T> getValue, TryParse<T> tryParse)
+        private object ReadJsonMultiNumberOrStringValue<T>(ref Utf8JsonReader reader, GetValue<T> getValue, TryParse<T> tryParse)
```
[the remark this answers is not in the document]
jovinson-ms: You can add the unit tests in the Core project like this PR: https://github.com/microsoft/dicom-server/pull/1942/files

forks/Microsoft.Health.FellowOakDicom/Serialization/JsonDicomConverter.cs
```diff
@@ -886,6 +894,11 @@ private static T[] ReadJsonMultiNumberOrStringValue<T>(ref Utf8JsonReader reader
                 {
                     childValues.Add(parsed);
                 }
+                else if (reader.TokenType == JsonTokenType.String && (_numberSerializationMode == NumberSerializationMode.AsString || _numberSerializationMode == NumberSerializationMode.PreferablyAsNumber))
```
jovinson-ms, line 897: For the case of reading, we just need to check the `_autovalidate` member - the `NumberSerializationMode` is for writing.

forks/Microsoft.Health.FellowOakDicom/Core/DicomElement.cs
```diff
@@ -0,0 +1,245 @@
...
+    public override DicomVR ValueRepresentation => DicomVR.UV;
+
+    #endregion
+}
+
+/// <summary>
+/// Base class to handle Multi String/Number VR Types
+/// e.g. DS, IS, SV, and UV
+/// </summary>
+public abstract class DicomMultiStringOrNumberElement<TType> : DicomMultiStringElement where TType : struct
```
jovinson-ms, line 152: Why `TType` instead of `T`?

forks/Microsoft.Health.FellowOakDicom/Core/DicomElement.cs
jovinson-ms: I looked into this as well, and there really isn't a way around overriding the `Get` method because it's used on writing. Even if we pass the IByteBuffer in successfully, the json writer will call `Get` on the element, which will not handle the value as a string.

forks/Microsoft.Health.FellowOakDicom/Core/DicomElement.cs
shivakumar-ms: "Get" override already uses "T", which is different from TType at the class level.

For e.g., the purpose of "T" in "Get" is to be able to ask for "Get<int>" on a VeryLong dicom element.

forks/Microsoft.Health.FellowOakDicom/Serialization/JsonDicomConverter.cs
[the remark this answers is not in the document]
shivakumar-ms: Done

forks/Microsoft.Health.FellowOakDicom/Serialization/JsonDicomConverter.cs
shivakumar-ms: Done.

jovinson-ms (approved):

PR #1964, opened 2022-09-23
Handling DataStoreExceptions when caused by a TaskCanceledException

jovinson-ms:
## Description
Currently, a request that is cancelled by the client will throw a `TaskCanceledException` in the blob libraries that is wrapped in a `DataStoreException`. This change will return 400 Bad Request in this case rather than a 500.

## Related issues
Addresses AB#95767

## Testing
Added a unit test.

### DIFF
--- a/src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
+++ b/src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
@@ -1,4 +1,4 @@
-﻿// -------------------------------------------------------------------------------------------------
+// -------------------------------------------------------------------------------------------------
 // Copyright (c) Microsoft Corporation. All rights reserved.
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
```diff
@@ -84,6 +84,7 @@ private IActionResult MapExceptionToResult(Exception exception)
             case AuditHeaderTooLargeException:
             case ConnectionResetException:
             case OperationCanceledException:
+            case DataStoreException e when e.InnerException is TaskCanceledException:
```
wsugarman, line 87: Should we way `OperationCanceledException` instead?

jovinson-ms: I think they come from different sources - the blob libraries throw `TaskCanceledException`.

PR #2014, opened 2022-10-15
Throw 400 on quotes character in query filter condition with fuzzymatch

smithago:
## Description
Qido query with PatienName=Thoma"s&FuzzyMatching=True throws 500 error

This because double quote is a invalid char in any fuzzy matching(word or phrase) in SQL.
Since this a fuzzy matching it needs to be a word or phrase and we should not replace the quote with any character.
 
Error will recommend customer to do exact match or try a simple word with fuzzy match. Which will help the customer to have better success.

## Related issues
[AB#96513](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/96513)

## Testing
Unit test added

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/DicomCoreResource.resx
+++ b/src/Microsoft.Health.Dicom.Core/DicomCoreResource.resx
@@ -627,4 +627,8 @@ For details on valid range queries, please refer to Search Matching section in C
   <data name="UnsupportedBuffering" xml:space="preserve">
     <value>Cannot buffer items if parallelism is unbounded.</value>
   </data>
+  <data name="InvalidTagValueWithFuzzyMatch" xml:space="preserve">
+    <value>Invalid query: specified tag '{0}' value '{1}' contains unsupported character '{2}'. Character is not supported with Fuzzy match, try exact match or removing the invalid character. </value>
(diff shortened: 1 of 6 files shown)


src/Microsoft.Health.Dicom.Core/DicomCoreResource.resx
```diff
@@ -627,4 +627,8 @@ For details on valid range queries, please refer to Search Matching section in C
   <data name="UnsupportedBuffering" xml:space="preserve">
     <value>Cannot buffer items if parallelism is unbounded.</value>
   </data>
+  <data name="InvalidTagValueWithFuzzyMatch" xml:space="preserve">
+    <value>Invalid query: specified tag '{0}' value '{1}' contains unsupported character '{2}'. Character is not supported with Fuzzy match, try exact match or removing the invalid character. </value>
```
alishahahmed, line 631: Do we want to include `value` in the message? Won't this be considered customer provided data?

alishahahmed (approved):

src/Microsoft.Health.Dicom.Core/DicomCoreResource.resx
smithago: I am hoping the trace filter will filter it for logs we get.

PR #2015, opened 2022-10-16
Move wado resource get without transcoding to new blob network api to reduce buffer copy and possible reduce cpu cycle

smithago:
## Description
Based on the doc https://github.com/microsoft/dicom-server/blob/main/src/Microsoft.Health.Dicom.Blob/AzureBlobAPIs.md

## Related issues
[AB#96577](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/96577)

## Testing
Existing tests pass.

### DIFF
--- a/test/Microsoft.Health.Dicom.Tests.Integration/Features/RetrieveResourceServiceTests.cs
+++ b/test/Microsoft.Health.Dicom.Tests.Integration/Features/RetrieveResourceServiceTests.cs
@@ -1,4 +1,4 @@
-﻿// -------------------------------------------------------------------------------------------------
+// -------------------------------------------------------------------------------------------------
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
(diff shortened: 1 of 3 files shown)


test/Microsoft.Health.Dicom.Tests.Integration/Features/RetrieveResourceServiceTests.cs
```diff
@@ -144,7 +144,7 @@ public async Task GivenStoredInstancesWithMissingFile_WhenRetrieveRequestForSeri
         await Assert.ThrowsAsync<ItemNotFoundException>(() => response.GetStreamsAsync());
     }
 
-    [Fact]
+    [Fact(Skip = "Fix getting stream")]
```
wsugarman, line 147: What happened here?

smithago: fixed, had to realize the stream and reset the position

PR #2042, opened 2022-10-31
Initial implementation to collect all the warnings and errors from the store tag validator

shivakumar-ms:
## Description
Initial implementation to collect all the warnings and errors from the store tag validator.

This PR enables collecting errors as part of the result. We will expose errors in result in a separate PR. This PR does not modify response of STOW.

## Related issues
Addresses [issue #].

## Testing
Updated unit tests


src/Microsoft.Health.Dicom.Core/Features/Store/StoreDatasetValidator.cs
```diff
@@ -112,16 +124,26 @@ string EnsureRequiredTagIsPresent(DicomTag dicomTag)
...
         foreach (QueryTag queryTag in queryTags)
         {
-            warning |= dicomDataset.ValidateQueryTag(queryTag, _minimumValidator);
+            try
+            {
+                var validationWarning = dicomDataset.ValidateQueryTag(queryTag, _minimumValidator);
+
+                validationResultBuilder.AddWarning(validationWarning, queryTag);
+            }
+            catch (ElementValidationException ex)
```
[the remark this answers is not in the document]
shivakumar-ms: catching here allows us to collect all the errors. to make use of in the Service layer.

We are re-throwing the 1st exception to keep the current behavior intact until we get the full implementation in place for the Validation Errors/Warnings in response.

src/Microsoft.Health.Dicom.Core/Features/Store/ValidationResultBuilderFactory.cs
```diff
@@ -0,0 +1,24 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.Health.Dicom.Core.Features.Store;
+
+
+internal class ValidationResultBuilderFactory : IValidationResultBuilderFactory
```
jovinson-ms, line 11: What are we using this factory for? I think we can just create the validation result.

src/Microsoft.Health.Dicom.Core/Features/Store/StoreDatasetValidator.cs
```diff
@@ -26,26 +27,37 @@ public class StoreDatasetValidator : IStoreDatasetValidator
...
     private readonly IQueryTagService _queryTagService;
+    private readonly IValidationResultBuilderFactory _validationResultBuilderFactory;
+
 
-    public StoreDatasetValidator(IOptions<FeatureConfiguration> featureConfiguration, IElementMinimumValidator minimumValidator, IQueryTagService queryTagService)
+    public StoreDatasetValidator(
+        IOptions<FeatureConfiguration> featureConfiguration,
+        IElementMinimumValidator minimumValidator,
+        IQueryTagService queryTagService,
+        IValidationResultBuilderFactory validationResultBuilderFactory)
```
jovinson-ms, line 37: I'm not sure we need DI for this - since this validation result is already specific to this usage I think we can just return the results as a data transfer object.

src/Microsoft.Health.Dicom.Core/DicomCoreResource.resx
```diff
@@ -631,4 +631,7 @@ For details on valid range queries, please refer to Search Matching section in C
     <value>Invalid query: specified tag '{0}' value '{1}' contains unsupported character '{2}'. Character is not supported with Fuzzy match, try exact match or removing the invalid character. </value>
     <comment>{0} Dicom Tag name, {1} Tag value to search. {2} Unsupported character.</comment>
   </data>
+  <data name="DatasetDoesNotMatchSOPClass" xml:space="preserve">
+    <value>Dataset does not match SOP Class.</value>
```
jovinson-ms, line 635: Is there not an existing resource string for this?

src/Microsoft.Health.Dicom.Core/Features/Store/ValidationResultBuilder.cs
```diff
@@ -0,0 +1,104 @@
...
+    {
+        get { return _errorMessages.Count > 0; }
+    }
+
+    public IEnumerable<string> Errors
+    {
+        get { return _errorMessages; }
+    }
+
+    public IEnumerable<string> Warnings
```
jovinson-ms, line 40: What's the disctinction between `Warnings` and `WarningCodes`?

src/Microsoft.Health.Dicom.Core/Features/Store/ValidationResultBuilderFactory.cs
shivakumar-ms: ValidationResult/Builder has to be transient. And I was trying to avoid "new..." in the code, to make our life easier with the unit test.

src/Microsoft.Health.Dicom.Core/Features/Store/ValidationResultBuilder.cs
shivakumar-ms: Warnings is the "list of warning messages". The only reason I am keeping WarningCodes is to keep the current behavior. We just need either one.

src/Microsoft.Health.Dicom.Core/Features/Store/ValidationResultBuilder.cs
```diff
@@ -0,0 +1,104 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using EnsureThat;
+using Microsoft.Health.Dicom.Core.Features.ExtendedQueryTag;
+
+namespace Microsoft.Health.Dicom.Core.Features.Store;
+
+internal sealed class ValidationResultBuilder : IValidationResultBuilder
```
wsugarman, line 13: nit: use expressions in this file

src/Microsoft.Health.Dicom.Core/Features/Store/ValidationResultBuilder.cs
```diff
@@ -0,0 +1,104 @@
...
+{
+    private readonly IList<string> _errorMessages;
```
wsugarman, line 15: nit: do not use interfaces for private members that do not leverage any polymorphism; no need to perform virtual dispatches

wsugarman (changes requested):
This chat more about this design offline

src/Microsoft.Health.Dicom.Core/DicomCoreResource.resx
shivakumar-ms: I couldn't find one that matches this value literal.

src/Microsoft.Health.Dicom.Core/Features/Store/ValidationResultBuilderFactory.cs
shivakumar-ms: Removed DI. Kept the builder, to abstract details on how the errors/warnings are collected. Builder.Build returns an immutable DTO instance.

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResultBuilder.cs
```diff
@@ -0,0 +1,89 @@
...
+{
+    public StoreValidatorResultBuilder()
+    {
+        ErrorMessages = new Collection<string>();
+        WarningMessages = new Collection<string>();
+        WarningCodes = ValidationWarnings.None;
+        FirstException = null;
+    }
+
+    private Collection<string> ErrorMessages { get; }
```
wsugarman, line 23: Please do not use `private` properties. Instead use fields.

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResultBuilder.cs
```diff
@@ -0,0 +1,89 @@
...
+using EnsureThat;
+using Microsoft.Health.Dicom.Core.Features.ExtendedQueryTag;
+
+namespace Microsoft.Health.Dicom.Core.Features.Store;
+
+internal sealed class StoreValidatorResultBuilder
+{
+    public StoreValidatorResultBuilder()
+    {
+        ErrorMessages = new Collection<string>();
```
wsugarman, line 17: Do not use `Collection<T>`. This is a base class used by other classes typically in the `System.Collections.ObjectModel` namespace. Please use a more typical collection like `List<T>`

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResultBuilder.cs
```diff
@@ -0,0 +1,89 @@
...
+        FirstException = null;
+    }
+
+    private Collection<string> ErrorMessages { get; }
+
+    private Collection<string> WarningMessages { get; }
+
+    private ValidationWarnings WarningCodes { get; set; }
+
+    // TODO: Remove this during the cleanup. *** Hack to support the existing validator behavior ***
```
wsugarman, line 29: Remove the "hack" comment and just indicate it is used to preserve the existing behavior

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResultBuilder.cs
```diff
@@ -0,0 +1,89 @@
...
+            ErrorMessages,
+            WarningMessages,
+            WarningCodes,
+            FirstException);
+    }
+
+    public void AddError(Exception ex, QueryTag queryTag = null)
+    {
+        // TODO: Remove this during the cleanup. *** Hack to support the existing validator behavior ***
+        if (null == FirstException)
```
wsugarman, line 44: nit: typically the variable is first in the comparison `FirstException == null` as we would say, "Is the 'FirstException' equal to null?"

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResultBuilder.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
```
wsugarman, line 45: nit: remove `{ ... }` here

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResultBuilder.cs
```diff
@@ -0,0 +1,89 @@
...
+            FirstException = ex;
+        }
+
+        ErrorMessages.Add(GetFormattedText(ex?.Message, queryTag));
+    }
+
+    public void AddError(string message, QueryTag queryTag = null)
+    {
+        if (string.IsNullOrWhiteSpace(message))
+            return;
```
wsugarman, line 55: Not an error?

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResultBuilder.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        ErrorMessages.Add(GetFormattedText(message, queryTag));
+    }
+
+    public void AddWarning(ValidationWarnings warningCode, QueryTag queryTag = null)
+    {
+        if (warningCode == ValidationWarnings.None)
+            return;
```
wsugarman, line 63: Not an error?

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResultBuilder.cs
```diff
@@ -0,0 +1,89 @@
...
+        WarningMessages = new Collection<string>();
+        WarningCodes = ValidationWarnings.None;
+        FirstException = null;
+    }
+
+    private Collection<string> ErrorMessages { get; }
+
+    private Collection<string> WarningMessages { get; }
+
+    private ValidationWarnings WarningCodes { get; set; }
```
wsugarman, line 27: What is the point of these codes? It is a bit confusing because we don't really record the exceptions, just messages, but we preserve these codes in addition to their corresponding messages?

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResultBuilder.cs
```diff
@@ -0,0 +1,89 @@
...
+    public StoreValidatorResult Build()
+    {
+        return new StoreValidatorResult(
+            ErrorMessages,
+            WarningMessages,
+            WarningCodes,
+            FirstException);
+    }
+
+    public void AddError(Exception ex, QueryTag queryTag = null)
```
wsugarman, line 41: I'm on the fence on these methods. A few questions:
1. Is there a single `StoreValidatorResult` per entry in the STOW request or one per STOW request?
2. Should we be recording the mapping of query tag -> exception(s)?

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResult.cs
```diff
@@ -0,0 +1,33 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Health.Dicom.Core.Features.Store;
+
+public sealed class StoreValidatorResult
```
wsugarman, line 11: Perhaps `StoreValidationResult` and `StoreValidationResultBuilder`

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResult.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    public StoreValidatorResult(
+        IReadOnlyCollection<string> errors,
```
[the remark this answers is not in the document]
wsugarman: I would say let's lean towards the final result as being immutable. Ideally each path simply has access to the builder. Alternatively, we could add some sort of aggregation method to the result class to combine them

wsugarman (changes requested):

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResultBuilder.cs
shivakumar-ms: 1) One per STOW request
2) My initial thought was to implement this as K, V (QueryTag vs. Error/Warning/Exception). But we deferred this to a later phase.

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResultBuilder.cs
shivakumar-ms: Tried using List<T>. But I get code analysis error:

CA1002	Change 'List<string>' in 'StoreValidationResultBuilder._errorMessages' to use 'Collection<T>', 'ReadOnlyCollection<T>' or 'KeyedCollection<K,V>'

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResultBuilder.cs
shivakumar-ms: Done.

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResultBuilder.cs
shivakumar-ms: Done.

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResultBuilder.cs
shivakumar-ms: Done

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResultBuilder.cs
shivakumar-ms: Done

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResultBuilder.cs
shivakumar-ms: it's not an error or warning.

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResultBuilder.cs
shivakumar-ms: We are keeping the "WarningCodes" to continue to support the current behavior.

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResult.cs
shivakumar-ms: Done.

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResult.cs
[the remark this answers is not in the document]
shivakumar-ms: @wsugarman @bcarthic @jovinson-ms could you please chime in on this?

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResultBuilder.cs
wsugarman: Are you sure about this? I could imagine you might get this error when they were properties, but not as private fields. Can you double-check? Otherwise, we can change our `.globalconfig`. We have the power

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResultBuilder.cs
wsugarman: Why would someone pass `None`? Perhaps `ArgumentOutOfRange`?

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResultBuilder.cs
wsugarman: Looking at the current behavior, they appear to be used only to populate more warnings. Now that we have a builder, can't we have these warnings added directly without having the codes as a middleman?

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResultBuilder.cs
wsugarman: Can we reduce the APIs to simply `public StoreValidatorResultBuilder AddError(Exception exception)` and `public StoreValidatorResultBuilder AddWarning(Exception exception)` (or perhaps `string` for warnings)? We could perhaps simply wrap messages in a generic validation exception.

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidatorResultBuilder.cs
shivakumar-ms: resolved

wsugarman (approved):

PR #2119, opened 2022-11-17
Update Project for .NET 7

brandonpollett:
## Description
- Update DICOM Web Server to .NET 7
- Multi-target shared libraries for both .NET 6 and .NET 7 as Azure Functions only supports LTS versions like .NET 6
- Add build properties for the latest version, the LTS version, and the STS version
- Update NuGet dependencies
- Use the latest .NET SDK images for Azure Functions, but include the .NET 6 SDK
- Address new analyzer rules
- Install .NET 6 SDK in pipelines to properly run Swashbuckle tool
- Fix API project dependencies to use proper Swashbuckle packages

## Related issues
N/A

## Testing
Existing pipelines


brandonpollett:
Still need to make a pass at items that are part of deployments and/or builds.

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core.UnitTests/Microsoft.Health.DicomCast.Core.UnitTests.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net7.0</TargetFramework>
```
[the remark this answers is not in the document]
brandonpollett: I actually got a warning in my IDE for only specifying one while using `TargetFrameworks`!

Directory.Build.props
```diff
@@ -11,16 +11,19 @@
...
     <LangVersion>Latest</LangVersion>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+    <NetStandardVersion>netstandard2.0</NetStandardVersion>
+    <NetVersion>net7.0</NetVersion>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <Product>Microsoft Health</Product>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
-    <SdkPackageVersion>6.0.0</SdkPackageVersion>
+    <SdkPackageVersion>7.0.0</SdkPackageVersion>
+    <TargetFramework>$(NetVersion)</TargetFramework>
```
wsugarman, line 26: We do need to keep .NET 6 for the Azure Functions, as they don't support .NET 7. So we'll need the same code that choses `<SdkPackageVersion>` based on the target framework. It also means we'll need to multiple target any package used by functions.

wsugarman, line 26: Maybe we should keep the `<TargetFramework>` or `<TargetFrameworks>` elements in each *.csproj, and instead just define properties for the versions here, like `LatestFramework`, `StsVersion`, and `LtsVersion`?

Directory.Build.props
```diff
@@ -11,16 +11,19 @@
...
     <FhirServerPackageVersion>2.0.55</FhirServerPackageVersion>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <HealthcareSharedPackageVersion>6.1.145</HealthcareSharedPackageVersion>
+    <HealthcareSharedPackageVersion>6.2.0</HealthcareSharedPackageVersion>
     <HighEntropyVA>true</HighEntropyVA>
     <Hl7FhirPackageVersion>4.3.0</Hl7FhirPackageVersion>
     <IdentityModelTokenPackageVersion>6.25.0</IdentityModelTokenPackageVersion>
     <LangVersion>Latest</LangVersion>
     <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+    <NetStandardVersion>netstandard2.0</NetStandardVersion>
```
wsugarman, line 20: I don't think we need a property for .NET Standard 2.0

wsugarman (approved):

PR #2151, opened 2022-12-02
start tracking indexable tag validation exceptions by vr code

esmadau:
## Description
Start tracking validation errors on index tags. See unit test to get idea of value of dimensions for metric.

## Related issues

## Testing
Added unit test.

### DIFF
--- a/src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/StoreDatasetValidatorTests.cs
+++ b/src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/StoreDatasetValidatorTests.cs
@@ -243,7 +243,6 @@ public class StoreDatasetValidatorTests
         await ExecuteAndValidateTagEntriesException<ElementValidationException>(ValidationFailedFailureCode);
     }
 
-
     [Fact]
     public async Task GivenDatasetWithInvalidIndexedTagValue_WhenValidating_ThenValidationExceptionMetricsShouldBeTracked()
     {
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.Dicom.Core/Features/Store/StoreDatasetValidator.cs
```diff
@@ -138,6 +142,16 @@ string EnsureRequiredTagIsPresent(DicomTag dicomTag)
             catch (ElementValidationException ex)
             {
                 validationResultBuilder.Add(ex, queryTag);
+                const int value = 1;
+                _telemetryClient.GetMetric(
```
wsugarman, line 146: nit: format

```csharp
_telemetryClient
    .GetMetric(
        "IndexTagValidationError",
        "ExceptionErrorCode",
        "ExceptionName",
        "VrCode")
    .TrackValue(
        value,
        ex.ErrorCode.ToString(),
        ex.Name,
        queryTag.VR.Code);
```

wsugarman, line 146: You can inline the `value` in source. I think it's clear either way

wsugarman, line 146: It might behoove us to write new metrics as [`EventCounters`](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/event-counters#consume-eventcounters) while we're considering the transition to OpenTelemetry. But we already have AppInsights metrics that need to be migrated, so perhaps one more isn't too bad lol

wsugarman (approved):

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/StoreDatasetValidatorTests.cs
```diff
@@ -236,6 +243,29 @@ public async Task GivenDatasetWithInvalidIndexedTagValue_WhenValidating_ThenVali
         await ExecuteAndValidateTagEntriesException<ElementValidationException>(ValidationFailedFailureCode);
     }
 
+
```
wsugarman, line 246: Extra newline here

PR #2167, opened 2022-12-06
track metadata json exceptions

esmadau:
## Description
To understand how often we have issues with validation that occurs as part of fo-dicom json de/serialization, we want to track a metric each time an error occurs.
We want to track these as metrics instead of logs to avoid having any leaking PI data as well as to be able to have long term analysis of issues that crop up.

## Related issues

## Testing

### DIFF
--- a/src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobMetadataStore.cs
+++ b/src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobMetadataStore.cs
@@ -36,6 +36,8 @@ public class BlobMetadataStore : IMetadataStore
 {
     private const string StoreInstanceMetadataStreamTagName = nameof(BlobMetadataStore) + "." + nameof(StoreInstanceMetadataAsync);
     private const string StoreInstanceFramesRangeTagName = nameof(BlobMetadataStore) + "." + nameof(StoreInstanceFramesRangeAsync);
+    private const string JsonDeserializationException = "JsonDeserializationException";
+    private const string JsonDeserializationExceptionTypeDimension = "JsonDeserializationExceptionTypeDimension";
     private readonly BlobContainerClient _container;
     private readonly JsonSerializerOptions _jsonSerializerOptions;
     private readonly RecyclableMemoryStreamManager _recyclableMemoryStreamManager;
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobMetadataStore.cs
```diff
@@ -106,6 +111,12 @@ public class BlobMetadataStore : IMetadataStore
         }
         catch (Exception ex)
         {
+            if (ex is NotSupportedException)
+            {
+                _telemetryClient
```
smithago, line 116: Interesting to see metric client used here. Can we just use the logger? If you want to track the original exceptions. I did not get a lot of details in the description. Can you add some description for the PR.

jovinson-ms: There are several advantages to recording a metric here instead of a log:
- We can track trends over a longer time range than our log retention period
- We can ensure no private data is collected

In order to collect metrics with Application Insights, this is the [documented method](https://learn.microsoft.com/en-us/azure/azure-monitor/app/get-metric). When we move to an Otel implementation this will need to change, but we are currently collecting metrics this way elsewhere.

docker/docker-compose.yml
```diff
@@ -43,6 +43,7 @@ services:
       AzureWebJobsStorage: "DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://azurite:10000/devstoreaccount1;QueueEndpoint=http://azurite:10001/devstoreaccount1;TableEndpoint=http://azurite:10002/devstoreaccount1;"
       AZURE_FUNCTIONS_ENVIRONMENT: "Development"
       WEBSITE_HOSTNAME: "localhost:8080"
+      APPINSIGHTS_INSTRUMENTATIONKEY: "dummy" # necessary to get host to configure telemetry client even when just running locally
```
jovinson-ms, line 46: Can this be an empty string?

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobMetadataStore.cs
```diff
@@ -92,6 +97,12 @@ public class BlobMetadataStore : IMetadataStore
...
+                    .GetMetric("JsonSerializationException", "ExceptionType")
```
jovinson-ms, line 103: Both of these should probably be constants, and I'd suggest just `Type` for the dimension name.

jovinson-ms (dismissed):

docker/docker-compose.yml
jovinson-ms: Let's use another string that's obviously not a valid instrumentation key, like empty GUID: `00000000-0000-0000-0000-000000000000`.

Policheck will flag this.

jovinson-ms (approved):

PR #2176, opened 2022-12-09
Handle parallel add partition calls

bcarthic:
## Description
Fix parallel Add partition call exception.

## Related issues
Addresses [issue [AB#96372](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/96372)].

## Testing


src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/AddPartition.sql
```diff
@@ -30,6 +30,13 @@ BEGIN
     DECLARE @createdDate DATETIME2(7) = SYSUTCDATETIME()
     DECLARE @partitionKey INT
 
+    SELECT @partitionKey = PartitionKey
```
smithago, line 33: This will reduce the occurance but won't elemenate it. You will have to hold a xLock to block any insertions after the check and before the insert.

Or you can do what STOW does in the c# store layer. Handle the conflict error and throw the exception

dicom-server\src\Microsoft.Health.Dicom.SqlServer\Features\Store\SqlIndexDataStoreV10.cs
catch (SqlException ex) when (ex.Number == SqlErrorCodes.Conflict

bcarthic: image

src/Microsoft.Health.Dicom.Core/Exceptions/DataPartitionsAlreadyExistsException.cs
```diff
@@ -0,0 +1,19 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System.Globalization;
+
+namespace Microsoft.Health.Dicom.Core.Exceptions;
+
+/// <summary>
+/// Exception that is thrown when partition name is not found
+/// </summary>
+public class DataPartitionsAlreadyExistsException : BadRequestException
```
jovinson-ms, line 13: nit: should be `Partition`

jovinson-ms (dismissed):

src/Microsoft.Health.Dicom.Core/Exceptions/DataPartitionsAlreadyExistsException.cs
bcarthic: Done

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/AddPartition.sql
smithago: Looks like I missed it. Makes sense

smithago (approved):

PR #2214, opened 2022-12-27
multiple accept header fix

esmadau:
## Description
- We sort accept headers by their quality preference as requested from the user.
- When multiple headers given, attempt to use the highest quality that we "accept"
- When highest quality header is not acceptable, attempt to use the next highest quality

_NOTE: This is slightly different from how the bug was initially filed, which requested that we pick the highest preference regardless of if it is accepted or not. When not accepted, we should return a "NotApplicable" syntax.
Example:
header 1- multipart/related;type=”image/jp2”;transfer-syntax=1.2.840.10008.1.2.5(RLELossless);q=0.7,
header 2 - multipart/related;type=”application/octet-stream”;transfer-syntax= 1.2.840.10008.1.2.1 (ExplicitVRLittleEndian);q=0.5

In the above case, header 1 would have been picked as it is highest quality. However, we do not support the transfer syntax requested of RLELossless, so we would have returned "NotAcceptable".__

I did not find any references in the Dicom standard or RFC indicating we should return "NotAcceptable" as originally noted instead of going to the next highest acceptable header.

- https://dicom.nema.org/medical/dicom/current/output/chtml/part18/sect_8.4.html#sect_8.4.1
- https://www.rfc-editor.org/rfc/rfc7231#section-5.3.2

Instead, I take the documentation to mean that if more than one header is provided, then any and each of the headers provided is acceptable for the user and we can use the "next most preferred" if the highest preferred is not supported by the service.

## Related issues
Addresses [[AB#75782](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/75782)].

## Testing
unit tests

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Features/Retrieve/AcceptHeaderHandler.cs
+++ b/src/Microsoft.Health.Dicom.Core/Features/Retrieve/AcceptHeaderHandler.cs
@@ -37,7 +37,7 @@ public class AcceptHeaderHandler : IAcceptHeaderHandler
     {
     }
 
-    public AcceptHeaderHandler(
+    private AcceptHeaderHandler(
         IReadOnlyDictionary<ResourceType, List<AcceptHeaderDescriptor>> acceptableDescriptors,
         ILogger<AcceptHeaderHandler> logger)
     {


src/Microsoft.Health.Dicom.Core/Messages/Retrieve/AcceptHeader.cs
```diff
@@ -10,12 +10,13 @@ namespace Microsoft.Health.Dicom.Core.Messages.Retrieve;
...
 public class AcceptHeader
 {
+    public const double DefaultQuality = 1.0;
     public AcceptHeader(StringSegment mediaType, PayloadTypes payloadType, StringSegment transferSyntax = default, double? quality = null)
     {
         MediaType = mediaType;
         PayloadType = payloadType;
         TransferSyntax = transferSyntax;
-        Quality = quality;
+        Quality = quality.GetValueOrDefault(DefaultQuality);
```
jovinson-ms, line 19: We can remove this change if the parameter is `double quality = DefaultQuality`

src/Microsoft.Health.Dicom.Core/Features/Retrieve/AcceptHeaderHandler.cs
```diff
@@ -0,0 +1,134 @@
...
+    /// <param name="resourceType">Used to understand if header properties are valid.</param>
+    /// <param name="acceptHeaders">One or more headers as requested by user.</param>
+    /// <returns>New accept header based on highest priority valid header requested.</returns>
+    /// <exception cref="NotAcceptableException"></exception>
+    public AcceptHeader GetValidAcceptHeader(ResourceType resourceType, IEnumerable<AcceptHeader> acceptHeaders)
+    {
+        EnsureArg.IsNotNull(acceptHeaders, nameof(acceptHeaders));
+        List<AcceptHeader> orderedHeaders = acceptHeaders.OrderByDescending(x => x.Quality).ToList();
+
+        _logger.LogInformation(
```
jovinson-ms, line 64: It might be useful to have this log statement first, so that invalid headers would be logged before throwing an exception.

jovinson-ms (commented):
We need to be able to handle [media ranges](https://dicom.nema.org/medical/dicom/current/output/html/part18.html#sect_8.4.1.1) (like `application/*`) as well as types.
I'd suggest looking at the general [Accept filter](https://github.com/microsoft/dicom-server/blob/fb172c126f442ba9aca44d3aa643b726b48b7c8e/src/Microsoft.Health.Dicom.Api/Features/Filters/AcceptContentFilterAttribute.cs#L66) where we cast the media type (or range) into a ASP.NET-provided `MediaTypeHeaderValue` which has a `MatchesMediaType` member which will handle media ranges and parameters.

jovinson-ms:
> we've been checking media-type and other things in [IsAcceptable](https://github.com/microsoft/dicom-server/blob/8d2500c4c880f3ef53633fc2fd5d8c2462589620/src/Microsoft.Health.Dicom.Core/Features/Retrieve/AcceptHeaderDescriptor.cs#L45) for each header descriptor, which we have a set of per resource type being requested. This does seem to be duplicating logic though - were you trying to call out the duplication?

I was noting that the existing implementation is flawed. If someone requests a study with an Accept header that includes a type of `*/*`, we should match with the only supported value of `application/dicom` and return that. But since we are doing string matching and not taking wildcards into account, we will return a 415 Not Acceptable. We can note this as another issue if it makes sense to not address in this PR.

brandonpollett:
Generally, I think we should probably think about a refactor that abstracts the concept of an "accept header" away from the core. It definitely feels more like a web api concern. It feels like for the core we should just be saying what we want returned.

brandonpollett (approved):

brandonpollett (approved):

PR #2220, opened 2022-12-29
Users/lagomesma/accept header refactor

esmadau:
## Description
Refactor to support multiple headers. This attempts to simplify the way by which we get a header to use based on if and when a user has requested any acceptable headers.
- we still have to use highest priority where we can
- we still have to except when user didn't request anything acceptable
- we have to set the default syntax when user didn't request it in their header and if it was not mandatory for them to do so

We now return a single value (a new accept header with appropriate transfer syntax set) so we do not have both a return value as well as multiple `out` params to track.

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
+++ b/src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
@@ -30,7 +30,7 @@ public class RetrieveResourceService : IRetrieveResourceService
     private readonly IInstanceStore _instanceStore;
     private readonly ITranscoder _transcoder;
     private readonly IFrameHandler _frameHandler;
-    private readonly IRetrieveTransferSyntaxHandler _retrieveTransferSyntaxHandler;
+    private readonly IAcceptHeaderHandler _acceptHeaderHandler;
     private readonly IDicomRequestContextAccessor _dicomRequestContextAccessor;
     private readonly IMetadataStore _metadataStore;
     private readonly RetrieveConfiguration _retrieveConfiguration;
@@ -43,7 +43,7 @@ public class RetrieveResourceService : IRetrieveResourceService
         IFileStore blobDataStore,
         ITranscoder transcoder,
         IFrameHandler frameHandler,
-        IRetrieveTransferSyntaxHandler retrieveTransferSyntaxHandler,
+        IAcceptHeaderHandler acceptHeaderHandler,
         IDicomRequestContextAccessor dicomRequestContextAccessor,
         IMetadataStore metadataStore,
         IInstanceMetadataCache instanceMetadataCache,
@@ -56,7 +56,7 @@ public class RetrieveResourceService : IRetrieveResourceService
         EnsureArg.IsNotNull(blobDataStore, nameof(blobDataStore));
         EnsureArg.IsNotNull(transcoder, nameof(transcoder));
         EnsureArg.IsNotNull(frameHandler, nameof(frameHandler));
-        EnsureArg.IsNotNull(retrieveTransferSyntaxHandler, nameof(retrieveTransferSyntaxHandler));
+        EnsureArg.IsNotNull(acceptHeaderHandler, nameof(acceptHeaderHandler));
         EnsureArg.IsNotNull(dicomRequestContextAccessor, nameof(dicomRequestContextAccessor));
         EnsureArg.IsNotNull(metadataStore, nameof(metadataStore));
         EnsureArg.IsNotNull(instanceMetadataCache, nameof(instanceMetadataCache));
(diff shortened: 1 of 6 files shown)


src/Microsoft.Health.Dicom.Core/Features/Retrieve/IRetrieveTransferSyntaxHandler.cs
```diff
@@ -11,5 +11,5 @@ namespace Microsoft.Health.Dicom.Core.Features.Retrieve;
 
 public interface IRetrieveTransferSyntaxHandler
```
brandonpollett, line 12: Should we change this interface name to reference `AcceptHeader`?

src/Microsoft.Health.Dicom.Tests.Common/AcceptHeaderHelpers.cs
```diff
@@ -49,6 +49,6 @@ public static AcceptHeader CreateAcceptHeaderForGetFrame(string transferSyntax =
 
     public static AcceptHeader CreateAcceptHeader(string transferSyntax = "*", PayloadTypes payloadType = PayloadTypes.MultipartRelated, string mediaType = KnownContentTypes.ApplicationOctetStream, double? quality = null)
     {
-        return new AcceptHeader(mediaType, payloadType, transferSyntax, quality);
+        return new AcceptHeader(mediaType, payloadType, transferSyntax, quality.GetValueOrDefault(AcceptHeader.DefaultQuality));
```
brandonpollett, line 52: Would it make more sense to do the `quality.GetValueOrDefault(AcceptHeader.DefaultQuality)` inside the `AcceptHeader` constructor so that we centralize that logic?

src/Microsoft.Health.Dicom.Core/Messages/Retrieve/AcceptHeaderQualityComparer.cs
```diff
@@ -24,8 +24,8 @@ public int Compare([AllowNull] AcceptHeader x, [AllowNull] AcceptHeader y)
             return x != null ? 1 : -1;
         }
 
-        double xQuality = x.Quality.GetValueOrDefault(DefaultQuality);
-        double yQuality = y.Quality.GetValueOrDefault(DefaultQuality);
+        double xQuality = x.Quality;
```
bcarthic, line 27: DefaultQuality can be removed

src/Microsoft.Health.Dicom.Core/Features/Retrieve/AcceptHeaderDescriptor.cs
```diff
@@ -64,18 +63,25 @@ public bool IsAcceptable(AcceptHeader acceptHeader, out string transferSyntax)
...
         {
-            transferSyntax = acceptHeader.TransferSyntax.Value;
             return true;
         }
 
         return false;
     }
+
+    public StringSegment GetTransferSyntax(AcceptHeader acceptHeader)
+    {
```
bcarthic, line 78: Should we do 
`EnsureArg.IsNotNull(acceptHeader, nameof(acceptHeader));`

src/Microsoft.Health.Dicom.Core/Features/Retrieve/IRetrieveTransferSyntaxHandler.cs
```diff
@@ -11,5 +11,5 @@ namespace Microsoft.Health.Dicom.Core.Features.Retrieve;
...
 {
```
bcarthic, line 13: Should the interface name changed?

src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
```diff
@@ -84,36 +84,37 @@ public async Task<RetrieveResourceResponse> GetInstanceResourceAsync(RetrieveRes
...
         try
         {
-            string requestedTransferSyntax = _retrieveTransferSyntaxHandler
-                .GetTransferSyntax(message.ResourceType, message.AcceptHeaders, out AcceptHeaderDescriptor acceptHeaderDescriptor, out AcceptHeader acceptedHeader);
+            AcceptHeader validAcceptHeader = _retrieveTransferSyntaxHandler.GetValidAcceptHeader(
+                message.ResourceType,
+                message.AcceptHeaders);
 
-            bool isOriginalTransferSyntaxRequested = DicomTransferSyntaxUids.IsOriginalTransferSyntaxRequested(requestedTransferSyntax);
+            bool isOriginalTransferSyntaxRequested = DicomTransferSyntaxUids.IsOriginalTransferSyntaxRequested(validAcceptHeader.TransferSyntax.Value);
```
brandonpollett, line 91: It may make sense to set a variable still for `string requestedTransferSyntax = validAcceptHeader.TransferSyntax.Value;` since it seems we do this a number of times.

src/Microsoft.Health.Dicom.Core/Messages/Retrieve/AcceptHeader.cs
```diff
@@ -10,21 +10,22 @@ namespace Microsoft.Health.Dicom.Core.Messages.Retrieve;
...
-        Quality = quality;
+        Quality = quality.GetValueOrDefault(DefaultQuality);
     }
 
     public StringSegment MediaType { get; }
 
     public PayloadTypes PayloadType { get; }
 
-    public StringSegment TransferSyntax { get; }
+    public StringSegment TransferSyntax { get; set; }
```
brandonpollett, line 26: I don't think we're calling the `set;` anywhere.

brandonpollett (approved):

PR #2226, opened 2023-01-05
Introduce option to drop attribute on read when invalid Dicom+JSON

esmadau:
## Description
Introduce option to drop attribute on read when invalid Dicom+JSON.
Does not handle parsing invalid JSON (such as invalid escape chars). Encountering an invalid JSON token will still result in a JSON exception

## Related issues
Addresses [[AB#99365](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/99365)].

## Testing
Added unit tests

PR #2227, opened 2023-01-07
drop invalid dicomelements

esmadau:
## Description
- drops invalid dicom elements from dicomDataset such that neither JSON metadata is written with invalid elements nor do we index them if any are indexable
- introduces feature flag

To be done in other prs:
- validation error formatting in comment sequences
- introducing flag at infra level
- adding a SQL column indicating data was dropped
- consideration of behavior when user asks to validate all items
- metrics and logging

## Related issues
Addresses [[AB#99359](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/99359)].

related draft pr on deserialization : https://github.com/microsoft/dicom-server/pull/2226

## Testing
code tests as well as manual tests on local server

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Features/Store/StoreResponseBuilder.cs
+++ b/src/Microsoft.Health.Dicom.Core/Features/Store/StoreResponseBuilder.cs
@@ -109,10 +109,10 @@ public class StoreResponseBuilder : IStoreResponseBuilder
                             error)))
                 .ToArray();
 
-            var failedSopSequence = new DicomSequence(
-                DicomTag.FailedSOPSequence,
+            var failedSequence = new DicomSequence(
+                DicomTag.FailedAttributesSequence,
                 warnings);
-            referencedSop.Add(failedSopSequence);
+            referencedSop.Add(failedSequence);
         }
 
         referencedSopSequence.Items.Add(referencedSop);
--- a/src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/DicomStoreServiceTests.cs
+++ b/src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/DicomStoreServiceTests.cs
@@ -194,7 +194,7 @@ public class DicomStoreServiceTests
     }
 
     [Fact]
-    public async Task GivenAValidationError_WhenDropDataEnabled_ThenSucceedsWithErrorsInCommentsSequence()
+    public async Task GivenAValidationError_WhenDropDataEnabled_ThenSucceedsWithErrorsInFailedAttributesSequence()
     {
         // setup
         IDicomInstanceEntry dicomInstanceEntry = Substitute.For<IDicomInstanceEntry>();
@@ -221,13 +221,13 @@ public class DicomStoreServiceTests
         DicomDataset firstInstance = refSopSequence.Items[0];
 
         // expect a comment sequence present
-        DicomSequence failedSOPSequence = firstInstance.GetSequence(DicomTag.FailedSOPSequence);
-        Assert.Single(failedSOPSequence);
+        DicomSequence failedAttributesSequence = firstInstance.GetSequence(DicomTag.FailedAttributesSequence);
+        Assert.Single(failedAttributesSequence);
 
         // expect comment sequence has single warning about single invalid attribute
         Assert.Equal(
             """(0008,0020) - StudyDate - Content "NotAValidStudyDate" does not validate VR DA: one of the date values does not match the pattern YYYYMMDD""",
-            failedSOPSequence.Items[0].GetString(DicomTag.ErrorComment)
+            failedAttributesSequence.Items[0].GetString(DicomTag.ErrorComment)
         );
 
         // expect that what we attempt to store has invalid attrs dropped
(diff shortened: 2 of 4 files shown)


src/Microsoft.Health.Dicom.Core/Features/Store/StoreResponseBuilder.cs
```diff
@@ -17,16 +20,21 @@ namespace Microsoft.Health.Dicom.Core.Features.Store;
...
+    private readonly bool _enableDropInvalidDicomJsonMetadata;
 
     private DicomDataset _dataset;
 
     private string _message;
 
-    public StoreResponseBuilder(IUrlResolver urlResolver)
+    public StoreResponseBuilder(
+        IUrlResolver urlResolver,
+        IOptions<FeatureConfiguration> featureConfiguration
```
bcarthic, line 31: Should the response builder know about feature stuffs? Or should we pass as a parameter?

bcarthic: I don't have strong opinion, but going with the name of the class. It feels like it should just build what the caller tells rather than importing configs. I see both the patterns followed in the code base.

src/Microsoft.Health.Dicom.Core/Features/Store/StoreValidationResultBuilder.cs
```diff
@@ -59,6 +63,15 @@ public void Add(ValidationWarnings warningCode, QueryTag queryTag = null)
         }
     }
 
+    /// <summary>
+    /// Adds a tag to a list representing invalid Dicom items.
+    /// </summary>
+    /// <param name="tag">Invalid item's tag to add.</param>
+    public void Add(DicomTag tag)
```
jovinson-ms, line 70: I don't think it's clear from the name of the method and the param what is happening here - how about changing the name of this method to `AddInvalidTag`?

src/Microsoft.Health.Dicom.Core/Features/Store/StoreService.cs
```diff
@@ -192,14 +202,25 @@ public class StoreService : IStoreService
 
         try
         {
+
+            if (_enableDropInvalidDicomJsonMetadata)
+            {
```
jovinson-ms, line 207: Can we add a metric here to capture the fact that we're dropping tags?

src/Microsoft.Health.Dicom.Core/Features/Store/StoreService.cs
```diff
@@ -192,14 +202,25 @@ public class StoreService : IStoreService
...
+                foreach (DicomTag tag in storeValidatorResult.InvalidTags)
+                {
+                    dicomDataset.Remove(tag);
+                }
+            }
+
             // Store the instance.
             long length = await _storeOrchestrator.StoreDicomInstanceEntryAsync(
                 dicomInstanceEntry,
+                dicomDataset, // this dataset has potentially been modified for metadata storing
```
jovinson-ms, line 218: We shouldn't need this extra param - since DicomDatasets are a reference type, any changes made in this context will be reflected in the object held by `dicomInstanceEntry`.

src/Microsoft.Health.Dicom.Core/Features/Store/IStoreOrchestrator.cs
```diff
@@ -18,10 +19,14 @@ public interface IStoreOrchestrator
     /// Asynchronously orchestrate the storing of a DICOM instance entry.
     /// </summary>
     /// <param name="dicomInstanceEntry">The DICOM instance entry to store.</param>
+    /// <param name="dicomDataset">Dataset to use to store JSON metadata</param>
```
jovinson-ms, line 22: See my comment where this is called in `StoreService` - we shouldn't need this parameter.

src/Microsoft.Health.Dicom.Core/Features/Store/StoreOrchestrator.cs
```diff
@@ -76,8 +78,6 @@ public class StoreOrchestrator : IStoreOrchestrator
 
         var partitionKey = _contextAccessor.RequestContext.GetPartitionKey();
 
-        DicomDataset dicomDataset = await dicomInstanceEntry.GetDicomDatasetAsync(cancellationToken);
```
jovinson-ms, line 79: We can add this back.

src/Microsoft.Health.Dicom.Core/Features/Store/StoreResponseBuilder.cs
jovinson-ms: Agree - it makes sense to just evaluate the feature in the `StoreService`.

src/Microsoft.Health.Dicom.Core/Features/Store/StoreResponseBuilder.cs
```diff
@@ -84,9 +96,28 @@ public void AddSuccess(DicomDataset dicomDataset, ushort? warningReasonCode = nu
...
+            if (warningReasonCode.HasValue)
+            {
+                referencedSop.Add(DicomTag.WarningReason, warningReasonCode.Value);
+            }
+        }
+        else
+        {
+            // add comment Sq / list of warnings here
+
+            var warningList = storeValidationResult.Errors.Select(
```
jovinson-ms, line 110: nit: seems odd to have a variable with `List` in the name which is actually just an array. Could call this variable simply `warnings`.

jovinson-ms (approved):

PR #2271, opened 2023-01-26
Fix MI Durable Task Connections

wsugarman:
## Description
Azure Functions allows for the use of `accountName` to specify the storage account as well as the specification of each storage service using `blobServiceUri`, `queueServiceUri`, and `tableServiceUri`. However, the `AzureComponentFactory` only expects either a connection string (via the section value or `connectionString`) or `serviceUri`. In this change, we add support for directly supplying the per-service URIs or the account name.

## Related issues
N/A

## Testing
Added new unit tests for parsing the configurations

### DIFF
--- a/src/Microsoft.Health.Dicom.Functions.Client/Extensions/AzureComponentFactoryExtensions.cs
+++ b/src/Microsoft.Health.Dicom.Functions.Client/Extensions/AzureComponentFactoryExtensions.cs
@@ -55,7 +55,7 @@ internal static class AzureComponentFactoryExtensions
                 return clientFactory(serviceUri, credential, options);
         }
 
-        return (TClient)factory.CreateClient(typeof(TClient), configuration, null, null);
+        return (TClient)factory.CreateClient(typeof(TClient), configuration, credential, options);
     }
 
     private sealed class StorageServiceUriOptions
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.Dicom.Functions.Client/Extensions/AzureComponentFactoryExtensions.cs
```diff
@@ -12,13 +17,76 @@ namespace Microsoft.Health.Dicom.Functions.Client.Extensions;
 
 internal static class AzureComponentFactoryExtensions
 {
-    public static TClient CreateClient<TOptions, TClient>(this AzureComponentFactory factory, IConfigurationSection configuration)
+    public static BlobServiceClient CreateBlobServiceClient(this AzureComponentFactory factory, IConfigurationSection configuration)
```
smithago, line 20: Un-related to this. I just reviewed the last commit on refactoring. 

Can you help me understand you comment and adding more testing of components?

I tried to reduce the amount of work we do in HealthCheck to improve performance and looks like the last commit is doing more now.

Also, looks like we are catching regressiong too late. Can you add MI tests here.

jnlycklama (approved):

jnlycklama (approved):

smithago (commented):
Will is going to move the logic to Init instead of health check in later PR

arunmk-ms (approved):

PR #2288, opened 2023-02-02
QIDO performance improvements

smithago:
## Description
Revision the API version to 2, to support smaller default QIDO response tags that can be returned by SQL wide tables(Study and Series)
Results will be pulled from blob if more than the default indexed fields are requested.

## Related issues
[AB#98125](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/98125)

## Testing
- UTs

## Remaining
- docs update


src/Microsoft.Health.Dicom.Core/Features/Query/Model/StudyResult.cs
```diff
@@ -61,4 +61,18 @@ public DicomDataset DicomDataset
             return _dicomDataset;
         }
     }
+
+    public static readonly IReadOnlyCollection<DicomTag> AvailableTags = new HashSet<DicomTag>()
```
wsugarman, line 65: I know that Josiah, Esma, Karthik, and I discussed standardizing how we refer to the different tag semantics. @josiahvinson - do you have our notes on what we'd call this scenario? lol

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
```diff
@@ -58,60 +62,129 @@ public class QueryService : IQueryService
...
+        var resultTime = stopwatch.ElapsedMilliseconds;
+        _logger.LogInformation("QueryService performance filterTimeMilliseconds:{FilterTime}, resultTimeMilliseconds:{ResultTime}", filterTime, resultTime);
+
+        var responseMetadata = instanceMetadata.Select(m => responseBuilder.GenerateResponseDataset(m));
+        return new QueryResourceResponse(responseMetadata, queryExpression.ErroneousTags);
+    }
 
-            if (queryExpression.IELevel == Messages.ResourceType.Study)
+    // Does not handle retrieving the extendedQueryTag indexes right now. Logs are in place to evaluate it in the future.
+    private async Task<IEnumerable<DicomDataset>> GetInstanceMetadataAsync(
```
wsugarman, line 90: This method is pretty dense. Could we walk through it offline?

src/Microsoft.Health.Dicom.Core/Features/Query/QueryResponseBuilder.cs
```diff
@@ -139,27 +138,36 @@ public IReadOnlyCollection<DicomTag> ReturnTags
...
+        if (apiMajorVersion != null && apiMajorVersion >= 2)
+        {
+            useNewDefaults = true;
+        }
+
         switch (queryExpression.QueryResource)
         {
-            case QueryResource.AllStudies:
-                _tagsToReturn = new HashSet<DicomTag>(queryExpression.IncludeFields.All ? AllStudyTags : DefaultStudyTags);
+            case (QueryResource.AllStudies):
```
wsugarman, line 154: Were the parentheses intentional?

src/Microsoft.Health.Dicom.Api/Features/Filters/ApiVersionValidatorAttribute.cs
```diff
@@ -0,0 +1,36 @@
...
+
+public sealed class ApiVersionValidatorAttribute : ActionFilterAttribute
+{
+    private readonly bool _isLatestApiVersionEnabled;
+
+    public ApiVersionValidatorAttribute(
+        IOptions<FeatureConfiguration> featureConfiguration)
+    {
+        EnsureArg.IsNotNull(featureConfiguration, nameof(featureConfiguration));
+        _isLatestApiVersionEnabled = featureConfiguration.Value.EnableLatestApiVersion;
```
wsugarman, line 24: Perhaps instead of simply "max," `EnableLatestApiVersion` is instead `LatestApiVersion` and it's a number

src/Microsoft.Health.Dicom.Api/Features/Exceptions/ApiVersionNotSupportedException.cs
```diff
@@ -0,0 +1,17 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Globalization;
+using Microsoft.Health.Dicom.Core.Exceptions;
+
+namespace Microsoft.Health.Dicom.Api.Features.Exceptions;
+
+public class ApiVersionNotSupportedException : BadRequestException
```
wsugarman, line 11: Is it sufficient to instead create a `BadRequestException` instead of creating a new exception type?

src/Microsoft.Health.Dicom.Core/Features/Query/Model/SeriesResult.cs
```diff
@@ -41,4 +42,14 @@ public DicomDataset DicomDataset
             return _dicomDataset;
         }
     }
+
+    public static readonly IReadOnlyCollection<DicomTag> AvailableTags = new HashSet<DicomTag>()
+    {
```
bcarthic, line 47: Instead of having this here, should we consolidate in QueryLimit.cs?

src/Microsoft.Health.Dicom.Api/Features/Filters/ApiVersionValidatorAttribute.cs
smithago: Not exactly, if you have a "preview" txt like we had before

src/Microsoft.Health.Dicom.Api/Features/Exceptions/ApiVersionNotSupportedException.cs
smithago: I prefer strong tying to consolidate the messaging for that error

src/Microsoft.Health.Dicom.Core/Features/Context/IDicomRequestContext.cs
```diff
@@ -23,4 +23,6 @@ public interface IDicomRequestContext : IRequestContext
     int PartCount { get; set; }
 
     PartitionEntry DataPartitionEntry { get; set; }
+
+    int? ApiMajorVersion { get; set; }
```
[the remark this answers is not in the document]
smithago: Perhaps I can change the name to just "version". Your suggestion is interesting. The mapping of the version to what gets enables happens inside the service. Like I did with "UseNewDefaults". Not sure what value the additional abstraction gets outside the service.

src/Microsoft.Health.Dicom.Core/Features/Query/QueryResponseBuilder.cs
smithago: Left over from trying to make the switch on multiple variables. Will clean up

src/Microsoft.Health.Dicom.Core/Features/Query/Model/SeriesResult.cs
smithago: These are the tags available from StudyResult Stored procedure.  Things in the QueryLimit are more like constants

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
smithago: ofcourse, will schedule sometime on Monday

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
```diff
@@ -58,60 +62,139 @@ public class QueryService : IQueryService
...
+
+        var responseMetadata = instanceMetadata.Select(m => responseBuilder.GenerateResponseDataset(m));
+        return new QueryResourceResponse(responseMetadata, queryExpression.ErroneousTags);
+    }
 
-        if (QueryLimit.ContainsComputedTag(queryExpression.IELevel, responseBuilder.ReturnTags))
+    private static bool ReturnNewTagDefaults(int? apiMajorVersion)
+    {
+        bool useNewDefaults = false;
+        if (apiMajorVersion != null && apiMajorVersion >= 2)
```
bcarthic, line 92: Shoudl we move the constants to core and use the same variable?

src/Microsoft.Health.Dicom.Core/Features/Query/Model/StudyResult.cs
smithago: I am not sure of the context. This is just exposing the indexed columns as tags

src/Microsoft.Health.Dicom.Api/Features/Filters/DicomRequestContextRouteDataPopulatingFilterAttribute.cs
```diff
@@ -34,9 +35,12 @@ public sealed class DicomRequestContextRouteDataPopulatingFilterAttribute : Acti
     public override void OnActionExecuting(ActionExecutingContext context)
     {
         EnsureArg.IsNotNull(context, nameof(context));
+
         IDicomRequestContext dicomRequestContext = _dicomRequestContextAccessor.RequestContext;
         dicomRequestContext.RouteName = context.ActionDescriptor?.AttributeRouteInfo?.Name;
 
+        dicomRequestContext.Version = context.HttpContext.GetRequestedApiVersion()?.MajorVersion;
```
bcarthic, line 42: Could we add a test for this scenario `DicomRequestContextDataPopulatingFilterAttributeTests`

src/Microsoft.Health.Dicom.Api/Features/Filters/ApiVersionValidatorAttribute.cs
```diff
@@ -0,0 +1,36 @@
+// -------------------------------------------------------------------------------------------------
```
jovinson-ms, line 1: The issue with using a custom filter and exception is that we break the existing versioning behavior:
- Without the flag enabled, when requesting version 2, the response doesn't include the `api-supported-versions` header as all other responses do. If you request a version that doesn't exist (e.g. 3), the `api-supported-versions` header will return `1.0-prerelease, 1, 2`.
- The new exception message doesn't match the [versioning library output](https://github.com/dotnet/aspnet-api-versioning/blob/5057b06414e5f278f33fcbd22a5cf880bf4e21fb/src/AspNet/WebApi/src/Asp.Versioning.WebApi/Dispatcher/HttpResponseExceptionFactory.cs#L45): 
  ```
    {
      "error": {
          "code": "UnsupportedApiVersion",
          "message": "The HTTP resource that matches the request URI 'https://localhost:63838/v3/studies/2.25.11973375897257723869094853169321604230' does not support the API version '3'."
      }
  }
  ```
The way around this is to remove the version 2 attribute from all controllers and then toggle adding version 2 by convention to all controllers in the [versioning setup](https://github.com/microsoft/dicom-server/blob/1e4e37175a84d23acf6f3758b295e73466d6064b/src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs#L116):
```dotnet
if (dicomServerConfiguration.Features.EnableLatestApiVersion)
{
    c.Conventions.Controller<ChangeFeedController>().HasApiVersion(ApiVersion.Parse("2"));
    // manually repeat for all controllers - I haven't found a simple way to add to all controllers dynamically
}
```
This will use the existing exception response and generate headers correctly.

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
```diff
@@ -58,60 +62,139 @@ public class QueryService : IQueryService
...
+                getSeriesResponse = true;
+            }
+        }
+        // if still remaining response tags, just get full metadata
+        if (remainingTags.Any())
+        {
+            _logger.LogInformation("QueryService tags returned from full metadata {FullMetadataTags}", string.Join(',', remainingTags));
+            getFullMetadata = true;
+            // exception of computed tags
+            if (QueryLimit.ContainsComputedTag(queryExpression.IELevel, returnTags))
```
jovinson-ms, line 132: This block shouldn't be needed, since the `StudyResult` and `SeriesResult` tag sets checked above include the computed tags.

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
```diff
@@ -58,60 +62,139 @@ public class QueryService : IQueryService
...
+
+    // Does not handle retrieving the extendedQueryTag indexes right now. Logs are in place to evaluate it in the future.
+    private async Task<IEnumerable<DicomDataset>> GetInstanceMetadataAsync(
+        int partitionKey,
+        QueryExpression queryExpression,
+        QueryResult queryResult,
+        IReadOnlyCollection<DicomTag> returnTags,
+        CancellationToken cancellationToken)
+    {
+        bool getStudyResponse = false, getSeriesResponse = false, getFullMetadata = false;
```
jovinson-ms, line 107: It might be worth commenting here that `getFullMetadata` is intended to cover any search for metadata at the Instance level, or any search with `includefield=all` at any level.

src/Microsoft.Health.Dicom.Api/Features/Filters/ApiVersionValidatorAttribute.cs
smithago: Good point on the api-supported-versions return header. I looked at conventions and decided to go with filters instead, because it followed the current pattern. 
I think the best mix will be to move all the version settinging into a custom convention and control the new behind a FF

bcarthic:
we need to update URSA scanning to test new version

src/Microsoft.Health.Dicom.Core/Features/Query/QueryResponseBuilder.cs
```diff
@@ -32,7 +32,19 @@ public class QueryResponseBuilder
         DicomTag.StudyID,
     };
 
-    private static readonly HashSet<DicomTag> AllStudyTags = new HashSet<DicomTag>(DefaultStudyTags)
+    internal static readonly HashSet<DicomTag> V2DefaultStudyTags = new HashSet<DicomTag>()
```
jovinson-ms, line 35: Now that we have the computed attributes available, like ModalitiesInStudy, should we include them here and for Series? They are required to be returned, and that could help simplify the logic in the QueryService.

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
```diff
@@ -58,60 +62,139 @@ public class QueryService : IQueryService
...
-        CancellationToken cancellationToken)
-    {
-        if (getFullMetadata)
+            if (instanceMetadata == null)
+            {
+                instanceMetadata = seriesComputedResults.Select(x => x.DicomDataset);
+            }
+            else
+            {
+                var map = seriesComputedResults.ToDictionary<SeriesResult, string>(a => a.StudyInstanceUid + "-" + a.SeriesInstanceUid, StringComparer.OrdinalIgnoreCase);
```
jovinson-ms, line 171: Would be good to add tests for the merge logic.

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
```diff
@@ -58,60 +62,139 @@ public class QueryService : IQueryService
...
+                instanceMetadata = instanceMetadata.Select(x =>
+                {
+                    var ds = new DicomDataset(x);
+                    return ds.AddOrUpdate(map[x.GetSingleValue<string>(DicomTag.StudyInstanceUID) + "-" + x.GetSingleValue<string>(DicomTag.SeriesInstanceUID)].DicomDataset);
```
jovinson-ms, line 175: I'm a little leery of using the `-` character as a separator here - currently we do not allow UIDs to have that character but if we do in the future as part of leniency this logic would break.

src/Microsoft.Health.Dicom.Api/Features/Filters/DicomRequestContextRouteDataPopulatingFilterAttribute.cs
smithago: I need to add an EnE test to validate it. I will add one E2E behind FF

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
smithago: Not sure of the value add.

src/Microsoft.Health.Dicom.Core/Features/Query/Model/SeriesResult.cs
```diff
@@ -41,4 +42,14 @@ public DicomDataset DicomDataset
             return _dicomDataset;
         }
     }
+
+    public static readonly ImmutableHashSet<DicomTag> AvailableTags = ImmutableHashSet.Create<DicomTag>
```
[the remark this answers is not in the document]
smithago: Not exactly. things in the QueryResponseBuilder are the contracts to customer response. This class just represent the tags that are indexed and avaliable from the Index/SQLServer

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
smithago: Same as above

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
smithago: It also includes anything custom specified in includeField. I will add more comments

src/Microsoft.Health.Dicom.Core/Features/Query/QueryResponseBuilder.cs
smithago: I want to keep the default to minimum and fastest and they can always add what they want in includeField

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
smithago: the separator character does not matter here.

src/Microsoft.Health.Dicom.Core/Features/Query/Model/StudyResult.cs
```diff
@@ -61,4 +61,18 @@ public DicomDataset DicomDataset
             return _dicomDataset;
         }
     }
+
+    public static readonly ImmutableHashSet<DicomTag> AvailableTags = ImmutableHashSet.Create<DicomTag>
```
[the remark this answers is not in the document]
smithago: same here

src/Microsoft.Health.Dicom.Api/Features/Filters/DicomRequestContextRouteDataPopulatingFilterAttribute.cs
smithago: > 

URSA needs to be added after the FF is enabled

src/Microsoft.Health.Dicom.Core/Features/Query/Model/SeriesResult.cs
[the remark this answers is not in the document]
jovinson-ms: Yes - my point is that we already have classes that define those sets: `QueryLimit` for tags that are indexed by default, and `QueryResponseBuilder` for the API response contract. Why not use the existing definitions of available tags in `QueryLimit`?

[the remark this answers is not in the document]
smithago: Not the same. Separation of concerns. 
QueryLimit: Contains things we support during filtering.
QueryResponseBuilder: Contains the tags that are returned by default.
StudyResult/SeriesResult: Contains the tags that are available through SQL

All these can independently change. What is available in SQL should not be tied to what the default tags are returned

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
wsugarman: Can we instead use my `DicomIdentifier` struct here instead? I implemented `GetHashCode` and `Equals` so it should be ok to use in a dictionary.

swagger/v1-prerelease/swagger.yaml
```diff
@@ -2052,6 +2052,57 @@ paths:
             application/json:
               schema:
                 type: string
+  /_schema/versions:
```
smithago, line 2055: SchemaController comes from healthcare-Shared component and does not have versions set. So it was using the config to use default which is V1. This is a good thing now we are explicitly putting a version on it and can deprecate v1

bcarthic:
I know we talked lot about this versioning, but I have a question around where we include the new changes that we are doing to the controller, should we add it to v1 or v2?

For e.g the new /rendered  endpoint should it work in v1 or v2? Don't we then add new version of the controller like StoreV2Controller to explicitly say it works in V2? Some features need a flag, or the feature needs to be reverted in case of issue with a customer so in that case, is all the features that are added under v2 gets blocked?

src/Microsoft.Health.Dicom.Core/Features/Query/Model/SeriesResult.cs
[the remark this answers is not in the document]
jovinson-ms: Ok - I would like to make sure I understand the distinction you're making here between 
> QueryLimit: Contains things we support during filtering.

and 

>StudyResult/SeriesResult: Contains the tags that are available through SQL

To me, those sound like the same thing - if we add a tag to be supported during filtering, it's by definition present in the SQL table, and vice versa.

Are you saying that the set of tags in StudyResult/SeriesResult model what is returned by the GetStudyResult / GetSeriesResult sprocs, and we want that to be able to independently change from what tags are stored in the tables?

swagger/v1-prerelease/swagger.yaml
jovinson-ms: Looks like it's still versionless in v2, like health/check.

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
```diff
@@ -58,60 +63,136 @@ public class QueryService : IQueryService
...
-        CancellationToken cancellationToken)
-    {
-        if (getFullMetadata)
+            if (instanceMetadata == null)
+            {
+                instanceMetadata = seriesComputedResults.Select(x => x.DicomDataset);
+            }
+            else
+            {
+                Dictionary<DicomIdentifier, SeriesResult> map = seriesComputedResults.ToDictionary<SeriesResult, DicomIdentifier>(a => new DicomIdentifier(a.StudyInstanceUid, a.SeriesInstanceUid, default));
```
wsugarman, line 169: Ah, you could also use a simple tuple too like `(string, string)` for the key, since you don't really need the instance part of the identifier here.

src/Microsoft.Health.Dicom.Core/Features/Query/Model/SeriesResult.cs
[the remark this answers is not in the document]
smithago: Yes. There are computed columns that I put in StudyResultComputedTags that together with CoreFilterTags that form the Result. The StudyResult class just represents the SQL columns, since the property is just in the class it will be easy to add it there.

swagger/v1-prerelease/swagger.yaml
jovinson-ms: Checked with @brandonpollett, we should be able to remove this now - perhaps we should do this as part of v2

smithago: It will have V2. I have not generated the swagger for V2 because it is behind FF. I have Feature 100502 to track all the work to make V2 public

src/Microsoft.Health.Dicom.Api/Features/Conventions/ApiVersionsConvention.cs
```diff
@@ -0,0 +1,93 @@
...
+                .Select(a => ((IntroducedInApiVersionAttribute)a).Version)
+                .SingleOrDefault();
+
+        var versions = _allSupportedVersions;
+        if (controllerIntroducedInVersion != null)
+        {
+            versions = GetAllSupportedVersions(controllerIntroducedInVersion.Value, _currentVersion);
+        }
+        // when upcomingVersion is ready for GA, move upcomingVerion to allSupportedVersion and remove this logic
+        versions = _isLatestApiVersionEnabled == true ? versions.Union(_upcomingVersion).ToList() : versions;
```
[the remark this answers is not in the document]
smithago: The types are different, one is ApiVersion and another is int. I can do the same ApiVersion.Parse. I would keep it explicit for now

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
smithago: I will keep the DicomIdentifier for now, considering it has GetHashCode Implemented.

src/Microsoft.Health.Dicom.Core/Features/Query/QueryService.cs
```diff
@@ -58,60 +63,136 @@ public class QueryService : IQueryService
...
+        QueryResult queryResult,
+        IReadOnlyCollection<DicomTag> returnTags,
+        CancellationToken cancellationToken)
+    {
+        bool getStudyResponse = false, getSeriesResponse = false, getFullMetadata = false;
+
+        IEnumerable<DicomTag> currentResponseTags = returnTags;
+        // are the expected responses available in StudyResult
+        var remainingTags = currentResponseTags.Except(StudyResult.AvailableTags).ToList();
+        if (currentResponseTags.Count() > remainingTags.Count)
```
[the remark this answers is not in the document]
smithago: I like it. Thanks for re0imagining it.

src/Microsoft.Health.Dicom.Api/Features/Conventions/IntroducedInApiVersionAttribute.cs
```diff
@@ -0,0 +1,23 @@
...
+
+/// <summary>
+/// Represents the starting API version the controller class was introduced.
+/// Don't use any attribute if you want to introduce the controller in all available versions.
+/// Did not use ApiVersionAttribute as base because it is a collection of versions and we need just one version here.
+/// </summary>
+[AttributeUsage(AttributeTargets.Class)]
+internal sealed class IntroducedInApiVersionAttribute : Attribute
+{
+    public int? Version { get; init; }
```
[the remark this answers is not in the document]
smithago: That could be one way, I was thinking of separate attribute to keep them orthogonal.

jovinson-ms (approved):

wsugarman (approved):

PR #2297, opened 2023-02-02
Fix STOW failing on SOPInstanceUID index

smithago:
## Description
Indexes on SOPInstanceUID and SeriesInstanceUID should not be unique.
All the 3 UIDs together are unique

## Related issues
[AB#100268](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/100268)

## Testing
Added new tests to catch this


src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/30.diff.sql
```diff
@@ -0,0 +1,41 @@
+SET XACT_ABORT ON
```
brandonpollett, line 1: Do we need to do an `IF EXISTS (DELETE/ALTER) INDEX` on the problematic indexes?

smithago: Yes, handled by DROP_EXISTING=ON

brandonpollett: `DROP_EXISTING=ON` allows for online updates

brandonpollett (approved):

PR #2300, opened 2023-02-03
Fix migration script

bcarthic:
## Description
Fix migration script to unblock deployment. 

- Updated the V28 migration script to drop index which are creating while retrying.

## Related issues
Addresses [issue #].

## Testing

### DIFF
--- a/src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/28.diff.sql
+++ b/src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/28.diff.sql
@@ -413,6 +413,8 @@ BEGIN
     )
     WITH (DROP_EXISTING=ON, ONLINE=ON)
 
+    DROP INDEX IF EXISTS IX_Instance_PartitionKey_Status_StudyInstanceUid_SeriesInstanceUid_SopInstanceUid on dbo.Instance
+
     CREATE NONCLUSTERED INDEX IX_Instance_PartitionKey_Status_StudyInstanceUid_SeriesInstanceUid_SopInstanceUid on dbo.Instance
     (
         PartitionKey,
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/28.diff.sql
```diff
@@ -429,6 +431,8 @@ BEGIN
     )
     WITH (DATA_COMPRESSION = PAGE, ONLINE=ON)
```
brandonpollett, line 432: We should be able to accomplish the same thing with the hint `DROP_EXISTING=ON`

bcarthic: I think it won't work if the index is not created yet

brandonpollett: Actually, it's probably not going to work the way I thought it would. Supposedly it doesn't work on an initial creation.

bcarthic: Yes

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/28.diff.sql
```diff
@@ -413,6 +413,8 @@ BEGIN
     )
     WITH (DROP_EXISTING=ON, ONLINE=ON)
 
+    DROP INDEX IF EXISTS IX_Instance_PartitionKey_Status_StudyInstanceUid_SeriesInstanceUid_SopInstanceUid on dbo.Instance
```
jovinson-ms, line 416: Don't we also need to drop the old indexes?

smithago (approved):

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/28.diff.sql
bcarthic: The problem is these indexes are already created and it failed on the UNIQUE index  creation.

PR #2322, opened 2023-02-09
Updating System.drawing.common to latest version

poadhika:
## Description
Upgrading nugets to latest version for functions.
Most of the libraries specified in the s360 vurenlability  has already been updated

## Related issues
Addresses [bug #100513](https://microsofthealth.visualstudio.com/Health/_boards/board/t/Medical%20Imaging/Stories/?workitem=100513)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,5 +1,4 @@
 <Project>
-
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


wsugarman (changes requested):
Let's sync offline about this

poadhika:
> Let's sync offline about this

Sure. Actually paas is already using 7.0.0.

PR #2323, opened 2023-02-10
leniency set status code and warningReasonCode on response

esmadau:
## Description
set status code to accepted when using leniency and set warningReasonCode

## Related issues
Addresses [[AB#100139](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/100139)].

## Testing
Tests updated and some new tests added

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Features/Store/StoreResponseBuilder.cs
+++ b/src/Microsoft.Health.Dicom.Core/Features/Store/StoreResponseBuilder.cs
@@ -76,8 +76,7 @@ public class StoreResponseBuilder : IStoreResponseBuilder
     {
         DicomSequence referencedSOPSequence = _dataset.GetSequence(DicomTag.ReferencedSOPSequence);
         return referencedSOPSequence
-            .Select(ds => ds.TryGetSequence(DicomTag.FailedAttributesSequence, out _) == true)
-            .Any();
+            .Any(ds => ds.TryGetSequence(DicomTag.FailedAttributesSequence, out _) == true);
     }
 
     /// <inheritdoc />
--- a/src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/StoreResponseBuilderTests.cs
+++ b/src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/StoreResponseBuilderTests.cs
@@ -97,7 +97,7 @@ public class StoreResponseBuilderTests
 
         StoreResponse response = _storeResponseBuilder.BuildResponse(null);
 
-        Assert.Equal(StoreResponseStatus.Success, response.Status);
+        Assert.Equal(StoreResponseStatus.PartialSuccess, response.Status);
         Assert.Single(response.Dataset);
 
         DicomSequence refSopSequence = response.Dataset.GetSequence(DicomTag.ReferencedSOPSequence);
@@ -137,7 +137,7 @@ public class StoreResponseBuilderTests
 
         StoreResponse response = _storeResponseBuilder.BuildResponse(null);
 
-        Assert.Equal(StoreResponseStatus.Success, response.Status);
+        Assert.Equal(StoreResponseStatus.PartialSuccess, response.Status);
         Assert.Single(response.Dataset);
 
         DicomSequence refSopSequence = response.Dataset.GetSequence(DicomTag.ReferencedSOPSequence);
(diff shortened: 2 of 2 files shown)


src/Microsoft.Health.Dicom.Core/Features/Store/StoreService.cs
```diff
@@ -184,19 +184,10 @@ TelemetryClient telemetryClient
...
-                    dicomDataset.Remove(tag);
-
-                    string message = storeValidatorResult.InvalidTagErrors[tag].Error;
-                    _telemetryClient.ForwardLogTrace(
-                        $"{message}. This attribute will not be present when retrieving study, series, or instance metadata resources, nor can it be used in searches." +
-                        "However, it will still be present when retrieving study, series, or instance resources.",
-                        identifier);
-                }
+                // set warning code if none set yet
+                warningReasonCode ??= WarningReasonCodes.DatasetHasValidationWarnings;
```
bcarthic, line 190: In case of _enableDropInvalidDicomJsonMetadata == true, we are not using this warning code in the response builder. It is only set when the flag is off. Also only when the warning code is IndexedDicomTagHasMultipleValues we set the warning header. Do we want to set the header if there is error in the attribute too?

src/Microsoft.Health.Dicom.Core/Features/Store/StoreService.cs
```diff
@@ -138,7 +138,7 @@ TelemetryClient telemetryClient
             }
         }
 
-        return _storeResponseBuilder.BuildResponse(requiredStudyInstanceUid);
+        return _storeResponseBuilder.BuildResponse(requiredStudyInstanceUid, _enableDropInvalidDicomJsonMetadata);
```
bcarthic, line 141: Instead of passing the feature flag status, we can determine the leniency applied or not here and depends on that the response builder can set the status

src/Microsoft.Health.Dicom.Core/Features/Store/StoreResponseBuilder.cs
```diff
@@ -64,6 +73,14 @@ public StoreResponse BuildResponse(string studyInstanceUid)
         return new StoreResponse(status, _dataset, _message);
     }
 
+    private bool LeniencyApplied()
+    {
+        DicomSequence referencedSOPSequence = _dataset.GetSequence(DicomTag.ReferencedSOPSequence);
+        return referencedSOPSequence
+            .Select(ds => ds.TryGetSequence(DicomTag.FailedAttributesSequence, out _) == true)
+            .Any();
```
bcarthic, line 81: If  storeValidationResult.InvalidTagErrors has any items then we can set to partial success right?

src/Microsoft.Health.Dicom.Core/Features/Store/WarningReasonCodes.cs
```diff
@@ -19,4 +19,12 @@ internal static class WarningReasonCodes
     /// </remarks>
     public const ushort DatasetDoesNotMatchSOPClass = 45063;
 
+    /// <summary>
+    /// Data Set has validation warnings
+    /// </summary>
+    /// <remarks>
+    /// The Studies Store Transaction (Section 10.5) observed that the Data Set has validation warnings.
+    /// </remarks>
+    public const ushort DatasetHasValidationWarnings = 1;
```
jovinson-ms, line 28: We should add this warning code to the conformance statement for V2. Right now, the STOW warning codes are spread out in text describing STOW - we may want to consolidate into a table like error codes.

src/Microsoft.Health.Dicom.Core/Features/Store/StoreService.cs
```diff
@@ -258,6 +249,22 @@ TelemetryClient telemetryClient
...
+        var identifier = dicomDataset.ToInstanceIdentifier();
+        foreach (DicomTag tag in storeValidatorResult.InvalidTagErrors.Keys)
+        {
+            // drop invalid metadata
+            dicomDataset.Remove(tag);
+
+            string message = storeValidatorResult.InvalidTagErrors[tag].Error;
+            _telemetryClient.ForwardLogTrace(
+                $"{message}. This attribute will not be present when retrieving study, series, or instance metadata resources, nor can it be used in searches." +
+                "However, it will still be present when retrieving study, series, or instance resources.",
```
jovinson-ms, line 263:
```suggestion
                " However, it will still be present when retrieving study, series, or instance resources.",
```

src/Microsoft.Health.Dicom.Core/Features/Store/StoreResponseBuilder.cs
jovinson-ms: agree - we shouldn't need to evaluate the feature flag because previous versions will just skip this.

PR #2324, opened 2023-02-10
doc changes for QIDO v2 in conformance

smithago:
## Description
Create a new V2 conformance statement

## Related issues
[AB#98125](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/98125)

## Testing
NA

### DIFF
--- a/docs/resources/v2-conformance-statement.md
+++ b/docs/resources/v2-conformance-statement.md
@@ -479,7 +479,7 @@ The query API will return one of the following status codes in the response:
 - Matching is case in-sensitive and accent in-sensitive for PN VR types.
 - Matching is case in-sensitive and accent sensitive for other string VR types.
 - Only the first value will be indexed of a single valued data element that incorrectly has multiple values.
-- For best performance default attributes or limited number of responses is recommended.
+- Using the default attributes or limiting the number of results requested will maximize performance.
 
 ## Delete
 
--- a/docs/resources/v2-conformance-statement.md
+++ b/docs/resources/v2-conformance-statement.md
@@ -1,4 +1,6 @@
-# DICOM Conformance Statement
+# Version 2 DICOM Conformance Statement
+
+> API version 2 is in **Preview** and should be used only for testing.
 
 The **Medical Imaging Server for DICOM** supports a subset of the DICOMweb&trade; Standard. Support includes:
 
@@ -370,47 +372,29 @@ The response will be an array of DICOM datasets. Depending on the resource, by *
 
 | Tag          | Attribute Name |
 | :----------- | :------------- |
-| (0008, 0005) | SpecificCharacterSet |
 | (0008, 0020) | StudyDate |
-| (0008, 0030) | StudyTime |
 | (0008, 0050) | AccessionNumber |
-| (0008, 0056) | InstanceAvailability |
+| (0008, 1030) | StudyDescription |
 | (0009, 0090) | ReferringPhysicianName |
-| (0008, 0201) | TimezoneOffsetFromUTC |
 | (0010, 0010) | PatientName |
 | (0010, 0020) | PatientID |
 | (0010, 0030) | PatientBirthDate |
-| (0010, 0040) | PatientSex |
-| (0020, 0010) | StudyID |
 | (0020, 000D) | StudyInstanceUID |
 
 #### Default Series tags
 
 | Tag          | Attribute Name |
 | :----------- | :------------- |
-| (0008, 0005) | SpecificCharacterSet |
 | (0008, 0060) | Modality |
-| (0008, 0201) | TimezoneOffsetFromUTC |
-| (0008, 103E) | SeriesDescription |
+| (0008, 1090) | ManufacturerModelName |
 | (0020, 000E) | SeriesInstanceUID |
 | (0040, 0244) | PerformedProcedureStepStartDate |
-| (0040, 0245) | PerformedProcedureStepStartTime |
-| (0040, 0275) | RequestAttributesSequence |
 
 #### Default Instance tags
 
 | Tag          | Attribute Name |
 | :----------- | :------------- |
-| (0008, 0005) | SpecificCharacterSet |
-| (0008, 0016) | SOPClassUID |
 | (0008, 0018) | SOPInstanceUID |
-| (0008, 0056) | InstanceAvailability |
-| (0008, 0201) | TimezoneOffsetFromUTC |
-| (0020, 0013) | InstanceNumber |
-| (0028, 0010) | Rows |
-| (0028, 0011) | Columns |
-| (0028, 0100) | BitsAllocated |
-| (0028, 0008) | NumberOfFrames |
 
 If `includefield=all`, below attributes are included along with default attributes. Along with default attributes, this is the full list of attributes supported at each resource level.
 
@@ -418,7 +402,10 @@ If `includefield=all`, below attributes are included along with default attribut
 
 | Tag          | Attribute Name |
 | :----------- | :------------- |
-| (0008, 1030) | Study Description |
+| (0008, 0005) | SpecificCharacterSet |
+| (0008, 0030) | StudyTime |
+| (0008, 0056) | InstanceAvailability |
+| (0008, 0201) | TimezoneOffsetFromUTC |
 | (0008, 0063) | AnatomicRegionsInStudyCodeSequence |
 | (0008, 1032) | ProcedureCodeSequence |
 | (0008, 1060) | NameOfPhysiciansReadingStudy |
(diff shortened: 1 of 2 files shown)


docs/resources/v2-conformance-statement.md
```diff
@@ -0,0 +1,818 @@
+# Version 2 DICOM Conformance Statement
+
+> API version 2 is in **Preview** and should be used only for testing.
```
mmitrik, line 3: Does this mean that that version is `/v2.0-prerelease`?

docs/resources/v2-conformance-statement.md
```diff
@@ -0,0 +1,818 @@
...
+### Store Failure Reason Codes
+
+| Code  | Description |
+| :---- | :---------- |
+| 272   | The store transaction did not store the instance because of a general failure in processing the operation. |
+| 43264 | The DICOM instance failed the validation. |
+| 43265 | The provided instance StudyInstanceUID did not match the specified StudyInstanceUID in the store request. |
+| 45070 | A DICOM instance with the same StudyInstanceUID, SeriesInstanceUID and SopInstanceUID has already been stored. If you wish to update the contents, delete this instance first. |
+| 45071 | A DICOM instance is being created by another process, or the previous attempt to create has failed and the cleanup process has not had chance to clean up yet. Please delete the instance first before attempting to create again. |
+| 45063 | A DICOM instance Data Set does not match SOP Class. The Studies Store Transaction (Section 10.5) observed that the Data Set did not match the constraints of the SOP Class during storage of the instance. |
```
mmitrik, line 185: Looks like this was moved from warnings to failure reasons, is that correct?

docs/resources/v2-conformance-statement.md
```diff
@@ -0,0 +1,818 @@
...
+| {dicomKeyword}   | StudyInstanceUID |
+
+Example query searching for instances:
+`../instances?Modality=CT&00280011=512&includefield=00280010&limit=5&offset=0`
+
+### Search Response
+
+The response will be an array of DICOM datasets. Depending on the resource, by *default* the following attributes are returned:
+
+#### Default Study tags
```
mmitrik, line 371: Were there intentional changes here?  This table is much smaller than the [published V1](https://learn.microsoft.com/en-us/azure/healthcare-apis/dicom/dicom-services-conformance-statement#default-study-tags) - I noted these are missing:

| Tag          | Attribute Name |
| :----------- | :------------- |
| (0008, 0005) | `SpecificCharacterSet` |
| (0008, 0030) | `StudyTime` |
| (0008, 0056) | `InstanceAvailability` |
| (0008, 0201) | `TimezoneOffsetFromUTC` |
| (0010, 0040) | `PatientSex` |
| (0020, 0010) | `StudyID` |

And this appears to be new in V2 (or missing in V1)

| Tag          | Attribute Name |
| :----------- | :------------- |
| (0008, 1030) | StudyDescription |

docs/resources/v2-conformance-statement.md
```diff
@@ -0,0 +1,818 @@
...
+| (0008, 0020) | StudyDate |
+| (0008, 0050) | AccessionNumber |
+| (0008, 1030) | StudyDescription |
+| (0009, 0090) | ReferringPhysicianName |
+| (0010, 0010) | PatientName |
+| (0010, 0020) | PatientID |
+| (0010, 0030) | PatientBirthDate |
+| (0020, 000D) | StudyInstanceUID |
+
+#### Default Series tags
```
mmitrik, line 384: Looks like changes here

docs/resources/v2-conformance-statement.md
```diff
@@ -0,0 +1,818 @@
...
+
+| Tag          | Attribute Name |
+| :----------- | :------------- |
+| (0008, 0060) | Modality |
+| (0008, 1090) | ManufacturerModelName |
+| (0020, 000E) | SeriesInstanceUID |
+| (0040, 0244) | PerformedProcedureStepStartDate |
+
+#### Default Instance tags
```
mmitrik, line 393: Another with changes

docs/resources/v2-conformance-statement.md
```diff
@@ -0,0 +1,818 @@
...
+
+| Tag          | Attribute Name |
+| :----------- | :------------- |
+| (0008, 0018) | SOPInstanceUID |
+
+If `includefield=all`, below attributes are included along with default attributes. Along with default attributes, this is the full list of attributes supported at each resource level.
+
+#### Additional Study tags
```
mmitrik, line 401: Above missing items appear to be moved here, which tells me that this was likely intentional, not an omission.  Is that right?

docs/resources/v2-conformance-statement.md
```diff
@@ -0,0 +1,818 @@
...
+
+| Tag          | Attribute Name |
+| :----------- | :------------- |
+| (0008, 0005) | SpecificCharacterSet |
+| (0008, 0201) | TimezoneOffsetFromUTC |
+| (0020, 0011) | SeriesNumber |
+| (0020, 0060) | Laterality |
+| (0008, 0021) | SeriesDate |
+| (0008, 0031) | SeriesTime |
+| (0008, 103E) | SeriesDescription |
```
mmitrik, line 432: How do we decide what is in the default vs. additional tags?  For example, we include `StudyDescription` in the study by default, but `SeriesDescription` is not included with the series.

docs/resources/v2-conformance-statement.md
```diff
@@ -0,0 +1,818 @@
...
+### Delete Response Payload
+
+The response body will be empty. The status code is the only useful information returned.
+
+# Worklist Service (UPS-RS)
+
+The DICOM service supports the Push and Pull SOPs of the [Worklist Service (UPS-RS)](https://dicom.nema.org/medical/dicom/current/output/html/part18.html#chapter_11). This service provides access to one Worklist containing Workitems, each of which represents a Unified Procedure Step (UPS).
+
+Throughout, the variable `{workitem}` in a URI template stands for a Workitem UID.
+
```
mmitrik, line 521: There is an entire table missing here - Available UPS-RS endpoints.  See https://learn.microsoft.com/en-us/azure/healthcare-apis/dicom/dicom-services-conformance-statement#worklist-service-ups-rs

mmitrik (commented):
Overall, it wasn't very clear to me what the scope of changes were for the conformance statement.  From the title and linked work item, I presume the changes to the default attributes were related to the perf changes (i.e. returning fewer things by default to improve response times) - but I'm really making a guess here.

I also spot checked some other areas and noted omissions in the Worklist section - was that intentional?  Or incidental?

Finally, there are a number of formatting inconsistencies in the tables compared to the Azure docs.  Most notable is that error codes and attribute names are styled with code tags in Azure, but not here.  It might be worth using the V1 conformance statement from the public docs as the V2 starting point and then applying the new changes to that document.  This is probably easier than changing the styling on every table in this document.  Having consistent styles will make future Azure updates simpler and less error prone.

docs/resources/v2-conformance-statement.md
smithago: currently it is /v2/ behind a Feature flag. They need to turn on a config setting to use it.

docs/resources/v2-conformance-statement.md
smithago: hmm. I did not remove UPS-RS. Since I created a new file, it is hard to see the diff. Best way is to download the file and compare with your local conformance file.

I see the diff in only in default reduction and adding them to All response.

But your assumption is right, we are reducing the default to make it more performant.

docs/resources/v2-conformance-statement.md
smithago: No, I did not.

docs/resources/v2-conformance-statement.md
```diff
@@ -0,0 +1,818 @@
+# Version 2 DICOM Conformance Statement
```
bcarthic, line 1: Is our plan to delete the v1 conformance after we enable v2 for all? Or follow the mandatory 6 months until then we update both.

docs/resources/v2-conformance-statement.md
smithago: Yes, reduce the default. Sorry, there was a Code change PR with more details and since this was a follow up did not put a lot in description. Will keep this in mind next time

docs/resources/v2-conformance-statement.md
smithago: yes

docs/resources/v2-conformance-statement.md
smithago: yes

docs/resources/v2-conformance-statement.md
smithago: Things we index by defaults stayed in as defaults

docs/resources/v2-conformance-statement.md
smithago: I think preview is 6 months. Released versions have a different lifecycle. There is no huge cost to supporting both. We may have to wait for another release before we start to deprecate V1. I will learn more as I onboard to Azure Advisory

docs/resources/v2-conformance-statement.md
smithago: Thanks for the feedback Esma, I am looking at this PR as incremental change targeting the change I made to V2 QIDO only. Rest needs to be tracked separately

docs/resources/v2-conformance-statement.md
```diff
@@ -0,0 +1,818 @@
...
+### Additional Notes
+
+- Querying using the `TimezoneOffsetFromUTC` (`00080201`) is not supported.
+- The query API will not return 413 (request entity too large). If the requested query response limit is outside of the acceptable range, a bad request will be returned. Anything requested within the acceptable range, will be resolved.
+- When target resource is Study/Series there is a potential for inconsistent study/series level metadata across multiple instances. For example, two instances could have different patientName. In this case latest will win and you can search only on the latest data.
+- Paged results are optimized to return matched *newest* instance first, this may result in duplicate records in subsequent pages if newer data matching the query was added.
+- Matching is case in-sensitive and accent in-sensitive for PN VR types.
+- Matching is case in-sensitive and accent sensitive for other string VR types.
+- Only the first value will be indexed of a single valued data element that incorrectly has multiple values.
+- For best performance default attributes or limited number of responses is recommended.
```
mmitrik, line 482: I suggest a slight re-wording for clarity:

```suggestion
- Using the default attributes or limiting the number of additional attributes requested will maximize performance.
```

mmitrik (dismissed):
Here is a diff I created for the v1 and v2: https://gist.github.com/mmitrik/1bec21e02f2c6eb3d59511fc85d58540/revisions

Changes look good, I had one minor suggestion.

mmitrik (approved):

PR #2343, opened 2023-02-14
add leniency to v2 api, remove feature flag

esmadau:
## Description
Folds leniency feature under the V2 API and removes feature flag.

## Related issues
Addresses [[AB#99848](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/99848)].

## Testing
Added unit tests and modified existing tests. Also tested with locally running server with version flag on and off and using v1 and v2.

### DIFF
--- a/test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/StoreTransactionTests.cs
+++ b/test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/StoreTransactionTests.cs
@@ -49,8 +49,8 @@ public class StoreTransactionTests : IClassFixture<HttpIntegrationTestFixture<St
         DicomWebException exception =
             await Assert.ThrowsAsync<DicomWebException>(() => _clientV2.StoreAsync(dicomFile1));
 
-        Assert.Equal(
-            """BadRequest: {"error":{"code":"UnsupportedApiVersion","message":"The HTTP resource that matches the request URI 'http://localhost/v2/studies' does not support the API version '2'."}}""",
+        Assert.Contains(
+            """BadRequest: {"error":{"code":"UnsupportedApiVersion""",
             exception.Message);
     }
 


src/Microsoft.Health.Dicom.Core/Features/Store/StoreService.cs
```diff
@@ -259,4 +259,8 @@ private async Task DisposeResourceAsync(int index)
             LogFailedToDisposeDelegate(_logger, index, ex);
         }
     }
+    private static bool EnableDropMetadata(int? version)
```
brandonpollett, line 262: nit: missing empty line before

docker/docker-compose.features.yml
```diff
@@ -4,4 +4,4 @@ services:
   dicomserver:
     environment:
       DicomServer__Features__EnableDataPartitions: "true"
-      DicomServer__Features__EnableDropInvalidDicomJsonMetadata: "true"
+      DicomServer__Features__EnableLatestApiVersion: "true"
```
brandonpollett, line 7: I realize this isn't your naming convention, but the `EnableLatestApiVersion` doesn't sound very exact. Especially because the `DicomApiVersion.cs` has a const called `Latest` which is set to `v1`.

src/Microsoft.Health.Dicom.Core/Features/Store/StoreDatasetValidator.cs
```diff
@@ -196,4 +199,8 @@ string EnsureRequiredTagIsPresent(DicomTag dicomTag)
             }
         }
     }
+    private static bool EnableDropMetadata(int? version)
```
brandonpollett, line 202: nit: missing new line.

test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/DropInvalidMetadataTests.cs
```diff
@@ -27,7 +27,7 @@ public class DropInvalidMetadataTests : IClassFixture<EnableDropInvalidDicomJson
     public DropInvalidMetadataTests(EnableDropInvalidDicomJsonMetadataHttpIntegrationTestFixture<Startup> fixture)
     {
         EnsureArg.IsNotNull(fixture, nameof(fixture));
-        _client = fixture.GetDicomWebClient();
+        _client = fixture.GetDicomWebClient(DicomApiVersions.V2);
```
brandonpollett, line 30: Do we need a test that checks if the client supports `v2` but the client sends `v1` that the expected behavior happens? Also, I assume we have a test somewhere that does the opposite (e.g. makes sure if you use `v2` with it disabled, that you get a failure?

brandonpollett (approved):

test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/StoreTransactionTests.cs
```diff
@@ -25,17 +25,31 @@ namespace Microsoft.Health.Dicom.Web.Tests.E2E.Rest;
...
         _recyclableMemoryStreamManager = fixture.RecyclableMemoryStreamManager;
     }
 
+    [Fact]
+    public async Task GivenV2NotEnabled_WhenAttemptingToUseV2_TheServerShouldReturnConflict()
+    {
+        var studyInstanceUID1 = TestUidGenerator.Generate();
+        DicomFile dicomFile1 = Samples.CreateRandomDicomFile(studyInstanceUid: studyInstanceUID1);
+
+        DicomWebException exception =
```
brandonpollett, line 49: Might want to assert that the exception is what we expect, or we can just throw away the response.

test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/DropInvalidMetadataTests.cs
```diff
@@ -21,17 +21,42 @@ namespace Microsoft.Health.Dicom.Web.Tests.E2E.Rest;
...
+
+        DicomDataset dicomDataset = new DicomDataset().NotValidated();
+
+        dicomDataset.Add(DicomTag.StudyDate, "NotAValidStudyDate");
+        dicomDataset.Add(DicomTag.PatientBirthDate, "20220315");
+
+        dicomFile.Dataset.Add(dicomDataset);
+
+        // run
+        await Assert.ThrowsAsync<DicomWebException>(() => _instancesManagerV1.StoreAsync(
```
brandonpollett, line 53: Same here.

brandonpollett (approved):

docker/docker-compose.features.yml
brandonpollett: Sounds good to me. I think also it allows us to have a cleaner path to cleanup once `v2` is enabled because there's no question on the use of the flag.

brandonpollett (dismissed):

brandonpollett (approved):

PR #2378, opened 2023-02-17
Fix startup issue

arunmk-ms:
## Description
DicomCast was failing at startup due to missing DicomCastMeter and AppInsights configuration inside service container. 

## Related issues
Addresses [issue [100846](https://microsofthealth.visualstudio.com/DefaultCollection/Health/_workitems/edit/100846)].

## Testing
Ran tests manually and verified the startup

### DIFF
--- a/converter/dicom-cast/src/Microsoft.Health.DicomCast.Hosting/Program.cs
+++ b/converter/dicom-cast/src/Microsoft.Health.DicomCast.Hosting/Program.cs
@@ -11,6 +11,8 @@ using Azure.Security.KeyVault.Secrets;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using Microsoft.Health.DicomCast.Core.Features.Worker;
 using Microsoft.Health.DicomCast.Core.Modules;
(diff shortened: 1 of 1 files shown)


converter/dicom-cast/src/Microsoft.Health.DicomCast.Hosting/Program.cs
```diff
@@ -82,6 +85,11 @@ private static void AddOpenTelemetryMetrics(IServiceCollection services, string
     /// </summary>
     private static void AddApplicationInsightsTelemetry(IServiceCollection services, string connectionString)
     {
-        services.AddApplicationInsightsTelemetry(aiOptions => aiOptions.ConnectionString = connectionString);
```
wsugarman, line 85: Why doesn't this work for us?

converter/dicom-cast/src/Microsoft.Health.DicomCast.Hosting/Program.cs
```diff
@@ -82,6 +85,11 @@ private static void AddOpenTelemetryMetrics(IServiceCollection services, string
...
+        services.AddApplicationInsightsTelemetryWorkerService(aiOptions => aiOptions.ConnectionString = connectionString);
```
wsugarman, line 88: Did the previous method not actually log?

wsugarman (approved):

PR #2398, opened 2023-02-22
Add APIVersion column to telemetry

smithago:
## Description
Extend ITelemetryInitializer to add a new custom column

## Related issues
[AB#100795](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/100795)

## Testing
Existing tests pass

### DIFF
--- a/src/Microsoft.Health.Dicom.Api/Logging/TelemetryInitializer.cs
+++ b/src/Microsoft.Health.Dicom.Api/Logging/TelemetryInitializer.cs
@@ -1,0 +1,49 @@
+// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using EnsureThat;
+using Microsoft.ApplicationInsights.Channel;
+using Microsoft.ApplicationInsights.DataContracts;
+using Microsoft.ApplicationInsights.Extensibility;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Microsoft.Health.Dicom.Api.Logging;
+
+/// <summary>
+/// Class extends AppInsights telemtry to include custom properties
+/// </summary>
+public class TelemetryInitializer : ITelemetryInitializer
+{
+    private readonly IHttpContextAccessor _httpContextAccessor;
+    private const string ApiVersionColumnName = "ApiVersion";
+
+    public TelemetryInitializer(IHttpContextAccessor httpContextAccessor)
+    {
+        _httpContextAccessor = EnsureArg.IsNotNull(httpContextAccessor, nameof(httpContextAccessor));
+    }
+
+    public void Initialize(ITelemetry telemetry)
+    {
+        AddApiVersionColumn(telemetry);
+    }
+
+    private void AddApiVersionColumn(ITelemetry telemetry)
+    {
+        var requestTelemetry = telemetry as RequestTelemetry;
+        if (requestTelemetry == null)
+        {
+            return;
+        }
+
+        string version = _httpContextAccessor.HttpContext?.GetRequestedApiVersion()?.ToString();
+        if (version == null)
+        {
+            return;
+        }
+
+        requestTelemetry.Properties[ApiVersionColumnName] = version;
+    }
+}
--- a/src/Microsoft.Health.Dicom.Api/Logging/TelemetryInitializer.cs
+++ b/src/Microsoft.Health.Dicom.Api/Logging/TelemetryInitializer.cs
@@ -15,7 +15,7 @@ namespace Microsoft.Health.Dicom.Api.Logging;
 /// <summary>
 /// Class extends AppInsights telemtry to include custom properties
 /// </summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Dicom.Api/Logging/TelemetryInitializer.cs
```diff
@@ -0,0 +1,49 @@
...
+using Microsoft.ApplicationInsights.Extensibility;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Microsoft.Health.Dicom.Api.Logging;
+
+/// <summary>
+/// Class extends AppInsights telemtry to include custom properties
+/// </summary>
+public class TelemetryInitializer : ITelemetryInitializer
```
wsugarman, line 18: I think we can keep this `internal`

wsugarman, line 18: Maybe a more specific name lol. Perhaps `[Api]VersionTelemetryInitializer`?

src/Microsoft.Health.Dicom.Api/Logging/TelemetryInitializer.cs
```diff
@@ -0,0 +1,49 @@
...
+    private const string ApiVersionColumnName = "ApiVersion";
+
+    public TelemetryInitializer(IHttpContextAccessor httpContextAccessor)
+    {
+        _httpContextAccessor = EnsureArg.IsNotNull(httpContextAccessor, nameof(httpContextAccessor));
+    }
+
+    public void Initialize(ITelemetry telemetry)
+    {
+        AddApiVersionColumn(telemetry);
```
wsugarman, line 30: Do we need this helper method?

src/Microsoft.Health.Dicom.Api/Logging/TelemetryInitializer.cs
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    private void AddApiVersionColumn(ITelemetry telemetry)
+    {
+        var requestTelemetry = telemetry as RequestTelemetry;
+        if (requestTelemetry == null)
```
wsugarman, line 36:
```suggestion
        if (telemetry is not RequestTelemetry requestTelemetry)
```

src/Microsoft.Health.Dicom.Api/Logging/TelemetryInitializer.cs
```diff
@@ -0,0 +1,49 @@
...
+            return;
+        }
+
+        string version = _httpContextAccessor.HttpContext?.GetRequestedApiVersion()?.ToString();
+        if (version == null)
+        {
+            return;
+        }
+
+        requestTelemetry.Properties[ApiVersionColumnName] = version;
```
wsugarman, line 47: Perhaps we can always assign, like `requestTelemetry.Properties[ApiVersionColumnName] = _httpContextAccessor.HttpContext?.GetRequestedApiVersion()?.ToString()`. Then this method can be simplified to "if the type is request telemetry, then assign the optional version"

src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
```diff
@@ -146,6 +148,8 @@ public static IDicomServerBuilder AddHostedServices(this IDicomServerBuilder ser
 
         services.TryAddSingleton<RecyclableMemoryStreamManager>();
 
+        services.AddSingleton<ITelemetryInitializer, TelemetryInitializer>();
```
wsugarman, line 151: Perhaps move this up next to the `AddApiVersioning` call

src/Microsoft.Health.Dicom.Api/Logging/TelemetryInitializer.cs
smithago: Keep it generic so that the same class can be used to extend other properties

src/Microsoft.Health.Dicom.Api/Logging/TelemetryInitializer.cs
smithago: Better readability and also I am thinking to reuse the class in the future for other things

src/Microsoft.Health.Dicom.Api/Logging/TelemetryInitializer.cs
smithago: We need the strongly types object, so I would prefer to do it once and check the conversion

src/Microsoft.Health.Dicom.Api/Logging/TelemetryInitializer.cs
smithago: I am assuming the thing I have will reduce the workload sent up when the value is null.

src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
smithago: I am looking at it as a generic telemetry extension which happens to do APIVersion column now.

wsugarman (approved):

PR #2404, opened 2023-02-23
Declare dependencies for docker-compose

jovinson-ms:
## Description
The docker-compose project requires the latest DicomWeb and Functions projects in order to build containers. When this dependency isn't declared, rebuilding the solution can fail on docker-compose.


wsugarman (approved):
Oh, good catch!

PR #2417, opened 2023-02-25
Support accept type with no quotes

smithago:
## Description
Asp.net core/RFC expects request header accept to have quotes in the type like
multipart/related; type="application/octet-stream"; transfer-syntax=*

The new OHIF viewer version and the DICOM standard supports below
multipart/related; type=application/octet-stream; transfer-syntax=*

Put in a hack to handle this in the conversion code.

## Related issues
Addresses [issue #].

## Testing
Added a unit test

### DIFF
--- a/src/Microsoft.Health.Dicom.Api/Extensions/MediaTypeHeaderValueExtensions.cs
+++ b/src/Microsoft.Health.Dicom.Api/Extensions/MediaTypeHeaderValueExtensions.cs
@@ -37,7 +37,9 @@ public static class MediaTypeHeaderValueExtensions
 
         bool isMultipartRelated = StringSegment.Equals(KnownContentTypes.MultipartRelated, mediaType, StringComparison.OrdinalIgnoreCase);
         // handle accept type with no quotes like "multipart/related; type=application/octet-stream; transfer-syntax=*"
-        // RFC 2045 is clear that any content type parameter value must be quoted if it contains at least one special character. However, the DICOMweb standard incorrectly omits quotes, so we need to compensate for it in our server
+        // RFC 2045 is clear that any content type parameter value must be quoted if it contains at least one special character. 
+        // However, RFC 2387 which defines `multipart/related` specifies in its ABNF definition of the `type` parameter that quotes are not allowed, although all examples include quotes (Errata 5048). 
+        // The DICOMweb standard currently requires quotes, but will soon (CP 1776) allow both forms, so we will allow both.
         bool? startsWithMultiPart = mediaType.Buffer?.StartsWith(KnownContentTypes.MultipartRelated, StringComparison.OrdinalIgnoreCase);
         if (isMultipartRelated)
         {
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.Dicom.Api/Extensions/MediaTypeHeaderValueExtensions.cs
```diff
@@ -34,11 +34,18 @@ public static AcceptHeader ToAcceptHeader(this MediaTypeHeaderValue headerValue)
     {
         EnsureArg.IsNotNull(headerValue, nameof(headerValue));
         StringSegment mediaType = headerValue.MediaType;
+
         bool isMultipartRelated = StringSegment.Equals(KnownContentTypes.MultipartRelated, mediaType, StringComparison.OrdinalIgnoreCase);
+        // handle accept type with no quotes like "multipart/related; type=application/octet-stream; transfer-syntax=*"
+        bool? startsWithMultiPart = mediaType.Buffer?.StartsWith(KnownContentTypes.MultipartRelated, StringComparison.OrdinalIgnoreCase);
```
wsugarman, line 40: The root of the problem is that the DICOMweb standard allows for the use of invalid content types. Should we instead attempt to fix those incoming content types so that downstream components don't need to worry? We could modify the header value with a regex that looks for this scenario, like:

```csharp
// As per RFC 2045, tokens may be any ASCII character except for space, control characters, and a set of special characters.
// [!#-:<-~]* - This is any printable ASCII character except for ' ', '"', and ';'
// [(),/:<=>?@[\\\\\\]] - This is the set of "special characters" that must be quoted, except for ';' and '"'
private static readonly Regex InvalidTypeParameterRegex = new Regex(
    "(\\s+|;)type=([!#-:<-~]*[(),/:<=>?@[\\\\\\]][!#-:<-~]*)( |;|$)",
    RegexOptions.IgnoreCase | RegexOptions.Compiled | RegexOptions.CultureInvariant);

// ...

string validString = invalidTypeParameter.Replace(invalidContentType, x => $"{x.Groups[1]}type=\"{x.Groups[2]}\"{x.Groups[3]}");
```

smithago: If there is a value add sure. Otherwise the hack is more readable and maintainable. We are letting ASP.Net to do best effort parsing and pining with a unit test if ASP.net changes

src/Microsoft.Health.Dicom.Api/Extensions/MediaTypeHeaderValueExtensions.cs
```diff
@@ -34,11 +34,19 @@ public static AcceptHeader ToAcceptHeader(this MediaTypeHeaderValue headerValue)
     {
         EnsureArg.IsNotNull(headerValue, nameof(headerValue));
         StringSegment mediaType = headerValue.MediaType;
+
         bool isMultipartRelated = StringSegment.Equals(KnownContentTypes.MultipartRelated, mediaType, StringComparison.OrdinalIgnoreCase);
+        // handle accept type with no quotes like "multipart/related; type=application/octet-stream; transfer-syntax=*"
+        // RFC 2045 is clear that any content type parameter value must be quoted if it contains at least one special character. However, the DICOMweb standard incorrectly omits quotes, so we need to compensate for it in our server
```
jovinson-ms, line 40:
```suggestion
        // RFC 2045 is clear that any content type parameter value must be quoted if it contains at least one special character. 
        // However, RFC 2387 which defines `multipart/related` specifies in its ABNF definition of the `type` parameter that quotes are not allowed, although all examples include quotes (Errata 5048). 
        // The DICOMweb standard currently requires quotes, but will soon (CP 1776) allow both forms, so we will allow both.
```

jovinson-ms (commented):
Looks good - just suggested some added context

jovinson-ms (approved):

PR #2431, opened 2023-03-02
add leniency to api v2 conformance statement

esmadau:
## Description
add leniency to api v2 conformance statement

## Related issues
Addresses [[AB#100991](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/100991)].

## Testing
N/A

### DIFF
--- a/docs/resources/v2-conformance-statement.md
+++ b/docs/resources/v2-conformance-statement.md
@@ -79,7 +79,7 @@ Each file stored must have a unique combination of StudyInstanceUID, SeriesInsta
 > Requests are limited to 2GB. No single DICOM file or combination of files may exceed this limit.
 
 ### Store Changes From V1
-A major change from V1 Store API is that attributes will continue to be validated, but only validation errors with [required attributes](#store-required-attributes) would cause Store to fail.
+In previous versions, a Store request would fail if any of the [required](#store-required-attributes) or [searchable attributes](#searchable-attributes) failed validation. Beginning with V2, the request will only fail if **required attributes** fail validation.
 
 Failed validation of attributes not required by the API will still result in the file being stored and a warning will be given about each failing attribute per instance.
 When a sequence contains an attribute that fails validation, or when there are multiple issues with a single attribute, only the first failing attribute reason will be noted.


docs/resources/v2-conformance-statement.md
```diff
@@ -75,21 +75,21 @@ The following DICOM elements are required to be present in every DICOM file atte
...
-| 503 (Service Unavailable)    | The service is unavailable or busy. Please try again later. |
+| Code                         | Description                                                                                                                                                                                                                         |
+| :--------------------------- |:------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| 200 (OK)                     | All the SOP instances in the request have been stored.                                                                                                                                                                              |
+| 202 (Accepted)               | The origin server stored some of the Instances and failures exist for others. Or origin server has warnings for Instances stored. Additional information regarding this error may be found in the response message body. |
```
mmitrik, line 85: I think we can tighten this up a little: 

```suggestion
| 202 (Accepted)               | The origin server stored some of the Instances and others have failed or returned warnings. Additional information regarding this error may be found in the response message body. |
```

docs/resources/v2-conformance-statement.md
```diff
@@ -173,6 +174,70 @@ An example response with `Accept` header `application/dicom+json`:
...
+            }
+          }
+        ]
+      }
+    }]
+  }
+}
+```
+
+In the case that leniency is applied and the file is stored, you will be notified with each failing attribute per instance. In case of a failing sequence or multiple issues of a single attribute, only the first failing attribute reason will be noted.
```
mmitrik, line 238: This bit about "leniency is applied" made me question - "what do they mean by leniency?" since it is not mentioned before this in the conformance statement.  

I wonder if we should rephrase this to talk about validation.  Maybe something like:

```suggestion
For attributes that are not required but are searchable by default, validation will be performed.  Failed validation of these attributes will still result in the file being stored and a warning will be given about each failing attribute per instance. In the case of a failing sequence or multiple issues of a single attribute, only the first failing attribute reason will be noted.
```

docs/resources/v2-conformance-statement.md
```diff
@@ -173,6 +174,70 @@ An example response with `Accept` header `application/dicom+json`:
...
+We will also output diagnostic logs per failed attribute per instance.
```
mmitrik, line 239: I think we should avoid using "we" here.  I know "we" is used in other places, but it would be a good practice start changing.  

```suggestion
Diagnostic logs will also list failed attributes per instance.
```

docs/resources/v2-conformance-statement.md
```diff
@@ -182,7 +247,13 @@ An example response with `Accept` header `application/dicom+json`:
...
 | 45071 | A DICOM instance is being created by another process, or the previous attempt to create has failed and the cleanup process has not had chance to clean up yet. Please delete the instance first before attempting to create again. |
-| 45063 | A DICOM instance Data Set does not match SOP Class. The Studies Store Transaction (Section 10.5) observed that the Data Set did not match the constraints of the SOP Class during storage of the instance. |
+
+### Store Warning Reason Codes
+
+| Code  | Description                                                                                                                                            |
+|:------|:-------------------------------------------------------------------------------------------------------------------------------------------------------|
+| 45063 | The Studies Store Transaction (Section 10.5) observed that the Data Set did not match the constraints of the SOP Class during storage of the instance. |
+| 1     | The Studies Store Transaction (Section 10.5) observed that the Data Set has validation warnings.                                                       |
```
mmitrik, line 256: This having a value of `1` surprised me.  Is that because there is no suitable code defined in the standard?  I see that this page only lists 3: https://dicom.nema.org/medical/dicom/current/output/chtml/part18/sect_i.2.html

docs/resources/v2-conformance-statement.md
```diff
@@ -309,6 +380,8 @@ The following `Accept` header(s) are supported for searching:
 
 - `application/dicom+json`
 
+Note that any instances stored with leniency applied to indexed attributes may not be retrievable by attributes which failed validation.
```
mmitrik, line 383: Similar to the above comment, the reader might not know what we mean by "leniency" in this context.  Being more explicit here might be helpful: 

```suggestion
Note if an instance returned validation warnings for searchable attributes at the time the instance was stored, those attributes may not be retrievable via search.
```

mmitrik (commented):
Overall looks good.  I made some suggestions for where I think can improve clarity.

docs/resources/v2-conformance-statement.md
```diff
@@ -61,7 +61,7 @@ The following `Content-Type` header(s) are supported:
 - `multipart/related; type="application/dicom"`
 - `application/dicom`
 
-> Note: The Server will <u>not</u> coerce or replace attributes that conflict with existing data. All data will be stored as provided.
+> Note: The Server will <u>not</u> coerce or replace existing data on the DICOM binary. The DICOM binary will be stored as provided.
```
jovinson-ms, line 64: Looking through the standard, `DICOM file` is probably the best term to use - and adding the `PS 3.10` part makes it extra clear. https://dicom.nema.org/medical/dicom/current/output/html/part18.html#sect_10.5.1.4

```suggestion
> Note: the Server will <u>not</u> coerce or replace data in the DICOM PS 3.10 file. The DICOM file will be stored as provided, except where otherwise noted.
```

docs/resources/v2-conformance-statement.md
```diff
@@ -75,21 +75,21 @@ The following DICOM elements are required to be present in every DICOM file atte
 
 Each file stored must have a unique combination of StudyInstanceUID, SeriesInstanceUID and SopInstanceUID. The warning code `45070` will be returned if a file with the same identifiers already exists.
 
-> DICOM File Size Limit: there is a size limit of 2GB for a DICOM file by default.
+> DICOM File Size Limit: there is a size limit of 2GB for request size, thus limiting the size of the DICOM file.
```
jovinson-ms, line 78:
```suggestion
> Requests are limited to 2GB. No single DICOM file or combination of files may exceed this limit.
```

docs/resources/v2-conformance-statement.md
jovinson-ms: I think we should have a whole heading discussing leniency on STOW since it's a major change from v1.

mmitrik: I agree with Josiah - it would be good to just have a section dedicated to this change and explain the nuances in the behavior.

docs/resources/v2-conformance-statement.md
```diff
@@ -75,21 +76,27 @@ The following DICOM elements are required to be present in every DICOM file atte
 
 Each file stored must have a unique combination of StudyInstanceUID, SeriesInstanceUID and SopInstanceUID. The warning code `45070` will be returned if a file with the same identifiers already exists.
 
-> DICOM File Size Limit: there is a size limit of 2GB for a DICOM file by default.
+> Requests are limited to 2GB. No single DICOM file or combination of files may exceed this limit.
+
+### Store Changes From V1
+A major change from V1 Store API is that attributes will continue to be validated, but only validation errors with [required attributes](#store-required-attributes) would cause for Store to fail.
```
mmitrik, line 82: Minor suggestion: 

```suggestion
A major change from V1 Store API is that attributes will continue to be validated, but only validation errors with [required attributes](#store-required-attributes) would cause Store to fail.
```

docs/resources/v2-conformance-statement.md
```diff
@@ -75,21 +76,27 @@ The following DICOM elements are required to be present in every DICOM file atte
 
 Each file stored must have a unique combination of StudyInstanceUID, SeriesInstanceUID and SopInstanceUID. The warning code `45070` will be returned if a file with the same identifiers already exists.
 
-> DICOM File Size Limit: there is a size limit of 2GB for a DICOM file by default.
+> Requests are limited to 2GB. No single DICOM file or combination of files may exceed this limit.
+
+### Store Changes From V1
```
mmitrik, line 81: Good addition 😄 Provides a lot of clarity about the changes.

docs/resources/v2-conformance-statement.md
```diff
@@ -309,6 +384,10 @@ The following `Accept` header(s) are supported for searching:
 
 - `application/dicom+json`
 
+### Search Changes From V1
+Note if an instance returned validation warnings for [searchable attributes](#searchable-attributes) at the time the [instance was stored](#store-changes-from-v1), those attributes may not be used to search for the stored instance.
```
mmitrik, line 388: Minor suggestion:
```suggestion
If an instance returned validation warnings for [searchable attributes](#searchable-attributes) at the time the [instance was stored](#store-changes-from-v1), those attributes may not be used to search for the stored instance.
```

mmitrik (approved):
Made a few minor suggestions to improve clarity/readability.  Overall the changes look great 😄

docs/resources/v2-conformance-statement.md
```diff
@@ -75,21 +76,27 @@ The following DICOM elements are required to be present in every DICOM file atte
...
 Each file stored must have a unique combination of StudyInstanceUID, SeriesInstanceUID and SopInstanceUID. The warning code `45070` will be returned if a file with the same identifiers already exists.
 
-> DICOM File Size Limit: there is a size limit of 2GB for a DICOM file by default.
+> Requests are limited to 2GB. No single DICOM file or combination of files may exceed this limit.
+
+### Store Changes From V1
+A major change from V1 Store API is that attributes will continue to be validated, but only validation errors with [required attributes](#store-required-attributes) would cause for Store to fail.
+
+Failed validation of attributes not required by the API will still result in the file being stored and a warning will be given about each failing attribute per instance.
+In the case of a failing sequence or multiple issues of a single attribute, only the first failing attribute reason will be noted.
```
jovinson-ms, line 85:
```suggestion
When a sequence contains an attribute that fails validation, or when there are multiple issues with a single attribute, only the first failing attribute reason will be noted.
```

jovinson-ms (approved):

mmitrik (approved):

PR #2478, opened 2023-03-13
Set response with accept status on warnings

esmadau:
## Description
Today, our response builder does not take directly into consideration if there's a warningCode set from this list:
[I.2 Store Instances Response Attribute Description (nema.org)](https://dicom.nema.org/medical/dicom/current/output/chtml/part18/sect_I.2.html#sect_I.2.1)

Based on Store Transaction documentation in the standard:
<html>
<body>

Status | Code | Description
-- | -- | --
Success | 200 (OK) | The origin server successfully stored all Instances.
202 (Accepted) | The origin server stored some of the Instances but warnings or failures exist for others.Additional information regarding this error may be found in the response message body.

</body>
</html>

We should always set to 202 any time there's even a warning, but no failures. Today, we set it to 200 even when there's warnings unless leniency was used. As this is a status code change, these changes should only be applied when api V2 is enabled.

Conformance statement V2 has already been updated to reflect these changes.
## Related issues
Addresses [[AB#100566](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/100566)].

## Testing
updated unit tests.

### DIFF
--- a/src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/StoreResponseBuilderTests.cs
+++ b/src/Microsoft.Health.Dicom.Core.UnitTests/Features/Store/StoreResponseBuilderTests.cs
@@ -70,7 +70,7 @@ public class StoreResponseBuilderTests
     }
 
     [Fact]
-    public void GivenOnlySuccessEntry_WhenApiV2EnabledAndHasWarningCode_ThenExpectPartialSuccessReturned()
+    public void GivenOnlySuccessEntry_WhenBuildWarningSequenceEnabledAndHasWarningCode_ThenExpectPartialSuccessReturned()
     {
         _storeResponseBuilder.AddSuccess(
             _dicomDataset1,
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.Dicom.Core/Features/Store/StoreService.cs
```diff
@@ -267,7 +268,7 @@ private async Task DisposeResourceAsync(int index)
         }
     }
 
-    private static bool EnableDropMetadata(int? version)
+    private static bool IsApiV2Enabled(int? version)
```
jovinson-ms, line 271: We may want to clarify this language - the feature is something like `ReturnWarning202` and the condition is it will be valid for all API versions greater than 1. Most of the naming is focused on the condition instead of the feature.

jovinson-ms (dismissed):

abdullah248 (approved):

PR #2489, opened 2023-03-15
Handle core tag validations issues as errors in V2

esmadau:
## Description
- When v2 and we are using full validation, we want to treat core tags validation failures as errors instead of just warnings

## Related issues
Addresses [[AB#101508](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/101508)].

## Testing
Added test to both v1 and v2 store

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Features/Store/StoreDatasetValidator.cs
+++ b/src/Microsoft.Health.Dicom.Core/Features/Store/StoreDatasetValidator.cs
@@ -38,6 +38,7 @@ public class StoreDatasetValidator : IStoreDatasetValidator
         DicomTag.SeriesInstanceUID,
         DicomTag.SOPInstanceUID,
         DicomTag.PatientID,
+        DicomTag.SOPClassUID,
     };
 
     public StoreDatasetValidator(


src/Microsoft.Health.Dicom.Core/Features/Store/StoreDatasetValidator.cs
```diff
@@ -32,6 +32,14 @@ public class StoreDatasetValidator : IStoreDatasetValidator
     private readonly StoreMeter _storeMeter;
     private readonly IDicomRequestContextAccessor _dicomRequestContextAccessor;
 
+    private static readonly HashSet<DicomTag> RequiredCoreTags = new HashSet<DicomTag>()
+    {
+        DicomTag.StudyInstanceUID,
+        DicomTag.SeriesInstanceUID,
+        DicomTag.SOPInstanceUID,
+        DicomTag.PatientID,
```
jovinson-ms, line 40: We'll want to add `SOPClassUID` to this list. Need an action item on deciding whether it should be required in V2 (isn't indexed).

jovinson-ms (dismissed):

jovinson-ms (approved):

PR #2519, opened 2023-03-29
Migrate Sample ARM Template to use Linux Web Apps

wsugarman:
## Description
- Use linux web app for sample ARM template... mostly for fun

See [here](https://learn.microsoft.com/en-us/azure/azure-functions/functions-infrastructure-as-code) for details on deploying functions and web apps

## Related issues
N/A

## Testing
Provisioned ad-hoc


bcarthic:
@wsugarman Will this break our CI environment? Since our app service is already created and published as a windows service, do we need to create a new CI environment?

jovinson-ms:
@wsugarman is this alive?

wsugarman:
Yes, I just need to test the transcoding cases

PR #2587, opened 2023-04-17
DICOM update durable function

bcarthic:
## Description
This PR includes changes to Azure durable function to update dicom instances for a study

## Related issues
Addresses [[AB#102008](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/102008)].

## Testing
New tests and E2E are added


src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Orchestration.cs
```diff
@@ -0,0 +1,149 @@
...
+                        _options.RetryOptions,
+                        new BatchUpdateArguments(input.PartitionKey, instanceWatermarks, input.ChangeDataset));
+            }
+
+            if (instanceWatermarks.Count > 0)
+            {
+                try
+                {
+                    await context.CallActivityWithRetryAsync(
+                        nameof(UpdateInstanceBatchAsync),
```
wsugarman, line 81: Should these methods be called `UpdateStudyAsync` and `CompleteUpdateStudyAsync`?

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Activity.cs
```diff
@@ -0,0 +1,235 @@
...
+
+        DicomDataset datasetToUpdate = GetDeserialzedDataset(arguments.ChangeDataset);
+
+        int processed = 0;
+
+        logger.LogInformation("Beginning to update all instance blobs, Total count {TotalCount}", arguments.InstanceWatermarks.Count);
+
+        while (processed < arguments.InstanceWatermarks.Count)
+        {
+            int batchSize = Math.Min(_options.BatchSize, arguments.InstanceWatermarks.Count - processed);
```
wsugarman, line 123: Oh, I see you're using the batch here. This is a bit different from the "fan out" pattern the other functions use -- hence my confusion. The "fan out" refers to the multiple activities that are spawn from a single orchestration to process data in parallel.

I think you can make this sort of batch setting a function-specific option from the config. The reason the other functions get batching options from the input as they have an input on the number of activities that are generated, and as such any change could potentially screw up the orchestration restoration.

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Activity.cs
```diff
@@ -0,0 +1,235 @@
...
+            var batch = arguments.InstanceWatermarks.Skip(processed).Take(batchSize).ToList();
```
wsugarman, line 124: It's unfortunate this isn't an extension method, but perhaps we can use something like [`List<T>.GetRange(int, int)`](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.getrange?view=net-8.0) or even make our own extension.

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Activity.cs
```diff
@@ -0,0 +1,235 @@
...
+                    batchSize);
+
+                await Parallel.ForEachAsync(
+                    batch,
+                    new ParallelOptions
+                    {
+                        CancellationToken = default,
+                        MaxDegreeOfParallelism = _options.MaxParallelThreads,
+                    },
+                    async (instance, token) =>
```
wsugarman, line 140: nit: you don't need `async` and `await:
```suggestion
                    (instance, token) => _updateInstanceService.UpdateInstanceBlobAsync(instance, datasetToUpdate, token)
```

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Activity.cs
```diff
@@ -0,0 +1,235 @@
...
+                    {
+                        await _updateInstanceService.UpdateInstanceBlobAsync(instance, datasetToUpdate, token);
+                    });
+
+                logger.LogInformation("Completed updating instance blobs starting with {StartingRange} and {EndingRange}. Total batchSize {BatchSize}.",
```
wsugarman, line 145:
```suggestion
                logger.LogInformation("Completed updating instance blobs starting for range [{Start}, {End}]. Total batch size {BatchSize}.",
```

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Activity.cs
```diff
@@ -0,0 +1,235 @@
...
+        logger.LogInformation("Beginning to update all instance blobs, Total count {TotalCount}", arguments.InstanceWatermarks.Count);
+
+        while (processed < arguments.InstanceWatermarks.Count)
+        {
+            int batchSize = Math.Min(_options.BatchSize, arguments.InstanceWatermarks.Count - processed);
+            var batch = arguments.InstanceWatermarks.Skip(processed).Take(batchSize).ToList();
+
+            if (batch.Count > 0)
+            {
+                logger.LogInformation("Beginning to update instance blobs starting with {StartingRange} and {EndingRange}. Total batchSize {BatchSize}.",
```
wsugarman, line 128:
```suggestion
                logger.LogInformation("Beginning to update instance blobs for range [{Start}, {End}]. Total batch size {BatchSize}.",
```

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Activity.cs
```diff
@@ -0,0 +1,235 @@
...
+    public async Task DeleteOldVersionBlobAsync([ActivityTrigger] IDurableActivityContext context, ILogger logger)
+    {
+        EnsureArg.IsNotNull(context, nameof(context));
+        EnsureArg.IsNotNull(logger, nameof(logger));
+
+        IReadOnlyList<InstanceFileIdentifier> fileIdentifiers = context.GetInput<IReadOnlyList<InstanceFileIdentifier>>();
+
+        logger.LogInformation("Begin deleting old blobs. Total size {TotalCount}", fileIdentifiers.Count);
+
+        await Parallel.ForEachAsync(
```
wsugarman, line 216: nit: you can filter the identifiers to avoid unnecessary threads:
```csharp
        await Parallel.ForEachAsync(
            fileIdentifiers.Where(f => f.OriginalVersion.HasValue),
            new ParallelOptions
            {
                CancellationToken = default,
                MaxDegreeOfParallelism = _options.MaxParallelThreads,
            },
            (fileIdentifier, token) => _updateInstanceService.DeleteInstanceBlobAsync(fileIdentifier.Version, token),
            });
```

src/Microsoft.Health.Dicom.Functions/Update/UpdateOptions.cs
```diff
@@ -0,0 +1,41 @@
...
+
+/// <summary>
+/// Represents the options for a "update" function.
+/// </summary>
+public class UpdateOptions
+{
+    internal const string SectionName = "Update";
+
+    /// <summary>
+    /// Gets or sets the number of DICOM instances processed by a single activity.
```
wsugarman, line 20: This comment is necessarily true -- these batches are within a single activity. So perhaps we should change the description and names to avoid ambiguity with the other functions that use these names

src/Microsoft.Health.Dicom.Functions/Update/UpdateOptions.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// Gets or sets the number of threads available for each batch.
+    /// </summary>
+    [Range(-1, int.MaxValue)]
+    public int MaxParallelThreads { get; set; } = -1;
+
+    /// <summary>
+    /// Gets or sets the maximum number of concurrent batches processed at a given time.
+    /// </summary>
+    [Range(1, int.MaxValue)]
+    public int MaxParallelBatches { get; set; } = 10;
```
wsugarman, line 35: Do you use this?

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -75,6 +77,79 @@ public class BlobFileStore : IFileStore
...
+    {
+        EnsureArg.IsNotNull(stream, nameof(stream));
+        EnsureArg.IsNotNull(blockLengths, nameof(blockLengths));
+        EnsureArg.IsGte(blockLengths.Count, 0, nameof(blockLengths.Count));
+
+        BlockBlobClient blobClient = GetInstanceBlockBlobClient(version);
+
+        try
+        {
+            stream.Seek(0, SeekOrigin.Begin);
```
wsugarman, line 95: I would say it's not up for the method to seek the `Stream`; the caller should do this. There's no guarantee a `Stream` is seekable

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -75,6 +77,79 @@ public class BlobFileStore : IFileStore
...
+
+            foreach (string blockId in blockLengths.Keys)
+            {
+                long blockSize = blockLengths[blockId];
+                byte[] blockData = new byte[blockSize];
```
wsugarman, line 100: Instead of creating an array, simply rent an array from an [`ArrayPool<T>`](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.arraypool-1?view=net-7.0). Then you can read into the buffer and create a stream from it.

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -75,6 +77,79 @@ public class BlobFileStore : IFileStore
...
+        EnsureArg.IsNotNull(stream, nameof(stream));
+        EnsureArg.IsNotNullOrWhiteSpace(blockId, nameof(blockId));
+
+        BlockBlobClient blobClient = GetInstanceBlockBlobClient(version);
+
+        BlockList blockList = await blobClient.GetBlockListAsync(BlockListTypes.Committed, snapshot: null, conditions: null, cancellationToken);
+
+        IEnumerable<string> blockIds = blockList.CommittedBlocks.Select(x => x.Name);
+
+        string blockToUpdate = blockIds.Where(x => x.Equals(blockId, StringComparison.OrdinalIgnoreCase)).First();
```
wsugarman, line 134: nit: combine with the line above and perhaps use `FirstOrDefault()`

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -75,6 +77,79 @@ public class BlobFileStore : IFileStore
...
+
+        if (blockToUpdate == null)
+            throw new ItemNotFoundException(null);
```
wsugarman, line 137: No message?

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -177,6 +254,54 @@ public async Task<Stream> GetFileFrameAsync(long version, FrameRange range, Canc
         return stream;
     }
 
+    /// <inheritdoc />
+    public async Task<BinaryData> GetFileInRangeAsync(long version, FrameRange range, CancellationToken cancellationToken)
```
wsugarman, line 258: Where is this used?

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -177,6 +254,54 @@ public async Task<Stream> GetFileFrameAsync(long version, FrameRange range, Canc
...
+
+    /// <inheritdoc />
+    public async Task<KeyValuePair<string, long>> GetFirstBlockPropertyAsync(long version, CancellationToken cancellationToken = default)
+    {
+        BlockBlobClient blobClient = GetInstanceBlockBlobClient(version);
+
+        BlockList blockList = await blobClient.GetBlockListAsync(BlockListTypes.Committed, snapshot: null, conditions: null, cancellationToken);
+
+        if (!blockList.CommittedBlocks.Any())
+            throw new ItemNotFoundException(null);
```
wsugarman, line 287: Message?

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobMetadataStore.cs
```diff
@@ -117,12 +117,12 @@ public async Task DeleteInstanceMetadataIfExistsAsync(long version, Cancellation
...
 
     /// <inheritdoc />
-    public Task<DicomDataset> GetInstanceMetadataAsync(long version, CancellationToken cancellationToken)
+    public async Task<DicomDataset> GetInstanceMetadataAsync(long version, CancellationToken cancellationToken)
     {
         try
         {
             BlockBlobClient blobClient = GetInstanceBlockBlobClient(version);
-            return ExecuteAsync(async t =>
+            return await ExecuteAsync(async t =>
```
wsugarman, line 125: Good catch!

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
bcarthic: There seemed to be pattern in other method, we need to change later

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
bcarthic: This is used UpdateInstanceService to get first block content

src/Microsoft.Health.Dicom.Core/Features/Update/UpdateInstanceService.cs
```diff
@@ -0,0 +1,193 @@
...
+            long blockSize = i == 0 ? initialBlockLength : Math.Min(StageBlockSizeInBytes, streamLength - bytesRead);
+            string blockId = Convert.ToBase64String(Guid.NewGuid().ToByteArray());
+            blockLengths.Add(blockId, blockSize);
+            bytesRead += blockSize;
+        }
+
+        return blockLengths;
+    }
+
+    private static void RemoveItemsFromDataset(DicomDataset dataset, DicomItem largeDicomItem)
```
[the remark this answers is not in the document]
bcarthic: Yes, renamed to RemoveItemsAfter. Whether pixeldata or non pixel data, if it is more than a certain length we would like to remove so that we can push the first set of items in the dataset to first block

src/Microsoft.Health.Dicom.Functions/Update/UpdateOptions.cs
bcarthic: I just used in backfill, I can remove those

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Orchestration.cs
```diff
@@ -0,0 +1,149 @@
...
+                    await context.CallActivityWithRetryAsync(
+                        nameof(CompleteUpdateInstanceAsync),
+                        _options.RetryOptions,
+                        new CompleteInstanceArguments(input.PartitionKey, studyInstanceUid, input.ChangeDataset));
+                }
+                catch (FunctionFailedException ex)
+                {
+                    // TODO: Need to call cleanup orchestration on failure after retries.
+                    logger.LogError(ex, "Failed to update instances for study", ex);
+                    var errors = new List<string>
```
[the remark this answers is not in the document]
bcarthic: I exactly tried this :). It didnt work since Errors is a readOnlyList and doesnt have Add method

src/Microsoft.Health.Dicom.Core/Features/Model/InstanceFileIdentifier.cs
```diff
@@ -0,0 +1,15 @@
+// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Health.Dicom.Core.Features.Model;
+
+public class InstanceFileIdentifier
```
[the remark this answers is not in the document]
bcarthic: Agree it needs some doc. And the reason I created this model to use the light weight version of InstanceMetadata. I could have used InstanceMetadata but then we hace to store all the instances for the activity input. we dont need the uids, this is to minimize the size.

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Orchestration.cs
```diff
@@ -0,0 +1,149 @@
...
+
+            if (instanceWatermarks.Count > 0)
+            {
+                instanceWatermarks = await context.CallActivityWithRetryAsync<IReadOnlyList<InstanceFileIdentifier>>(
+                        nameof(UpdateInstanceWatermarkAsync),
+                        _options.RetryOptions,
+                        new BatchUpdateArguments(input.PartitionKey, instanceWatermarks, input.ChangeDataset));
+            }
+
+            if (instanceWatermarks.Count > 0)
```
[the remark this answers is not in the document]
bcarthic: I thought about that while development, since I was doing batching I didnt go with that logic. Now I removed the batching in DB, seems like we get rid of the GetWatermark activity.

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Activity.cs
bcarthic: Not sure if that works, I cannot convert a Task to ValueTask

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Activity.cs
bcarthic: Good catch!!

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Orchestration.cs
```diff
@@ -0,0 +1,149 @@
...
+
+            context.ContinueAsNew(
+                new UpdateCheckpoint
+                {
+                    Batching = input.Batching,
+                    StudyInstanceUids = studyUids,
+                    ChangeDataset = input.ChangeDataset,
+                    PartitionKey = input.PartitionKey,
+                    TotalNumberOfStudies = input.TotalNumberOfStudies,
+                    NumberOfStudyCompleted = input.NumberOfStudyCompleted + 1,
```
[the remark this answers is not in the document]
bcarthic: Great catch!!

src/Microsoft.Health.Dicom.Core/Features/Update/UpdateInstanceService.cs
```diff
@@ -0,0 +1,198 @@
...
+
+        _logger.LogInformation("Updating metadata file {OrignalFileIdentifier} - {NewFileIdentifier} completed successfully", originFileIdentifier, newFileIdentifier);
+    }
+
+    private async Task UpdateDatasetInFileAsync(long newFileIdentifier, DicomDataset datasetToUpdate, KeyValuePair<string, long> block = default, CancellationToken cancellationToken = default)
+    {
+        _logger.LogInformation("Begin updating new file {NewFileIdentifier}", newFileIdentifier);
+
+        // If the block is not provided, then we need to get the first block to update the dataset
+        if (block.Key is null)
```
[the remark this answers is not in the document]
bcarthic: When we are trying to update the instance for the second time, that time we need to access the first block directly, but when we update for the first time, we already knew what the first block is.

src/Microsoft.Health.Dicom.Functions/Update/Models/CompleteInstanceArguments.cs
```diff
@@ -0,0 +1,25 @@
...
+/// <summary>
+///  Represents input to <see cref="UpdateDurableFunction.CompleteUpdateInstanceAsync"/>
+/// </summary>
+public sealed class CompleteInstanceArguments
+{
+    public int PartitionKey { get; }
+
+    public string StudyInstanceUid { get; }
+
+    public string ChangeDataset { get; set; }
```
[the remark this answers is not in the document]
bcarthic: Unfortunately, I need a set otherwise the property is not while reading in the activity

src/Microsoft.Health.Dicom.Functions/Update/UpdateOptions.cs
bcarthic: Updated comment

src/Microsoft.Health.Dicom.Core/Features/Update/UpdateInstanceService.cs
```diff
@@ -0,0 +1,198 @@
...
+    {
+        var blockLengths = new Dictionary<string, long>();
+        long fileSizeWithoutFirstBlock = streamLength - initialBlockLength;
+        int numStagesWithoutFirstBlock = (int)Math.Ceiling((double)fileSizeWithoutFirstBlock / StageBlockSizeInBytes) + 1;
+
+        long bytesRead = 0;
+        for (int i = 0; i < numStagesWithoutFirstBlock; i++)
+        {
+            long blockSize = i == 0 ? initialBlockLength : Math.Min(StageBlockSizeInBytes, streamLength - bytesRead);
+            string blockId = Convert.ToBase64String(Guid.NewGuid().ToByteArray());
```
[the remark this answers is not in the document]
bcarthic: I dont want to translate back, just want to store as base64 string instead of actual Guid.

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Activity.cs
bcarthic: I will just keep this as a constant in UpdateOptions and not from a host.json.

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Activity.cs
bcarthic: I created one, but I will do that in future PRs.

src/Microsoft.Health.Dicom.Core/Features/Model/InstanceFileIdentifier.cs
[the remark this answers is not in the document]
bcarthic: Added some docs, explained why I created this class in the below comment.

src/Microsoft.Health.Dicom.Functions.App/host.json
```diff
@@ -51,6 +51,16 @@
         "FirstRetryInterval": "00:01:00",
         "MaxNumberOfAttempts": 4
       }
+    },
+    "Update": {
```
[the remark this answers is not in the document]
bcarthic: Done

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -75,6 +78,85 @@ public class BlobFileStore : IFileStore
...
+        try
+        {
+            foreach (string blockId in blockLengths.Keys)
+            {
+                long blockSize = blockLengths[blockId];
+
+                await stream.ReadAsync(buffer.AsMemory(0, (int)blockSize), cancellationToken);
+
+                using var blockStream = new MemoryStream(buffer);
+                await blobClient.StageBlockAsync(blockId, blockStream, cancellationToken: cancellationToken);
```
smithago, line 106: can this be parallelized similar to Azure client sdk? Like a configurable 4/5 limit and upload all the stages in parallel and commit?

Another option is to use Azure client SDK with the blocksize option. For that all the blocks need to be same.

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -75,6 +78,85 @@ public class BlobFileStore : IFileStore
...
+
+                Array.Clear(buffer);
+            }
+
+            await blobClient.CommitBlockListAsync(blockLengths.Keys, cancellationToken: cancellationToken);
+
+            return blobClient.Uri;
+        }
+        catch (Exception ex)
```
smithago, line 115: can we use ExecuteAsync pattern for this?

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -75,6 +78,85 @@ public class BlobFileStore : IFileStore
...
+
+        try
+        {
+            stream.Seek(0, SeekOrigin.Begin);
+
+            await blobClient.StageBlockAsync(blockId, stream, cancellationToken: cancellationToken);
+
+            await blobClient.CommitBlockListAsync(blockIds, cancellationToken: cancellationToken);
+        }
+        catch (Exception ex)
```
smithago, line 154: same here

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -177,6 +261,54 @@ public async Task<Stream> GetFileFrameAsync(long version, FrameRange range, Canc
         return stream;
     }
 
+    /// <inheritdoc />
+    public async Task<BinaryData> GetFileInRangeAsync(long version, FrameRange range, CancellationToken cancellationToken)
```
smithago, line 265: GetFileContentInRangeAsync?

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -177,6 +261,54 @@ public async Task<Stream> GetFileFrameAsync(long version, FrameRange range, Canc
...
+    public async Task<KeyValuePair<string, long>> GetFirstBlockPropertyAsync(long version, CancellationToken cancellationToken = default)
+    {
+        BlockBlobClient blobClient = GetInstanceBlockBlobClient(version);
+
+        BlockList blockList = await blobClient.GetBlockListAsync(BlockListTypes.Committed, snapshot: null, conditions: null, cancellationToken);
+
+        if (!blockList.CommittedBlocks.Any())
+            throw new DataStoreException(DicomBlobResource.BlockListNotFound);
+
+        BlobBlock firstBlock = blockList.CommittedBlocks.FirstOrDefault();
```
smithago, line 296: Should this be just First() since you are expecting atleast 1 by here?

src/Microsoft.Health.Dicom.Core/Features/Update/UpdateInstanceService.cs
```diff
@@ -0,0 +1,202 @@
...
+namespace Microsoft.Health.Dicom.Core.Features.Update;
+
+public class UpdateInstanceService : IUpdateInstanceService
+{
+    private readonly IFileStore _fileStore;
+    private readonly IMetadataStore _metadataStore;
+    private readonly ILogger<UpdateInstanceService> _logger;
+    private readonly RecyclableMemoryStreamManager _recyclableMemoryStreamManager;
+    private const int LargeObjectsizeInBytes = 1024 * 1024;
+    private const int StageBlockSizeInBytes = 4 * 1024 * 1024;
```
smithago, line 30: comment 4MB? Do these need to be configurable?

src/Microsoft.Health.Dicom.Core/Features/Update/UpdateInstanceService.cs
```diff
@@ -0,0 +1,202 @@
...
+    /// <inheritdoc />
+    public async Task UpdateInstanceBlobAsync(InstanceFileIdentifier instanceFileIdentifier, DicomDataset datasetToUpdate, CancellationToken cancellationToken)
+    {
+        EnsureArg.IsNotNull(datasetToUpdate, nameof(datasetToUpdate));
+        EnsureArg.IsNotNull(instanceFileIdentifier, nameof(instanceFileIdentifier));
+        EnsureArg.IsTrue(instanceFileIdentifier.NewVersion.HasValue, nameof(instanceFileIdentifier.NewVersion.HasValue));
+
+        Task updateInstanceFileTask = UpdateInstanceFileAsync(instanceFileIdentifier, datasetToUpdate, cancellationToken);
+        Task updateInstanceMetadataTask = UpdateInstanceMetadataAsync(instanceFileIdentifier, datasetToUpdate, cancellationToken);
+        await Task.WhenAll(updateInstanceFileTask, updateInstanceMetadataTask);
```
smithago, line 53: maybe logs here also

src/Microsoft.Health.Dicom.Core/Features/Update/UpdateInstanceService.cs
```diff
@@ -0,0 +1,202 @@
...
+
+    private async Task UpdateInstanceFileAsync(InstanceFileIdentifier instanceFileIdentifier, DicomDataset datasetToUpdate, CancellationToken cancellationToken)
+    {
+        long originFileIdentifier = instanceFileIdentifier.Version;
+        long newFileIdentifier = instanceFileIdentifier.NewVersion.Value;
+
+        KeyValuePair<string, long> block = default;
+
+        // If the file is already updated, then we can copy the file and just update the first block
+        if (instanceFileIdentifier.OriginalVersion.HasValue)
```
smithago, line 77: So I am assuming that originalVersion is only updated when we are ready to commit the update? So does that mean the condition is to copy when this is first update? 
so is'nt the condition then Not of the expression?

src/Microsoft.Health.Dicom.Core/Features/Update/UpdateInstanceService.cs
```diff
@@ -0,0 +1,202 @@
...
+        {
+            _logger.LogInformation("Begin copying instance file {OrignalFileIdentifier} - {NewFileIdentifier}", originFileIdentifier, newFileIdentifier);
+            await _fileStore.CopyFileAsync(originFileIdentifier, newFileIdentifier, cancellationToken);
+        }
+        else
+        {
+            _logger.LogInformation("Begin downloading original file {OrignalFileIdentifier} - {NewFileIdentifier}", originFileIdentifier, newFileIdentifier);
+
+            using Stream stream = await _fileStore.GetFileAsync(originFileIdentifier, cancellationToken);
```
smithago, line 86: This is bringing the entire file in-memory. You need a FileStream

src/Microsoft.Health.Dicom.Core/Features/Update/UpdateInstanceService.cs
```diff
@@ -0,0 +1,202 @@
...
+            DicomFile dcmFile = await DicomFile.OpenAsync(stream, FileReadOption.ReadLargeOnDemand);
```
smithago, line 87: can we skip reading largeObjects?

src/Microsoft.Health.Dicom.Core/Features/Update/UpdateInstanceService.cs
smithago, line 77: I am little confused on this if and else logic. 

Do I read it like

if (previouslyUpdated)
{
 createAnotherCLone()
}
else
{
 fixTheBlocksInPlace()
}

Are'nt you suppose to copy file for the first time of update and fix the blocks in the copied version of the file?

src/Microsoft.Health.Dicom.Core/Features/Update/UpdateInstanceService.cs
```diff
@@ -0,0 +1,202 @@
...
+
+            _logger.LogInformation("Begin uploading instance file in blocks {OrignalFileIdentifier} - {NewFileIdentifier}", originFileIdentifier, newFileIdentifier);
+
+            stream.Seek(0, SeekOrigin.Begin);
+
+            await _fileStore.StoreFileInBlocksAsync(newFileIdentifier, stream, blockLengths, cancellationToken);
+
+            block = blockLengths.First();
+
+            _logger.LogInformation("Uploading instance file in blocks {FileIdentifier} completed successfully", newFileIdentifier);
```
smithago, line 118: can we track how long this takes?

src/Microsoft.Health.Dicom.Core/Features/Update/UpdateInstanceService.cs
bcarthic: I thought it opens the connection and buffers 4MB chunks to the stream, since we are opening the file as large file on demand, we may not read the whole file right? And we also upload the the content to clone blob in 4 MB blobs so that means is the file entirely stored in memory? I will run some tests to understand

src/Microsoft.Health.Dicom.Core/Features/Update/UpdateInstanceService.cs
bcarthic: We need to read large objects to create a copy of the file

src/Microsoft.Health.Dicom.Core/Features/Update/UpdateInstanceService.cs
bcarthic: Yes, I am doing by calling the method StoreFileInBlocksAsync

src/Microsoft.Health.Dicom.Core/Features/Update/UpdateInstanceService.cs
bcarthic: Yes, it can be.

src/Microsoft.Health.Dicom.Core/Features/Update/UpdateInstanceService.cs
bcarthic: Discussed offline. 
The blob api indeed downloads only in chunks so it is not downloading the entire file, but the DicomFile.OpenAsync(stream, ReadOption.LargeFileOnDemand) puts everything in memory.

If we use ReadOption.SkipLargeTags then we will not know when to stop to get the cursor before the large file. but if we use

DicomFile.OpenAsync(stream, ReadOption.LargeFileOnDemand, largeObjectSize: 1000) then fo-dicom buffers it while calling GetBytes and I am still able to get the offset

Overall the memory allocated was max 75MB for this application

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
bcarthic: This method is already called in parallel from a activity. 5 threads are parallely calling this method. Should we measure now and add the improvements later?

src/Microsoft.Health.Dicom.Core/Features/Update/UpdateInstanceService.cs
smithago: Sorry I am still confused. Can we sync on this tomorrow Karthik?

smithago: I was expected

If (!OriginialVersion.HasValue)
{
  Copy()
  UpdateBlocksInTheCopy()
}

UpdatePatientModuleDataSet()

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Update/UpdateInstanceServiceTests.cs
```diff
@@ -0,0 +1,283 @@
...
+using Microsoft.Health.Dicom.Core.Features.Update;
+using Microsoft.Health.Dicom.Core.UnitTests.Features.Retrieve;
+using Microsoft.Health.Dicom.Tests.Common;
+using Microsoft.IO;
+using NSubstitute;
+using Xunit;
+using DicomFileExtensions = Microsoft.Health.Dicom.Core.Features.Retrieve.DicomFileExtensions;
+
+namespace Microsoft.Health.Dicom.Core.UnitTests.Features.Update;
+public class UpdateInstanceServiceTests
```
wsugarman, line 29: nit: newline

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
```diff
@@ -574,6 +575,51 @@ public static void ValidateRequirement(this DicomDataset dataset, DicomTag tag,
...
 
+    /// <summary>
+    /// Returns a dicom item if it matches a large object size criteria.
+    /// </summary>
+    /// <param name="dataset"></param>
+    /// <param name="minLargeObjectsizeInBytes"></param>
+    /// <param name="maxLargeObjectsizeInBytes"></param>
+    /// <param name="largeDicomItem"></param>
+    /// <returns>dicom item</returns>
+    public static bool TryGetLargeDicomItem(
```
wsugarman, line 586: I think I had a related question, but I cannot find it now. Do we have code to ensure that we're not dropping any private tags from the metadata when we extract the pixel data?

src/Microsoft.Health.Dicom.Core/Features/Common/IFileStore.cs
```diff
@@ -68,4 +69,50 @@ public interface IFileStore
...
+    Task<Uri> StoreFileInBlocksAsync(long version, Stream stream, IDictionary<string, long> blockLengths, CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Asynchronously gets a file content from the file store. The file content will be in memory. Use only for small files
+    /// </summary>
+    /// <param name="version">The DICOM instance version.</param>
+    /// <param name="range">Byte range in Httprange format with offset and length</param>
+    /// <param name="cancellationToken">The cancellation token.</param>
+    /// <returns></returns>
+    Task<BinaryData> GetFileContentInRangeAsync(long version, FrameRange range, CancellationToken cancellationToken);
```
wsugarman, line 90: nit: missing default token

src/Microsoft.Health.Dicom.Core/Features/Common/IFileStore.cs
```diff
@@ -68,4 +69,50 @@ public interface IFileStore
...
+    Task<KeyValuePair<string, long>> GetFirstBlockPropertyAsync(long version, CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Asynchronously copies file from the same container
+    /// </summary>
+    /// <param name="originalVersion">The DICOM instance original version.</param>
+    /// <param name="newVersion">The DICOM instance new version.</param>
+    /// <param name="cancellationToken"></param>
+    /// <returns></returns>
+    Task CopyFileAsync(long originalVersion, long newVersion, CancellationToken cancellationToken);
```
wsugarman, line 117: nit: missing default token

src/Microsoft.Health.Dicom.Core/Features/Model/InstanceFileIdentifier.cs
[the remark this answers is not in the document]
wsugarman: Maybe rename the class then to like `InstanceFileStatus`? Or something to suggest that it's state is potentially in-flight? `InstanceFileState`?

src/Microsoft.Health.Dicom.Core/Features/Retrieve/DicomFileExtensions.cs
```diff
@@ -64,4 +69,27 @@ public static void ValidateFrames(DicomPixelData pixelData, IEnumerable<int> fra
...
         }
     }
+
+    /// <summary>
+    /// Given a dicom file, the method will return the length of the dataset.
+    /// </summary>
+    /// <param name="dcmFile"> Dicom file</param>
+    /// <param name="recyclableMemoryStreamManager">RecyclableMemoryStreamManager to get Memory stream</param>
+    /// <returns>Dataset size</returns>
+    public static async Task<long> GetByteLengthAsync(DicomFile dcmFile, RecyclableMemoryStreamManager recyclableMemoryStreamManager)
```
wsugarman, line 79: nit: extension method should have `this` keyword:
```suggestion
    public static async Task<long> GetByteLengthAsync(this DicomFile dcmFile, RecyclableMemoryStreamManager recyclableMemoryStreamManager)
```

src/Microsoft.Health.Dicom.Core/Features/Store/IIndexDataStore.cs
```diff
@@ -134,6 +134,15 @@ public interface IIndexDataStore
...
     Task<IEnumerable<InstanceMetadata>> BeginUpdateInstanceAsync(int partitionKey, IReadOnlyCollection<long> versions, CancellationToken cancellationToken = default);
 
+    /// <summary>
+    /// Asynchronously updates a DICOM instance NewWatermark
+    /// </summary>
+    /// <param name="partitionKey">The partition key.</param>
+    /// <param name="studyInstanceUid">StudyInstanceUID to update</param>
+    /// <param name="cancellationToken">The cancellation token.</param>
+    /// <returns>A task that with list of instance metadata with new watermark.</returns>
+    Task<IEnumerable<InstanceMetadata>> BeginUpdateInstanceAsync(int partitionKey, string studyInstanceUid, CancellationToken cancellationToken = default);
```
wsugarman, line 144: I know we use `IEnumerable<T>` here, but we should use `IAsyncEnumerable<T>` in the future as appropriate (or even `IReadOnlyCollection<T>`)

wsugarman, line 144: nit: `BeginUpdateInstancesAsync` instead? (Plural)

src/Microsoft.Health.Dicom.Core/Features/Update/UpdateInstanceService.cs
```diff
@@ -0,0 +1,218 @@
...
+            if (stream.Length > _largeDicomItemSizeInBytes)
+            {
+                _logger.LogInformation("Found bigger DICOM item, splitting the file into multiple blocks. {OrignalFileIdentifier} - {NewFileIdentifier}", originFileIdentifier, newFileIdentifier);
+
+                if (dcmFile.Dataset.TryGetLargeDicomItem(_largeDicomItemSizeInBytes, _stageBlockSizeInBytes, out DicomItem largeDicomItem))
+                {
+                    RemoveItemsAfter(dcmFile.Dataset, largeDicomItem.Tag);
+                }
+
+                firstBlockLength = await DicomFileExtensions.GetByteLengthAsync(dcmFile, _recyclableMemoryStreamManager);
```
wsugarman, line 115: nit: use as extension method

src/Microsoft.Health.Dicom.Core/Features/Update/UpdateInstanceService.cs
```diff
@@ -0,0 +1,218 @@
...
+
+            IDictionary<string, long> blockLengths = GetBlockLengths(stream.Length, firstBlockLength);
+
+            _logger.LogInformation("Begin uploading instance file in blocks {OrignalFileIdentifier} - {NewFileIdentifier}", originFileIdentifier, newFileIdentifier);
+
+            stream.Seek(0, SeekOrigin.Begin);
+
+            await _fileStore.StoreFileInBlocksAsync(newFileIdentifier, stream, blockLengths, cancellationToken);
+
+            block = blockLengths.First();
```
wsugarman, line 126: Isn't this always non-`null` because `blockLengths` always have count >= 1

src/Microsoft.Health.Dicom.Functions/Update/Models/UpdateInstanceWatermarkArguments.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Microsoft.Health.Dicom.Functions.Update.Models;
+
+/// <summary>
+/// Represents input to <see cref="UpdateDurableFunction.UpdateInstanceWatermarkAsync"/>
+/// </summary>
+public class UpdateInstanceWatermarkArguments
+{
+    public int PartitionKey { get; }
+    public string StudyInstanceUid { get; }
```
wsugarman, line 14: nit: missing newline

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Activity.cs
```diff
@@ -0,0 +1,194 @@
...
+        DicomDataset datasetToUpdate = GetDeserialzedDataset(arguments.ChangeDataset);
+
+        int processed = 0;
+
+        logger.LogInformation("Beginning to update all instance blobs, Total count {TotalCount}", arguments.InstanceWatermarks.Count);
+
+        while (processed < arguments.InstanceWatermarks.Count)
+        {
+            int batchSize = Math.Min(_options.BatchSize, arguments.InstanceWatermarks.Count - processed);
+            var batch = arguments.InstanceWatermarks.Skip(processed).Take(batchSize).ToList();
```
wsugarman, line 86: For both the errors list and the even instance watermarks, I won't be too upset if these internal JSON contracts use mutable types for ease-of-use

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Activity.cs
```diff
@@ -0,0 +1,194 @@
...
+
+            if (batch.Count > 0)
```
wsugarman, line 88: This is never `false`, right?

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Orchestration.cs
```diff
@@ -0,0 +1,130 @@
...
+
+                    input.Errors = errors;
+                }
+            }
+
+            var numberOfStudyCompleted = input.NumberOfStudyCompleted + 1;
+
+            if (input.TotalNumberOfStudies != numberOfStudyCompleted)
+            {
+                logger.LogInformation("Completed updating the instances for a study. {TotalInstanceUpdatedInaStudy}. Continuing with new execution...", instanceWatermarks.Count);
```
wsugarman, line 92:
```suggestion
                logger.LogInformation("Completed updating the instances for a study. {Updated}. Continuing with new execution...", instanceWatermarks.Count);
```

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Orchestration.cs
```diff
@@ -0,0 +1,130 @@
...
+                    PartitionKey = input.PartitionKey,
+                    NumberOfStudyCompleted = numberOfStudyCompleted,
+                    TotalNumberOfInstanceUpdated = input.TotalNumberOfInstanceUpdated + instanceWatermarks.Count,
+                    Errors = input.Errors,
+                    CreatedTime = input.CreatedTime ?? await context.GetCreatedTimeAsync(_options.RetryOptions),
+                });
+        }
+        else
+        {
+            if (input.Errors != null && input.Errors.Count > 0)
```
wsugarman, line 117:
```suggestion
            if (input.Errors?.Count > 0)
```

src/Microsoft.Health.Dicom.Web/appsettings.json
```diff
@@ -45,6 +45,13 @@
         "MaxParallelCount": 2,
         "Size": 100
       }
+    },
+    "Update": {
+      "Name": "UpdateInstancesAsync",
+      "Batching": {
```
wsugarman, line 51: Are these still used?

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -75,6 +78,85 @@ public class BlobFileStore : IFileStore
...
+
+            return blobClient.Uri;
+        }
+        catch (Exception ex)
+        {
+            throw new DataStoreException(ex);
+        }
+        finally
+        {
+            ArrayPool<byte>.Shared.Return(buffer, clearArray: true);
```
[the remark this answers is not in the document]
bcarthic: Not needed here, anyway not reusing it.

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -75,6 +78,85 @@ public class BlobFileStore : IFileStore
...
+        int maxBufferSize = (int)blockLengths.Max(x => x.Value);
+        byte[] buffer = ArrayPool<byte>.Shared.Rent(maxBufferSize);
+
+        try
+        {
+            foreach (string blockId in blockLengths.Keys)
+            {
+                long blockSize = blockLengths[blockId];
+
+                await stream.ReadAsync(buffer.AsMemory(0, (int)blockSize), cancellationToken);
```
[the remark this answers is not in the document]
bcarthic: Its because of this suggestion from VS, https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1835

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -75,6 +78,79 @@ public class BlobFileStore : IFileStore
...
+        await ExecuteAsync(async () =>
+        {
+            byte[] buffer = ArrayPool<byte>.Shared.Rent(maxBufferSize);
+            try
+            {
+                foreach (string blockId in blockLengths.Keys)
+                {
+                    long blockSize = blockLengths[blockId];
+
+                    await stream.ReadAsync(buffer.AsMemory(0, (int)blockSize), cancellationToken);
```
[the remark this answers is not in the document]
bcarthic: Because of this https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1835. Suppressed it

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -75,6 +78,79 @@ public class BlobFileStore : IFileStore
...
+
+        if (blockToUpdate == null)
+            throw new DataStoreException(DicomBlobResource.BlockNotFound);
+
+        stream.Seek(0, SeekOrigin.Begin);
+
+        await ExecuteAsync(async () =>
+        {
+            await blobClient.StageBlockAsync(blockId, stream, cancellationToken: cancellationToken);
+            await blobClient.CommitBlockListAsync(blockIds, cancellationToken: cancellationToken);
```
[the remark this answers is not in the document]
bcarthic: Yeah, we have to commit all the old blockids, else it will overwrite with one block. Strange API.

```
The [StageBlockAsync(String, Stream, BlockBlobStageBlockOptions, CancellationToken)](https://learn.microsoft.com/en-us/dotnet/api/azure.storage.blobs.specialized.blockblobclient.stageblockasync?view=azure-dotnet#azure-storage-blobs-specialized-blockblobclient-stageblockasync(system-string-system-io-stream-azure-storage-blobs-models-blockblobstageblockoptions-system-threading-cancellationtoken)) operation creates a new block as part of a block blob's "staging area" to be eventually committed via the [CommitBlockListAsync(IEnumerable<String>, CommitBlockListOptions, CancellationToken)](https://learn.microsoft.com/en-us/dotnet/api/azure.storage.blobs.specialized.blockblobclient.commitblocklistasync?view=azure-dotnet#azure-storage-blobs-specialized-blockblobclient-commitblocklistasync(system-collections-generic-ienumerable((system-string))-azure-storage-blobs-models-commitblocklistoptions-system-threading-cancellationtoken)) operation.
```

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -177,6 +255,60 @@ public async Task<Stream> GetFileFrameAsync(long version, FrameRange range, Canc
...
+
+    /// <inheritdoc />
+    public async Task<KeyValuePair<string, long>> GetFirstBlockPropertyAsync(long version, CancellationToken cancellationToken = default)
+    {
+        BlockBlobClient blobClient = GetInstanceBlockBlobClient(version);
+        KeyValuePair<string, long> result = new KeyValuePair<string, long>();
+
+        await ExecuteAsync(async () =>
+        {
+            BlockList blockList = await blobClient.GetBlockListAsync(BlockListTypes.Committed, snapshot: null, conditions: null, cancellationToken);
```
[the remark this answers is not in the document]
bcarthic: Yes, the GetBlockListAsync method returns the list of committed blocks in the order in which they were committed.
So, the committed blocks should be returned in an ordered array. The order of the committed blocks is determined by the order in which they were added to the block blob.

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -177,6 +255,60 @@ public async Task<Stream> GetFileFrameAsync(long version, FrameRange range, Canc
...
+
+            BlobBlock firstBlock = blockList.CommittedBlocks.First();
+            result = new KeyValuePair<string, long>(firstBlock.Name, firstBlock.Size);
+        });
+
+        return result;
+    }
+
+    /// <inheritdoc />
+    public async Task CopyFileAsync(long originalVersion, long newVersion, CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
bcarthic: Good thought, I am thinking of throwing an exception if the copy hasnt completed.

src/Microsoft.Health.Dicom.Core/Extensions/DicomDatasetExtensions.cs
bcarthic: We are not extracting the pixel data, we are only finding the first 4MB or the first set of dicom item as part of initial block. I have unit tests to make sure the block lengths is equal to the original stream length

src/Microsoft.Health.Dicom.Core/Features/Store/IIndexDataStore.cs
bcarthic: Changed to IReadOnlyList

src/Microsoft.Health.Dicom.Core/Features/Update/UpdateInstanceService.cs
bcarthic: Yes, this is always not null when we are updating first time, but when we already updated, we wont go through this code. So at that time, it will be null

src/Microsoft.Health.Dicom.Web/appsettings.json
bcarthic: No, forgot to remove these

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
bcarthic: Also the committed blocks are in the order similar to order we commit. If I make this parallel, we may need to preserve the first block name somewhere so we know how to retrieve it

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Activity.cs
bcarthic: I'll leave it as of now, may be I will come back later

src/Microsoft.Health.Dicom.Api/Controllers/StoreController.cs
```diff
@@ -47,7 +47,7 @@ public StoreController(IMediator mediator, ILogger<StoreController> logger, IOpt
 
         _mediator = mediator;
         _logger = logger;
-        _dicomUpdateEnabled = featureConfiguration.Value.EnableDicomUpdate;
+        _dicomUpdateEnabled = featureConfiguration.Value.EnableUpdate;
```
smithago, line 50: nit: looks like not used

bcarthic: It is used here
```
public async Task<IActionResult> UpdateAsync([FromBody][Required] UpdateSpecification updateSpecification)
    {
        if (!_dicomUpdateEnabled)
        {
            throw new DicomUpdateFeatureDisabledException();
        }
```

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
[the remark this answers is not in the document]
wsugarman: Presumably this is here in case we re-run the activity, right? So perhaps it'd be nice to log a message saying it already exists and we're skipping inside of the activity?

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -75,6 +78,77 @@ public class BlobFileStore : IFileStore
...
+        int maxBufferSize = (int)blockLengths.Max(x => x.Value);
+
+        await ExecuteAsync(async () =>
+        {
+            byte[] buffer = ArrayPool<byte>.Shared.Rent(maxBufferSize);
+            try
+            {
+                foreach ((string blockId, long blockSize) in blockLengths)
+                {
+#pragma warning disable CA1835 // Prefer the 'Memory'-based overloads for 'ReadAsync' and 'WriteAsync'
```
wsugarman, line 103: You can also update the `.globalconfig` file as presumably we'll use these APIs if we have the appropriate type:

```
dotnet_diagnostic.CA1835.severity = none
```

src/Microsoft.Health.Dicom.Core/Features/Update/UpdateInstanceService.cs
wsugarman: Oh my gosh, I didn't notice the bracket!

wsugarman (approved):

smithago (approved):

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
bcarthic: Will do it in my next PR

src/Microsoft.Health.Dicom.Core/Features/Update/UpdateInstanceService.cs
```diff
@@ -0,0 +1,231 @@
...
+    {
+        long originFileIdentifier = instanceFileIdentifier.Version;
+        long newFileIdentifier = instanceFileIdentifier.NewVersion.Value;
+        var stopwatch = new Stopwatch();
+
+        KeyValuePair<string, long> block = default;
+
+        bool isPreUpdated = instanceFileIdentifier.OriginalVersion.HasValue;
+
+        // If the file is already updated, then we can copy the file and just update the first block
```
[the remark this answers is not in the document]
bcarthic: Its not for race condition, currently we block multiple update operation running parallel. We may allow only for different studyuid. This is for updating more than once, like if you trigger update second time, we dont need to read the whole file but only the first block

src/Microsoft.Health.Dicom.Core/Features/Update/UpdateInstanceService.cs
```diff
@@ -0,0 +1,231 @@
...
+            DicomFile dcmFile = await DicomFile.OpenAsync(stream, FileReadOption.ReadLargeOnDemand, LargeObjectSizeInBytes);
+
+            long firstBlockLength = stream.Length;
+
+            // If the file is greater than 1MB try to upload in different blocks.
+            // Since we are supporting updating only Patient demographic and identification module,
+            // we assume patient attributes are at the very beginning of the dataset.
+            // If in future, we support updating other attributes like private tags, which could appear at the end
+            // We need to read the whole file instead of first block.
+            if (stream.Length > _largeDicomItemSizeInBytes)
```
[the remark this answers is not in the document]
bcarthic: This is very rare scenario I dont expect to happen, but if the file doesn't have large item but multiple small items less than 1 MB then we dont want to have the first block bloated more than 4 MB, so performing this check

PR #2601, opened 2023-04-20
Added metric for Dicom update API

arunmk-ms:
## Description
Added metric for recording the number of instances updated

## Related issues
Addresses [issue #].

## Testing


src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Activity.cs
```diff
@@ -140,6 +140,7 @@ public async Task UpdateInstanceBatchAsync([ActivityTrigger] BatchUpdateArgument
                     async (instance, token) =>
                     {
                         await _updateInstanceService.UpdateInstanceBlobAsync(instance, datasetToUpdate, token);
+                        _updateMeter.UpdatedInstances.Add(1);
```
bcarthic, line 143: Let's do this after we successfully committed. Move this to the orchestration

PR #2603, opened 2023-04-24
Migrate frame range files using orchestration

bcarthic:
## Description
This PR includes changes to migrate frame range files from one blob format to another

## Related issues
Addresses [[AB#102675](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/102675)].

## Testing
New tests are added


src/Microsoft.Health.Dicom.Core/Features/Operations/IDicomOperationsClient.cs
```diff
@@ -118,4 +117,17 @@ public interface IDicomOperationsClient
     /// </exception>
     /// <exception cref="OperationCanceledException">The <paramref name="cancellationToken"/> was canceled.</exception>
     Task<OperationReference> StartUpdateOperationAsync(Guid operationId, UpdateSpecification updateSpecification, int partitionKey, CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Asynchronously migrating instance frame range files.
```
wsugarman, line 122:
```suggestion
    /// Asynchronously migrate instance frame range files.
```

src/Microsoft.Health.Dicom.Core/Features/Operations/IDicomOperationsClient.cs
```diff
@@ -118,4 +117,17 @@ public interface IDicomOperationsClient
...
+    /// </summary>
+    /// <param name="operationId">The desired ID for the clenup operation.</param>
```
wsugarman, line 124:
```suggestion
    /// <param name="operationId">The desired ID for the cleanup operation.</param>
```

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Tables/Instance.sql
```diff
@@ -157,3 +157,19 @@ INCLUDE
     OriginalWatermark
 )
 WITH (DATA_COMPRESSION = PAGE)
+
+-- Used in migrating files
+CREATE NONCLUSTERED INDEX IX_Instance_Watermark_Status_CreatedDate on dbo.Instance
```
wsugarman, line 162: What does the query plan look like with this index?

src/Microsoft.Health.Dicom.Functions/Migration/Models/MigrationBatchCreationArguments.cs
```diff
@@ -0,0 +1,65 @@
...
+
+    /// <summary>
+    /// Gets or sets the maximum number of concurrent batches processed at a given time.
+    /// </summary>
+    public int MaxParallelBatches { get; }
+
+    /// <summary>
+    /// Gets or sets the start filter stamp
+    /// </summary>
+    public DateTime StartFilterTimeStamp { get; set; }
```
wsugarman, line 34: nit: don't need `set` here and below if you have ctor

src/Microsoft.Health.Dicom.Functions/Migration/MigrationFilesOptions.cs
```diff
@@ -0,0 +1,41 @@
...
+/// </summary>
+public class MigrationFilesOptions
+{
+    internal const string SectionName = "MigratingFiles";
+
+    /// <summary>
+    /// Gets or sets the number of DICOM instances processed by a single activity.
+    /// </summary>
+    [Range(1, int.MaxValue)]
+    public int BatchSize { get; set; } = 100;
```
wsugarman, line 23: You don't need `BatchSize` and `MaxParallelBatches` as config options because these are presumably function inputs

src/Microsoft.Health.Dicom.Functions/Migration/MigrationFilesDurableFunction.Activity.cs
```diff
@@ -0,0 +1,110 @@
...
+        var versions = instanceIdentifiers.Select(x => x.Version).ToList();
+
+        await Parallel.ForEachAsync(
+            versions,
+            new ParallelOptions
+            {
+                CancellationToken = default,
+                MaxDegreeOfParallelism = _options.MaxParallelThreads,
+            },
+            async (version, token) =>
```
wsugarman, line 89: nit: these functions don't need `async` and `await` if they return `Task<T>`. Otherwise they can be converted to `Task<T>` or left as-is

src/Microsoft.Health.Dicom.Functions.Client/DicomAzureFunctionsClient.cs
```diff
@@ -204,6 +205,23 @@ public async Task<OperationReference> StartUpdateOperationAsync(Guid operationId
         return new OperationReference(operationId, _urlResolver.ResolveOperationStatusUri(operationId));
     }
 
+    public async Task StartMigratingFrameRangeBlobAsync(Guid operationId, DateTime startFilterTimeStamp, DateTime endFilterTimeStamp, CancellationToken cancellationToken = default)
```
wsugarman, line 208: Should this method check that `endFilterTimeStamp > startFilterTimeStamp`

src/Microsoft.Health.Dicom.Functions.App/host.json
```diff
@@ -52,6 +52,16 @@
         "MaxNumberOfAttempts": 4
       }
     },
+    "MigratingFiles": {
+      "BatchSize": 100,
+      "MaxParallelThreads": 5,
```
wsugarman, line 57: Remove `BatchSize` and `MaxParallelBatches`

src/Microsoft.Health.Dicom.Functions/Migration/Models/MigrationFilesBatchArguments.cs
```diff
@@ -0,0 +1,28 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using Microsoft.Health.Dicom.Core.Features.Model;
+
+namespace Microsoft.Health.Dicom.Functions.Migration.Models;
+
+/// <summary>
+///  Represents input to <see cref="MigrationFilesDurableFunction.MigrateFrameRangeFilesAsync"/>
+/// </summary>
+public sealed class MigrationFilesBatchArguments
```
wsugarman, line 13: You could skip this class altogether too if you wanted such that argument type `WatermarkRange` directly

src/Microsoft.Health.Dicom.Functions.Abstractions/Migration/MigratingFilesInput.cs
```diff
@@ -0,0 +1,17 @@
...
+
+using System;
+
+namespace Microsoft.Health.Dicom.Functions.Migration;
+
+public class MigratingFilesInput
+{
+    public BatchingOptions Batching { get; set; }
+
+    public DateTime StartFilterTimeStamp { get; set; }
```
wsugarman, line 14: nit: same comment about `DateTimeOffset`

src/Microsoft.Health.Dicom.Functions.Abstractions/Migration/MigratingFilesCheckpoint.cs
```diff
@@ -0,0 +1,34 @@
...
+            if (Completed.HasValue)
+            {
+                WatermarkRange range = Completed.GetValueOrDefault();
+                return range.End == 1 ? 100 : (int)((double)(range.End - range.Start + 1) / range.End * 100);
+            }
+            return 0;
+        }
+    }
+    public IReadOnlyCollection<string> ResourceIds => null;
+    public object GetResults(JToken output) => null;
```
wsugarman, line 33: nit: missing newline

src/Microsoft.Health.Dicom.Functions/Migration/Models/MigrationFilesBatchArguments.cs
bcarthic: True, initially I had extra property then removed and missed it can be simplified.

src/Microsoft.Health.Dicom.Functions.App/host.json
bcarthic: BatchSize and MaxParallelBatches are used to get instance batches similar to re-indexing. I could make it as constant. Is that what you are suggesting?

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobMetadataStore.cs
```diff
@@ -218,6 +218,30 @@ public async Task DeleteInstanceFramesRangeAsync(long version, CancellationToken
...
+        }
+    }
+
+    /// <inheritdoc />
+    public async Task DeleteMigratedFramesRangeIfExistsAsync(long version, CancellationToken cancellationToken)
+    {
+        var blobClientWithSpace = GetInstanceFramesRangeBlobClient(version, fallBackClient: true);
+        var blobClient = GetInstanceFramesRangeBlobClient(version);
+
+        if (await blobClient.ExistsAsync(cancellationToken))
```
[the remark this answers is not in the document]
bcarthic: Here i am ensuring the new blob is present before deleting the old one.

src/Microsoft.Health.Dicom.Functions/Migration/MigrationFilesDurableFunction.Activity.cs
bcarthic: In order to do that, I would need wrap in a ValueTask.

src/Microsoft.Health.Dicom.Functions/Migration/Models/MigrationBatchCreationArguments.cs
bcarthic: No I need the set, Otherwise the value won't be set for the activity

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Tables/Instance.sql
bcarthic: image

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobMetadataStore.cs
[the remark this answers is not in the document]
wsugarman: Oh my mistake -- these are different clients

src/Microsoft.Health.Dicom.Functions/Migration/Models/MigrationBatchCreationArguments.cs
wsugarman: Presumably that's because you named the ctor argument differently from the property. I think for newtonsoft if you name them the same, JSON.NET should figure it out.

src/Microsoft.Health.Dicom.Functions/Migration/MigrationFilesOptions.cs
wsugarman: No they aren't, right? You use the value from the input which is correct. You shouldn't use these, as a developer may inadvertently change them and affect the execution on running operations. That could potentially crash the long-running operations as it would change the number of activities and/or their expected output. I looked at the use of `_options` and it's correctly only used for getting the concurrency within a single activity

src/Microsoft.Health.Dicom.Functions/Migration/MigrationFilesDurableFunction.Activity.cs
wsugarman: There's a method like `AsTask()` or something like that if you wanted

src/Microsoft.Health.Dicom.Functions/Migration/Models/MigrationBatchCreationArguments.cs
bcarthic: oh nice catch. I totally missed that.

src/Microsoft.Health.Dicom.Functions.App/host.json
bcarthic: Removed

src/Microsoft.Health.Dicom.Functions/Migration/MigrationFilesOptions.cs
bcarthic: yeah you are right. Removed the unwanted properties and config

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/34.diff.sql
```diff
@@ -31,8 +31,8 @@ CREATE OR ALTER PROCEDURE dbo.GetInstanceBatchesByTimeStamp
     @batchSize INT,
     @batchCount INT,
     @status TINYINT,
-    @startTimeStamp DATETIME,
-    @endTimeStamp DATETIME,
+    @startTimeStamp DATETIMEOFFSET(0),
```
wsugarman, line 34: Do you need any more granularity than this?

bcarthic: I don't think so, I already gave plus or minus one day. So the dates are not that exact.

src/Microsoft.Health.Dicom.Functions.Client/DicomAzureFunctionsClient.cs
```diff
@@ -205,8 +205,10 @@ public async Task<OperationReference> StartUpdateOperationAsync(Guid operationId
         return new OperationReference(operationId, _urlResolver.ResolveOperationStatusUri(operationId));
     }
 
-    public async Task StartMigratingFrameRangeBlobAsync(Guid operationId, DateTime startFilterTimeStamp, DateTime endFilterTimeStamp, CancellationToken cancellationToken = default)
+    public async Task StartMigratingFrameRangeBlobAsync(Guid operationId, DateTimeOffset startFilterTimeStamp, DateTimeOffset endFilterTimeStamp, CancellationToken cancellationToken = default)
     {
+        EnsureArg.IsTrue(endFilterTimeStamp > startFilterTimeStamp, nameof(endFilterTimeStamp));
```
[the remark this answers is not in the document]
bcarthic: Done

PR #2605, opened 2023-04-24
Support external store for dcm files

smithago:
## Description
- New config and FF for External store
- Support SxS Internal and External store with IBlobClient which also initializes the externalStore after the app initialization to support post configuration
- Handle exceptions from external store differently than internal store

## Related issues
[AB#102847](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/102847)
[AB#102848](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/102848)

## Testing
New unit tests added. 
Next iteration will add a new run when the FF is on for EnE tests


src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
```diff
@@ -90,6 +90,10 @@ private IActionResult MapExceptionToResult(Exception exception)
             case IOException io when io.Message.Equals("The request stream was aborted.", StringComparison.OrdinalIgnoreCase):
                 statusCode = HttpStatusCode.BadRequest;
                 break;
+            case DataStoreException ex when ex.IsExternal == true:
```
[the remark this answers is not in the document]
smithago: Yes we can, I thought about it too, but the code to check the condition and throw different exception types, when they are the same got too verbose and felt this was more tight

src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalBlobClient.cs
```diff
@@ -0,0 +1,86 @@
...
+using Microsoft.Health.Dicom.Blob.Utilities;
+using Microsoft.Health.Dicom.Core.Features.Common;
+using Microsoft.Health.Dicom.Blob.Features.Storage;
+using Microsoft.Health.Dicom.Core.Exceptions;
+using System;
+
+namespace Microsoft.Health.Dicom.Blob.Features.ExternalStore;
+
+/// Represents the blob container created by the user and initialized JIT
+internal class ExternalBlobClient : IBlobClient
```
jovinson-ms, line 22: Should this live in `healthcare-shared-components` as FHIR will need the same?

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -130,15 +119,20 @@ public class BlobFileStore : IFileStore
         EnsureArg.IsNotNullOrWhiteSpace(blockId, nameof(blockId));
 
         BlockBlobClient blobClient = GetInstanceBlockBlobClient(version);
+        _logger.LogInformation("Trying to read block list for DICOM instance file with watermark '{Version}'.", version);
```
jovinson-ms, line 122: Should we set the log level on these statements to `Debug`? This will be a lot of logs.

src/Microsoft.Health.Dicom.Blob/Registration/DicomFunctionBuilderMetadataRegistrationExtensions.cs
```diff
@@ -56,6 +59,25 @@ public static class DicomFunctionsBuilderRegistrationExtensions
...
+            ExternalBlobDataStoreConfiguration externalBlobData = new ExternalBlobDataStoreConfiguration();
+            configuration.GetSection(ExternalBlobDataStoreConfiguration.SectionName).Bind(externalBlobData);
+            functionsBuilder.Services.AddSingleton(Options.Options.Create(externalBlobData));
+
+            functionsBuilder.Services.Add<ExternalBlobClient>()
+            .Singleton()
+            .AsSelf()
+            .AsImplementedInterfaces();
+        }
+        else
```
jovinson-ms, line 74: Don't we need both internal and external?

src/Microsoft.Health.Dicom.Blob/Registration/DicomServerBuilderBlobRegistrationExtensions.cs
```diff
@@ -12,10 +12,13 @@
 using Microsoft.Health.Dicom.Blob.Features.Storage;
 using Microsoft.Health.Dicom.Blob.Features.Telemetry;
 using Microsoft.Health.Dicom.Blob.Utilities;
+using Microsoft.Health.Dicom.Core.Configs;
 using Microsoft.Health.Dicom.Core.Features.Common;
+using Microsoft.Health.Dicom.Blob.Features.ExternalStore;
 using Microsoft.Health.Dicom.Core.Features.Workitem;
 using Microsoft.Health.Dicom.Core.Registration;
 using Microsoft.Health.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
```
jovinson-ms, line 21: nit: sort usings

src/Microsoft.Health.Dicom.Blob/Registration/DicomServerBuilderBlobRegistrationExtensions.cs
```diff
@@ -37,6 +40,26 @@ public static IDicomServerBuilder AddBlobDataStores(this IDicomServerBuilder ser
...
+            ExternalBlobDataStoreConfiguration externalBlobData = new ExternalBlobDataStoreConfiguration();
+            configuration.GetSection(ExternalBlobDataStoreConfiguration.SectionName).Bind(externalBlobData);
+            serverBuilder.Services.AddSingleton(Options.Options.Create(externalBlobData));
+
+            serverBuilder.Services.Add<ExternalBlobClient>()
+            .Singleton()
+            .AsSelf()
+            .AsImplementedInterfaces();
+        }
+        else
```
jovinson-ms, line 55: Same here, don't we always need internal and add external if specified?

src/Microsoft.Health.Dicom.Api.UnitTests/Features/Exceptions/ExceptionHandlingMiddlewareTests.cs
```diff
@@ -57,6 +57,8 @@ public static IEnumerable<object[]> GetExceptionToStatusCodeMapping()
         yield return new object[] { new TaskCanceledException(), HttpStatusCode.BadRequest };
         yield return new object[] { new InvalidOperationException(), HttpStatusCode.BadRequest };
         yield return new object[] { new PayloadTooLargeException(1), HttpStatusCode.RequestEntityTooLarge };
+        yield return new object[] { new DataStoreException(new Exception(), isExternal: true), HttpStatusCode.FailedDependency };
```
jovinson-ms, line 60: Should we add tests that capture the `innerException` handling of BYOS exceptions as defined [here](https://github.com/microsoft/api-guidelines/blob/vNext/azure/Guidelines.md#bring-your-own-storage-byos)?

smithago: Thanks for the link. I have in the BlobFileTests

src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalBlobClient.cs
```diff
@@ -0,0 +1,86 @@
...
+                    if (_blobContainerClient == null)
+                    {
+                        try
+                        {
+                            if (_externalStoreOptions.UseManagedIdentity)
+                            {
+                                TokenCredential credential = _credentialProvider.GetTokenCredential();
+                                if (credential == null)
+                                {
+                                    throw new InvalidOperationException(DicomBlobResource.MissingServerIdentity);
```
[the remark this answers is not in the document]
smithago: good point, let me cleanit up to pass null and catch everything as requestFailedException

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
smithago: Good point. There is no limit right on logs right now. Lets dial it down when we have to.

src/Microsoft.Health.Dicom.Blob.UnitTests/BlobFileTests.cs
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    private class TestInternalBlobClient : IBlobClient
+    {
+        public TestInternalBlobClient()
+        {
+            BlobContainerClient = Substitute.For<BlobContainerClient>();
+            BlockBlobClient = Substitute.For<BlockBlobClient>();
+            BlobContainerClient.GetBlockBlobClient(Arg.Any<string>()).Returns(BlockBlobClient);
+
```
brandonpollett, line 125: nit: extra new line

src/Microsoft.Health.Dicom.Blob.UnitTests/BlobFileTests.cs
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    private class TestExternalBlobClient : IBlobClient
+    {
+        public TestExternalBlobClient()
+        {
+            BlobContainerClient = Substitute.For<BlobContainerClient>();
+            BlockBlobClient = Substitute.For<BlockBlobClient>();
+            BlobContainerClient.GetBlockBlobClient(Arg.Any<string>()).Returns(BlockBlobClient);
+
```
brandonpollett, line 102: nit: extra new line

src/Microsoft.Health.Dicom.Blob.UnitTests/BlobFileTests.cs
```diff
@@ -0,0 +1,141 @@
...
+        {
+            get
+            {
+                return false;
+            }
+        }
+
+        public BlockBlobClient BlockBlobClient { get; private set; }
+    }
+
```
brandonpollett, line 140: nit: extra new line

src/Microsoft.Health.Dicom.Blob.UnitTests/BlobFileTests.cs
```diff
@@ -0,0 +1,141 @@
...
+        {
+            BlobContainerClient = Substitute.For<BlobContainerClient>();
+            BlockBlobClient = Substitute.For<BlockBlobClient>();
+            BlobContainerClient.GetBlockBlobClient(Arg.Any<string>()).Returns(BlockBlobClient);
+
+        }
+
+        public virtual BlobContainerClient BlobContainerClient { get; private set; }
+
+        public bool IsExternal
```
brandonpollett, line 130: ` public bool IsExternal => false;`

src/Microsoft.Health.Dicom.Blob.UnitTests/BlobFileTests.cs
```diff
@@ -0,0 +1,141 @@
...
```
brandonpollett, line 107: `public bool IsExternal => true;`

src/Microsoft.Health.Dicom.Blob.UnitTests/BlobFileTests.cs
```diff
@@ -0,0 +1,141 @@
...
+using NSubstitute.ExceptionExtensions;
+using Microsoft.Extensions.Logging.Abstractions;
+using System.Threading.Tasks;
+using System.Globalization;
+using Microsoft.Health.Dicom.Core;
+using Azure;
+
+namespace Microsoft.Health.Dicom.Blob.UnitTests;
+
+public class BlobFileTests
```
brandonpollett, line 26: maybe: `BlobFileStoreTests`?

src/Microsoft.Health.Dicom.Blob.UnitTests/BlobFileTests.cs
```diff
@@ -0,0 +1,141 @@
...
+
+        Assert.False(ex.IsExternal);
+        Assert.Equal(DicomCoreResource.DataStoreOperationFailed, ex.Message);
+    }
+
+    [Fact]
+    public async Task GivenExternalStore_WhenGetFails_ThenThrowExceptionWithRightMessageAndProperty()
+    {
+        InitializeExternalBlobFileStore(out BlobFileStore blobFileStore, out TestExternalBlobClient client);
+        RequestFailedException requestFailedException = new RequestFailedException(status: 404, message: "test", errorCode: BlobErrorCode.BlobNotFound.ToString(), innerException: null);
```
brandonpollett, line 56: Do we need a test with an non `404` `RequestFailedException`?

src/Microsoft.Health.Dicom.Blob.UnitTests/BlobFileTests.cs
```diff
@@ -0,0 +1,141 @@
...
+        Assert.Equal(DicomCoreResource.DataStoreOperationFailed, ex.Message);
+    }
+
+    private static void InitializeInternalBlobFileStore(out BlobFileStore blobFileStore, out TestInternalBlobClient externalBlobClient)
+    {
+        externalBlobClient = new TestInternalBlobClient();
+        var options = Substitute.For<IOptions<BlobOperationOptions>>();
+        options.Value.Returns(Substitute.For<BlobOperationOptions>());
+        blobFileStore = new BlobFileStore(externalBlobClient, Substitute.For<DicomFileNameWithPrefix>(), options, NullLogger<BlobFileStore>.Instance);
+
```
brandonpollett, line 83: nit: extra new line

src/Microsoft.Health.Dicom.Blob/Features/Storage/InternalBlobClient.cs
```diff
@@ -0,0 +1,36 @@
...
+namespace Microsoft.Health.Dicom.Blob.Features.Storage;
+
+/// <summary>
+/// Represents the blob container created by the service and initialized during app startup
+/// </summary>
+internal class InternalBlobClient : IBlobClient
+{
+    private readonly BlobServiceClient _client;
+    private readonly string _containerName;
+    public InternalBlobClient(BlobServiceClient blobServiceClient,
```
brandonpollett, line 20: nit: missing new line

src/Microsoft.Health.Dicom.Blob/Features/Storage/InternalBlobClient.cs
```diff
@@ -0,0 +1,36 @@
...
+        IOptionsMonitor<BlobContainerConfiguration> namedBlobContainerConfigurationAccessor)
+    {
+        _client = EnsureArg.IsNotNull(blobServiceClient, nameof(blobServiceClient));
+        _containerName = EnsureArg.IsNotNull(namedBlobContainerConfigurationAccessor.Get(Constants.BlobContainerConfigurationName).ContainerName, nameof(namedBlobContainerConfigurationAccessor));
+    }
+
+    public bool IsExternal => false;
+
+    public BlobContainerClient BlobContainerClient
```
brandonpollett, line 29: `public BlobContainerClient BlobContainerClient => _client.GetBlobContainerClient(_containerName);`

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -130,15 +119,20 @@ public class BlobFileStore : IFileStore
...
 
-        BlockList blockList = await blobClient.GetBlockListAsync(BlockListTypes.Committed, snapshot: null, conditions: null, cancellationToken);
+        BlockList blockList = null;
+        await ExecuteAsync(async () =>
+        {
+            blockList = await blobClient.GetBlockListAsync(BlockListTypes.Committed, snapshot: null, conditions: null, cancellationToken);
+        });
 
         IEnumerable<string> blockIds = blockList.CommittedBlocks.Select(x => x.Name);
```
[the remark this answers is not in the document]
bcarthic: I also verified with file created using DataLakeStorage API and use blobAPi to get BlockList. It is not null but blocks return empty result.
image

brandonpollett (dismissed):

src/Microsoft.Health.Dicom.Blob/Utilities/ExternalBlobDataStoreConfiguration.cs
```diff
@@ -0,0 +1,21 @@
+// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.Health.Dicom.Blob.Utilities;
+internal class ExternalBlobDataStoreConfiguration
```
wsugarman, line 9: nit: missing newline

src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalBlobClient.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            if (_blobContainerClient == null)
+            {
+                lock (_lockObj)
+                {
+                    if (_blobContainerClient == null)
+                    {
+                        try
+                        {
+                            if (_externalStoreOptions.UseManagedIdentity)
```
wsugarman, line 53: We could also just check the presence of `BlobContainerUri`

src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalBlobClient.cs
```diff
@@ -0,0 +1,73 @@
...
+        IOptions<BlobServiceClientOptions> blobClientOptions)
+    {
+        _credentialProvider = EnsureArg.IsNotNull(credentialProvider, nameof(credentialProvider));
+        _blobClientOptions = EnsureArg.IsNotNull(blobClientOptions?.Value, nameof(blobClientOptions));
+        _externalStoreOptions = EnsureArg.IsNotNull(externalStoreOptions?.Value, nameof(externalStoreOptions));
+    }
+
+    public bool IsExternal => true;
+
+    public BlobContainerClient BlobContainerClient
```
wsugarman, line 41: Why lazily create this?

test/Microsoft.Health.Dicom.Tests.Integration/Persistence/DataStoreTestsFixture.cs
```diff
@@ -90,4 +90,16 @@ public async Task DisposeAsync()
...
     }
+
+    private class TestInternalBlobClient : IBlobClient
+    {
+        public TestInternalBlobClient(BlobServiceClient blobClient, string containerName)
+        {
+            BlobContainerClient = blobClient.GetBlobContainerClient(containerName);
+        }
+
+        public virtual BlobContainerClient BlobContainerClient { get; private set; }
```
wsugarman, line 101: nit: can remove `private set`

src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
wsugarman, line 93: Should we instead have a separate `Exception` like `ExternalStorageException` whose `InnerException` varies. That way we don't need to update our existing `Exception` types to accommodate. Then we just special-case this new exception here

wsugarman: Although, I admit that by keeping the types we can preserve any logic that relies on `DataStoreException` and `ItemNotFoundException`. An alternative idea could be an interface that indicates the `Exception` may be from an external source like `IExternalException`

src/Microsoft.Health.Dicom.Blob/Registration/DicomServerBuilderBlobRegistrationExtensions.cs
```diff
@@ -37,6 +39,25 @@ public static IDicomServerBuilder AddBlobDataStores(this IDicomServerBuilder ser
             .AddOptions<BlobOperationOptions>()
             .Bind(blobConfig.GetSection(nameof(BlobServiceClientOptions.Operations)));
 
+        FeatureConfiguration featureConfiguration = new FeatureConfiguration();
+        configuration.GetSection("DicomServer").GetSection("Features").Bind(featureConfiguration);
+        if (featureConfiguration.EnableExternalStore)
```
jovinson-ms, line 44: Currently with external store enabled, `DcmHealthCheck` still runs against the internal blob store:

```
info: Azure.Core[1]
      Request [89d96890-0e0d-4d3f-a66a-3d2b7c3b63df] GET http://127.0.0.1:10000/devstoreaccount1/dicomwebcontainer?restype=container
```

Do we expect the health check to run against the external store in this case, or not run?

src/Microsoft.Health.Dicom.Blob/Features/Storage/IBlobClient.cs
```diff
@@ -0,0 +1,20 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using Azure.Storage.Blobs;
+
+namespace Microsoft.Health.Dicom.Blob.Features.Storage;
+
+/// <summary>
+/// Responsible to get the right BlobContainerClient based on the configuration
+/// </summary>
+public interface IBlobClient
```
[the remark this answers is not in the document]
smithago: ContainerClient is just one of the property it provides. Class is more generic

src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalBlobClient.cs
smithago: To allow container to initialize even when the BYOS is down

src/Microsoft.Health.Dicom.Blob/Registration/DicomServerBuilderBlobRegistrationExtensions.cs
smithago: Good point. I will condition it

src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
smithago: Esma had the same suggestion. The code to throw these exceptions will get very verbose.

src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalBlobClient.cs
wsugarman: I don't think any remote operation happens within the ctor for the client, so I think you're safe to create the client (unless the URL is not in the proper format).

src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
wsugarman: At the very least, using an interface allows you to just have a case like:
```csharp
case IExternalException:
   // TODO
```

Instead of listing each implementation

src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalBlobClient.cs
smithago: We will try to get token which access their key valut that can fail

src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
smithago: I have not seen interfaces used in Exceptions, only classes to allow for easy disambiguation, even though in this case we may not catch and throw based on the interface type.

Maybe we can reduce the need of specific exceptions like ItemNotFoundException. DataStoreException has the innerexception for details.
Only difference today is ItemNotFound is throwing 500 and DataStoreException is throwing  503. 

what do you think?

smithago: Same idea as yours. Just used based class instead of interface

https://github.com/microsoft/dicom-server/pull/2605/commits/413c96606b583d7611436da624ce6fffe3598dc5

src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalBlobClient.cs
wsugarman: We're only fetching the [`TokenCredential`](https://learn.microsoft.com/en-us/dotnet/api/azure.core.tokencredential?view=azure-dotnet) I think which confusingly does not actually have the token. Instead, it just has a method to fetch the token like `GetTokenAsync`

src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
wsugarman: Yeah, that makes sense. An alternative may just be to make `DataStoreException` the base exception, as we're really saying, "There was a problem with the data store." Optionally users can check the `IsExternal` property on the exception. Then `ItemNotFoundException` can derived from `DataStoreException`.

smithago: Yeah maybe, right now DataStore as some additional properties that I don't want to touch.

src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalBlobClient.cs
smithago: I have also see Verify checks on the BlobContainerClient https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/storage/Azure.Storage.Blobs/src/BlobContainerClient.cs.

This maybe pessimestic, but I would prefer this code to be lazy and catch and throw rather than failing the app intialization. I don't think there is any con with being lazy other than the lock logic.

jovinson-ms: Linked AB#102834 for related design discussion.

brandonpollett (approved):

wsugarman (approved):

PR #2611, opened 2023-04-25
Enable Time Range for Change Feed

wsugarman:
## Description
This Pull Request updates the Change Feed API. Not only does it add APIs for v2 that include time/temporal semantics, but it also fixes a latent issue where users may fetch events from the Change Feed while entries are still being committed. This may lead to problem where a user may see a sequence number 'n' while 'n - 1' has yet to be committed; this response may incorrect lead users to believe there are no additional events below 'n' remaining.

- Add new `startTime` and `endTime` query string parameters for the `/changefeed` endpoint where users can further scope their queries to particular times starting from version `v2`
  - Enables easier batch processing of changes that do not require maintaining the offset
  - Uses custom `IModelBinder` that requires time zone
- Add new v35 SQL version that includes:
  - Add `GetChangeFeedByTime` and `GetChangeFeedLatestByTime` sprocs to order data by `Timestamp` column
  - Updates to sprocs that paginate to use `BIGINT` instead of `INT`
  - Use `HOLDLOCK` when accessing the `dbo.ChangeFeed` table to ensure writes and reads cannot happen concurrently
  - Move `INSERT` commands into `dbo.ChangeFeed` to occur at the end of add instance, delete instance, and update instance sprocs
  - Generate `Timestamp` column in `dbo.ChangeFeed` instead of pre-generating it for consistency between tables. That consistency is not necessary, and may lead to inconsistent orders between `Sequence` and `Timestamp`
- Normalize default pagination options to align with QIDO for v2 (default = 100, max = 200)
- Update API versioning to disable actions if their version isn't active
- Refactor change feed service test
- Fix local test runs for update operation
- Update Change Feed documentation

## Related issues
[AB#102633](https://microsofthealth.visualstudio.com/Health/_workitems/edit/102633)
[AB#103724](https://microsofthealth.visualstudio.com/Health/_workitems/edit/103724/)
[AB#89771](https://microsofthealth.visualstudio.com/Health/_workitems/edit/89771/)

## Testing
Add new unit, integration, and E2E tests


src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/GetChangeFeedV33.sql
```diff
@@ -0,0 +1,48 @@
...
+        PartitionName,
+        StudyInstanceUid,
+        SeriesInstanceUid,
+        SopInstanceUid,
+        OriginalWatermark,
+        CurrentWatermark
+    FROM dbo.ChangeFeed c
+    INNER JOIN dbo.Partition p
+    ON p.PartitionKey = c.PartitionKey
+    WHERE c.Timestamp >= @startTime AND c.Timestamp <@endTime
```
smithago, line 44: I like that the single prc can server both the old and new behavior.

But this SQL is not performant. There is no index on Timestamp, so this will a index scan on cluster index.
Even with the new index, we should look at the query plan and make sure only the limit rows are read when starttime and endtime is infinite

src/Microsoft.Health.Dicom.Api/Controllers/ChangeFeedController.cs
```diff
@@ -68,10 +88,26 @@ public async Task<IActionResult> GetChangeFeedLatest([FromQuery] bool includeMet
...
 
-        var response = await _mediator.GetChangeFeedLatest(
+        ChangeFeedLatestResponse response = await _mediator.GetChangeFeedLatest(
             includeMetadata,
             cancellationToken: HttpContext.RequestAborted);
 
         return StatusCode((int)HttpStatusCode.OK, response.Entry);
     }
+
+    private static int GetLimit(int? userValue, int majorVersion)
```
smithago, line 98: tests to make sure backcompat works?

src/Microsoft.Health.Dicom.SqlServer/Microsoft.Health.Dicom.SqlServer.csproj
```diff
@@ -68,6 +68,8 @@
     <TSqlScript Include="Features\Schema\Sql\Scripts\Sequences.sql" />
     <TSqlScript Include="Features\Schema\Sql\Tables\*.sql" />
     <TSqlScript Include="Features\Schema\Sql\Types\*.sql" />
+    <None Remove="features\schema\migrations\33.diff.sql" />
```
smithago, line 71: cleanup this adds

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/GetChangeFeedV33.sql
```diff
@@ -0,0 +1,48 @@
...
+        Action,
+        PartitionName,
+        StudyInstanceUid,
+        SeriesInstanceUid,
+        SopInstanceUid,
+        OriginalWatermark,
+        CurrentWatermark
+    FROM dbo.ChangeFeed c
+    INNER JOIN dbo.Partition p
+    ON p.PartitionKey = c.PartitionKey
```
smithago, line 43: Please check if Timestamp can come out of order in the changefeed table

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/GetChangeFeedV33.sql
wsugarman: I've altered the clustered index to include an ascending order on `Timestamp`. This is the new query plan for fetching change feed:

image

wsugarman: While the previous v1 latest entry sproc has a decent query plan that scans the new nonclustered index for a single row (since it's ordered), I think the other get change feed sproc query plan can be improved.

### GetChangeFeedLatestV6
Seems ok

image

### GetChangeFeedV6
Seems less than ideal... with a lot of focus on grabbing the changes per partition:

image

Naively I would have thought the query plan for `GetChangeFeedV6` could be the same as `GetChangeFeedLatestV6` except that the top left `Index Scan` is replaced with an `Index Seek` + the sequence predicate

wsugarman:
I have simplified the approach by using the `Timestamp` sort for all API versions. The query plans are as follows:

build/common/versioning.yml
```diff
@@ -39,7 +39,7 @@ jobs:
 
   - task: PowerShell@2
     displayName: 'Check for breaking API / Swagger changes'
-    enabled: true
+    enabled: false
```
wsugarman, line 42: I am knowingly creating "breaking changes" as I am changing the type from `int` to `long` for `offset`. However, this doesn't practically affect any user as to them they can simply pass bigger values now

src/Microsoft.Health.Dicom.Api/Features/Conventions/ApiVersionsConvention.cs
```diff
@@ -85,4 +92,24 @@ internal static IReadOnlyList<ApiVersion> GetAllSupportedVersions(int start, int
             .ToList();
     }
 
+    // TODO: Can remove once _isLatestApiVersionEnabled is removed
```
wsugarman, line 95: Actually... this will need to stay lol

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/36.diff.sql
```diff
@@ -0,0 +1,748 @@
...
+GO
+
+DROP INDEX IF EXISTS IXC_ChangeFeed ON dbo.ChangeFeed
+DROP INDEX IF EXISTS IX_ChangeFeed_Sequence ON dbo.ChangeFeed
+
+CREATE UNIQUE CLUSTERED INDEX IXC_ChangeFeed ON dbo.ChangeFeed
+(
+    Timestamp,
+    Sequence
+)
```
smithago, line 25: You can do with(ONLINE=ON and DROP_EXISTING=TRUE)

There should be samples in other diff files

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/DeleteInstanceV36.sql
```diff
@@ -0,0 +1,266 @@
...
+--     @createdStatus
+--         * Status value representing the created state.
+--     @studyInstanceUid
+--         * The study instance UID.
+--     @seriesInstanceUid
+--         * The series instance UID.
+--     @sopInstanceUid
+--         * The SOP instance UID.
+/***************************************************************************************/
+CREATE OR ALTER PROCEDURE dbo.DeleteInstanceV36
```
smithago, line 25: We don't need tp version if the params are not changing

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/GetChangeFeedByTime.sql
```diff
@@ -0,0 +1,48 @@
...
+        SeriesInstanceUid,
+        SopInstanceUid,
+        OriginalWatermark,
+        CurrentWatermark
+    FROM dbo.ChangeFeed c WITH (HOLDLOCK)
+    INNER JOIN dbo.Partition p
+    ON p.PartitionKey = c.PartitionKey
+    WHERE c.Timestamp >= @startTime AND c.Timestamp < @endTime
+    ORDER BY Timestamp, Sequence
+    OFFSET @offset ROWS
```
smithago, line 46: nit: Can we add a comment on the fact that this will progressively get worse>

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/SqlExtendedQueryTagStoreV36.cs
```diff
@@ -0,0 +1,70 @@
...
+internal class SqlExtendedQueryTagStoreV36 : SqlExtendedQueryTagStoreV16
+{
+    public SqlExtendedQueryTagStoreV36(SqlConnectionWrapperFactory sqlConnectionWrapperFactory, ILogger<SqlExtendedQueryTagStoreV36> logger)
+        : base(sqlConnectionWrapperFactory, logger)
+    {
+    }
+
+    public override SchemaVersion Version => SchemaVersion.V36;
+
+    public override async Task<IReadOnlyList<ExtendedQueryTagStoreJoinEntry>> GetExtendedQueryTagsAsync(int limit, long offset = 0, CancellationToken cancellationToken = default)
```
smithago, line 32: same here

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStoreV36.cs
```diff
@@ -0,0 +1,148 @@
...
+internal class SqlIndexDataStoreV36 : SqlIndexDataStoreV35
+{
+    public SqlIndexDataStoreV36(SqlConnectionWrapperFactory sqlConnectionWrapperFactory)
+        : base(sqlConnectionWrapperFactory)
+    {
+    }
+
+    public override SchemaVersion Version => SchemaVersion.V36;
+
+    public override async Task EndCreateInstanceIndexAsync(
```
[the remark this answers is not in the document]
wsugarman: I got anxious because I updated these sprocs in two ways:
1. Moved the `ChangeFeed` `INSERT` to the end of the transaction
2. Use the new default column constraint to generate the timestamps upon `INSERT` instead of relying on one computed at the start of the transaction

(2) made me nervous about whether there was some race condition between the sproc being updated but the table not having the constraint yet... but I don't need to worry given the order of operations in the migration script

src/Microsoft.Health.Dicom.Core/Features/ChangeFeed/ChangeFeedService.cs
```diff
@@ -3,86 +3,77 @@
...
-
-        _changeFeedStore = changeFeedStore;
-        _metadataStore = metadataStore;
+        _changeFeedStore = EnsureArg.IsNotNull(changeFeedStore, nameof(changeFeedStore));
+        _metadataStore = EnsureArg.IsNotNull(metadataStore, nameof(metadataStore));
+        _options = EnsureArg.IsNotNull(options?.Value, nameof(metadataStore));
     }
 
-    public async Task<IReadOnlyCollection<ChangeFeedEntry>> GetChangeFeedAsync(long offset, int limit, bool includeMetadata, CancellationToken cancellationToken)
+    public async Task<IReadOnlyList<ChangeFeedEntry>> GetChangeFeedAsync(TimeRange range, long offset, int limit, ChangeFeedOrder order, bool includeMetadata, CancellationToken cancellationToken = default)
```
smithago, line 35: Are we logging the offset somewhere? Hopefully this helps to investigate SQL perf issues?

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/36.diff.sql
bcarthic: I think, it doesn't work for clustered index. I may be wrong

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/DeleteInstanceV36.sql
wsugarman: Done

src/Microsoft.Health.Dicom.Core/Features/ChangeFeed/ChangeFeedService.cs
wsugarman: We do log that information in the controller

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/GetExtendedQueryTagErrorsV36.sql
```diff
@@ -0,0 +1,57 @@
...
+--         * The TagPath for the extended query tag for which we retrieve error(s).
+--     @limit
+--         * The maximum number of results to retrieve.
+--     @offset
+--         * The offset from which to retrieve paginated results.
+--
+-- RETURN VALUE
+--     The tag error fields and the corresponding instance UIDs.
+/***************************************************************************************/
+CREATE OR ALTER PROCEDURE dbo.GetExtendedQueryTagErrorsV36
```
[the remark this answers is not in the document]
wsugarman: For the extended query tag sprocs (and in the generated workitem/qido queries), I've standardized the APIs to use `BIGINT` instead of `INT`:
image

src/Microsoft.Health.Dicom.SqlServer/Features/ExtendedQueryTag/SqlExtendedQueryTagStoreV36.cs
wsugarman: Same change from `INT` to `BIGINT`:
image

smithago (dismissed):

docs/concepts/change-feed.md
```diff
@@ -36,10 +38,20 @@ current  | This instance is the current version.
 replaced | This instance has been replaced by a new version.
 deleted  | This instance has been deleted and is no longer available in the service.
 
-### Read Change Feed
+## Change Feed
+The Change Feed resource is a collection of events that have occurred within the DICOM server.
+
+### Version 2
```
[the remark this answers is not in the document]
wsugarman: The conformance statement links to this document, but the public docs will need to be updated separately

docs/concepts/change-feed.md
```diff
@@ -62,26 +74,82 @@ deleted  | This instance has been deleted and is no longer available in the serv
...
+        }
+    },
+    {
+        "Sequence": 2,
+        "StudyInstanceUid": "{uid}",
+        "SeriesInstanceUid": "{uid}",
+        "SopInstanceUid": "{uid}",
+        "Action": "create|delete",
+        "Timestamp": "2020-03-05T07:13:16.4834Z",
+        "State": "current|replaced|deleted",
```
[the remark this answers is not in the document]
wsugarman: There is a table at the top of this doc for the value descriptions

docs/concepts/change-feed.md
```diff
@@ -62,26 +74,82 @@ deleted  | This instance has been deleted and is no longer available in the serv
...
+        }
+    },
+    // ...
+]
+```
 
+#### Parameters
+Name            | Type     | Description | Default | Min | Max |
+:-------------- | :------- | :---------- | :------ | :-- | :-- |
+offset          | int      | The exclusive starting sequence number for events | `0` | `0` | |
```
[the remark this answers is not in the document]
wsugarman: Maxiumum is left unspecified here, but it is really 9,223,372,036,854,775,807

docs/concepts/change-feed.md
```diff
@@ -62,26 +74,82 @@ deleted  | This instance has been deleted and is no longer available in the serv
...
+limit           | int      | The maximum number of records to return. There may be more results, even if the returned number is less than the limit | `10` | `1` | `100` |
```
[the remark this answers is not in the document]
wsugarman: I cleared up the defintion based on the existing semantics

src/Microsoft.Health.Dicom.Api.UnitTests/Controllers/ChangeFeedControllerTests.cs
```diff
@@ -0,0 +1,201 @@
...
+
+    [Fact]
+    public async Task GivenNoUserValues_WhenFetchingChangeFeed_ThenUseProperDefaults()
+    {
+        var expected = new List<ChangeFeedEntry>();
+
+        _mediator
+            .Send(
+                Arg.Is<ChangeFeedRequest>(x =>
+                    x.Range == TimeRange.MaxValue &&
```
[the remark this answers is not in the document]
wsugarman: Went with `GivenNoUserValues_WhenFetchingChangeFeed_ThenUseDefaultValues` if that sounds reasonable

src/Microsoft.Health.Dicom.Api.UnitTests/Controllers/ChangeFeedControllerTests.cs
```diff
@@ -0,0 +1,201 @@
...
+                    x.Limit == 10 &&
+                    x.IncludeMetadata),
+                _controller.HttpContext.RequestAborted)
+            .Returns(new ChangeFeedResponse(expected));
+
+        IActionResult result = await _controller.GetChangeFeedAsync();
+        var actual = result as ObjectResult;
+        Assert.Same(expected, actual.Value);
+
+        await _mediator
```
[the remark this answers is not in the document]
wsugarman: I'm not sure I follow. I am calling the method such that the defaults are populated, and then I assert that mediatr received the proper default values I am expecting

src/Microsoft.Health.Dicom.Api.UnitTests/Features/Conventions/ApiVersionsConventionTests.cs
```diff
@@ -16,99 +17,112 @@
...
+        };
+        var builder = new ControllerApiVersionConventionBuilder(typeof(TestController));
+        var featuresOptions = Options.Create(new FeatureConfiguration { EnableLatestApiVersion = enableLatest });
+        var convention = new ApiVersionsConvention(featuresOptions);
+
+        // act
+        convention.Apply(builder, controllerModel);
+
+        // assert
+        if (enableLatest)
```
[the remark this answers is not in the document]
wsugarman: Yeah, that makes sense. But I think I'll punt that to the PR that inevitably has to address the impact of updating our api convention logic

src/Microsoft.Health.Dicom.Api/Models/WindowedPaginationOptions.cs
```diff
@@ -0,0 +1,41 @@
...
+    public DateTimeOffset? EndTime { get; set; }
+
+    public TimeRange Window => new TimeRange(StartTime.GetValueOrDefault(DateTimeOffset.MinValue), EndTime.GetValueOrDefault(DateTimeOffset.MaxValue));
+
+    public IEnumerable<ValidationResult> Validate(ValidationContext validationContext)
+    {
+        DateTimeOffset start = StartTime.GetValueOrDefault(DateTimeOffset.MinValue);
+        DateTimeOffset end = EndTime.GetValueOrDefault(DateTimeOffset.MaxValue);
+
+        if (end <= start)
```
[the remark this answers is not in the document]
wsugarman: No, we don't care about the specifics of the time -- only that the start time is strictly before the end time. As a side note, the minimum time for `DateTimeOffset` in .NET is `0001-01-01T00:00:00Z`

src/Microsoft.Health.Dicom.Core.UnitTests/Features/ChangeFeed/ChangeFeedServiceTests.cs
```diff
@@ -18,140 +21,106 @@ namespace Microsoft.Health.Dicom.Core.UnitTests.Features.ChangeFeed;
...
-
-        _changeFeedStore.GetChangeFeedLatestAsync(cancellationToken: default)
-            .ReturnsForAnyArgs(callInfo => _changeFeedEntries.Last());
-
-        _metadataStore = Substitute.For<IMetadataStore>();
-        _changeFeedService = new ChangeFeedService(_changeFeedStore, _metadataStore);
+        _changeFeedService = new ChangeFeedService(
+            _changeFeedStore,
+            _metadataStore,
+            Options.Create(new RetrieveConfiguration { MaxDegreeOfParallelism = -1 }));
```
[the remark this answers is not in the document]
wsugarman: -1 is unbounded. I'll leave a comment

src/Microsoft.Health.Dicom.Core/Features/ChangeFeed/ChangeFeedService.cs
```diff
@@ -3,86 +3,77 @@
...
-    private async Task PopulateMetadata(ChangeFeedEntry entry, CancellationToken cancellationToken)
+    private async ValueTask PopulateMetadata(ChangeFeedEntry entry, CancellationToken cancellationToken)
     {
-        if (entry.State == ChangeFeedState.Deleted || entry.CurrentVersion == null)
-        {
+        if (entry.CurrentVersion == null)
             return;
-        }
 
         var identifier = new VersionedInstanceIdentifier(entry.StudyInstanceUid, entry.SeriesInstanceUid, entry.SopInstanceUid, entry.CurrentVersion.Value);
```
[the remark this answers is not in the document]
wsugarman: Me too! I'll do that lol

src/Microsoft.Health.Dicom.Core/Models/TimeRange.cs
```diff
@@ -0,0 +1,50 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.Health.Dicom.Core.Models;
+
+public readonly struct TimeRange : IEquatable<TimeRange>
+{
+    public static readonly TimeRange MaxValue = new TimeRange(DateTimeOffset.MinValue, DateTimeOffset.MaxValue);
```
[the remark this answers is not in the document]
wsugarman: Strangely no 😢

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/36.diff.sql
```diff
@@ -0,0 +1,754 @@
+SET XACT_ABORT ON
+
+CREATE UNIQUE CLUSTERED INDEX IXC_ChangeFeed ON dbo.ChangeFeed
```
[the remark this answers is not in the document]
wsugarman: I don't think `ALTER INDEX` supports adding columns

smithago (approved):

smithago (approved):

PR #2623, opened 2023-04-28
Exception handling for existing update operation

arunmk-ms:
## Description
Exception handling for existing update operation

## Related issues
Addresses [issue #103043](https://microsofthealth.visualstudio.com/Health/_workitems/edit/103043/).

## Testing
Manual testing


src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
```diff
@@ -106,6 +106,7 @@ private IActionResult MapExceptionToResult(Exception exception)
             case InstanceAlreadyExistsException:
             case ExtendedQueryTagsAlreadyExistsException:
             case ExtendedQueryTagsOutOfDateException:
+            case ExistingUpdateOperationException:
```
bcarthic, line 109: Can we add a test?

wsugarman, line 109: Instead of adding more things here, just have `ExistingUpdateOperationException` and `ExistingReindexOperationException` derive from `InvalidOperationException` which is already tracked here (and is semantically appropriate).

wsugarman (changes requested):

src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
arunmk-ms: InvalidOperationException sends a 400 whereas the other two should be a conflict.

wsugarman: I am unsure if a conflict makes the most sense. A conflict implies that a resource at the URL already exists, but for update I don't see any semantic reason why you couldn't submit two updates in quick succession other than performance (and perhaps ordering complexities on the backend). If the second update request will succeed if I just wait long enough, a "conflict" doesn't seem like the reason we rejected it.

arunmk-ms: Synced offline. We already have a test for the same.

arunmk-ms: As per our current design, we allow only one request to be queued and there is no check on the study instance ids that we receive in the request. I feel a conflict on the same update request makes more sense at this time, than a bad request.

wsugarman: I suppose we can argue that the Conflict is appropriate because the state of the resource is in flux during update, and an additional update may conflict with what is being applied. In this case, can you please update re-indexing's exception to perhaps share some common exception that we can catch in the `ExceptionHandlingMiddleware`. Or better yet create a generic shared exception that is used in both scenarios and get rid of the operation-specific exception. E.g. `OngoingOperationException` or something like that.

arunmk-ms: Created a new parent `ExisitngOperationException` class and added to exception middleware.

src/Microsoft.Health.Dicom.Core/Exceptions/ExistingReindexException.cs
```diff
@@ -13,32 +13,19 @@ namespace Microsoft.Health.Dicom.Core.Exceptions;
...
     /// Initializes a new instance of the <see cref="ExistingReindexException"/> class.
     /// </summary>
     /// <param name="operation">The operation reference for the existing re-index operation.</param>
     /// <exception cref="ArgumentNullException"><paramref name="operation"/> is <see langword="null"/>.</exception>
     public ExistingReindexException(OperationReference operation)
         : base(
+            operation,
             string.Format(
                 CultureInfo.CurrentCulture,
                 DicomCoreResource.ExistingReindexOperation,
```
wsugarman, line 28: Do we need unique messages, or can they combined since we know the operation name based on the exception

arunmk-ms: I think it is better to give unique messages for the customer as below
`There is already an active update operation with ID '{0}'`
or
`There is already an active re-index operation with ID '{0}'`
We are not giving any of the exception details in the response.

arunmk-ms: Let me know if you have a different thought.

wsugarman: Couldn't we make the names `"update"` and `"re-index"` placeholders in the format string like `"There is already an active {0} operation with ID '{1}'"`

src/Microsoft.Health.Dicom.Core/Exceptions/ExistingReindexException.cs
```diff
@@ -13,32 +13,20 @@ namespace Microsoft.Health.Dicom.Core.Exceptions;
...
     /// <param name="operation">The operation reference for the existing re-index operation.</param>
     /// <exception cref="ArgumentNullException"><paramref name="operation"/> is <see langword="null"/>.</exception>
     public ExistingReindexException(OperationReference operation)
         : base(
+            operation,
             string.Format(
                 CultureInfo.CurrentCulture,
-                DicomCoreResource.ExistingReindexOperation,
+                DicomCoreResource.ExistingOperation,
+                "re-index",
```
wsugarman, line 29: Couldn't this be an argument?

wsugarman (changes requested):

wsugarman (approved):

PR #2641, opened 2023-05-05
allow null padding on attributes for v2

esmadau:
## Description
- Allow null padding on all values when using API V2
- When null padding detected, we will not drop data from metadata json, and we will go ahead and index it, as is
- No warnings will be produced
- Update to conformance statement

## Related issues
Addresses [[AB#103254](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/103254)].

## Testing
unit and e2e tests

### DIFF
--- a/docs/resources/conformance-statement.md
+++ b/docs/resources/conformance-statement.md
@@ -21,7 +21,7 @@ Additionally, the following non-standard API(s) are supported:
 
 All paths below include an implicit base URL of the server, such as `https://localhost:63838` when running locally.
 
-The service makes use of REST Api versioning. Do note that the version of the REST API must be explicitly specified as part of the base URL as in the following example: 
+The service makes use of REST Api versioning. Do note that the version of the REST API must be explicitly specified as part of the base URL as in the following example:
 
 `https://localhost:63838/v1/studies`
 
@@ -69,7 +69,7 @@ The following DICOM elements are required to be present in every DICOM file atte
 - SOPClassUID
 - PatientID
 
-> Note: All identifiers must be between 1 and 64 characters long, and only contain alpha numeric characters or the following special characters: `.`, `-`.
+> Note: All UIDs must be between 1 and 64 characters long, and only contain alpha numeric characters or the following special characters: `.`, `-`. PatientID is validated based on its LO VR type.
 
 Each file stored must have a unique combination of StudyInstanceUID, SeriesInstanceUID and SopInstanceUID. The warning code `45070` will be returned if a file with the same identifiers already exists.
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Dicom.Core/Features/Store/StoreDatasetValidator.cs
```diff
@@ -81,9 +81,13 @@ public class StoreDatasetValidator : IStoreDatasetValidator
         }
 
         // validate input data elements
-        if (EnableDropMetadata(_dicomRequestContextAccessor.RequestContext.Version) || _enableFullDicomItemValidation)
+        if (_enableFullDicomItemValidation)
```
smithago, line 84: We don't have this FF enabled for anyone. Should we just remove this? We started with this 2 years hoping someone will ask for it. 
We are not even test this case I think.

If it simplifies we should consider removing it.

There are too many if else in this method

src/Microsoft.Health.Dicom.Core/Features/Store/StoreDatasetValidator.cs
```diff
@@ -194,36 +197,64 @@ string EnsureRequiredTagIsPresent(DicomTag dicomTag)
...
+                    TelemetryDimension(item, isIndexableTag));
+            }
         }
     }
 
+    private void ValidateWithoutNullPadding(string value, DicomItem item, IReadOnlyCollection<QueryTag> queryTags)
+    {
+        // we still want to save the original value when validation passes so create new ds to validate with
+        var newValue = value.TrimEnd('\0');
+        DicomDataset ds = new DicomDataset().NotValidated();
```
smithago, line 240: Do you need to create the DicomDataset object here? Looks like you just need DicomItem with the tag and trimmed value to do the validation.

smithago: I agree, I would not go to reflection, How about this, change your for - loop

//Change to DicomElement, it is better object DicomItem just as tag.
foreach (DicomElement item in dicomDataset)
{

try
{
IgnoredNullPaddingValidation(DicomElement)
}
catch
{
}
}

Now change your method
private void IgnoredNullPaddingValidation(DicomElement de)
{
       var stringValue = de.Get<string>(-1);
    
       if (stringValue.EndsWith('\0'))
       {
           //log
           de.ValueRepresentation.ValidateString(stringValue.TrimEnd('\0'));
       }
       else
       {
             de.Validate()
        }
}

smithago (approved):

docs/resources/conformance-statement.md
```diff
@@ -69,7 +69,7 @@ The following DICOM elements are required to be present in every DICOM file atte
 - SOPClassUID
 - PatientID
 
-> Note: All identifiers must be between 1 and 64 characters long, and only contain alpha numeric characters or the following special characters: `.`, `-`.
+> Note: All UIDs must be between 1 and 64 characters long, and only contain alpha numeric characters or the following special characters: `.`, `-`. PatientID is validated based on its LO VR type.
```
jovinson-ms, line 72: since we're updating this language, we should check this - I think UIDs can only contain numeric and `.`. Just ran some tests where UIDs with alpha or `-` failed.

PR #2672, opened 2023-05-11
Use custom dicom implementation and version name

bcarthic:
## Description
This PR includes changes to use microsoft specific implementation class uid and version name

## Related issues
Addresses [[AB#103060](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/103060)].

## Testing
Added new unit tests and E2E tests


src/Microsoft.Health.Dicom.Core/Features/FellowOakDicom/CustomDicomImplementation.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// <summary>
+    /// This method sets the DICOM implementation class UID and version.
+    /// ImplementationClassUID and ImplementationVersion are used to identify the software that generated or last touched the data.
+    /// </summary>
+    public static void SetDicomImplementationClassUIDAndVersion()
+    {
+        Version version = typeof(CustomDicomImplementation).GetTypeInfo().Assembly.GetName().Version;
+
+        DicomImplementation.ClassUID = new DicomUID(ImplementationClassUid, "Implementation Class UID", DicomUidType.Unknown);
+        DicomImplementation.Version = $"{version.Major}.{version.Minor}.{version.Build}";
```
wsugarman, line 32: Should this be `{Major}.{Minor}.{Patch}+{Build}` like https://semver.org/? Or should it just be `{Major}.{Minor}.{Patch}`?

bcarthic: I don't have an opinion here, just followed the fo-dicom. But I do agree we should do similar to semver. I also see there is a restriction on number of characters which is 16. Sometimes our custom build will take more characters. I will stick to {Major}.{Minor}.{Patch}

wsugarman (approved):

PR #2697, opened 2023-05-16
Ensure all validators handle null same way

esmadau:
## Description
Check for value being null before checking if it ends in null padding in string.
MinimumValidator also checks for nulls before calling validate to ensure we have consistency.

Issue submitted to fo-dicom. Depending on how we call validation, sometimes nulls are handled gracefully: https://github.com/fo-dicom/fo-dicom/issues/1590

## Related issues
Addresses [[AB#103623](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/103623)].

## Testing
added unit tests


src/Microsoft.Health.Dicom.Core.UnitTests/Features/Validation/EncodedStringElementValidationTests.cs
```diff
@@ -45,6 +45,7 @@ public void GivenDicomStringElementWithMultipleValues_WhenValidating_ThenShouldV
         new object[] { new DicomDateTime(DicomTag.EffectiveDateTime, DateTimeOffset.UtcNow.ToString("yyyyMMddHHmmss'.'ffffff'+'0000", CultureInfo.InvariantCulture)) },
         new object[] { new DicomIntegerString(DicomTag.PixelAspectRatio, "0012345") },
         new object[] { new DicomTime(DicomTag.Time, DateTime.UtcNow.ToString("HHmmss'.'fffff", CultureInfo.InvariantCulture)) },
+        new object[] { new DicomTime(DicomTag.Time, (string)null )},
```
jovinson-ms, line 48: I don't think we need to cast the `null` here.

PR #2706, opened 2023-05-18
External Store - start using instancekey

esmadau:
## Description
start retrieving instanceKey when index created on stow and use it to populate FileProperty table

## Related issues
Addresses [issue #].

## Testing

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Features/Store/StoreOrchestrator.cs
+++ b/src/Microsoft.Health.Dicom.Core/Features/Store/StoreOrchestrator.cs
@@ -74,27 +74,23 @@ public class StoreOrchestrator : IStoreOrchestrator
         var partitionKey = _contextAccessor.RequestContext.GetPartitionKey();
 
         IReadOnlyCollection<QueryTag> queryTags = await _queryTagService.GetQueryTagsAsync(cancellationToken: cancellationToken);
-        long watermark = await _indexDataStore.BeginCreateInstanceIndexAsync(partitionKey, dicomDataset, queryTags, cancellationToken);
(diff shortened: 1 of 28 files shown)


src/Microsoft.Health.Dicom.Core/Features/Store/StoreOrchestrator.cs
```diff
@@ -74,27 +74,23 @@ CancellationToken cancellationToken
...
-        var versionedInstanceIdentifier = dicomDataset.ToVersionedInstanceIdentifier(watermark);
+        InstanceProperties instanceProperties = await _indexDataStore.BeginCreateInstanceIndexAsync(partitionKey, dicomDataset, queryTags, cancellationToken);
+        var versionedInstanceIdentifier = dicomDataset.ToVersionedInstanceIdentifier((long)instanceProperties.NewVersion);
 
         try
         {
             // We have successfully created the index, store the files.
             Task<FileProperties> storeFileTask = StoreFileAsync(versionedInstanceIdentifier, dicomInstanceEntry, cancellationToken);
-            Task<bool> frameRangeTask = StoreFileFramesRangeAsync(dicomDataset, watermark, cancellationToken);
+            Task<bool> frameRangeTask = StoreFileFramesRangeAsync(dicomDataset, (long)instanceProperties.NewVersion, cancellationToken);
```
bcarthic, line 84: I wouldn't use New version for this. Newversion is created for intermediate storing and it could be null.

bcarthic:
And one more suggestion is to put all the db schema changes in one PR to avoid multiple schema changes.

PR #2734, opened 2023-05-26
Diagnostic logging for Dicom update

arunmk-ms:
## Description
Enable diagnostic logging for Dicom update
Updated tests

## Related issues
Addresses [103949](https://microsofthealth.visualstudio.com/Health/_workitems/edit/103949)

## Testing
Manually testing Dicom update feature.

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Features/Diagnostic/LogForwarderExtensions.cs
+++ b/src/Microsoft.Health.Dicom.Core/Features/Diagnostic/LogForwarderExtensions.cs
@@ -22,6 +22,8 @@ internal static class LogForwarderExtensions
     private const string StudyInstanceUID = $"{Prefix}studyInstanceUID";
     private const string SeriesInstanceUID = $"{Prefix}seriesInstanceUID";
     private const string SOPInstanceUID = $"{Prefix}sopInstanceUID";
+    private const string InputPayload = $"{Prefix}input";
+    private const string OperationId = $"{Prefix}operationId";
 
     /// <summary>
     /// Emits a trace log with forwarding flag set and adds properties from instanceIdentifier as properties to telemetry.
@@ -66,11 +68,11 @@ internal static class LogForwarderExtensions
         EnsureArg.IsNotNull(message, nameof(message));
         EnsureArg.IsNotNull(jsonSerializerOptions?.Value, nameof(jsonSerializerOptions));
 
-        string property = JsonSerializer.Serialize(value, jsonSerializerOptions.Value);
+        string input = JsonSerializer.Serialize(value, jsonSerializerOptions.Value);
 
         var telemetry = new TraceTelemetry(message);
-        telemetry.Properties.Add(nameof(T), property);
-        telemetry.Properties.Add("operation_id", operationId);
+        telemetry.Properties.Add(InputPayload, input);
+        telemetry.Properties.Add(OperationId, operationId);
         telemetry.Properties.Add(ForwardLogFlag, bool.TrueString);
 
         telemetryClient.TrackTrace(telemetry);
--- a/src/Microsoft.Health.Dicom.Core/Features/Diagnostic/LogForwarderExtensions.cs
+++ b/src/Microsoft.Health.Dicom.Core/Features/Diagnostic/LogForwarderExtensions.cs
@@ -3,6 +3,7 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System;
 using EnsureThat;
 using Microsoft.ApplicationInsights;
 using Microsoft.ApplicationInsights.DataContracts;
@@ -15,6 +16,7 @@ namespace Microsoft.Health.Dicom.Core.Features.Diagnostic;
 /// </summary>
 internal static class LogForwarderExtensions
 {
+    private const int MaxShoeboxPropertySize = 32 * 1024;
     private const string ForwardLogFlag = "forwardLog";
     private const string Prefix = "dicomAdditionalInformation_";
     private const string StudyInstanceUID = $"{Prefix}studyInstanceUID";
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.Dicom.Core/Features/Diagnostic/LogForwarderExtensions.cs
```diff
@@ -44,4 +45,33 @@ internal static class LogForwarderExtensions
...
+        this TelemetryClient telemetryClient,
+        string message,
+        string operationId,
+        T value)
+    {
+        EnsureArg.IsNotNull(telemetryClient, nameof(telemetryClient));
+        EnsureArg.IsNotNull(message, nameof(message));
+
+        string property = JsonConvert.SerializeObject(value, Formatting.None,
+            new JsonSerializerSettings
```
bcarthic, line 66: there is a serialization options from DI  which you can use

src/Microsoft.Health.Dicom.Core/Features/Diagnostic/LogForwarderExtensions.cs
```diff
@@ -44,4 +46,33 @@ internal static class LogForwarderExtensions
...
+        IOptions<JsonSerializerOptions> jsonSerializerOptions)
+    {
+        EnsureArg.IsNotNull(telemetryClient, nameof(telemetryClient));
+        EnsureArg.IsNotNull(message, nameof(message));
+        EnsureArg.IsNotNull(jsonSerializerOptions?.Value, nameof(jsonSerializerOptions));
+
+        string property = JsonSerializer.Serialize(value, jsonSerializerOptions.Value);
+
+        var telemetry = new TraceTelemetry(message);
+        telemetry.Properties.Add(nameof(T), property);
```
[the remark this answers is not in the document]
arunmk-ms: Agree, I changed it to `input` now.

src/Microsoft.Health.Dicom.Core/Features/Update/UpdateInstanceOperationService.cs
```diff
@@ -88,6 +100,9 @@ public class UpdateInstanceOperationService : IUpdateInstanceOperationService
         try
         {
             var operation = await _client.StartUpdateOperationAsync(operationId, updateSpecification, partitionKey, cancellationToken);
+
+            string input = JsonSerializer.Serialize(updateSpecification, _jsonSerializerOptions.Value);
+            _telemetryClient.ForwardLogTrace("Dicom update operation", operationId.ToString(), input);
```
bcarthic, line 105: As you mentioned earlier, how are we going handle if the payload size larger than 32KB?

src/Microsoft.Health.Dicom.Core/Features/Diagnostic/LogForwarderExtensions.cs
```diff
@@ -44,4 +46,28 @@ internal static class LogForwarderExtensions
 
         telemetryClient.TrackTrace(telemetry);
     }
+
+    /// <summary>
+    /// Emits a trace log with forwarding flag set and adds properties from instanceIdentifier as properties to telemetry.
```
bcarthic, line 51: Its not instanceidentitfier

src/Microsoft.Health.Dicom.Core/Features/Diagnostic/LogForwarderExtensions.cs
```diff
@@ -44,4 +46,28 @@ internal static class LogForwarderExtensions
...
+    /// </summary>
+    /// <param name="telemetryClient">client to use to emit the trace</param>
+    /// <param name="message">message to set on the trace log</param>
+    /// <param name="operationId">operation id</param>
+    /// <param name="input">Input payload to pass to the forward logger</param>
+    public static void ForwardLogTrace(
```
bcarthic, line 57: Since we include operationId, this doesnt feel like generic method. Should we scope the method name to be like ForwardOperationLogTrace? Or do you want to include the operationId as part of the input?

src/Microsoft.Health.Dicom.Core/Features/Diagnostic/LogForwarderExtensions.cs
```diff
@@ -44,4 +46,28 @@ internal static class LogForwarderExtensions
...
+        this TelemetryClient telemetryClient,
+        string message,
+        string operationId,
+        string input)
+    {
+        EnsureArg.IsNotNull(telemetryClient, nameof(telemetryClient));
+        EnsureArg.IsNotNull(message, nameof(message));
+
+        var telemetry = new TraceTelemetry(message);
+        telemetry.Properties.Add(InputPayload, input);
```
wsugarman, line 67: Can this really be so generic to include input? The operation input contains data the function uses for checkpointing and data the user doesn't know about. Do we really want to expose this via the logs?

Do we include the input for the web apis?

wsugarman: What is the requirement? We need to record what is being updated? If so:
- Record the user's input as part of the audit log AND/OR
- Record an "input" payload to telemetry that is tailored to someone reading it via the logs (i.e. it isn't coupled with the function input)

src/Microsoft.Health.Dicom.Core/Features/Diagnostic/LogForwarderExtensions.cs
```diff
@@ -44,4 +48,44 @@ internal static class LogForwarderExtensions
...
+    {
+        EnsureArg.IsNotNull(telemetryClient, nameof(telemetryClient));
+        EnsureArg.IsNotNull(message, nameof(message));
+
+        // Shoebox property size has a limitation of 32 KB which is why the diagnostic log is split into multiple messages
+        int startIndex = 0, offset = 0, inputSize = input.Length;
+        while (startIndex < inputSize)
+        {
+            offset = Math.Min(MaxShoeboxPropertySize, input.Length - startIndex);
+            ForwardOperationLogTraceWithSizeLimit(telemetryClient, message, operationId, input.Substring(startIndex, offset));
```
wsugarman, line 73: What is the usability of this? lol
A bunch of copies of the log with partial inputs? Is there an alternative?

arunmk-ms: There is a 32KB limitation for the property added, which is why we split the input message into multiple diagnostic messages. We discussed multiple approaches like either truncate anything after 32KB or split the parameters and add them separately. But the one we are currently doing seems to be the better approach.

PR #2752, opened 2023-06-01
Skip emitting metric during function replay

arunmk-ms:
## Description
Dicom update function was emitting the metric more than once. As part of this PR, the metric will be sent only once and skipped during replay.

## Related issues
Addresses [issue [102011].](https://microsofthealth.visualstudio.com/Health/_workitems/edit/102011)

## Testing
Tested and verified the metrics manually.


src/Microsoft.Health.Dicom.Functions/Registration/ReplaySafeUpdateMeter.cs
```diff
@@ -0,0 +1,36 @@
...
+// -------------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using EnsureThat;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask;
+using Microsoft.Health.Dicom.Core.Features.Telemetry;
+
+namespace Microsoft.Health.Dicom.Functions.Registration;
+
+public class ReplaySafeUpdateMeter
```
wsugarman, line 13: I don't think this is the abstraction we want to be "replay-safe" as it's not reusable. Instead, this should probably be a Counter

arunmk-ms: Thanks for pointing it out. I have made changed it to a counter.

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Orchestration.cs
```diff
@@ -101,8 +103,7 @@ public partial class UpdateDurableFunction
 
             if (instanceWatermarks.Count > 0)
             {
-                _updateMeter.UpdatedInstances.Add(instanceWatermarks.Count,
-                    new KeyValuePair<string, object>("ExecutionId", context.NewGuid()));
+                replaySafeCounter.Add(instanceWatermarks.Count);
```
bcarthic, line 106: This is great, how do we know for a particular orchestration if the counter is called once or twice?

src/Microsoft.Health.Dicom.Functions/Registration/ReplaySafeCounter.cs
```diff
@@ -0,0 +1,34 @@
...
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System.Diagnostics.Metrics;
+using EnsureThat;
+using Microsoft.Azure.WebJobs.Extensions.DurableTask;
+
+namespace Microsoft.Health.Dicom.Functions.Registration;
+
+public class ReplaySafeCounter
```
wsugarman, line 12: Shouldn't this be generic for Counter?

src/Microsoft.Health.Dicom.Functions/Registration/ReplaySafeCounter.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    private readonly IDurableOrchestrationContext _context;
+
+    private readonly Counter<int> _counter;
+
+    internal ReplaySafeCounter(IDurableOrchestrationContext context, Counter<int> counter)
+    {
+        EnsureArg.IsNotNull(context, nameof(context));
```
wsugarman, line 20: nit: combine assignment and null check

src/Microsoft.Health.Dicom.Functions/Registration/ReplaySafeCounter.cs
wsugarman, line 12: It's unfortunate there's no interface/class that can be used to make our counter look like the underlying counter. Perhaps a comment here to stay that the Counter class is sealed and other Counter methods can be added as necessary

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Orchestration.cs
wsugarman: This is about replay -- for when the orchestration is rebuilding state -- not necessarily for failure. There is no signal for whether the orchestration restarted and already sent a counter that is not part of a replay

wsugarman (approved):

bcarthic (approved):

PR #2753, opened 2023-06-02
Prioritize original transfer syntax when multiple accepts with no quality provided

smithago:
## Description
SLIM viewer sends multiple accept headers to get frames. Original is the last, this causes us to pick another accept and fail because today our code needs entire file to do transcoding and will throw when file size is greater than 100M.

This fix will make us prioritize picking original if available in the accept list.

## Related issues
https://github.com/ImagingDataCommons/dicom-microscopy-viewer/issues/95

## Testing
Added unit tests

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Features/Retrieve/AcceptHeaderHandler.cs
+++ b/src/Microsoft.Health.Dicom.Core/Features/Retrieve/AcceptHeaderHandler.cs
@@ -59,7 +59,7 @@ public class AcceptHeaderHandler : IAcceptHeaderHandler
     public AcceptHeader GetValidAcceptHeader(ResourceType resourceType, IReadOnlyCollection<AcceptHeader> acceptHeaders)
     {
         EnsureArg.IsNotNull(acceptHeaders, nameof(acceptHeaders));
-        List<AcceptHeader> orderedHeaders = acceptHeaders.OrderByDescending(x => x.Quality).ToList();
+        List<AcceptHeader> orderedHeaders = acceptHeaders.OrderByDescending(x => x.Quality ?? 0.000).ToList();
 
         _logger.LogInformation(
             "Getting transfer syntax for retrieving {ResourceType} with accept headers {AcceptHeaders}.",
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.Dicom.Core/Messages/Retrieve/AcceptHeader.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.Health.Dicom.Core.Messages.Retrieve;
 public class AcceptHeader
 {
     public const double DefaultQuality = 1.0;
-    public AcceptHeader(StringSegment mediaType, PayloadTypes payloadType, StringSegment transferSyntax = default, double? quality = DefaultQuality)
+    public AcceptHeader(StringSegment mediaType, PayloadTypes payloadType, StringSegment transferSyntax = default, double? quality = null)
```
jovinson-ms, line 14: The default quality should be set at `1.0` per the standard

src/Microsoft.Health.Dicom.Core/Features/Retrieve/AcceptHeaderHandler.cs
```diff
@@ -68,26 +68,42 @@ public AcceptHeader GetValidAcceptHeader(ResourceType resourceType, IReadOnlyCol
 
         List<AcceptHeaderDescriptor> descriptors = _acceptableDescriptors[resourceType];
 
+        AcceptHeader selectedHeader = null;
         // since these are ordered by quality already, we can return the first acceptable header
```
jovinson-ms, line 72:
```suggestion
        // we will return the highest priority media type we support, breaking ties by favoring the original media type.
```

src/Microsoft.Health.Dicom.Core/Features/Retrieve/AcceptHeaderHandler.cs
```diff
@@ -68,26 +68,42 @@ public AcceptHeader GetValidAcceptHeader(ResourceType resourceType, IReadOnlyCol
...
+
         //  none were valid
         throw new NotAcceptableException(DicomCoreResource.NotAcceptableHeaders);
     }
 
+    // if no quality provided prioritize returning original transfer syntax
+    private static bool IsHigherPriorityTransferSyntax(AcceptHeader header, AcceptHeader selectedHeader)
+    {
+        bool isQualityGreater = (header.Quality ?? 0.000) >= (selectedHeader.Quality ?? 0.000);
+        return (header.TransferSyntax.Value == DicomTransferSyntaxUids.Original && isQualityGreater);
```
jovinson-ms, line 104: This naming is misleading: `Original` should really be called `Any`. If we want to prioritize the original, we should look up the original transfer syntax from the DB.

smithago: Hmm, I am not seeing the difference. It is still saying the same without knowing the exact transfer syntax UID.

jovinson-ms: This logic will prioritize the original syntax if the request specifies any transfer syntax for the media type via `*`. 

However, if we have an image stored in `Explicit VR Little Endian` and a caller sends the following in the `Accept` header:
- media type `application/dicom`, transfer syntax `JPEG 2000`, no qvalue
- media type `application/dicom`, transfer syntax `Explicit VR Little Endian`, no qvalue

Then this tie-breaking logic will pick `JPEG 2000` as the transfer syntax to return and we will transcode.

smithago: Correct. The goal is not to avoid transcoding, it is to prioritize * when given in a list of media type with no q value

jovinson-ms (approved):

PR #2766, opened 2023-06-06
Correcting naming of Original to Any

jovinson-ms:
## Description
When transfer syntax is requested, the client can send `*` to specify that any transfer syntax is acceptable:
>The wildcard value "*" indicates that the user agent will accept any Transfer Syntax. [Standard](https://dicom.nema.org/medical/dicom/current/output/html/part18.html#sect_8.7.3.5.2)

Our service currently returns the original transfer syntax in this case, but we should use the same semantics as the standard in our code to minimize confusion and be able to handle different behavior in the future.

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Features/Retrieve/IFrameHandler.cs
+++ b/src/Microsoft.Health.Dicom.Core/Features/Retrieve/IFrameHandler.cs
@@ -11,5 +11,5 @@ namespace Microsoft.Health.Dicom.Core.Features.Retrieve;
 
 public interface IFrameHandler
 {
-    Task<IReadOnlyCollection<Stream>> GetFramesResourceAsync(Stream stream, IEnumerable<int> frames, bool originalTransferSyntaxRequested, string requestedRepresentation);
+    Task<IReadOnlyCollection<Stream>> GetFramesResourceAsync(Stream stream, IEnumerable<int> frames, bool originalTransferSyntaxSelected, string requestedTransferSyntax);
 }
--- a/src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
+++ b/src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
@@ -89,7 +89,7 @@ public class RetrieveResourceService : IRetrieveResourceService
                 message.AcceptHeaders);
 
             string requestedTransferSyntax = validAcceptHeader.TransferSyntax.Value;
-            bool isOriginalTransferSyntaxRequested = DicomTransferSyntaxUids.IsOriginalTransferSyntaxRequested(requestedTransferSyntax);
+            bool isOriginalTransferSyntaxRequested = DicomTransferSyntaxUids.IsAnyTransferSyntaxRequested(requestedTransferSyntax);
 
             if (message.ResourceType == ResourceType.Frames)
             {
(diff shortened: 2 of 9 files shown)


src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
```diff
@@ -89,7 +89,7 @@ public async Task<RetrieveResourceResponse> GetInstanceResourceAsync(RetrieveRes
                 message.AcceptHeaders);
 
             string requestedTransferSyntax = validAcceptHeader.TransferSyntax.Value;
-            bool isOriginalTransferSyntaxRequested = DicomTransferSyntaxUids.IsOriginalTransferSyntaxRequested(requestedTransferSyntax);
+            bool isOriginalTransferSyntaxRequested = DicomTransferSyntaxUids.IsAnyTransferSyntaxRequested(requestedTransferSyntax);
```
smithago, line 92: I think Any is confusing. Original is more accurate. Any means we can transcode and do anything. Original means we are returning what was stored.

here is a example of another repo where same naming convesion is used,

jovinson-ms: There's two concepts here that I think should be distinguished, customer input and our choice: 
- The customer, by sending `*`, is requesting any transfer syntax according to the standard.
- Our service is choosing to break ties by returning the original transfer syntax, but this is not strictly what the customer is requesting - we are making that decision.  

The goal here is to clarify what is the customer input and what is our choice, because someone looking at the code to understand how we pick return syntax will need to understand that distinction.

src/Microsoft.Health.Dicom.Core/Features/Retrieve/IFrameHandler.cs
```diff
@@ -11,5 +11,5 @@ namespace Microsoft.Health.Dicom.Core.Features.Retrieve;
 
 public interface IFrameHandler
 {
-    Task<IReadOnlyCollection<Stream>> GetFramesResourceAsync(Stream stream, IEnumerable<int> frames, bool originalTransferSyntaxRequested, string requestedRepresentation);
+    Task<IReadOnlyCollection<Stream>> GetFramesResourceAsync(Stream stream, IEnumerable<int> frames, bool originalTransferSyntaxSelected, string requestedTransferSyntax);
```
jovinson-ms, line 14: Changed the names of these parameters to clarify intent - at this point, the `RetrieveResourceService` has evaluated the request, and selected to return the original transfer syntax.

src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
```diff
@@ -89,15 +89,14 @@ public async Task<RetrieveResourceResponse> GetInstanceResourceAsync(RetrieveRes
...
-            bool isOriginalTransferSyntaxRequested = DicomTransferSyntaxUids.IsOriginalTransferSyntaxRequested(requestedTransferSyntax);
+            bool isAnyTransferSyntaxRequested = DicomTransferSyntaxUids.IsAnyTransferSyntaxRequested(requestedTransferSyntax);
 
             if (message.ResourceType == ResourceType.Frames)
             {
                 return await GetFrameResourceAsync(
                     message,
                     partitionKey,
                     requestedTransferSyntax,
-                    isOriginalTransferSyntaxRequested,
```
jovinson-ms, line 100: Simplifies method signatures by doing an evaluation at the end of the call chain rather than passing the value down.

PR #2767, opened 2023-06-06
iterate through sequences and allow for null padding with inner elements

esmadau:
## Description
Given a DicomSequence in a dataset, when the sequence contains an element with a string value padded with null, we want to validate without the null padding.

## Related issues
Addresses [[AB#104213](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/104213)].

## Testing
Added unit tests

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Features/Store/StoreDatasetValidator.cs
+++ b/src/Microsoft.Health.Dicom.Core/Features/Store/StoreDatasetValidator.cs
@@ -216,38 +216,43 @@ public class StoreDatasetValidator : IStoreDatasetValidator
             // add to stack to keep iterating when SQ type, otherwise validate
             foreach (DicomItem item in ds)
             {
-                try
+                if (item.ValueRepresentation == DicomVR.SQ)
                 {
-                    if (item.ValueRepresentation == DicomVR.SQ)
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.Dicom.Core/Features/Store/StoreDatasetValidator.cs
```diff
@@ -207,32 +207,43 @@ string EnsureRequiredTagIsPresent(DicomTag dicomTag)
...
     {
         IReadOnlyCollection<QueryTag> queryTags = await _queryTagService.GetQueryTagsAsync();
-        foreach (DicomElement de in dicomDataset)
+        foreach (DicomItem item in dicomDataset)
         {
             try
             {
-                string value = dicomDataset.GetString(de.Tag);
-                if (value != null && value.EndsWith('\0'))
+                if (item.ValueRepresentation == DicomVR.SQ)
```
wsugarman, line 214: While not likely, it could be that we've been given a DICOM dataset with more than expected depth, so I think we have to make this iterative. Although I like the use of recursion, and it makes the intent quite clear

wsugarman (approved):

PR #2768, opened 2023-06-07
when item can not be cast to dicomElement, skip validation

esmadau:
## Description
when item can not be cast to dicomElement, skip validation

## Related issues
Addresses [[AB#104213](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/104213)].

## Testing
added unit test

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Features/Store/StoreDatasetValidator.cs
+++ b/src/Microsoft.Health.Dicom.Core/Features/Store/StoreDatasetValidator.cs
@@ -229,14 +229,13 @@ public class StoreDatasetValidator : IStoreDatasetValidator
             {
                 if (item is DicomSequence sequence)
                 {
-                    foreach (DicomDataset childDs in ((DicomSequence)item).Items)
+                    foreach (DicomDataset childDs in sequence)
                     {
                         stack.Push(childDs);
                     }
                 }
                 else if (item is DicomElement de)
                 {
-                    DicomElement de = (DicomElement)item;
                     string value = ds.GetString(de.Tag);
                     try
                     {


src/Microsoft.Health.Dicom.Core/Features/Store/StoreDatasetValidator.cs
```diff
@@ -225,7 +225,17 @@ string EnsureRequiredTagIsPresent(DicomTag dicomTag)
                 }
                 else
                 {
-                    DicomElement de = (DicomElement)item;
+                    DicomElement de;
```
smithago, line 228: I am curious about this whole method. We are validating all items now, why? I though we only validate things that we index or during json conversion.

src/Microsoft.Health.Dicom.Core/Features/Store/StoreDatasetValidator.cs
```diff
@@ -216,14 +227,14 @@ string EnsureRequiredTagIsPresent(DicomTag dicomTag)
             // add to stack to keep iterating when SQ type, otherwise validate
             foreach (DicomItem item in ds)
             {
-                if (item.ValueRepresentation == DicomVR.SQ)
+                if (item is DicomSequence)
```
wsugarman, line 230: nit: you can cast it inline like:
```suggestion
                if (item is DicomSequence sequence)
```

src/Microsoft.Health.Dicom.Core/Features/Store/StoreDatasetValidator.cs
```diff
@@ -216,14 +227,14 @@ string EnsureRequiredTagIsPresent(DicomTag dicomTag)
...
                 {
                     foreach (DicomDataset childDs in ((DicomSequence)item).Items)
                     {
                         stack.Push(childDs);
                     }
                 }
-                else
+                else if (item is DicomElement)
```
wsugarman, line 237:
```suggestion
                else if (item is DicomElement de)
```

PR #2785, opened 2023-06-10
external store - propagate partitions

esmadau:
## Description
start propagating partition down to blobFileStore's GetInstanceBlockBlobClient so we can use it to build a path

## Related issues
Addresses [[AB#104372](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/104372)].

## Testing
All tests updated as they were and just include partitions. No logic has changed, just signatures.

### DIFF
--- a/src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
+++ b/src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
@@ -48,13 +48,13 @@ public class BlobFileStore : IFileStore
     /// <inheritdoc />
     public async Task<FileProperties> StoreFileAsync(
         long version,
-        long partition,
+        int partitionKey,
         Stream stream,
         CancellationToken cancellationToken)
     {
         EnsureArg.IsNotNull(stream, nameof(stream));
 
-        BlockBlobClient blobClient = GetInstanceBlockBlobClient(version, partition);
+        BlockBlobClient blobClient = GetInstanceBlockBlobClient(version, partitionKey);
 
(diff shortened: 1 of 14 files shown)


src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Orchestration.cs
```diff
@@ -98,7 +98,7 @@ public partial class UpdateDurableFunction
                 await context.CallActivityWithRetryAsync(
                     nameof(DeleteOldVersionBlobAsync),
                     _options.RetryOptions,
-                    instanceWatermarks);
+                    new UpdateInstanceBlobArguments(input.PartitionKey, instanceWatermarks, input.ChangeDataset));
```
wsugarman, line 101: You cannot change the arguments to an activity/function without versioning the function too. Otherwise it's a breaking change and could break functions that are "in-flight" during an update.

Furthermore, let's use a separate type that way we're not needlessly passing the change dataset too

wsugarman (changes requested):

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -48,12 +48,13 @@ public class BlobFileStore : IFileStore
     /// <inheritdoc />
     public async Task<FileProperties> StoreFileAsync(
         long version,
+        long partition,
```
smithago, line 51: PartitionKey to be consistent in naming. And it is Int and not long.

PR #2787, opened 2023-06-14
Don't overwrite Study and Series value on a new instance when values are null

smithago:
## Description
SLIM viewer and similar scenarios that generate DCM files that have annotation information generally have some missing Study and Series information. Today DICOM services lets latest win. We will still continue to do that only if the value is not null.

## Related issues
https://github.com/ImagingDataCommons/slim/issues/153

## Testing
Added unit tests


src/Microsoft.Health.Dicom.SqlServer/Microsoft.Health.Dicom.SqlServer.csproj
```diff
@@ -68,6 +68,7 @@
     <TSqlScript Include="Features\Schema\Sql\Scripts\Sequences.sql" />
     <TSqlScript Include="Features\Schema\Sql\Tables\*.sql" />
     <TSqlScript Include="Features\Schema\Sql\Types\*.sql" />
+    <None Remove="Features\Schema\Migrations\38.diff.sql" />
```
bcarthic, line 71: nit: we can remove this

bcarthic (dismissed):

PR #2807, opened 2023-06-16
Start requiring PartitionEntry, pt1

esmadau:
## Description
Today, we sometimes use partition name and sometimes key. We often have partition in our code available, but the value is not always set.
This is a refactor that gets us closer to being able to rely on Partition being set and accessing its values always produces a value.
With IDP work continuing and supporting Update operation next, this work sets us up to have partition names that we can rely on to then use for blob paths in implementing that. It gets us closer to being able to use partition in other features as well without a heavier uplift per feature.
Also, we want to stop using context accessor below the api level so a propagation of Partition will also get us closer to this.
This came up as a request for cleanup while working on 104372. It is also an issue that impacted us where partition did not have values and so we saw NPEs ( see ticket # 105059).

The big changes are:
- Renaming PartitionEntry to Partition and inner fields, but maintaining PartitionEntry and its CreatedTime where needed for swagger and backwards compatibility
- Beginning of enforcing through use of ctors that partition name and key are always available when partition used in other classes such as InstanceIdentifier

## Related issues
Addresses [[AB#105251](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/105251)].

## Testing
Updated all tests to use Partition

### DIFF
--- a/src/Microsoft.Health.Dicom.Tests.Common/Extensions/IIndexDataStoreExtensions.cs
+++ b/src/Microsoft.Health.Dicom.Tests.Common/Extensions/IIndexDataStoreExtensions.cs
@@ -16,8 +16,8 @@ namespace Microsoft.Health.Dicom.Tests.Common.Extensions;
 
 public static class IIndexDataStoreExtensions
 {
-    public static Task<long> BeginCreateInstanceIndexAsync(this IIndexDataStore indexDataStore, int partitionKey, DicomDataset dicomDataset, CancellationToken cancellationToken = default)
-        => indexDataStore.BeginCreateInstanceIndexAsync(new Partition(partitionKey, Partition.DefaultName), dicomDataset, Array.Empty<QueryTag>(), cancellationToken);
+    public static Task<long> BeginCreateInstanceIndexAsync(this IIndexDataStore indexDataStore, Partition partition, DicomDataset dicomDataset, CancellationToken cancellationToken = default)
+        => indexDataStore.BeginCreateInstanceIndexAsync(partition, dicomDataset, Array.Empty<QueryTag>(), cancellationToken);
 
     public static Task EndCreateInstanceIndexAsync(this IIndexDataStore indexDataStore, int partitionKey, DicomDataset dicomDataset, long watermark, FileProperties fileProperties = null, bool hasFrameMetadata = false, CancellationToken cancellationToken = default)
         => indexDataStore.EndCreateInstanceIndexAsync(partitionKey, dicomDataset, watermark, Array.Empty<QueryTag>(), fileProperties, true, hasFrameMetadata, cancellationToken);
--- a/test/Microsoft.Health.Dicom.Tests.Integration/Features/RetrieveResourceServiceTests.cs
+++ b/test/Microsoft.Health.Dicom.Tests.Integration/Features/RetrieveResourceServiceTests.cs
@@ -180,7 +180,7 @@ public class RetrieveResourceServiceTests : IClassFixture<DataStoreTestsFixture>
 
     private async Task StoreDatasetsAndInstances(DicomDataset dataset, bool flagToStoreInstance)
     {
-        long version = await _indexDataStore.BeginCreateInstanceIndexAsync(1, dataset);
+        long version = await _indexDataStore.BeginCreateInstanceIndexAsync(new Partition(1, "clinic-one"), dataset);
 
         if (flagToStoreInstance)
         {
--- a/test/Microsoft.Health.Dicom.Tests.Integration/Persistence/ChangeFeedTests.cs
+++ b/test/Microsoft.Health.Dicom.Tests.Integration/Persistence/ChangeFeedTests.cs
@@ -227,7 +227,7 @@ public class ChangeFeedTests : IClassFixture<ChangeFeedTestsFixture>
             { DicomTag.PatientID, TestUidGenerator.Generate() },
         };
 
-        var version = await _fixture.DicomIndexDataStore.BeginCreateInstanceIndexAsync(1, newDataSet);
+        var version = await _fixture.DicomIndexDataStore.BeginCreateInstanceIndexAsync(new Partition(1, "clinic-one"), newDataSet);
 
         var versionedIdentifier = newDataSet.ToVersionedInstanceIdentifier(version, Partition.Default);
 
--- a/test/Microsoft.Health.Dicom.Tests.Integration/Persistence/DeleteServiceTests.cs
+++ b/test/Microsoft.Health.Dicom.Tests.Integration/Persistence/DeleteServiceTests.cs
@@ -53,7 +53,7 @@ public class DeleteServiceTests : IClassFixture<DeleteServiceTestsFixture>
     {
         var newDataSet = CreateValidMetadataDataset();
 
-        var version = await _fixture.IndexDataStore.BeginCreateInstanceIndexAsync(Partition.DefaultKey, newDataSet);
+        var version = await _fixture.IndexDataStore.BeginCreateInstanceIndexAsync(Partition.Default, newDataSet);
         var versionedDicomInstanceIdentifier = newDataSet.ToVersionedInstanceIdentifier(version, Partition.Default);
 
         if (persistMetadata)
--- a/test/Microsoft.Health.Dicom.Tests.Integration/Persistence/ExtendedQueryTagErrorStoreTests.cs
+++ b/test/Microsoft.Health.Dicom.Tests.Integration/Persistence/ExtendedQueryTagErrorStoreTests.cs
@@ -486,7 +486,7 @@ public class ExtendedQueryTagErrorStoreTests : IClassFixture<SqlDataStoreTestsFi
     private async Task<long> AddInstanceAsync(string studyId, string seriesId, string sopInstanceId)
     {
         DicomDataset dataset = Samples.CreateRandomInstanceDataset(studyId, seriesId, sopInstanceId);
-        long watermark = await _indexDataStore.BeginCreateInstanceIndexAsync(1, dataset);
+        long watermark = await _indexDataStore.BeginCreateInstanceIndexAsync(new Partition(1, "clinic-one"), dataset);
         await _indexDataStore.EndCreateInstanceIndexAsync(1, dataset, watermark);
         return watermark;
     }
--- a/test/Microsoft.Health.Dicom.Tests.Integration/Persistence/IndexDataStoreTests.cs
+++ b/test/Microsoft.Health.Dicom.Tests.Integration/Persistence/IndexDataStoreTests.cs
@@ -66,7 +66,7 @@ public partial class IndexDataStoreTests : IClassFixture<SqlDataStoreTestsFixtur
         string accessionNumber = dataset.GetString(DicomTag.AccessionNumber);
         string modality = dataset.GetString(DicomTag.Modality);
 
-        long version = await _indexDataStore.BeginCreateInstanceIndexAsync(1, dataset);
+        long version = await _indexDataStore.BeginCreateInstanceIndexAsync(new Partition(1, "clinic-one"), dataset);
 
         IReadOnlyList<StudyMetadata> studyMetadataEntries = await _testHelper.GetStudyMetadataAsync(studyInstanceUid);
 
@@ -124,8 +124,8 @@ public partial class IndexDataStoreTests : IClassFixture<SqlDataStoreTestsFixtur
         var partition1 = await _partitionStore.AddPartitionAsync(partitionName1);
         var partition2 = await _partitionStore.AddPartitionAsync(partitionName2);
 
-        await _indexDataStore.BeginCreateInstanceIndexAsync(partition1.Key, dataset);
-        await _indexDataStore.BeginCreateInstanceIndexAsync(partition2.Key, dataset);
+        await _indexDataStore.BeginCreateInstanceIndexAsync(partition1, dataset);
+        await _indexDataStore.BeginCreateInstanceIndexAsync(partition2, dataset);
 
         IReadOnlyList<Instance> instances1 = await _testHelper.GetInstancesAsync(studyInstanceUid, seriesInstanceUid, sopInstanceUid, partition1.Key);
         IReadOnlyList<Instance> instances2 = await _testHelper.GetInstancesAsync(studyInstanceUid, seriesInstanceUid, sopInstanceUid, partition2.Key);
@@ -158,7 +158,7 @@ public partial class IndexDataStoreTests : IClassFixture<SqlDataStoreTestsFixtur
         dataset.AddOrUpdate(DicomTag.Modality, conflictModality);
         dataset.AddOrUpdate(DicomTag.SOPInstanceUID, newInstance);
 
-        await _indexDataStore.BeginCreateInstanceIndexAsync(Partition.DefaultKey, dataset);
+        await _indexDataStore.BeginCreateInstanceIndexAsync(Partition.Default, dataset);
 
         IReadOnlyList<StudyMetadata> studyMetadataEntries = await _testHelper.GetStudyMetadataAsync(studyInstanceUid);
         IReadOnlyList<SeriesMetadata> seriesMetadataEntries = await _testHelper.GetSeriesMetadataAsync(seriesInstanceUid);
@@ -177,7 +177,7 @@ public partial class IndexDataStoreTests : IClassFixture<SqlDataStoreTestsFixtur
         DicomDataset dataset = CreateTestDicomDataset();
         string studyInstanceUid = dataset.GetString(DicomTag.StudyInstanceUID);
         string seriesInstanceUid = dataset.GetString(DicomTag.SeriesInstanceUID);
-        long watermark = await _indexDataStore.BeginCreateInstanceIndexAsync(Partition.DefaultKey, dataset);
+        long watermark = await _indexDataStore.BeginCreateInstanceIndexAsync(Partition.Default, dataset);
         await _indexDataStore.EndCreateInstanceIndexAsync(1, dataset, watermark, Array.Empty<QueryTag>());
 
         // add another instance in the same study+series with null patientName and modality and validate previous data wins
@@ -188,7 +188,7 @@ public partial class IndexDataStoreTests : IClassFixture<SqlDataStoreTestsFixtur
         newdataset.AddOrUpdate(DicomTag.PatientName, conflictPatientName);
         newdataset.AddOrUpdate(DicomTag.Modality, conflictModality);
         newdataset.AddOrUpdate(DicomTag.StudyDescription, conflictStudyDescription);
-        await _indexDataStore.BeginCreateInstanceIndexAsync(Partition.DefaultKey, newdataset);
+        await _indexDataStore.BeginCreateInstanceIndexAsync(Partition.Default, newdataset);
 
         IReadOnlyList<StudyMetadata> studyMetadataEntries = await _testHelper.GetStudyMetadataAsync(studyInstanceUid);
         IReadOnlyList<SeriesMetadata> seriesMetadataEntries = await _testHelper.GetSeriesMetadataAsync(seriesInstanceUid);
@@ -444,9 +444,9 @@ public partial class IndexDataStoreTests : IClassFixture<SqlDataStoreTestsFixtur
 
         DicomDataset dataset = Samples.CreateRandomDicomFile(studyInstanceUid, seriesInstanceUid, sopInstanceUid).Dataset;
 
-        await _indexDataStore.BeginCreateInstanceIndexAsync(Partition.DefaultKey, dataset);
+        await _indexDataStore.BeginCreateInstanceIndexAsync(Partition.Default, dataset);
 
-        await Assert.ThrowsAsync<PendingInstanceException>(() => _indexDataStore.BeginCreateInstanceIndexAsync(Partition.DefaultKey, dataset));
+        await Assert.ThrowsAsync<PendingInstanceException>(() => _indexDataStore.BeginCreateInstanceIndexAsync(Partition.Default, dataset));
     }
 
     [Fact]
@@ -458,10 +458,10 @@ public partial class IndexDataStoreTests : IClassFixture<SqlDataStoreTestsFixtur
 
         DicomDataset dataset = Samples.CreateRandomDicomFile(studyInstanceUid, seriesInstanceUid, sopInstanceUid).Dataset;
 
-        long version = await _indexDataStore.BeginCreateInstanceIndexAsync(Partition.DefaultKey, dataset);
+        long version = await _indexDataStore.BeginCreateInstanceIndexAsync(Partition.Default, dataset);
         await _indexDataStore.EndCreateInstanceIndexAsync(Partition.DefaultKey, dataset, version);
 
-        await Assert.ThrowsAsync<InstanceAlreadyExistsException>(() => _indexDataStore.BeginCreateInstanceIndexAsync(Partition.DefaultKey, dataset));
+        await Assert.ThrowsAsync<InstanceAlreadyExistsException>(() => _indexDataStore.BeginCreateInstanceIndexAsync(Partition.Default, dataset));
     }
 
     [Fact]
@@ -473,7 +473,7 @@ public partial class IndexDataStoreTests : IClassFixture<SqlDataStoreTestsFixtur
 
         DicomDataset dataset = Samples.CreateRandomDicomFile(studyInstanceUid, seriesInstanceUid, sopInstanceUid).Dataset;
 
-        long version = await _indexDataStore.BeginCreateInstanceIndexAsync(Partition.DefaultKey, dataset);
+        long version = await _indexDataStore.BeginCreateInstanceIndexAsync(Partition.Default, dataset);
 
         Instance instance = await _testHelper.GetInstanceAsync(studyInstanceUid, seriesInstanceUid, sopInstanceUid, version);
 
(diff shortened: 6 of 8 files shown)


src/Microsoft.Health.Dicom.Core/Features/Model/VersionedInstanceIdentifier.cs
```diff
@@ -27,7 +27,7 @@ public class VersionedInstanceIdentifier : InstanceIdentifier
         long version,
         int partitionKey = default,
         string partitionName = default)
-        : base(studyInstanceUid, seriesInstanceUid, sopInstanceUid, partitionKey, partitionName)
+        : base(studyInstanceUid, seriesInstanceUid, sopInstanceUid, new PartitionEntry(partitionKey, partitionName))
```
wsugarman, line 30: Should this ctors be symmetrical to the `InstanceIdentifier`?

src/Microsoft.Health.Dicom.Core/Features/Model/VersionedInstanceIdentifier.cs
```diff
@@ -25,9 +25,9 @@ public class VersionedInstanceIdentifier : InstanceIdentifier
         string seriesInstanceUid,
         string sopInstanceUid,
         long version,
-        int partitionKey = default,
-        string partitionName = default)
-        : base(studyInstanceUid, seriesInstanceUid, sopInstanceUid, partitionKey, partitionName)
+        int partitionKey = Partition.DefaultKey,
```
wsugarman, line 28: Will we delete this ctor?

src/Microsoft.Health.Dicom.Core/Features/Partitioning/IPartitionStore.cs
```diff
@@ -7,13 +7,13 @@
 using System.Threading;
 using System.Threading.Tasks;
 
-namespace Microsoft.Health.Dicom.Core.Features.Partition;
+namespace Microsoft.Health.Dicom.Core.Features.Partitioning;
 
 public interface IPartitionStore
 {
-    Task<PartitionEntry> AddPartitionAsync(string partitionName, CancellationToken cancellationToken = default);
+    Task<Partitioning.Partition> AddPartitionAsync(string partitionName, CancellationToken cancellationToken = default);
```
wsugarman, line 14: Does this conflict with something?

src/Microsoft.Health.Dicom.Core/Features/Partitioning/Partition.cs
```diff
@@ -0,0 +1,27 @@
...
+
+    public int Key { get; }
+
+    public string Name { get; }
+
+    public static Partition Default => new(DefaultKey, DefaultName);
+
+    public Partition(int key, string name)
+    {
+        Key = key;
```
wsugarman, line 24: `> 0`?

src/Microsoft.Health.Dicom.Core/Features/Partitioning/Partition.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    public const string DefaultName = "Microsoft.Default";
+
+    public const int DefaultKey = 1;
+
+    public int Key { get; }
+
+    public string Name { get; }
+
+    public static Partition Default => new(DefaultKey, DefaultName);
```
wsugarman, line 20: Singleton:

```suggestion
    public static Partition Default { get; } = new(DefaultKey, DefaultName);
```

src/Microsoft.Health.Dicom.Core/Features/Partitioning/PartitionCache.cs
```diff
@@ -8,9 +8,9 @@
 using Microsoft.Health.Dicom.Core.Configs;
 using Microsoft.Health.Dicom.Core.Features.Common;
 
-namespace Microsoft.Health.Dicom.Core.Features.Partition;
+namespace Microsoft.Health.Dicom.Core.Features.Partitioning;
 
-public class PartitionCache : EphemeralMemoryCache<string, PartitionEntry>
+public class PartitionCache : EphemeralMemoryCache<string, Partitioning.Partition>
```
wsugarman, line 13: Is there a type collision?

src/Microsoft.Health.Dicom.Core/Features/Store/IStoreResponseBuilder.cs
```diff
@@ -28,8 +29,10 @@ public interface IStoreResponseBuilder
     /// <param name="storeValidationResult">Store validation errors and warnings</param>
     /// <param name="warningReasonCode">The warning reason code.</param>
     /// <param name="buildWarningSequence">Whether to build response warning sequence or not.</param>
+    /// <param name="partition">Data Partition entry</param>
```
wsugarman, line 32: nit: this is in the wrong order

src/Microsoft.Health.Dicom.Core/Features/Store/StoreResponseBuilder.cs
```diff
@@ -97,7 +99,7 @@ private bool HasWarningReasonCode()
             _dataset.Add(referencedSopSequence);
         }
 
-        var dicomInstance = dicomDataset.ToInstanceIdentifier();
+        var dicomInstance = dicomDataset.ToInstanceIdentifier(partition);
```
wsugarman, line 102: This is good, and it further highlights some of the inconsistencies where the identifier is scoped to a partition, but the method below `ResolveRetrieveInstanceUri` fetches the partition from the request context.

wsugarman: It may be worth including a `Debug.Assert` in that method to assert the partitions match

src/Microsoft.Health.Dicom.Core/Messages/Partitioning/AddPartitionResponse.cs
```diff
@@ -3,16 +3,15 @@
...
 {
-    public AddPartitionResponse(PartitionEntry partitionEntry)
+    public AddPartitionResponse(Features.Partitioning.Partition partition)
     {
-        PartitionEntry = partitionEntry;
+        Partition = partition;
     }
 
-    public PartitionEntry PartitionEntry { get; }
+    public Features.Partitioning.Partition Partition { get; }
```
wsugarman, line 16: Why is this necessary?

src/Microsoft.Health.Dicom.Core/Messages/Partitioning/GetPartitionResponse.cs
```diff
@@ -3,16 +3,15 @@
...
 {
-    public GetPartitionResponse(PartitionEntry partitionEntry)
+    public GetPartitionResponse(Features.Partitioning.Partition partition)
     {
-        PartitionEntry = partitionEntry;
+        Partition = partition;
     }
 
-    public PartitionEntry PartitionEntry { get; }
+    public Features.Partitioning.Partition Partition { get; }
```
wsugarman, line 16: What's this for?

swagger/v1-prerelease/swagger.yaml
```diff
@@ -3219,18 +3219,15 @@ components:
...
       properties:
-        partitionKey:
+        key:
           type: integer
           format: int32
-        partitionName:
+        name:
           type: string
           nullable: true
-        createdDate:
```
wsugarman, line 3231: Yeah.... we'll need to add this back lol

wsugarman: Oh dang, all of the names will need to be updated too, but I think we can isolate these changes to this code path

test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/ExtendedQueryTagTests.cs
```diff
@@ -20,6 +20,7 @@
 using Microsoft.Health.Operations;
 using Xunit;
 using FunctionsStartup = Microsoft.Health.Dicom.Functions.App.Startup;
+using Partition = Microsoft.Health.Dicom.Core.Features.Partitioning.Partition;
```
wsugarman, line 23: This is because of the client type collision?

wsugarman (changes requested):

src/Microsoft.Health.Dicom.Api.UnitTests/Features/Routing/UrlResolverTests.cs
```diff
@@ -123,7 +124,7 @@ public void GivenAnInstance_WhenRetrieveInstanceUriIsResolved_ThenCorrectUrlShou
 
         _httpContext.Request.RouteValues.Add(KnownActionParameterNames.Version, version);
 
-        var instance = new InstanceIdentifier(studyInstanceUid, seriesInstanceUid, sopInstanceUid);
+        var instance = new InstanceIdentifier(studyInstanceUid, seriesInstanceUid, sopInstanceUid, Partition.Default);
```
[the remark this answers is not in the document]
wsugarman: Oh that's right. Nevermind!

src/Microsoft.Health.Dicom.Client/Models/Partition.cs
```diff
@@ -3,25 +3,33 @@
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
+using System;
+using System.Text.Json.Serialization;
 using EnsureThat;
 
 namespace Microsoft.Health.Dicom.Client.Models;
 
+[Serializable]
```
wsugarman, line 12: This shouldn't be necessary.

src/Microsoft.Health.Dicom.Client/Models/PartitionEntry.cs
```diff
@@ -3,11 +3,16 @@
...
 
+using System;
+
 namespace Microsoft.Health.Dicom.Client.Models;
 
-public static class DefaultPartition
+/// <summary>
+/// Used to ensure API contract does not change while we change the underlying property naming.
+/// </summary>
+public class PartitionEntry : Partition
```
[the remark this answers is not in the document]
wsugarman: Oh I left these comments in client. The function stuff is only relevant for Core lol

src/Microsoft.Health.Dicom.Core/Features/Routing/IUrlResolver.cs
```diff
@@ -45,8 +45,9 @@ public interface IUrlResolver
     /// Resovles the URI to retrieve an instance.
     /// </summary>
     /// <param name="instanceIdentifier">The identifier to the instance.</param>
+    /// <param name="isPartitionEnabled">Whether or not partition feature is enabled</param>
     /// <returns>An instance of <see cref="Uri"/> pointing to where the instance can be retrieved.</returns>
-    Uri ResolveRetrieveInstanceUri(InstanceIdentifier instanceIdentifier);
+    Uri ResolveRetrieveInstanceUri(InstanceIdentifier instanceIdentifier, bool isPartitionEnabled);
```
wsugarman, line 50: Unsure how I feel about this. I like bubbling up the partition concept to the signature as opposed to the hidden HttpContext resolution. There are concepts like "studyInstanceUid` above that are UIDs without the context of a partition. Should there be an "identifier" class that is agnostic of partition? Or perhaps the better question is: how should the URI resolver consistently across its methods know whether to use a partition? Perhaps that is something determined by DI (e.g. a partitioned "UrlResolver" versus a non-partitioned?)

wsugarman: Also would the method above `ResolveRetrieveStudyUri` need partition information?

src/Microsoft.Health.Dicom.Client/Models/Partition.cs
```diff
@@ -9,17 +9,16 @@
...
     public const string DefaultName = "Microsoft.Default";
 
     public const int DefaultKey = 1;
 
-    [JsonPropertyName("PartitionKey")]
+    [JsonPropertyName("partitionKey")] // these explicit names are here for the REST schema
     public int Key { get; }
 
-    [JsonPropertyName("PartitionName")]
+    [JsonPropertyName("partitionName")] // these explicit names are here for the REST schema
```
wsugarman, line 21: There must be a way for us to incorporate camelCase names as a setting for Swagger gen (like we can for JSON.NET and System.Text.Json), but I won't block on that here

src/Microsoft.Health.Dicom.Core/Features/Store/IStoreResponseBuilder.cs
wsugarman: This still needs to move 😛

src/Microsoft.Health.Dicom.Client/Models/PartitionEntry.cs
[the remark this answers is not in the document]
wsugarman: At the very least, you must add `JsonProperty` attributes to the class for `Partition` as it's used in `ExportInput`. Functions JSON.NET and changing the name of the property would be a breaking change that would affect any in-progress functions.

src/Microsoft.Health.Dicom.Functions.UnitTests/Update/UpdateDurableFunctionTests.Activity.cs
```diff
@@ -36,15 +36,15 @@ public async Task GivenInstanceMetadata_WhenUpdatingInstanceWatermark_ThenShould
 
         var versions = expected.Select(x => x.Version).ToList();
 
-        _indexStore.BeginUpdateInstancesAsync(Partition.DefaultKey, studyInstanceUid, CancellationToken.None).Returns(identifiers);
+        _indexStore.BeginUpdateInstancesAsync(Arg.Any<Partition>(), studyInstanceUid, CancellationToken.None).Returns(identifiers);
```
wsugarman, line 39: nit: if this is the default partition, you should be able to pass it here as it will use reference equality on the type.

src/Microsoft.Health.Dicom.Core/Features/Partitioning/Partition.cs
```diff
@@ -13,6 +13,8 @@ public class Partition
 {
     public const string DefaultName = "Microsoft.Default";
 
+    public const string UnknownName = "Unknown"; // use for older binaries where we are unable to get partition name and it is not known
```
wsugarman, line 16: nit: maybe `Unspecified`?

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStoreV35.cs
```diff
@@ -62,7 +63,7 @@ public override async Task<IReadOnlyList<InstanceMetadata>> RetrieveDeletedInsta
                                 seriesInstanceUid,
                                 sopInstanceUid,
                                 watermark,
-                                partitionKey),
+                                new Partition(partitionKey, Partition.UnknownName)),
```
wsugarman, line 66: nit: perhaps we can have a utility method for these legacy scenarios

smithago:
For others looking at the review, can you help understand excatly want is the change that is resulting in enforcement? There are too many renames and it is unable to pin point the exact change

src/Microsoft.Health.Dicom.Client/Models/PartitionEntry.cs
```diff
@@ -4,22 +4,15 @@
...
 using System;
-using EnsureThat;
 
 namespace Microsoft.Health.Dicom.Client.Models;
 
-public class PartitionEntry
+/// <summary>
+/// Used to ensure API contract does not change while we change the underlying property naming.
+/// </summary>
+public class PartitionEntry : Partition
```
wsugarman, line 13: What's the point of this class now? I don't think anything operates on this type, nor does the type name matter here because we only cared about it for Swagger (even though that's not such a big deal either lol)

src/Microsoft.Health.Dicom.Api/Features/Routing/UrlResolver.cs
```diff
@@ -109,7 +109,7 @@ public Uri ResolveRetrieveWorkitemUri(string workitemInstanceUid)
     }
 
     /// <inheritdoc />
-    public Uri ResolveRetrieveInstanceUri(InstanceIdentifier instanceIdentifier)
+    public Uri ResolveRetrieveInstanceUri(InstanceIdentifier instanceIdentifier, bool isPartitionEnabled)
```
bcarthic, line 112: Is this change required?

src/Microsoft.Health.Dicom.Blob/Features/Export/AzureBlobExportSink.cs
```diff
@@ -63,7 +63,7 @@ public async Task<bool> CopyAsync(ReadResult value, CancellationToken cancellati
             return false;
         }
 
-        using Stream sourceStream = await _source.GetStreamingFileAsync(value.Identifier.Version, value.Identifier.PartitionName, cancellationToken);
+        using Stream sourceStream = await _source.GetStreamingFileAsync(value.Identifier.Version, value.Identifier.Partition.Name, cancellationToken);
```
bcarthic, line 66: Are we sure that the Partition is populated in Export?

src/Microsoft.Health.Dicom.Core/Features/Partitioning/Partition.cs
```diff
@@ -0,0 +1,39 @@
...
+using Newtonsoft.Json;
+
+namespace Microsoft.Health.Dicom.Core.Features.Partitioning;
+
+public class Partition
+{
+    public const string DefaultName = "Microsoft.Default";
+
+    public const string UnknownName = "Unknown"; // use for older binaries where we are unable to get partition name and it is not known
+
```
bcarthic, line 18: Not sure why we want Unknown name. It should always be Microsoft.Default. There is no entry in DB that corresponds to Unknown Partition entry

src/Microsoft.Health.Dicom.Core/Features/Model/InstanceIdentifier.cs
```diff
@@ -17,34 +17,12 @@ public class InstanceIdentifier
...
-        EnsureArg.IsNotNullOrWhiteSpace(seriesInstanceUid, nameof(seriesInstanceUid));
-        EnsureArg.IsNotNullOrWhiteSpace(sopInstanceUid, nameof(sopInstanceUid));
-
-        StudyInstanceUid = studyInstanceUid;
-        SeriesInstanceUid = seriesInstanceUid;
-        SopInstanceUid = sopInstanceUid;
-        PartitionKey = partitionKey;
-        PartitionName = partitionName;
-        PartitionEntry = new PartitionEntry(PartitionKey, PartitionName);
+        Partition = EnsureArg.IsNotNull(partition, nameof(partition));
```
bcarthic, line 22: Checking for null partition is correct, but are we ensuring that the key is not default?

src/Microsoft.Health.Dicom.Core/Features/Model/VersionedInstanceIdentifier.cs
```diff
@@ -14,20 +14,8 @@ public class VersionedInstanceIdentifier : InstanceIdentifier
...
-
-    public VersionedInstanceIdentifier(
-        string studyInstanceUid,
-        string seriesInstanceUid,
-        string sopInstanceUid,
-        long version,
-        int partitionKey = DefaultPartition.Key,
-        string partitionName = DefaultPartition.Name)
-        : base(studyInstanceUid, seriesInstanceUid, sopInstanceUid, partitionKey, partitionName)
+        Partition partition = null)
```
bcarthic, line 17: In which scenario do you want to accept null partition?

src/Microsoft.Health.Dicom.Core/Features/Routing/IUrlResolver.cs
bcarthic: Why do we need this? Already the existing code populates the partition based on route values

src/Microsoft.Health.Dicom.Core/Features/Store/StoreResponseBuilder.cs
```diff
@@ -97,12 +106,12 @@ private bool HasWarningReasonCode()
...
 
         var referencedSop = new DicomDataset
         {
             { DicomTag.ReferencedSOPInstanceUID, dicomDataset.GetSingleValue<string>(DicomTag.SOPInstanceUID) },
-            { DicomTag.RetrieveURL, _urlResolver.ResolveRetrieveInstanceUri(dicomInstance).ToString() },
+            { DicomTag.RetrieveURL, _urlResolver.ResolveRetrieveInstanceUri(dicomInstance, _isPartitionEnabled).ToString() },
```
bcarthic, line 114: Do we need this? Even without this it populates the partition right?

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Activity.cs
```diff
@@ -42,7 +43,7 @@ public async Task<IReadOnlyList<InstanceFileState>> UpdateInstanceWatermarkAsync
 
         logger.LogInformation("Beginning to update all instance watermarks");
 
-        IEnumerable<InstanceMetadata> instanceMetadata = await _indexStore.BeginUpdateInstancesAsync(arguments.PartitionKey, arguments.StudyInstanceUid, CancellationToken.None);
+        IEnumerable<InstanceMetadata> instanceMetadata = await _indexStore.BeginUpdateInstancesAsync(new Partition(arguments.PartitionKey, Partition.UnknownName), arguments.StudyInstanceUid, CancellationToken.None);
```
bcarthic, line 47: Is the unknown Partition Name correct? Even though the underlying method doesn't use the name, in future if its used will have issues. What is the problem in just accept the key in this case?

wsugarman (dismissed):

src/Microsoft.Health.Dicom.SqlServer/Features/Partitioning/SqlPartitionStoreV6.cs
```diff
@@ -85,18 +84,17 @@ public override async Task<IEnumerable<PartitionEntry>> GetPartitionsAsync(Cance
                        VLatest.Partition.PartitionName,
                        VLatest.Partition.CreatedDate);
 
-                    results.Add(new PartitionEntry(
+                    results.Add(new Partition(
```
bcarthic, line 87: If createdDate is not fetched, or you not changing the sproc?

src/Microsoft.Health.Dicom.Core/Features/Store/IIndexDataStore.cs
```diff
@@ -139,20 +140,20 @@ public interface IIndexDataStore
     /// <summary>
     /// Asynchronously updates a DICOM instance NewWatermark
     /// </summary>
-    /// <param name="partitionKey">The partition key.</param>
+    /// <param name="partition">The partition.</param>
     /// <param name="versions">List of instances watermark to update</param>
     /// <param name="cancellationToken">The cancellation token.</param>
     /// <returns>A task that with list of instance metadata with new watermark.</returns>
-    Task<IEnumerable<InstanceMetadata>> BeginUpdateInstanceAsync(int partitionKey, IReadOnlyCollection<long> versions, CancellationToken cancellationToken = default);
+    Task<IEnumerable<InstanceMetadata>> BeginUpdateInstanceAsync(Partition partition, IReadOnlyCollection<long> versions, CancellationToken cancellationToken = default);
```
bcarthic, line 147: I am not sure why we needed this change?

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStoreV6.cs
```diff
@@ -197,7 +198,7 @@ public override async Task<IEnumerable<VersionedInstanceIdentifier>> RetrieveDel
                             seriesInstanceUid,
                             sopInstanceUid,
                             watermark,
-                            partitionKey));
+                            new Partition(partitionKey, Partition.UnknownName)));
```
bcarthic, line 201: This is incorrect, since the corresponding partitionKey will not match Unknown Name. Either we need to pass the name which we have from the requestcontext or remove the old files

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStoreV35.cs
```diff
@@ -62,7 +63,7 @@ public override async Task<IReadOnlyList<InstanceMetadata>> RetrieveDeletedInsta
...
                             new InstanceProperties()
```
bcarthic, line 67: Same here, Defaulting to uknown is wrong. Either populate the name or remove the files

src/Microsoft.Health.Dicom.Core/Features/Store/IIndexDataStore.cs
bcarthic: If that is the case, please pass the name why pass wrong value. You have the value in the FIlterAttribute, we just need to populate and pass on here

src/Microsoft.Health.Dicom.Core/Features/Model/InstanceIdentifier.cs
bcarthic: Because there is some inconsistency in the model, you check for null for name but for key you dont check if it greater than 0

src/Microsoft.Health.Dicom.Core/Features/Model/VersionedInstanceIdentifier.cs
bcarthic: Is there any repercussion of making this change? we are introducing a new change which wasn't there before

wsugarman (approved):

PR #2823, opened 2023-06-22
Dispose Blob Network Stream when Transcoding

smithago:
## Description
- Dispose `Stream` returned from `BlockBlobClient` when transcoding after loading DICOM file
- Refactor `RecyclableMemoryStreamManager` factory to allow for run-time configuration
- Normalize stream tags

## Related issues
[AB#104597](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/104597)

## Testing
Local tests


jnlycklama (approved):

src/Microsoft.Health.Dicom.Core/Features/Common/RecyclableMemoryStreamOptions.cs
```diff
@@ -0,0 +1,38 @@
...
+
+using Microsoft.Extensions.Configuration;
+using Microsoft.IO;
+
+namespace Microsoft.Health.Dicom.Core.Features.Common;
+
+/// <summary>
+/// Represents the settings that may be configured for the shared instance of <see cref="RecyclableMemoryStreamManager"/>.
+/// </summary>
+public class RecyclableMemoryStreamOptions
```
smithago, line 14: Are we planning to set this in the internal project?

src/Microsoft.Health.Dicom.Core/Registration/RecyclableMemoryStreamRegistrationExtensions.cs
```diff
@@ -0,0 +1,53 @@
...
+            .Configure(configure);
+    }
+
+    public static IServiceCollection AddRecyclableMemoryStreamManager(this IServiceCollection services)
+    {
+        EnsureArg.IsNotNull(services, nameof(services));
+        return services.AddSingleton(
+            sp =>
+            {
+                RecyclableMemoryStreamOptions options = sp.GetRequiredService<IOptions<RecyclableMemoryStreamOptions>>().Value;
```
smithago, line 41: Are we okay if these are not set and result in default values? Meaning we will not override the better defaults if any they have?

src/Microsoft.Health.Dicom.Core/Features/Common/RecyclableMemoryStreamOptions.cs
wsugarman: Yeah, I was planning to modify these options for PaaS over time (and for potential investigations). But this way we don't need further binary changes.

src/Microsoft.Health.Dicom.Core/Registration/RecyclableMemoryStreamRegistrationExtensions.cs
wsugarman: I double-checked their defaults and aligned them with the options. Their defaults are all `0` and `false`:  https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream/blob/master/src/RecyclableMemoryStreamManager.cs

PR #2831, opened 2023-06-23
M1a - delete from FileProperty on instance delete

esmadau:
## Description
Delete matching rows in FileProperty table when deleting instance.

## Related issues
Addresses [[AB#104629](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/104629)].

## Testing
Updated integration test to ensure rows deleted and ensure that instance continues to be deleted even when no file properties present.


src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/DeleteInstanceV40.sql
```diff
@@ -0,0 +1,272 @@
...
+--     @createdStatus
+--         * Status value representing the created state.
+--     @studyInstanceUid
+--         * The study instance UID.
+--     @seriesInstanceUid
+--         * The series instance UID.
+--     @sopInstanceUid
+--         * The SOP instance UID.
+/***************************************************************************************/
+CREATE OR ALTER PROCEDURE dbo.DeleteInstanceV40
```
wsugarman, line 25: I don't think this is changing the signature for the deletion sproc, so you can instead update the existing one

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStoreV40.cs
```diff
@@ -0,0 +1,100 @@
...
+using Microsoft.Health.Dicom.SqlServer.Features.Schema.Model;
+using Microsoft.Health.SqlServer.Features.Client;
+using Microsoft.Health.SqlServer.Features.Storage;
+
+namespace Microsoft.Health.Dicom.SqlServer.Features.Store;
+
+/// <summary>
+/// Sql IndexDataStore version 6.
+/// </summary>
+internal class SqlIndexDataStoreV40 : SqlIndexDataStoreV37
```
wsugarman, line 23: Without a new sproc, you won't need this either

wsugarman (changes requested):

wsugarman (approved):

PR #2838, opened 2023-06-23
Handle unknown fo-dicom validation exceptions

esmadau:
## Description
When on attempting to get value or validate an unknown exception type is thrown from fo-dicom rethrow to bubble up as system error.

Example of this is fo-dicom attempting to get parse a value to validate and excepting.

## Related issues
Addresses [[AB#104779](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/104779)].

## Testing
Added unit test


src/Microsoft.Health.Dicom.Core/Features/Store/StoreDatasetValidator.cs
```diff
@@ -236,18 +236,22 @@ string EnsureRequiredTagIsPresent(DicomTag dicomTag)
                 }
                 else if (item is DicomElement de)
                 {
-                    string value = ds.GetString(de.Tag);
                     try
                     {
-                        ValidateItemWithLeniency(value, de, queryTags);
+                        if (de.ValueRepresentation.ValueType == typeof(string))
```
wsugarman, line 241: Ah nice! I like it

wsugarman (approved):

PR #2878, opened 2023-07-07
idp-when partitions not enabled, do not use default ms partition

esmadau:
## Description
IDP - when partitions not enabled, do not use default ms partition in blob path.

## Related issues
Addresses [[AB#105115](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/105115)].

## Testing
Updated unit test

### DIFF
--- a/src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalBlobClient.cs
+++ b/src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalBlobClient.cs
@@ -97,6 +97,6 @@ internal class ExternalBlobClient : IBlobClient
     {
         return _isPartitionEnabled ?
             _externalStoreOptions.StorageDirectory + partitionName + "/" :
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalBlobClient.cs
```diff
@@ -98,6 +97,6 @@ public string GetServiceStorePath(string partitionName)
     {
         return _isPartitionEnabled ?
             _externalStoreOptions.StorageDirectory + partitionName + "/" :
-            _externalStoreOptions.StorageDirectory + PartitionEntry.Default.PartitionName + "/";
+            _externalStoreOptions.StorageDirectory + "/";
```
wsugarman, line 100: Does `StorageDirectory` already end with a `/`? Should we do some massaging of this value?

wsugarman (approved):

PR #2943, opened 2023-07-20
update conformance statements to declare V2 as latest version

esmadau:
## Description
update conformance statements to declare V2 as latest version and encourage using it instead of V1

## Related issues
Addresses [[AB#105496](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/105496)].

## Testing
n/a

### DIFF
--- a/docs/resources/conformance-statement.md
+++ b/docs/resources/conformance-statement.md
@@ -1,6 +1,6 @@
 # DICOM Conformance Statement
 
-> API version 2 is the latest API version and should be used as API version 1 will be deprecated soon.
+> API version 2 is the latest API version and should be used.
 > Go to [API version 2](v2-conformance-statement.md) for the latest conformance statement.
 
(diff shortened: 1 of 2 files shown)


docs/resources/conformance-statement.md
```diff
@@ -1,5 +1,8 @@
 # DICOM Conformance Statement
 
+> API version 2 is the latest API version and should be used as API version 1 will be deprecated soon.
```
mmitrik, line 3: > version 1 will be deprecated soon.

Do we have a timetable for V1 deprecation?

mmitrik (commented):
The notes to direct users to V2 of our API makes sense, however, is the plan to actually deprecate V1?  Is there a cost to keeping it around?  What if there are customers that do not want to move?  Does the Azure program we're onboarding to (can't remember the name 😅)  have guidelines for how to communicate and manage API version changes?

mmitrik (approved):

PR #2959, opened 2023-07-26
Refactor partition logic to service.

brandonpollett:
## Description
This PR takes the logic found in the partition attribute and moves it to the `PartitionService`.

## Related issues
Addresses [issue #].

## Testing


src/Microsoft.Health.Dicom.Core/Messages/Partitioning/GetOrAddPartitionRequest.cs
```diff
@@ -7,13 +7,19 @@
...
     {
-        PartitionName = paritionName;
+        PartitionName = partitionName;
+        AddIfNotExists = addIfNotExists;
     }
 
+    /// <summary>
+    /// Should the request attempt add if the partition doesn't exist.
+    /// </summary>
+    public bool AddIfNotExists { get; set; }
```
brandonpollett, line 21: I'm not 100% in love with this pattern, but couldn't come up with anything better. The goal is that only on some paths do we actually want to do the `AddIfNotExists` path, but it seemed weird to have 2 paths for messages through the code.

brandonpollett: Let's move to two paths to make sure the auth story stays correct.

src/Microsoft.Health.Dicom.Core/Features/Partitioning/PartitionService.cs
```diff
@@ -25,15 +28,31 @@ public PartitionService(PartitionCache partitionCache, IPartitionStore partition
...
 
-    public async Task<AddPartitionResponse> AddPartitionAsync(string partitionName, CancellationToken cancellationToken = default)
+    public async Task<GetOrAddPartitionResponse> GetOrAddPartitionAsync(string partitionName, CancellationToken cancellationToken = default)
     {
         EnsureArg.IsNotNull(partitionName, nameof(partitionName));
 
-        _logger.LogInformation("Creating partition with name '{PartitionName}'.", partitionName);
+        _logger.LogInformation("Getting partition with name '{PartitionName}'.", partitionName);
+
+        PartitionNameValidator.Validate(partitionName);
```
wsugarman, line 37: This is definitely a more appropriate place for this logic

src/Microsoft.Health.Dicom.Core/Features/Partitioning/PartitionService.cs
```diff
@@ -25,15 +28,31 @@ public PartitionService(PartitionCache partitionCache, IPartitionStore partition
...
+
+        Partition partition = await _partitionCache.GetAsync(partitionName, partitionName, _partitionStore.GetPartitionAsync, cancellationToken);
```
wsugarman, line 39: Would it be simpler if the `_partitionStore` had a `GetOrAddAsync` like [`ConcurrentDictionary<TKey, TValue>`](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd?view=net-7.0)?

brandonpollett: Simpler in the long term yes, but I'm not sure the complexity of that change in this PR is worth the effort since the goal was just to move the existing logic to a place that could be used in other processes like functions.

wsugarman: I'd feel a lot more comfortable if we did this work sooner than later to ensure our more concurrent operations are done at the source, like adding a new sproc. If we don't do it now, I'd like to see it done very soon haha

PR #2962, opened 2023-07-27
Electron tool - Add v2 support and remove pre-release

bcarthic:
## Description
Add support for v2 api and remove v1-prerelease

## Related issues
Addresses [issue #].

## Testing


wsugarman (approved):

wsugarman:
> we can remove v1 as well, right?

I think we may want to keep v1 as long as we support it -- perhaps just make v2 the default

bcarthic:
> > we can remove v1 as well, right?
> 
> I think we may want to keep v1 as long as we support it -- perhaps just make v2 the default

V2 is default

PR #2995, opened 2023-08-16
Encode String in error message

abdullah248:
## Description
Encode string in error message to prevent cross scripting attacks

When validating a frame number passed in as part of the url we check if the input is a valid number. If its not we throw an error message that contains the invalid input that they passed directly. Since we did not encode the response we were susceptible to sending anything malicious back to the user 

_When a web application is vulnerable to this type of attack, it will pass unvalidated input sent through requests to the client. Reflected attacks are delivered to the victim in various ways, such as in an e-mail message, or through some specially crafted URI. When a user is tricked into clicking on the malicious link, the injected code travels to the vulnerable web site, which reflects the attack back to the user’s browser. The browser then executes the offending code because it came from a "trusted" server._

By encoding the response we escape characters in their input before returning error message back to the user

## Related issues
[AB#105701](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/105701)

## Testing

### DIFF
--- a/src/Microsoft.Health.Dicom.Api/Features/Filters/QueryModelStateValidatorAttribute.cs
+++ b/src/Microsoft.Health.Dicom.Api/Features/Filters/QueryModelStateValidatorAttribute.cs
@@ -1,9 +1,11 @@
-﻿// -------------------------------------------------------------------------------------------------
+// -------------------------------------------------------------------------------------------------
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 // -------------------------------------------------------------------------------------------------
 
 using System.Linq;
+using System.Text.RegularExpressions;
+using System.Web;
 using EnsureThat;
 using Microsoft.AspNetCore.Mvc.Filters;
 using Microsoft.AspNetCore.Mvc.ModelBinding;
@@ -19,7 +21,14 @@ public sealed class QueryModelStateValidatorAttribute : ActionFilterAttribute
         if (!context.ModelState.IsValid)
         {
             (string key, ModelStateEntry value) = context.ModelState.Where(x => x.Value.Errors.Count > 0).First();
-            throw new InvalidQueryStringValuesException(key, value.Errors[0].ErrorMessage);
+
+            string errorMessage = value.Errors[0].ErrorMessage;
+            if (Regex.IsMatch(errorMessage, @"<[^>]*>"))
+            {
+                errorMessage = HttpUtility.HtmlEncode(errorMessage);
+            }
+
+            throw new InvalidQueryStringValuesException(key, errorMessage);
         }
     }
 }
--- a/src/Microsoft.Health.Dicom.Api/Features/Filters/QueryModelStateValidatorAttribute.cs
+++ b/src/Microsoft.Health.Dicom.Api/Features/Filters/QueryModelStateValidatorAttribute.cs
@@ -10,12 +10,11 @@ using EnsureThat;
 using Microsoft.AspNetCore.Mvc.Filters;
 using Microsoft.AspNetCore.Mvc.ModelBinding;
 using Microsoft.Health.Dicom.Core.Exceptions;
-using Microsoft.IdentityModel.Tokens;
 
 namespace Microsoft.Health.Dicom.Api.Features.Filters;
 
 public sealed class QueryModelStateValidatorAttribute : ActionFilterAttribute
-{ 
+{
     private static readonly Regex HtmlCharacters = new Regex("<[^>]*>", RegexOptions.Compiled);
 
     public override void OnActionExecuting(ActionExecutingContext context)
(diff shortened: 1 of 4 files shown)


wsugarman:
Could you elaborate more about this PR and how it prevents a vulnerability?

KnicKnic (approved):

src/Microsoft.Health.Dicom.Core/Exceptions/InvalidQueryStringValuesException.cs
```diff
@@ -1,16 +1,17 @@
...
 using System.Globalization;
+using System.Web;
 
 namespace Microsoft.Health.Dicom.Core.Exceptions;
 
 public class InvalidQueryStringValuesException : ValidationException
 {
     public InvalidQueryStringValuesException(string key, string error)
-        : base(string.Format(CultureInfo.InvariantCulture, DicomCoreResource.InvalidQueryStringValue, key, error))
+        : base(string.Format(CultureInfo.InvariantCulture, DicomCoreResource.InvalidQueryStringValue, key, HttpUtility.UrlEncode(error)))
```
KnicKnic, line 14: Shouldn't key be encoded then?

KnicKnic (changes requested):

bcarthic:
> Could you elaborate more about this PR and how it prevents a vulnerability?

We return the output message without encoding, so because of that, anyone can inject a custom javascript to route to any url or make the customer click any button in their website. Encoding the output string is one way of preventing this reflected XSS attack

bcarthic:
Could you add a unit test in the controller code?

bcarthic (changes requested):
Could you add a unit test in the controller code?

src/Microsoft.Health.Dicom.Core/Exceptions/InvalidQueryStringValuesException.cs
abdullah248: no, the key is the part of the url path so it wouldn't match our router and would return a 404 (ex. in https://localhost:63838/v2/studies/1.2.332323232 `studies` is the key and if it were changed to something else we would just return a 404 instead of saying the key is invalid.)

src/Microsoft.Health.Dicom.Api/Features/Filters/QueryModelStateValidatorAttribute.cs
```diff
@@ -19,7 +21,14 @@ public override void OnActionExecuting(ActionExecutingContext context)
         if (!context.ModelState.IsValid)
         {
             (string key, ModelStateEntry value) = context.ModelState.Where(x => x.Value.Errors.Count > 0).First();
-            throw new InvalidQueryStringValuesException(key, value.Errors[0].ErrorMessage);
+
+            string errorMessage = value.Errors[0].ErrorMessage;
+            if (Regex.IsMatch(errorMessage, @"<[^>]*>"))
```
bcarthic, line 26: Can you make this Regex compiled. So we dont spend time.

You can make this a constant.
```
string pattern = @"<[^>]*>";
Regex regex = new Regex(pattern, RegexOptions.Compiled);
```

And then you can use.

```regex.IsMatch(errorMessage)```

Comment (abdullah248) [reply src/Microsoft.Health.Dicom.Api/Features/Filters/QueryModelStateValidatorAttribute.cs:26]: updated

File: src/Microsoft.Health.Dicom.Api/Features/Filters/QueryModelStateValidatorAttribute.cs [hunk 3]
```diff
@@ -13,13 +15,22 @@ namespace Microsoft.Health.Dicom.Api.Features.Filters;
...
     public override void OnActionExecuting(ActionExecutingContext context)
     {
         EnsureArg.IsNotNull(context, nameof(context));
         if (!context.ModelState.IsValid)
         {
             (string key, ModelStateEntry value) = context.ModelState.Where(x => x.Value.Errors.Count > 0).First();
-            throw new InvalidQueryStringValuesException(key, value.Errors[0].ErrorMessage);
+
+            string errorMessage = value.Errors[0].ErrorMessage;
+            if (HtmlCharacters.IsMatch(errorMessage))
```
bcarthic, line 28: If the errorMessage is null, then this will thrown an exception.

bcarthic:
```
if (!string.IsNullOrEmpty() && HtmlCharacters.IsMatch(errorMessage))
```

src/Microsoft.Health.Dicom.Api/Features/Filters/QueryModelStateValidatorAttribute.cs
```diff
@@ -1,25 +1,37 @@
...
     public override void OnActionExecuting(ActionExecutingContext context)
     {
         EnsureArg.IsNotNull(context, nameof(context));
         if (!context.ModelState.IsValid)
         {
             (string key, ModelStateEntry value) = context.ModelState.Where(x => x.Value.Errors.Count > 0).First();
-            throw new InvalidQueryStringValuesException(key, value.Errors[0].ErrorMessage);
+
+            string errorMessage = value.Errors[0].ErrorMessage;
+            if (!errorMessage.IsNullOrEmpty() && HtmlCharacters.IsMatch(errorMessage))
```
bcarthic, line 29: is this even compile?

abdullah248: fixed the typo

bcarthic (approved):

KnicKnic (approved):

wsugarman (approved):

PR #3014, opened 2023-08-30
Add CMK Health Check to Blob and SQL checks; Add health check publisher cache

jnlycklama:
## Description
Incorporates the changes from https://github.com/microsoft/healthcare-shared-components/pull/777 to the dicom service. This will add a check for the customer-managed key (if configured) to our blob and sql health checks. The overall health check state is also run periodically and cached by the health check publisher.

## Related issues
Addresses [#104759] & [#104763].

## Testing
Ran locally, configured to use a real key & key vault, tested disabling and reenabling


src/Microsoft.Health.Dicom.Api/Registration/DicomServerServiceCollectionExtensions.cs
```diff
@@ -45,12 +48,24 @@ public static class DicomServerServiceCollectionExtensions
...
+    /// <param name="configuration">The configuration for the DICOM server.</param>
     /// <returns>The DICOM server builder instance.</returns>
-    public static IDicomServerBuilder AddBackgroundWorkers(this IDicomServerBuilder serverBuilder)
+    public static IDicomServerBuilder AddBackgroundWorkers(this IDicomServerBuilder serverBuilder, IConfiguration configuration)
     {
         EnsureArg.IsNotNull(serverBuilder, nameof(serverBuilder));
         serverBuilder.Services.AddScoped<DeletedInstanceCleanupWorker>();
         serverBuilder.Services.AddHostedService<DeletedInstanceCleanupBackgroundService>();
+
+        serverBuilder.Services.AddCustomerKeyValidationBackgroundService(options =>
```
wsugarman, line 59: nit: Let's ditch the code blocks and use expressions. Perhaps something like this:
```csharp
serverBuilder.Services
    .AddCustomerKeyValidationBackgroundService(options => configuration
        .GetSection(CustomerManagedKeyOptions.CustomerManagedKey)
        .Bind(options))
    .AddHealthCheckPublisher(options => configuration
        .GetSection("HealthCheckPublisher")
        .Bind(options));
```

src/Microsoft.Health.Dicom.Web/appsettings.json
```diff
@@ -124,6 +124,9 @@
       }
     }
   },
+  "HealthCheckPublisher": {
+    "Period": "00:00:10"
```
wsugarman, line 128: This is pretty aggressive, isn't it? 10 seconds? Or because this is OSS, we won't worry about CMK and so it can be a lot shorter? Still, I think like `00:01:00` or at a minimum `00:00:30` is not unreasonable

jnlycklama: Spoke offline - 10s was chosen because of the period of k8s readiness checks, but since that is not relevant in OSS we will leave this as the default 30 seconds here. Removed the app settings

wsugarman (approved):

PR #3015, opened 2023-09-01
Replace external operation credential provider with generic

jnlycklama:
## Description
Replace ExternalOperationCredentialProvider with ExternalCredentialProvider - it does exactly the same thing with a more generic name


src/Microsoft.Health.Dicom.Core/Features/Common/DefaultExternalCredentialProvider.cs
```diff
@@ -1,14 +1,15 @@
...
 // -------------------------------------------------------------------------------------------------
 
 using Azure.Core;
 using Azure.Identity;
+using Microsoft.Health.Core.Features.Identity;
 
 namespace Microsoft.Health.Dicom.Core.Features.Common;
 
-internal sealed class DefaultExternalOperationCredentialProvider : IExternalOperationCredentialProvider
+internal sealed class DefaultExternalCredentialProvider : IExternalCredentialProvider
```
wsugarman, line 12: Don't we have this type in core? Do we have a method to register it as a service?

jnlycklama: Added registration method (previously it was only a part of CMK registration)

arunmk-ms (approved):

PR #3018, opened 2023-09-02
Add tests to validate HasFrameData

bcarthic:
## Description
Add tests to validate HasFrameData.

To support unit testing, added logic to include CompositeByteBuffer to determine position and size

## Related issues
Addresses [issue #].

## Testing

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Features/Store/StoreOrchestrator.cs
+++ b/src/Microsoft.Health.Dicom.Core/Features/Store/StoreOrchestrator.cs
@@ -97,7 +97,7 @@ public class StoreOrchestrator : IStoreOrchestrator
 
             bool hasFrameMetadata = await frameRangeTask;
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.Health.Dicom.Core/Features/Store/StoreOrchestrator.cs
```diff
@@ -255,6 +255,17 @@ private static bool TryGetBufferPosition(IByteBuffer buffer, out long position,
             size = fileByteBuffer.Size;
             result = true;
         }
+        else if (buffer is CompositeByteBuffer compositeByteBuffer)
```
smithago, line 258: Is this valid? I forgotten the logic maybe, but it could be that the single frame is in multiple buffers, this won't work right? I had a comment to support that correctly later.

bcarthic: In our current case, this won't hit at all. Only if we create sample file using stream or memory byte buffer this hits.

bcarthic: I thought of calculating the total size, but we don't use the size. We only use the position and add it with the offset.

bcarthic:
Closing this PR

PR #3024, opened 2023-09-07
add logging and metrics on transcoding instances and their size

esmadau:
## Description
add logging and metrics on transcoding instances and their size

## Related issues
Addresses [[AB#106470](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/106470)].

## Testing
n/a

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
+++ b/src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
@@ -127,6 +127,7 @@ public class RetrieveResourceService : IRetrieveResourceService
             // transcoding of single instance
             if (needsTranscoding)
             {
+                _logger.LogInformation("Transcoding Instance");
                 FileProperties fileProperties = await RetrieveHelpers.CheckFileSize(_blobDataStore, _retrieveConfiguration.MaxDicomFileSize, version, partition.Name, false, cancellationToken);
                 LogFileSize(fileProperties.ContentLength, version, needsTranscoding, instance.InstanceProperties.HasFrameMetadata);
                 SetTranscodingBillingProperties(fileProperties.ContentLength);
@@ -216,11 +217,10 @@ public class RetrieveResourceService : IRetrieveResourceService
 
             return new RetrieveResourceResponse(fastFrames, mediaType, isSinglePart);
         }
+        _logger.LogInformation("Downloading the entire instance for frame parsing");
 
         FileProperties fileProperties = await RetrieveHelpers.CheckFileSize(_blobDataStore, _retrieveConfiguration.MaxDicomFileSize, instance.VersionedInstanceIdentifier.Version, partition.Name, false, cancellationToken);
         LogFileSize(fileProperties.ContentLength, instance.VersionedInstanceIdentifier.Version, needsTranscoding, instance.InstanceProperties.HasFrameMetadata);
-
-        _logger.LogInformation("Downloading the entire instance for frame parsing");
 
         // eagerly doing getFrames to validate frame numbers are valid before returning a response
         Stream stream = await _blobDataStore.GetFileAsync(instance.VersionedInstanceIdentifier.Version, partition.Name, cancellationToken);


src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
```diff
@@ -123,9 +127,10 @@ public async Task<RetrieveResourceResponse> GetInstanceResourceAsync(RetrieveRes
             // transcoding of single instance
             if (needsTranscoding)
             {
-                _logger.LogInformation("Transcoding Instance");
                 FileProperties fileProperties = await RetrieveHelpers.CheckFileSize(_blobDataStore, _retrieveConfiguration.MaxDicomFileSize, version, instance.VersionedInstanceIdentifier.Partition.Name, false, cancellationToken);
                 SetTranscodingBillingProperties(fileProperties.ContentLength);
+                _logger.LogInformation("Transcoding Instance {Path} of size {ContentLength}", fileProperties.ContentLength, fileProperties.Path);
```
bcarthic, line 132: Do we want to log path? Today it is just watermark but once IDP is enabled it can contain any information

src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
```diff
@@ -123,9 +127,10 @@ public async Task<RetrieveResourceResponse> GetInstanceResourceAsync(RetrieveRes
...
+                _retrieveMeter.TranscodeInstanceCount.Add(1, TelemetryDimension(fileProperties.ContentLength, _retrieveConfiguration.MaxDicomFileSize));
```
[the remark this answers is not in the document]
bcarthic: Also line 288. In this case, we will know how much we sent back for normal WADO

src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveHelpers.cs
```diff
@@ -32,4 +32,18 @@ public static async Task<FileProperties> CheckFileSize(IFileStore blobDataStore,
...
+
+        // limit the file size that can be read in memory
+        if (fileProperties.ContentLength > maxDicomFileSize)
+        {
+            if (render)
+            {
+                throw new NotAcceptableException(string.Format(CultureInfo.CurrentCulture, DicomCoreResource.RenderFileTooLarge, maxDicomFileSize));
+            }
+            throw new NotAcceptableException(string.Format(CultureInfo.CurrentCulture, DicomCoreResource.RetrieveServiceFileTooBig, maxDicomFileSize));
+        }
```
bcarthic, line 47: Should you put this in a private method so both methods can use?

src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
```diff
@@ -120,11 +124,12 @@ public async Task<RetrieveResourceResponse> GetInstanceResourceAsync(RetrieveRes
                     string.Format(CultureInfo.CurrentCulture, DicomCoreResource.RetrieveServiceMultiInstanceTranscodingNotSupported, requestedTransferSyntax));
             }
 
+            FileProperties fileProperties = await GetFileProperties(partition, instance, version, needsTranscoding, cancellationToken: cancellationToken);
```
bcarthic, line 127: This is going to degrade in current retrieve performance if there is no transcoding. Are we okay with that?

bcarthic: Same for frame resources too.

bcarthic: If we don't do transcoding, we return the stream using DownloadStreamingAsync method, which may have Content.Length to get the whole file size. But may not work, could you check?

src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
```diff
@@ -338,10 +359,12 @@ yield return
         foreach (int frame in frames)
         {
             FrameRange frameRange = framesRange[frame];
-            Stream frameStream = await _blobDataStore.GetFileFrameAsync(version, _dicomRequestContextAccessor.RequestContext.GetPartitionName(), frameRange, cancellationToken);
+            (Stream frameStream, long streamLength) = await _blobDataStore.GetFileFrameAsync(version, _dicomRequestContextAccessor.RequestContext.GetPartitionName(), frameRange, cancellationToken);
```
bcarthic, line 362: When you already have FrameRange.Length why do you need streamLength here?

src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
```diff
@@ -123,8 +127,8 @@ public async Task<RetrieveResourceResponse> GetInstanceResourceAsync(RetrieveRes
             // transcoding of single instance
             if (needsTranscoding)
             {
-                _logger.LogInformation("Transcoding Instance");
```
bcarthic, line 126: Do you think this message is useful. If checkFileSize throws an exception, we may not know what part the error is thrown

jnlycklama (approved):

src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
bcarthic: the stack trace will show. But earlier log would tell us how many calls made to transcoding with and without eroor

src/Microsoft.Health.Dicom.Core/Features/Telemetry/RetrieveMeter.cs
```diff
@@ -15,11 +16,24 @@ public sealed class RetrieveMeter : IDisposable
...
 
+    public Counter<long> RetrieveInstanceMetadataCount { get; }
+
     public Counter<long> RetrieveInstanceCount { get; }
 
+    public static KeyValuePair<string, object>[] RetrieveInstanceCountTelemetryDimension(long contentLength, long maxFileSize, bool isTranscoding = false, bool hasFrameMetadata = false, bool isRendered = false) =>
+        new[]
+        {
+            new KeyValuePair<string, object>("ContentLength", contentLength),
+            new KeyValuePair<string, object>("MaxFileSize", maxFileSize),
```
bcarthic, line 31: nit: It could be MaxFileSizeForTranscoding. But not sure why we want to log this info

src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
```diff
@@ -280,19 +296,23 @@ private static bool NeedsTranscoding(bool isOriginalTransferSyntaxRequested, str
...
             FileProperties fileProperties = await _blobDataStore.GetFilePropertiesAsync(version, _dicomRequestContextAccessor.RequestContext.GetPartitionName(), cancellationToken);
             Stream stream = await _blobDataStore.GetStreamingFileAsync(version, _dicomRequestContextAccessor.RequestContext.GetPartitionName(), cancellationToken);
+            streamTotalLength += fileProperties.ContentLength;
             yield return
                 new RetrieveResourceInstance(
                     stream,
                     GetResponseTransferSyntax(isOriginalTransferSyntaxRequested, requestedTransferSyntax, instanceMetadata),
                     fileProperties.ContentLength);
         }
+        LogFileSize(streamTotalLength, requestedVersion, needsTranscoding: false, hasFrameMetadata: true);
```
bcarthic, line 315: Is this statement TRUE? hasFrameMetadata: true. How do you know that?

PR #3048, opened 2023-09-19
Update FHIR Client to Fix Component Governance

wsugarman:
## Description
- Update FHIR Client to remove dependency on legacy app authentication library

## Related issues
N/A

## Testing
Unit tests


converter/dicom-cast/src/Microsoft.Health.DicomCast.Core.UnitTests/Features/Fhir/FhirTransactionExecutorTests.cs
```diff
@@ -1,4 +1,4 @@
-﻿// -------------------------------------------------------------------------------------------------
+// -------------------------------------------------------------------------------------------------
```
abdullah248, line 1: why is it flagging this as changed

abdullah248 (approved):

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core.UnitTests/Features/Fhir/FhirTransactionExecutorTests.cs
wsugarman: File encoding based on our `.gitattributes` file

PR #3071, opened 2023-09-28
File Store metrics

esmadau:
## Description
Start emitting a log and metric on each file store operation and when each client type registered.
Start emitting stream i/o per operation

## Related issues
Addresses [[AB#105845](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/105845)].

## Testing
Updated existing tests

### DIFF
--- a/src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
+++ b/src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
@@ -123,7 +123,7 @@ public class BlobFileStore : IFileStore
 
                 BlobContentInfo info = await blobClient.CommitBlockListAsync(blockLengths.Keys, cancellationToken: cancellationToken);
 
-                EmitTelemetry(nameof(StoreFileInBlocksAsync), OperationType.Input, info.Length);
+                EmitTelemetry(nameof(StoreFileInBlocksAsync), OperationType.Input, stream.Length);
 
                 fileProperties = new FileProperties
                 {


src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -172,9 +183,39 @@ public async Task DeleteFileIfExistsAsync(long version, string partitionName, Ca
         BlockBlobClient blobClient = GetInstanceBlockBlobClient(version, partitionName);
         _logger.LogInformation("Trying to delete DICOM instance file with watermark '{Version}'.", version);
 
+        EmitTelemetry("DeleteFileIfExistsAsync");
+
         await ExecuteAsync(() => blobClient.DeleteIfExistsAsync(DeleteSnapshotsOption.IncludeSnapshots, conditions: null, cancellationToken));
     }
 
+    private static readonly Action<ILogger, string, Exception> LogBlobClientOperationDelegate =
```
bcarthic, line 191: Why this pattern? we wanted to avoid this. You can just log in the method itself

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -189,7 +230,12 @@ public async Task<Stream> GetFileAsync(long version, Partition partition, FilePr
         // We should either remove fo-dicom parsing for transcoding or make SDK change to support Length property on RetrievableStream
         //Response<BlobDownloadStreamingResult> result = await blobClient.DownloadStreamingAsync(range: default, conditions: null, rangeGetContentHash: false, cancellationToken);
         //stream = result.Value.Content;
-        return await ExecuteAsync(async () => await blobClient.OpenReadAsync(blobOpenReadOptions, cancellationToken));
+        return await ExecuteAsync(async () =>
+        {
+            Stream stream = await blobClient.OpenReadAsync(blobOpenReadOptions, cancellationToken);
+            EmitTelemetry("GetFileAsync", stream.Length);
```
bcarthic, line 236: didnt you just added a metric for retrieve?

src/Microsoft.Health.Dicom.Blob/Features/Telemetry/BlobFileStoreMeter.cs
```diff
@@ -0,0 +1,71 @@
...
+using System.Collections.Generic;
+using System.Diagnostics.Metrics;
+using Microsoft.Health.Dicom.Core.Features.Telemetry;
+
+namespace Microsoft.Health.Dicom.Blob.Features.Telemetry;
+
+/// <summary>
+/// Represents whether operation is input (write) or output(read) from perspective of I/O on the blob store
+/// </summary>
+public enum OperationType
```
wsugarman, line 16: Let's not define two public types in the same file

src/Microsoft.Health.Dicom.Blob/Features/Storage/InternalBlobClient.cs
```diff
@@ -19,12 +20,17 @@ internal class InternalBlobClient : IBlobClient
...
-        IOptionsMonitor<BlobContainerConfiguration> optionsMonitor)
+    public InternalBlobClient(
+        BlobServiceClient blobServiceClient,
+        IOptionsMonitor<BlobContainerConfiguration> optionsMonitor,
+        ILogger<InternalBlobClient> logger)
     {
         _client = EnsureArg.IsNotNull(blobServiceClient, nameof(blobServiceClient));
         _containerName = EnsureArg.IsNotNull(optionsMonitor.Get(Constants.BlobContainerConfigurationName).ContainerName, nameof(optionsMonitor));
+        _logger = EnsureArg.IsNotNull(logger, nameof(logger));
+        _logger.LogInformation("Internal blob client registered.");
```
wsugarman, line 33: I take it we're trying to log whether IDP is in use? Can we log this earlier when registering the service?

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -172,9 +183,43 @@ public async Task DeleteFileIfExistsAsync(long version, string partitionName, Ca
...
+        {
+            LogBlobClientOperationDelegate(_logger, operationName, _blobClient.IsExternal, null);
+        }
+        else
+        {
+            var lenght = streamLength.Value;
+            LogBlobClientOperationWithStreamDelegate(_logger, operationName, streamLength.Value, null);
+            _blobFileStoreMeter.BlobFileStoreOperationStreamSize.Add(
+                lenght,
+                BlobFileStoreMeter.BlobFileStoreOperationTelemetryDimension(operationName, operationType, _blobClient.IsExternal));
```
wsugarman, line 219: This API is not the most appealing. Is there a way to statically type this?

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -172,9 +183,43 @@ public async Task DeleteFileIfExistsAsync(long version, string partitionName, Ca
...
+
+    private static readonly Action<ILogger, string, long, Exception> LogBlobClientOperationWithStreamDelegate =
+        LoggerMessage.Define<string, long>(
+            LogLevel.Information,
+            default,
+            "Operation '{OperationName}' processed stream length '{StreamLength}'.");
+
+    private void EmitTelemetry(string operationName, OperationType operationType, long? streamLength = null)
+    {
+        _blobFileStoreMeter.BlobFileStoreOperationCount.Add(
```
wsugarman, line 205: Could we instead inject these sorts of basic metrics inside of the blob client's request pipeline as a policy?

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -172,9 +183,43 @@ public async Task DeleteFileIfExistsAsync(long version, string partitionName, Ca
...
+            1,
+            BlobFileStoreMeter.BlobFileStoreOperationTelemetryDimension(operationName, operationType, _blobClient.IsExternal));
+
+        if (streamLength == null)
+        {
+            LogBlobClientOperationDelegate(_logger, operationName, _blobClient.IsExternal, null);
+        }
+        else
+        {
+            var lenght = streamLength.Value;
```
wsugarman, line 215: Nit: typo

wsugarman (changes requested):

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
wsugarman: Sure, we don't need to do it if it looks like a bunch of work. It may be something of interest in the future such that each method doesn't need to explicit write the metrics.

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -172,9 +183,43 @@ public async Task DeleteFileIfExistsAsync(long version, string partitionName, Ca
         BlockBlobClient blobClient = GetInstanceBlockBlobClient(version, partitionName);
         _logger.LogInformation("Trying to delete DICOM instance file with watermark '{Version}'.", version);
 
+        EmitTelemetry("DeleteFileIfExistsAsync", OperationType.Input);
+
         await ExecuteAsync(() => blobClient.DeleteIfExistsAsync(DeleteSnapshotsOption.IncludeSnapshots, conditions: null, cancellationToken));
     }
 
+    private static readonly Action<ILogger, string, bool, Exception> LogBlobClientOperationDelegate =
```
wsugarman, line 191: nit: can you move these to the top with other fields?

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
wsugarman: This could potentially be very out-of-scope.

If there is no such API, and only key-value pairs are accepted, then we can put some thought into later coming up with some sort of nifty type-safe API like:

```csharp
_blobFileStoreMeter.BlobFileStoreOperationStreamSize.Add(
    length,
    operationName,
    operationType,
    _blobClient.IsExternal);

// OR

_blobFileStoreMeter.BlobFileStoreOperationStreamSize.Add<OperationContext>(
    length,
    new OperationContext(operationName, operationType, _blobClient.IsExternal);

// OR
using OperationContext context = new(operationName, operationType, _blobClient.IsExternal);

_blobFileStoreMeter.BlobFileStoreOperationStreamSize.Add(length, context);
```

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -104,6 +110,9 @@ public class BlobFileStore : IFileStore
                 }
 
                 BlobContentInfo info = await blobClient.CommitBlockListAsync(blockLengths.Keys, cancellationToken: cancellationToken);
+
+                EmitTelemetry("StoreFileInBlocksAsync", OperationType.Input, stream.Length);
```
wsugarman, line 114: nit: use `nameof()` for the operation names

wsugarman, line 114: Are we sure this stream's length can always be read? This is not a required property for all streams, like for ones whose length is not known ahead of time. I think for us this is always the case as we buffer, but please double-check. We can always compute it above too

wsugarman: Although, I see us accessing it in other parts of this file, so this comment may be moot

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -189,7 +234,12 @@ public async Task<Stream> GetFileAsync(long version, Partition partition, FilePr
         // We should either remove fo-dicom parsing for transcoding or make SDK change to support Length property on RetrievableStream
         //Response<BlobDownloadStreamingResult> result = await blobClient.DownloadStreamingAsync(range: default, conditions: null, rangeGetContentHash: false, cancellationToken);
         //stream = result.Value.Content;
-        return await ExecuteAsync(async () => await blobClient.OpenReadAsync(blobOpenReadOptions, cancellationToken));
+        return await ExecuteAsync(async () =>
+        {
+            Stream stream = await blobClient.OpenReadAsync(blobOpenReadOptions, cancellationToken);
+            EmitTelemetry("GetFileAsync", OperationType.Output, stream.Length);
```
wsugarman, line 240: I think you're ok with this `Length` too: https://github.com/Azure/azure-sdk-for-net/blob/fba9f87d8e6c44ad213c7f6b9844d32d86ea8edb/sdk/storage/Azure.Storage.Blobs/src/BlobBaseClient.cs#L3167

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -316,7 +386,9 @@ public async Task SetBlobToColdAccessTierAsync(long version, Partition partition
 
         await ExecuteAsync(async () =>
         {
-            return await blobClient.SetAccessTierAsync(AccessTier.Cold, cancellationToken: cancellationToken);
+            Response response = await blobClient.SetAccessTierAsync(AccessTier.Cold, cancellationToken: cancellationToken);
+            EmitTelemetry("SetBlobToColdAccessTierAsync", OperationType.Input);
```
wsugarman, line 390: Should there be some other `OperationType` for this sort of method?

src/Microsoft.Health.Dicom.Blob/Features/Telemetry/BlobFileStoreMeter.cs
```diff
@@ -0,0 +1,71 @@
...
+    public Counter<long> BlobFileStoreOperationStreamSize { get; }
+
+    /// <summary>
+    /// Sets telemetry dimensions on meter
+    /// </summary>
+    /// <param name="operationName">Name of operation being hit</param>
+    /// <param name="operationType">Represents whether operation is input (write) or output(read) </param>
+    /// <param name="isExternal">Whether or not this metric is being emitted for an external store</param>
+    /// <returns></returns>
+    public static KeyValuePair<string, object>[] BlobFileStoreOperationTelemetryDimension(string operationName, OperationType operationType, bool isExternal) =>
```
wsugarman, line 61: nit: rename this method to something with a verb like `CreateDimension`

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
wsugarman: There's actually something in R9 about this, so we'll get to it later

wsugarman (approved):

PR #3078, opened 2023-09-30
Check for SQL inaccessible state in the BackgroundServiceHealthCheck before reporting unhealthy

jnlycklama:
## Description
Similar to this PR  - https://github.com/microsoft/healthcare-shared-components/pull/820
We want to cover this state:
- cmk is enabled
- customer misconfigures cmk, and leaves it misconfigured for 30 mins
- we report Degraded status because we cannot access the key
- customer fixes their cmk configuration, but does not reprovision the service in order to revalidate the key on the SQL DB
- we are able to access the key, so health checks will pass the CMK check, but we will still get `DataStoreNotReadyException` because the schema is not yet initialized

The BackgroundServiceHealthCheck will always check the schema first, by getting the version here https://github.com/microsoft/dicom-server/blob/04de4ff8168099be54396624aafaf26faded48f2/src/Microsoft.Health.Dicom.SqlServer/Features/Schema/PassthroughSchemaVersionResolver.cs#L36
This won't fail when the DB is Inaccessible, just return null. So instead, I added an additional call after the `DataStoreNotReadyException` is caught, so we can determine if the DB is inaccessible or if its failing for some other reason. 

## Related issues
Addresses [#110155].

## Testing

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Features/HealthCheck/BackgroundServiceHealthCheck.cs
+++ b/src/Microsoft.Health.Dicom.Core/Features/HealthCheck/BackgroundServiceHealthCheck.cs
@@ -78,6 +78,16 @@ public class BackgroundServiceHealthCheck : IHealthCheck
             _deleteMeter.OldestRequestedDeletion.Add((await oldestWaitingToBeDeleted).ToUnixTimeSeconds());
             _deleteMeter.CountDeletionsMaxRetry.Add(await numReachedMaxedRetry);
         }
+        // Error: Can not connect to the database in its current state. This error can be for various DB states (recovering, inacessible) but we assume that our DB will only hit this for Inaccessible state
+        catch (SqlException ex) when (ex.ErrorCode == 40925)
+        {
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.Dicom.Core/Microsoft.Health.Dicom.Core.csproj
```diff
@@ -14,6 +14,7 @@
     <PackageReference Include="fo-dicom.Imaging.ImageSharp" />
     <PackageReference Include="MediatR" />
     <PackageReference Include="Microsoft.ApplicationInsights" />
+    <PackageReference Include="Microsoft.Data.SqlClient" />
```
wsugarman, line 17: This is probably a bad sign. This project shouldn't know about SQL.

wsugarman: Are there some other ways to do this? For example, is there a way to avoid having each to know about CMK? Incoming requests in ASP.NET Core and outgoing HTTP requests in the Azure SDKs use a pipeline concept where there are discreet steps. Could our health checks leverage a similar pattern?

wsugarman (changes requested):

PR #3089, opened 2023-10-09
Exception handling for  unique constraint exception

poadhika:
## Description
This will handle the sql exception that will throw 2601 error code if an operation tries to insert an existing partition key.

## Related issues
Addresses [issue #110438](https://microsofthealth.visualstudio.com/Health/_workitems/edit/110438)

## Testing
Manually tested. Also updated integration test.


src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/AddPartition.sql
```diff
@@ -31,7 +31,7 @@ BEGIN
     DECLARE @partitionKey INT
 
     SELECT @partitionKey = PartitionKey
-    FROM dbo.Partition
+    FROM dbo.Partition WITH (UPDLOCK)
```
bcarthic, line 34: We have to upgrade the version of sql

bcarthic: We should not update this file, this is a generated one

src/Microsoft.Health.Dicom.SqlServer/Features/Partitioning/SqlPartitionStoreV6.cs
```diff
@@ -57,7 +57,7 @@ public override async Task<Partition> AddPartitionAsync(string partitionName, Ca
             }
             catch (SqlException ex)
             {
-                if (ex.Number == SqlErrorCodes.Conflict)
+                if (ex.Number == SqlErrorCodes.Conflict || ex.Number == SqlErrorCodes.UniqueKeyConstraintVoilation)
```
bcarthic, line 60: If you are doing this we dont want the sql changes. We can make the sql changes in another PR

bcarthic: Or if you would like it you can do it here and we dont have to check for unique constraint

bcarthic (changes requested):
SQL changes are not needed, but if you want to do in SQL we have to upgrade the sql version

test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/DataPartitionEnabledTests.cs
```diff
@@ -57,18 +57,35 @@ public async Task WhenRetrievingPartitions_TheServerShouldReturnAllPartitions()
...
 
-        using DicomWebResponse<DicomDataset> response = await _instancesManager.StoreAsync(new[] { dicomFile }, partition: newPartition);
+        DicomWebResponse<DicomDataset> response = await _instancesManager.StoreAsync(new[] { dicomFile }, partition: newPartition);
 
         Assert.True(response.IsSuccessStatusCode);
 
         ValidationHelpers.ValidateReferencedSopSequence(
             await response.GetValueAsync(),
             ConvertToReferencedSopSequenceEntry(dicomFile.Dataset, newPartition.Name));
+        return response;
```
bcarthic, line 88: Why do we want to return response now?

poadhika: There is another test that is also using the same code. check line 60.

bcarthic: I get that but that test is not using the response value right?

PR #3108, opened 2023-10-16
delete files based on file properties in sql

esmadau:
## Description
- start deleting files based on file properties in sql
- ensure we delete original files when using external store and instance has gone through update operation
- added additional null check on fp when using external store to capture the temporary possibility of V2 activities not passing in fp as they run. In that case, we will default to using partition and version. We can undo this after this pr has been pushed.

## Related issues
Addresses [[AB#110009](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/110009)].

## Testing
updated existing and added new tests
- e2e test added on UpdateInstanceTests will run with external flag being set with both true and false
- integration tests added on deleteService to ensure the delete service logic switches to using originalWatermark when not using externalStore and uses file properties when using external store to test both paths continue to work and we continue to store and retrieve both sets of info

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Features/Delete/DeleteService.cs
+++ b/src/Microsoft.Health.Dicom.Core/Features/Delete/DeleteService.cs
@@ -134,7 +134,9 @@ public class DeleteService : IDeleteService
                                 cancellationToken)
                         };
 
-                        // only need to delete by "original watermark" to catch updates if not IDP
+                        // NOTE: in the input deletedInstanceIdentifiers we're going to have a row for each version in IDP,
+                        // but for non-IDP we'll have a single row whose original version needs to be explicitly deleted below.
+                        // To that end, we only need to delete by "original watermark" to catch changes from Update operation if not IDP.
                         if (!_isExternalStoreEnabled && deletedInstanceIdentifier.InstanceProperties.OriginalVersion.HasValue)
                         {
                             tasks.Add(_fileStore.DeleteFileIfExistsAsync(deletedInstanceIdentifier.InstanceProperties.OriginalVersion.Value, deletedInstanceIdentifier.VersionedInstanceIdentifier.Partition, deletedInstanceIdentifier.InstanceProperties.fileProperties, cancellationToken));


src/Microsoft.Health.Dicom.SqlServer/Features/Schema/SchemaVersion.cs
```diff
@@ -58,4 +58,5 @@ public enum SchemaVersion
     V45 = 45,
     V46 = 46,
     V47 = 47,
+    V48 = 48,
```
bcarthic, line 61: Sorry I took 48, you may have to rebase

test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/UpdateInstanceTests.cs
```diff
@@ -91,6 +91,26 @@ public async Task WhenUpdatingDicomMetadataForMultipleStudy_ThenItShouldUpdateCo
         await VerifyDeleteStudyAsync(studyInstanceUid1, dicomFile1, true);
     }
 
+    [Fact]
```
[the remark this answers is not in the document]
bcarthic: just checking, how this test is different from others

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/48.diff.sql
```diff
@@ -0,0 +1,37 @@
...
+CREATE OR ALTER PROCEDURE dbo.RetrieveDeletedInstanceV42
+    @count          INT,
+    @maxRetries     INT
+    AS
+BEGIN
+    SET NOCOUNT ON
+
+SELECT  TOP (@count) p.PartitionName, d.PartitionKey, d.StudyInstanceUid, d.SeriesInstanceUid, d.SopInstanceUid, d.Watermark, d.OriginalWatermark, d.FilePath, d.ETag
+FROM    dbo.DeletedInstance as d WITH (UPDLOCK, READPAST)
+    INNER JOIN dbo.Partition as p
```
bcarthic, line 30: nit: Is indent off?

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -189,15 +189,45 @@ public class BlobFileStore : IFileStore
...
 
     /// <inheritdoc />
-    public async Task DeleteFileIfExistsAsync(long version, string partitionName, CancellationToken cancellationToken)
+    public async Task DeleteFileIfExistsAsync(long version, Partition partition, FileProperties fileProperties, CancellationToken cancellationToken)
     {
+        EnsureArg.IsNotNull(partition);
 
-        BlockBlobClient blobClient = GetInstanceBlockBlobClient(version, partitionName);
-        _logger.LogInformation("Trying to delete DICOM instance file with watermark '{Version}'.", version);
+        BlockBlobClient blobClient = GetInstanceBlockBlobClient(version, partition, fileProperties);
```
bcarthic, line 196: Just wanted to be sure, in some place we pass fileproperties as null explicitly, hope we dont get null reference

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -189,15 +189,45 @@ public class BlobFileStore : IFileStore
...
+        _logger.LogInformation("Trying to delete DICOM instance file with watermark: '{Version}' and PartitionKey: {PartitionKey}.", version, partition.Key);
 
         EmitTelemetry(nameof(DeleteFileIfExistsAsync), OperationType.Input);
 
-        await ExecuteAsync(() => blobClient.DeleteIfExistsAsync(DeleteSnapshotsOption.IncludeSnapshots, conditions: null, cancellationToken));
+        await ExecuteAsync(async () =>
+        {
+            try
+            {
+                bool exists = await blobClient.ExistsAsync(cancellationToken);
```
bcarthic, line 205: I think you don't have to do 2 calls. DeleteExistsAsync `Returns true if container exists and was marked for deletion, return false otherwise.`

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -189,15 +189,45 @@ public class BlobFileStore : IFileStore
...
+                        version, partition.Key);
+                }
+
+                return exists;
+            }
+            catch (RequestFailedException ex) when (ex.ErrorCode == BlobErrorCode.ConditionNotMet &&
+                                                    _blobClient.IsExternal)
+            {
+                // when using external store and condition on etag match not met but blob exists, rethrow
+                throw ex;
```
bcarthic, line 228: So in this case, the user will see DataStoreException instead of ItemNotFoundException right?

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -189,15 +189,45 @@ public class BlobFileStore : IFileStore
...
 
         EmitTelemetry(nameof(DeleteFileIfExistsAsync), OperationType.Input);
 
-        await ExecuteAsync(() => blobClient.DeleteIfExistsAsync(DeleteSnapshotsOption.IncludeSnapshots, conditions: null, cancellationToken));
+        await ExecuteAsync(async () =>
+        {
+            try
+            {
+                bool exists = await blobClient.ExistsAsync(cancellationToken);
+                // todo change this back to DeleteIfExistsAsync when bug fixed story#110757
```
wsugarman, line 206: Did we file an issue for this?

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
wsugarman: What exactly happens in this case? I assume you get a `412` back, right? Is there a separate error code or inner exception we could use to differentiate? I ask because if we check and delete in two separate calls, there is always the chance something happened in between

wsugarman: In fact -- do we care why it failed outside of logs? If it does fail for either reason, we can just query the state of the blob afterwards, right?

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Activity.cs
```diff
@@ -198,12 +199,94 @@ public async Task DeleteOldVersionBlobV2Async([ActivityTrigger] CleanupBlobArgum
...
+    /// <paramref name="arguments"/> or <paramref name="logger"/> is <see langword="null"/>.
+    /// </exception>
+    [FunctionName(nameof(DeleteOldVersionBlobV3Async))]
+    public async Task DeleteOldVersionBlobV3Async([ActivityTrigger] CleanupBlobArgumentsV2 arguments, ILogger logger)
+    {
+        EnsureArg.IsNotNull(arguments, nameof(arguments));
+        EnsureArg.IsNotNull(arguments.Partition, nameof(arguments.Partition));
+        EnsureArg.IsNotNull(logger, nameof(logger));
+        EnsureArg.IsNotNull(arguments.Instances, nameof(arguments.Instances));
+
```
wsugarman, line 226: nit: extra line

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Activity.cs
```diff
@@ -198,12 +199,94 @@ public async Task DeleteOldVersionBlobV2Async([ActivityTrigger] CleanupBlobArgum
...
+    /// <exception cref="ArgumentNullException">
+    /// <paramref name="arguments"/> or <paramref name="logger"/> is <see langword="null"/>.
+    /// </exception>
+    [FunctionName(nameof(CleanupNewVersionBlobV3Async))]
+    public async Task CleanupNewVersionBlobV3Async([ActivityTrigger] CleanupBlobArgumentsV2 arguments, ILogger logger)
+    {
+        EnsureArg.IsNotNull(arguments, nameof(arguments));
+        EnsureArg.IsNotNull(arguments.Partition, nameof(arguments.Partition));
+        EnsureArg.IsNotNull(logger, nameof(logger));
+
```
wsugarman, line 266: nit: extra newline

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Orchestration.cs
```diff
@@ -165,13 +166,158 @@ public partial class UpdateDurableFunction
...
+    /// </remarks>
+    /// <param name="context">The context for the orchestration instance.</param>
+    /// <param name="logger">A diagnostic logger.</param>
+    /// <returns>A task representing the <see cref="UpdateInstancesV3Async"/> operation.</returns>
+    /// <exception cref="ArgumentNullException">
+    /// <paramref name="context"/> or <paramref name="logger"/> is <see langword="null"/>.
+    /// </exception>
+    /// <exception cref="FormatException">Orchestration instance ID is invalid.</exception>
+    [FunctionName(nameof(UpdateInstancesV3Async))]
+    public async Task UpdateInstancesV3Async(
```
wsugarman, line 184: Is the only difference the calls to v3 activities?

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStoreV48.cs
```diff
@@ -0,0 +1,99 @@
...
+                            VLatest.DeletedInstance.StudyInstanceUid,
+                            VLatest.DeletedInstance.SeriesInstanceUid,
+                            VLatest.DeletedInstance.SopInstanceUid,
+                            VLatest.DeletedInstance.Watermark,
+                            VLatest.DeletedInstance.OriginalWatermark,
+                            VLatest.FileProperty.ETag.AsNullable(),
+                            VLatest.FileProperty.FilePath.AsNullable());
+
+                        results.Add(
+                        new InstanceMetadata(
```
wsugarman, line 61: nit: indent

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStoreV48.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        using (SqlConnectionWrapper sqlConnectionWrapper = await SqlConnectionWrapperFactory.ObtainSqlConnectionWrapperAsync(cancellationToken, true))
+        using (SqlCommandWrapper sqlCommandWrapper = sqlConnectionWrapper.CreateRetrySqlCommand())
+        {
+            VLatest.RetrieveDeletedInstanceV42.PopulateCommand(
+                sqlCommandWrapper,
+                batchSize,
+                maxRetries);
+
+            using (var reader = await sqlCommandWrapper.ExecuteReaderAsync(CommandBehavior.SequentialAccess, cancellationToken))
```
wsugarman, line 43: nit: use `using` statements

src/Microsoft.Health.Dicom.SqlServer/Features/Store/SqlIndexDataStoreV48.cs
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        return results;
+    }
+
+    private static InstanceProperties CreateInstanceProperties(string eTag, string filePath, long? originalWatermark)
+    {
+        if (eTag != null && filePath != null)
+        {
+            return new InstanceProperties()
```
wsugarman, line 86: nit: remove `()`

src/Microsoft.Health.Dicom.Core/Features/Delete/DeleteService.cs
```diff
@@ -119,18 +119,18 @@ public async Task<(bool Success, int RetrievedInstanceCount)> CleanupDeletedInst
...
-                        {
-                            _fileStore.DeleteFileIfExistsAsync(deletedInstanceIdentifier.VersionedInstanceIdentifier.Version, deletedInstanceIdentifier.VersionedInstanceIdentifier.Partition.Name, cancellationToken),
+                        Task[] tasks = {
+                            _fileStore.DeleteFileIfExistsAsync(deletedInstanceIdentifier.VersionedInstanceIdentifier.Version, deletedInstanceIdentifier.VersionedInstanceIdentifier.Partition, deletedInstanceIdentifier.InstanceProperties.fileProperties, cancellationToken),
                             _metadataStore.DeleteInstanceMetadataIfExistsAsync(deletedInstanceIdentifier.VersionedInstanceIdentifier.Version, cancellationToken),
                             _metadataStore.DeleteInstanceFramesRangeAsync(deletedInstanceIdentifier.VersionedInstanceIdentifier.Version, cancellationToken),
                         };
 
-                        if (deletedInstanceIdentifier.InstanceProperties.OriginalVersion.HasValue)
+                        // only need to delete by "original watermark" to catch updates if not IDP
```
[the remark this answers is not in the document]
wsugarman: Oh, so in the input `deletedInstanceIdentifiers` we're going to have a row for each version in IDP, but for non-IDP we'll have a single row whose original version needs to be explicitly deleted below? Can you add a comment to that effect then?

wsugarman (dismissed):

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Orchestration.cs
wsugarman: Oh, I'm not going to ask that lolol (don't fret). There was just a lot of "new" code and I wanted to confirm that the differences are pretty insignificant

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
wsugarman: Sure, but the message can be pretty straight-forward if we don't check. That would significantly simplify this and remove any race conditions.

```
"There was a problem deleting the DICOM file with SOP instance ID foo. Expected the file to be at '/foo/bar/baz.dcm' with eTag 0x12345, but the file has since been changed or deleted."
```

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -189,15 +189,45 @@ public class BlobFileStore : IFileStore
...
+        EnsureArg.IsNotNull(partition);
 
-        BlockBlobClient blobClient = GetInstanceBlockBlobClient(version, partitionName);
-        _logger.LogInformation("Trying to delete DICOM instance file with watermark '{Version}'.", version);
+        BlockBlobClient blobClient = GetInstanceBlockBlobClient(version, partition, fileProperties);
+        _logger.LogInformation("Trying to delete DICOM instance file with watermark: '{Version}' and PartitionKey: {PartitionKey}.", version, partition.Key);
 
         EmitTelemetry(nameof(DeleteFileIfExistsAsync), OperationType.Input);
 
-        await ExecuteAsync(() => blobClient.DeleteIfExistsAsync(DeleteSnapshotsOption.IncludeSnapshots, conditions: null, cancellationToken));
```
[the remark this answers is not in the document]
wsugarman: Sure, see my other comment for unifying it

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -202,31 +209,29 @@ public async Task DeleteFileIfExistsAsync(long version, Partition partition, Fil
...
+                    cancellationToken);
             }
             catch (RequestFailedException ex) when (ex.ErrorCode == BlobErrorCode.ConditionNotMet &&
                                                     _blobClient.IsExternal)
             {
-                // when using external store and condition on etag match not met but blob exists, rethrow
-                throw ex;
+                string message = string.Format(
+                    CultureInfo.InvariantCulture,
+                    DicomCoreResource.ExternalDataStoreOperationFailed,
```
wsugarman, line 223: Should this be a different resource string, like:

```xml
<data name="ExternalDataStoreBlobModified" xml:space="preserve">
  <value>The blob '{0}' with eTag '{1}' has been modified or deleted.</value>
</data>
```

Is it worth promoting this new message to `ExecuteAsync` where we can reuse this message for things like WADO as well. Shouldn't all of the `Get*` methods in this class attach an eTag condition to their respective API calls?

wsugarman (approved):

PR #3111, opened 2023-10-16
handle private tag null default vrs

esmadau:
## Description
When a private tag is being validated, we use its vr to emit telemetry to understand issues that come up. Using DefaultVR returns nulls for private tags even when a VR is available. Instead of using defaultVR, use the VR available off of the DicomItem.
Also, add extra checks so telemetry doesn't cause issues when data not present

## Related issues
Addresses [[AB#110733](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/110733)].

## Testing
Added unit test

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Features/Store/StoreDatasetValidator.cs
+++ b/src/Microsoft.Health.Dicom.Core/Features/Store/StoreDatasetValidator.cs
@@ -256,7 +256,7 @@ public class StoreDatasetValidator : IStoreDatasetValidator
                 validationResultBuilder.Add(ex, requiredCoreTag, isCoreTag: true);
                 _logger.LogInformation("Dicom instance validation failed with error on required core tag {Tag} with id {Id}", requiredCoreTag.DictionaryEntry.Keyword, requiredCoreTag);
                 _storeMeter.V2ValidationError.Add(1,
-                    TelemetryDimension(requiredCoreTag, IsIndexableTag(queryTags, requiredCoreTag)));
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.Dicom.Core/Features/Store/StoreDatasetValidator.cs
```diff
@@ -365,14 +365,14 @@ private void ValidateWithoutNullPadding(string value, DicomElement de, IReadOnly
...
 
-    private static KeyValuePair<string, object>[] TelemetryDimension(DicomTag tag, bool isIndexableTag) =>
+    private static KeyValuePair<string, object>[] TelemetryDimension(DicomTag tag, DicomVR vr, bool isIndexableTag) =>
         new[]
         {
-            new KeyValuePair<string, object>("TagKeyword", tag.DictionaryEntry.Keyword),
-            new KeyValuePair<string, object>("VR", tag.GetDefaultVR().ToString()),
-            new KeyValuePair<string, object>("Tag", tag.ToString()),
+            new KeyValuePair<string, object>("TagKeyword", tag?.DictionaryEntry?.Keyword),
+            new KeyValuePair<string, object>("VR", vr?.ToString()),
```
bcarthic, line 374: I think if you just do tag.GetDefaultVR() instead of ToString() that is enough right. We dont want all this change?

bcarthic (approved):

PR #3129, opened 2023-10-23
Update renovate.json to skip IdentityMode 7.x

bcarthic:
## Description
Update renovate.json to skip IdentityMode 7.x

## Related issues
Addresses [issue #].

## Testing

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -48,8 +48,8 @@
       "matchPackagePatterns": [
         "IdentityModel.Tokens"
       ],
-      "ignoreVersions": [
-        ">=7.0.0"
+      "allowedVersions": [
+        "<7.0.0"
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -47,6 +47,9 @@
       "groupName": "IdentityModel.Tokens",
       "matchPackagePatterns": [
         "IdentityModel.Tokens"
+      ],
+      "allowedVersions": [
```
wsugarman, line 51: `"allowedVersions"` does not appear to be an `array` but a `string`: https://docs.renovatebot.com/configuration-options/#allowedversions

Also could you alphabetize this?

wsugarman (changes requested):

renovate.json
bcarthic: ah, missed that. I like to read the package name first and then the rest, but anyways

wsugarman (approved):

PR #3130, opened 2023-10-24
GetFileFrameAsync now uses file properties in sql

esmadau:
## Description
GetFileFrameAsync now uses file properties in sql and uses etag condition matching

## Related issues
Addresses [[AB#109237](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/109237)].

## Testing
Added unit tests to exercise when we do not need transcoding and frames present path to use GetFileFrameAsync  and ensure file properties get passed in. E2E tests will run with external store enabled and will exercise this path already.

### DIFF
--- a/src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
+++ b/src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
@@ -78,7 +78,7 @@ public class BlobFileStore : IFileStore
     {
         EnsureArg.IsNotNull(stream, nameof(stream));
 
-        BlockBlobClient blobClient = GetInstanceBlockBlobClient(version, partitionName);
+        BlockBlobClient blobClient = GetNewInstanceBlockBlobClient(version, partitionName);
 
         var blobUploadOptions = new BlobUploadOptions { TransferOptions = _options.Upload };
         stream.Seek(0, SeekOrigin.Begin);
@@ -108,7 +108,7 @@ public class BlobFileStore : IFileStore
         EnsureArg.IsNotNull(blockLengths, nameof(blockLengths));
         EnsureArg.IsGte(blockLengths.Count, 0, nameof(blockLengths.Count));
 
-        BlockBlobClient blobClient = GetInstanceBlockBlobClient(version, partition.Name);
+        BlockBlobClient blobClient = GetNewInstanceBlockBlobClient(version, partition.Name);
 
         int maxBufferSize = (int)blockLengths.Max(x => x.Value);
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -438,6 +445,14 @@ protected virtual BlockBlobClient GetInstanceBlockBlobClient(long version, strin
...
     }
 
+    /// <summary>
+    /// Get client based on watermark/version and partition when using internal store and based on file properties
+    /// when using external store.
+    /// </summary>
+    /// <param name="version">Version of file to get</param>
+    /// <param name="partition">Partition within which the file should live in</param>
+    /// <param name="fileProperties">File properties to use for external store. If not using external store, set to null.</param>
+    /// <remarks>Always use on existing files whether using for external or internal store.</remarks>
```
bcarthic, line 455: Do you think can we simplify into one method? We can return the client based on fileProperties being null or not. Some one may use wrong method.

bcarthic: I get that, but I still don't understand why we can't have in one method. 

```
protected virtual BlockBlobClient GetExistingInstanceBlockBlobClient(long version, Partition partition, FileProperties fileProperties)
{
	EnsureArg.IsNotNull(partition, nameof(partition));
	if (_blobClient.IsExternal && fileProperties != null)
	{
		// does not throw, just appends uri with blobName
		return _blobClient.BlobContainerClient.GetBlockBlobClient(fileProperties.Path);
	}
	string blobName = _nameWithPrefix.GetInstanceFileName(version);
	string fullPath = _blobClient.GetServiceStorePath(partition.Name) + blobName;
	// does not throw, just appends uri with blobName
	return _blobClient.BlobContainerClient.GetBlockBlobClient(fullPath);
}
```

bcarthic: I am fine keeping it separate, it is just it may lead to wrong usage.

src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
```diff
@@ -213,6 +213,7 @@ public async Task<RetrieveResourceResponse> GetInstanceResourceAsync(RetrieveRes
                 framesRange,
                 message.Frames,
                 responseTransferSyntax,
+                instance.InstanceProperties.fileProperties,
```
bcarthic, line 216: nit: shouldn't "f" be capital?

bcarthic (approved):

PR #3145, opened 2023-10-26
Emit Delete Metrics from Background Service

wsugarman:
## Description
This PR moves the egress of deletion-related metrics from a health check to the deletion background service itself. It also lightly refactors the `DeleteService` to reduce its complexity without affecting the code paths. In the future, the deletion mechanism should be reassessed.

## Related issues
N/A

## Testing
New tests


src/Microsoft.Health.Dicom.Core/Features/HealthCheck/BackgroundServiceHealthCheck.cs
```diff
@@ -1,87 +0,0 @@
...
-            {
-                // if the customer-managed key is inaccessible, the data store will also be inaccessible
-                return new HealthCheckResult(
-                    HealthStatus.Degraded,
-                    DegradedDescription,
-                    cmkStatus.Exception,
-                    new Dictionary<string, object> { { "Reason", cmkStatus.Reason } });
-            }
-
-            Task<DateTimeOffset> oldestWaitingToBeDeleted = _backgroundServiceHealthCheckCache.GetOrAddOldestTimeAsync(_indexDataStore.GetOldestDeletedAsync, cancellationToken);
```
wsugarman, line 72: The only noticeable logic in this health check is the querying and logging of these metrics

jnlycklama (dismissed):

src/Microsoft.Health.Dicom.Api/Features/BackgroundServices/DeletedInstanceCleanupWorker.cs
```diff
@@ -42,15 +44,17 @@ public async Task ExecuteAsync(CancellationToken stoppingToken)
...
 
-                bool success;
-                int retrievedInstanceCount;
+                DeleteSummary summary;
                 do
                 {
-                    (success, retrievedInstanceCount) = await _deleteService.CleanupDeletedInstancesAsync(stoppingToken);
+                    summary = await _deleteService.CleanupDeletedInstancesAsync(stoppingToken);
+
+                    if (summary.Metrics != null)
```
KnicKnic, line 54: I would suggest writting metrics of delete time first, as they don't request as much logic, and if there is any error they still write

PR #3153, opened 2023-10-30
Fix Workload Identity Argument

wsugarman:
## Description
Fix usage of the new `--enable-workload-identity` and `-wi` arguments. Without the new option, the arguments are flagged as invalid and the app fails to run

## Related issues
[AB#111048](https://microsofthealth.visualstudio.com/Health/_workitems/edit/111048/)

## Testing
Tested locally and equivalent changes in PaaS

### DIFF
--- a/src/Microsoft.Health.Dicom.SchemaManager.Console/Properties/launchSettings.json
+++ b/src/Microsoft.Health.Dicom.SchemaManager.Console/Properties/launchSettings.json
@@ -2,7 +2,7 @@
   "profiles": {
     "Microsoft.Health.Dicom.SchemaManager": {
       "commandName": "Project",
-      "commandLineArgs": "apply --connection-string \"server=(local);Initial Catalog=DICOM3;TrustServerCertificate=True;Integrated Security=True\" --version 20",
+      "commandLineArgs": "apply --connection-string \"server=(local);Initial Catalog=DICOM3;TrustServerCertificate=True;Integrated Security=True\" --version 20 -wi=false",
       "environmentVariables": {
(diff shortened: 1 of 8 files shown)


src/Microsoft.Health.Dicom.SchemaManager.Console/Properties/launchSettings.json
```diff
@@ -2,7 +2,7 @@
   "profiles": {
     "Microsoft.Health.Dicom.SchemaManager": {
       "commandName": "Project",
-      "commandLineArgs": "apply --connection-string \"server=(local);Initial Catalog=DICOM3;TrustServerCertificate=True;Integrated Security=True\" --version 20",
+      "commandLineArgs": "apply --connection-string \"server=(local);Initial Catalog=DICOM3;TrustServerCertificate=True;Integrated Security=True\" --version 20 -wi=false",
```
wsugarman, line 5: I would have liked to use a switch argument, but this sort of argument works better with our usage in PaaS (and the underlying command line config source doesn't work with a switch)

src/Microsoft.Health.Dicom.SchemaManager/CommandLineOptions.cs
```diff
@@ -19,6 +19,4 @@ public class CommandLineOptions
 
     [Obsolete("Use Connection String instead for different authentication types.")]
     public string? ManagedIdentityClientId { get; set; }
-
-    public bool EnableWorkloadIdentity { get; set; }
```
wsugarman, line 23: This was unused. I read the value directly from the config

PR #3172, opened 2023-11-03
start writing contentLength on file properties

esmadau:
start writing contentLength and conditionally retrieve FP if content length not available

## Description

## Related issues
Addresses [issue #].

## Testing

### DIFF
--- a/src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/50.sql
+++ b/src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Migrations/50.sql
@@ -284,11 +284,10 @@ CREATE NONCLUSTERED INDEX IX_ExtendedQueryTagString_PartitionKey_ResourceType_So
     ON dbo.ExtendedQueryTagString(PartitionKey, ResourceType, SopInstanceKey1, SopInstanceKey2, SopInstanceKey3) WITH (DATA_COMPRESSION = PAGE);
 
 CREATE TABLE dbo.FileProperty (
-    InstanceKey   BIGINT          NOT NULL,
-    Watermark     BIGINT          NOT NULL,
-    FilePath      NVARCHAR (4000) NOT NULL,
-    ETag          NVARCHAR (4000) NOT NULL,
-    ContentLength BIGINT          NOT NULL
+    InstanceKey BIGINT          NOT NULL,
+    Watermark   BIGINT          NOT NULL,
+    FilePath    NVARCHAR (4000) NOT NULL,
+    ETag        NVARCHAR (4000) NOT NULL
 )
 WITH (DATA_COMPRESSION = PAGE);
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -296,10 +296,25 @@ public async Task<Stream> GetStreamingFileAsync(long version, Partition partitio
...
+            return fileProperties.ContentLength;
+        }
+        FileProperties fp = await GetFilePropertiesAsync(version, partition, fileProperties, cancellationToken);
+        return fp.ContentLength;
+    }
+
     /// <inheritdoc />
     public async Task<FileProperties> GetFilePropertiesAsync(long version, Partition partition, FileProperties fileProperties, CancellationToken cancellationToken)
     {
+        // until we have backfilled contentLength, we have to continue to support retrieving file properties off of
```
wsugarman, line 313: How are we going to backfill this entirely so that this logic is no longer necessary? Will we say that all of these preview IDP services are invalid? Will we write a function like we're done in the past to enumerate the data?

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -296,10 +296,25 @@ public async Task<Stream> GetStreamingFileAsync(long version, Partition partitio
         });
     }
 
+    /// <inheritdoc />
+    public async Task<long> GetFilePropertiesContentLengthAsync(long version, Partition partition, FileProperties fileProperties, CancellationToken cancellationToken)
```
wsugarman, line 300: Should this instead be an extension method that unifies both `IFileStore` and `IInstanceStore`? Essentially we want to join `GetInstanceIdentifierWithPropertiesAsync` and `GetFilePropertiesAsync`

wsugarman: Things I can think of off the top of my head:
- Some new abstraction that leverages both `IFileStore` and `IInstanceStore`
- An extension method on `IInstanceStore` that does `GetInstanceIdentifierWithPropertiesAsync` with an `IFileStore` as an argument
- An extension method of `FileProperties` that uses an `IFileStore`

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Sprocs/UpdateInstanceStatusV50.sql
```diff
@@ -0,0 +1,100 @@
...
+    INSERT INTO dbo.ChangeFeed
+        (Timestamp, Action, PartitionKey, StudyInstanceUid, SeriesInstanceUid, SopInstanceUid, OriginalWatermark, FilePath)
+    VALUES
+        (@currentDate, 0, @partitionKey, @studyInstanceUid, @seriesInstanceUid, @sopInstanceUid, @watermark, @path)
+
+    -- Update existing instance currentWatermark to latest
+    UPDATE dbo.ChangeFeed
+    SET CurrentWatermark      = @watermark
+    WHERE PartitionKey = @partitionKey
+        AND StudyInstanceUid    = @studyInstanceUid
```
wsugarman, line 95: I think this indentation needs work. If we're indenting, then it should be align and so should the query above. Otherwise, skip it

src/Microsoft.Health.Dicom.SqlServer/Features/Schema/Sql/Tables/FileProperty.sql
```diff
@@ -12,7 +12,9 @@ CREATE TABLE dbo.FileProperty (
     FilePath        NVARCHAR (4000) NOT NULL,
     -- ETag is of unspecified size and its generation is opaque to us. In practice, these are typically short hashes 
     -- no more than 100 characters long. We will use max available as it is a varchar.
-    ETag            NVARCHAR (4000)  NOT NULL
+    ETag            NVARCHAR (4000)  NOT NULL,
+    -- ContentLength represents length os stream of blob we ended up storing
```
wsugarman, line 16: typo

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
wsugarman: Sounds good to me!

PR #3174, opened 2023-11-03
refactor blobFileStore so we emit telemetry as part of execute async

esmadau:
## Description
refactor blobFileStore so we emit telemetry as part of execute async

## Related issues
Addresses [[AB#111889](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/111889)].

## Testing
Ensured all existing tests continue to run

### DIFF
--- a/src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
+++ b/src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
@@ -208,7 +208,7 @@ public class BlobFileStore : IFileStore
     }
 
     /// <inheritdoc />
-    public Task DeleteFileIfExistsAsync(
+    public async Task DeleteFileIfExistsAsync(
         long version,
         Partition partition,
         FileProperties fileProperties,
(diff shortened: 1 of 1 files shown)


src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -362,24 +395,32 @@ public class BlobFileStore : IFileStore
             Conditions = _blobClient.GetConditions(fileProperties),
         };
 
-        Response<BlobDownloadResult> result = await ExecuteAsync(
-            action: async () => await blob.DownloadContentAsync(blobDownloadOptions, cancellationToken),
+        BlobDownloadResult result = await ExecuteAsync(
+            func: async () =>
+            {
+                Response<BlobDownloadResult> result = await blob.DownloadContentAsync(blobDownloadOptions, cancellationToken);
```
wsugarman, line 401: If you did want to remove the code block, you could rely on the [`implicit` conversion operation of `Response<T>` to `T`](https://github.com/Azure/azure-sdk-for-net/blob/c284046de29a1561373843a0f4f33600c474348a/sdk/core/Azure.Core/src/ResponseOfT.cs#L32). You'll need to provide a hint to the generic method, but I think this should compile:

```csharp
BlobDownloadResult result = await ExecuteAsync<BlobDownloadResult>(
    func: async () => await blob.DownloadContentAsync(blobDownloadOptions, cancellationToken),
    operationName: nameof(GetFileContentInRangeAsync),
    operationType: OperationType.Output,
    extractLength: long? (result) => result.Details.ContentLength);
```

wsugarman (approved):

PR #3190, opened 2023-11-10
Get correct file path on copying files

esmadau:
## Description
Get correct file path on copying files by always passing in null file properties when getting file properties for a copied file as it is a new file and we need the new path

## Related issues
Addresses [[AB#112069](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/112069)].

## Testing
Updated unit test to assert we are passing in null for fp


jnlycklama (approved):

bcarthic (commented):
Looks good to me

PR #3192, opened 2023-11-11
idp handle host unknown

esmadau:
## Description
when host is not known, handle exception so we give back a better error message

## Related issues
Addresses [[AB#112091](https://microsofthealth.visualstudio.com/f8da5110-49b1-4e9f-9022-2f58b6124ff9/_workitems/edit/112091)].

## Testing

### DIFF
--- a/src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
+++ b/src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
@@ -558,7 +558,7 @@ public class BlobFileStore : IFileStore
             _logger.LogError(ex, message: "Access to storage account failed with ErrorCode: {ErrorCode}", ex.ErrorCode);
             throw new DataStoreRequestFailedException(ex, _blobClient.IsExternal);
         }
-        catch (AggregateException ex) when (_blobClient.IsExternal && ex.InnerException?.GetType() == typeof(RequestFailedException))
+        catch (AggregateException ex) when (_blobClient.IsExternal && ex.InnerException is RequestFailedException)
         {
             var innerEx = ex.InnerException as RequestFailedException;
             _logger.LogError(innerEx,


KnicKnic (dismissed):

src/Microsoft.Health.Dicom.Core/Exceptions/DataStoreRequestFailedException.cs
```diff
@@ -17,10 +18,33 @@ public class DataStoreRequestFailedException : ConditionalExternalException
...
+                    CultureInfo.InvariantCulture,
+                    DicomCoreResource.ExternalDataStoreOperationFailed,
+                    ex?.ErrorCode)
+                : GetFormattedExternalStoreMessageWithoutErrorCode(ex)
+            ;
+    }
+
+    private static string GetFormattedExternalStoreMessageWithoutErrorCode(RequestFailedException ex)
+    {
+        if (ex.Message.Contains("No such host is known.", StringComparison.OrdinalIgnoreCase))
```
bcarthic, line 41: nit: I would remove "." at the end, if in case the sentence changes

src/Microsoft.Health.Dicom.Blob/Features/Storage/BlobFileStore.cs
```diff
@@ -558,6 +558,14 @@ protected virtual BlockBlobClient GetNewInstanceBlockBlobClient(long version, st
             _logger.LogError(ex, message: "Access to storage account failed with ErrorCode: {ErrorCode}", ex.ErrorCode);
             throw new DataStoreRequestFailedException(ex, _blobClient.IsExternal);
         }
+        catch (AggregateException ex) when (_blobClient.IsExternal && ex.InnerException?.GetType() == typeof(RequestFailedException))
```
bcarthic, line 561: In a aggregate exception, the inner Exception could be a list. Do you want to do something like

`aggEx.InnerExceptions.Any(x => x is RequestFailedException)`

bcarthic: okay, atleast you could do ex.InnerException is RequestFailedException to simplify

bcarthic (approved):

PR #3232, opened 2023-11-29
pass along partition information to export checkpoint 

esmadau:
## Description
pass along partition information to checkpoint so we can access it on larger than batch failures

This checkpoint isn't above reply specifically. As an optimization for storage, we break up our activities into multiple orchestration instance executions via ContinueAsNew. If the operation replays, due to a failure of some kind, it will re-read the input. However, the failure here seems to instead be in export operations that operate on data larger than the batch size, resulting in multiple executions. In this case, we fail to propagate the partition information to the next orchestration instance.

## Related issues
Addresses [AB#112619].

## Testing
updated unit test

### DIFF
--- a/src/Microsoft.Health.Dicom.Functions.UnitTests/Export/ExportDurableFunctionTests.Orchestration.cs
+++ b/src/Microsoft.Health.Dicom.Functions.UnitTests/Export/ExportDurableFunctionTests.Orchestration.cs
@@ -205,9 +205,14 @@ public partial class ExportDurableFunctionTests
             .Received(1)
             .ContinueAsNew(
                 Arg.Is<ExportCheckpoint>(x =>
-                    x.Source == null &&
+                    x.Batching == checkpoint.Batching &&
+                    x.CreatedTime == checkpoint.CreatedTime &&
+                    x.Destination == checkpoint.Destination &&
+                    x.ErrorHref == checkpoint.ErrorHref &&
                     x.Progress == new ExportProgress(1236, 56) &&
-                    x.Partition == checkpoint.Partition),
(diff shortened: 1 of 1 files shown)


bcarthic (approved):

src/Microsoft.Health.Dicom.Functions.UnitTests/Export/ExportDurableFunctionTests.Orchestration.cs
```diff
@@ -204,7 +204,15 @@ await context
...
-                Arg.Is<ExportCheckpoint>(x => x.Source == null && x.Progress == new ExportProgress(1236, 56)),
+                Arg.Is<ExportCheckpoint>(x =>
+                    x.Batching == checkpoint.Batching &&
+                    x.CreatedTime == checkpoint.CreatedTime &&
+                    x.Destination == checkpoint.Destination &&
+                    x.ErrorHref == checkpoint.ErrorHref &&
+                    x.Progress == new ExportProgress(1236, 56) &&
+                    x.Source == null &&
+                    x.Partition == checkpoint.Partition
+                    ),
```
wsugarman, line 215: nit: move `);` above

wsugarman (approved):
As a note: this checkpoint isn't above reply specifically. As an optimization for storage, we break up our activities into multiple orchestration instance executions via `ContinueAsNew`. If the operation replays, due to a failure of some kind, it will re-read the input. However, the failure here seems to instead be in export operations that operate on data larger than the batch size, resulting in multiple executions. In this case, we fail to propagate the partition information to the next orchestration instance.

PR #3238, opened 2023-12-01
Handle exceptions when CMK is misconfigured

jnlycklama:
## Description
Return 424 FailedDependency when a request fails due to an issue with CMK set up

## Related issues
Addresses #109384

## Testing
Tested via unit tests. These exceptions are known exceptions that the data plane of SQL & Storage account will return when they do not have access to the key


src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
```diff
@@ -160,6 +166,12 @@ private static bool IsTaskCanceledException(Exception ex)
         return ex is TaskCanceledException || (ex is AggregateException aggEx && aggEx.InnerExceptions.Any(x => x is TaskCanceledException));
     }
 
+    private static bool IsCMKException(Exception ex)
+    {
+        return ex is SqlException sqlEx && sqlEx.Number is 40981 or 33183 or 33184 or 40925 ||
+            ex is RequestFailedException rfEx && rfEx.ErrorCode == "KeyVaultEncryptionKeyNotFound";
```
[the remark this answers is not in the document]
jnlycklama: Created consts & added a comment about where the error codes come from. 

I don't know if theres a great way to test the SQL side - SqlException is a sealed class with no public constructor, so I'd have to do some ugly reflection stuff to get an instance of it to be able to add it to the unit tests. I didn't add it so far because I'm not sure its worth it, since we would just be injecting the #s anyways.

I was thinking of updating the CMK e2e test to call dicom while its in a bad state (part of the test is disabling the key) and making sure we get back the correct error code. That wouldn't test each SQL number, but it would be something

src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
```diff
@@ -160,6 +172,12 @@ private static bool IsTaskCanceledException(Exception ex)
         return ex is TaskCanceledException || (ex is AggregateException aggEx && aggEx.InnerExceptions.Any(x => x is TaskCanceledException));
     }
 
+    private static bool IsCMKException(Exception ex)
+    {
+        return ex is SqlException sqlEx && sqlEx.Number is SQLKeyVaultCriticalError or SQLKeyVaultEncounteredError or SQLKeyVaultErrorObtainingInfo or SQLCannotConnectToDBInCurrentState ||
```
bcarthic, line 177: Most of our SQL Exception are transformed to DataStoreException. May be you need to do 

```
var exception = ex is SqlException ? ex : (ex.InnerException is SqlException ? ex.InnerException : null)
if(exception != null) 
{
return exception .Number is SQLKeyVaultCriticalError or SQLKeyVaultEncounteredError or SQLKeyVaultErrorObtainingInfo or SQLCannotConnectToDBInCurrentState
}
```

src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
```diff
@@ -25,6 +27,12 @@ namespace Microsoft.Health.Dicom.Api.Features.Exceptions;
 
 public class ExceptionHandlingMiddleware
 {
+    // SQL Errors taken from https://learn.microsoft.com/en-us/sql/relational-databases/errors-events/database-engine-events-and-errors-31000-to-41399?view=sql-server-ver16
+    private const int SQLKeyVaultCriticalError = 40981;
```
bcarthic, line 31: As of now all the sql error codes are kept in health care shared components. Does it make sense to add here - https://github.com/microsoft/healthcare-shared-components/blob/main/src/Microsoft.Health.SqlServer/Features/Storage/SqlErrorCodes.cs

src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
bcarthic: And we have missed in some places, I have a PR to add a try catch in all the GET calls. #3239

bcarthic: Yes, but I believe this is enough and we can improve it later if we find it

src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
jnlycklama: Ah thanks I didn't know about this - will add it there. We are referencing these error codes in that repo anyways.

src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
```diff
@@ -94,6 +102,10 @@ private IActionResult MapExceptionToResult(Exception exception)
             case ConditionalExternalException ex when ex.IsExternal == true:
                 statusCode = HttpStatusCode.FailedDependency;
```
bcarthic, line 103: nit: You can remove line 103 and 104

src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
bcarthic: there is already a static method `IsCMKError(this SqlException sqlEx)`

bcarthic (approved):

bcarthic (approved):

PR #3242, opened 2023-12-05
Add failed instances while update to shoebox logs

bcarthic:
## Description
Add failed instances while update to shoebox logs

## Related issues
Addresses AB#111978

## Testing


src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Orchestration.cs
```diff
@@ -327,8 +185,17 @@ public partial class UpdateDurableFunction
         {
             // We don't want to populate all the errors in Azure Table Storage, DTFx may attempt to compress the entry as needed using GZip and storing in blob storage
             // But I think we should also be wary of what the user experience is for this via the response, so restricting to 5 errors for now. We can update based on feedback.
-            // TODO: Inform the user that the remaining failures can be found in the logs.
             errors.AddRange(instanceErrors.Take(5));
+
+            if (instanceErrors.Count > 5)
+            {
+                errors.Add("There are more instances failed to update than listed above. Please check the diagnostics logs for the complete list.");
```
wsugarman, line 192: Is it worth even having this "top 5" logic in the response? Should we just put all of the daa in the logs?

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Orchestration.cs
```diff
@@ -327,8 +185,17 @@ public partial class UpdateDurableFunction
...
+            }
+
+            foreach (string error in instanceErrors)
+            {
+                _telemetryClient.ForwardLogTrace(error);
```
wsugarman, line 197: So can these logs be correlated with the ongoing update operation? Are there other properties that should be included?

src/Microsoft.Health.Dicom.Functions.UnitTests/Update/UpdateDurableFunctionTests.cs
```diff
@@ -20,6 +21,7 @@
 using NSubstitute;
 using OpenTelemetry;
 using OpenTelemetry.Metrics;
+using Metric = OpenTelemetry.Metrics.Metric;
```
wsugarman, line 24: What type does this conflict with? There's something called `Metric` in `Microsoft.ApplicationInsights`?

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Orchestration.cs
bcarthic: We are not sure, how many enabled diagnostic logging. I would think this will be a step for them to do if they didnt do

src/Microsoft.Health.Dicom.Functions.UnitTests/Update/UpdateDurableFunctionTests.cs
bcarthic: Yes, there is Metric in Application Insights

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Orchestration.cs
bcarthic: This is a very good point. I will include the operation Id

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Orchestration.cs
```diff
@@ -223,4 +230,14 @@ private async Task TryCleanupActivityV3(IDurableOrchestrationContext context, IR
         }
         catch (Exception) { }
     }
+
+    private string GetSerializedInput(UpdateCheckpoint checkpoint) => JsonSerializer.Serialize(new
```
wsugarman, line 234: So we're re-using the `Input` column in the diagnostic table?

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Orchestration.cs
```diff
@@ -223,4 +230,14 @@ private async Task TryCleanupActivityV3(IDurableOrchestrationContext context, IR
...
+    {
```
wsugarman, line 235: I would use a block body expression for multiple lines

src/Microsoft.Health.Dicom.Functions/Update/UpdateDurableFunction.Orchestration.cs
wsugarman, line 234: nit: this isn't exactly the "input"

bcarthic: Yes we log while we create the operation, I want to log when we complete the operator or when we error

bcarthic:
Follow up PR #3249 to fix severity level

wsugarman (approved):

PR #3245, opened 2023-12-06
get fp content length directly from getting instances

esmadau:
## Description

## Related issues
Addresses [issue #].

## Testing


src/Microsoft.Health.Dicom.Core/Features/Retrieve/InstanceStoreExtensions.cs
```diff
@@ -19,21 +20,37 @@ public static class InstanceStoreExtensions
...
+        
+        // backfill each instance wit content length if none present and using external store
+        if (instancesToRetrieve.Count > 0 && instancesToRetrieve[0].InstanceProperties.FileProperties != null)
+        {
+
+            foreach (InstanceMetadata instance in instancesToRetrieve)
+            {
+                if (instance.InstanceProperties.FileProperties.ContentLength == 0)
+                {
+                    instance.InstanceProperties.FileProperties.ContentLength = (await blobFileStore.GetFilePropertiesAsync(instance.GetVersion(isOriginalVersionRequested), partition, instance.InstanceProperties.FileProperties, cancellationToken)).ContentLength;
```
bcarthic, line 50: I wouldn't do it in GET call. We can utilize our job framework to populate the content length if it needs to be back filled. And let's say if I get 1000 instances we are making 1000 extra calls to blob store.

bcarthic (changes requested):
I believe the back-fill should be done in a job rather than in a GET call

PR #3265, opened 2023-12-19
Catch operation cancelled inner exceptions in middleware

jnlycklama:
## Description
Catch msft health exceptions where the inner exception is OperationCanceledException (TaskCanceledException derives from this)

## Related issues
Addresses [113536].

## Testing
Added new unit tests

### DIFF
--- a/src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
+++ b/src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
@@ -5,6 +5,7 @@
 
 using System;
 using System.IO;
+using System.Linq;
 using System.Net;
 using System.Runtime.ExceptionServices;
 using System.Text.Json;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
```diff
@@ -90,8 +90,7 @@ private IActionResult MapExceptionToResult(Exception exception)
             case AuditHeaderTooLargeException:
             case ConnectionResetException:
             case OperationCanceledException:
-            case DicomImageException ex when IsTaskCanceledException(ex.InnerException):
-            case DataStoreException e when IsTaskCanceledException(e.InnerException):
+            case Exception e when IsOperationCanceledException(e.InnerException):
```
bcarthic, line 93: Instead of doing for all Exception, may be limit to either `DicomServerException` or `MicrosoftHealthException`

jnlycklama: Talked offline; initially did this because the RBAC exception base was just "Exception". Updated to be MicrosoftHealthException so we can reduce scope here

src/Microsoft.Health.Dicom.Api/Features/Exceptions/ExceptionHandlingMiddleware.cs
```diff
@@ -90,8 +90,8 @@ private IActionResult MapExceptionToResult(Exception exception)
             case AuditHeaderTooLargeException:
             case ConnectionResetException:
             case OperationCanceledException:
-            case DicomImageException ex when IsTaskCanceledException(ex.InnerException):
-            case DataStoreException e when IsTaskCanceledException(e.InnerException):
+            case MicrosoftHealthException ex when IsOperationCanceledException(ex.InnerException):
+            case DataStoreException e when IsOperationCanceledException(e.InnerException):
```
bcarthic, line 94: DataStoreException is also MicrosoftHealthException

bcarthic (approved):

PR #3272, opened 2024-01-02
Do not log cmk errors at critical level

jnlycklama:
## Description
When we see CMK misconfigurations, the delete instance background worker will either:
1. Throw a DataStoreNotReadyException. There is no polling wait in the while loop here, so this will log infinitely until we are throttled, which will kill the pod (at least I think so, its hard to see the behavior here because its overwhelmed with logs
2. Throw a CMK specific exception, which is not handled specially and logged as a critical error.
 Ex: {"Timestamp":"2023-12-20T05:08:31.4623481+00:00","Level":"Fatal","MessageTemplate":"Unhandled exception in the deleted instance cleanup worker.","Exception":"Microsoft.Data.SqlClient.SqlException (0x80131904): Can not connect to the database in its current state.\n   at Microsoft.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)\n   at Microsoft.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)\n 
...

Instead log CMK specific errors as Information, and add polling delay even if we encounter an error

## Related issues
Addresses #109382

## Testing
Ensure unit tests pass

### DIFF
--- a/src/Microsoft.Health.Dicom.Api/Features/BackgroundServices/DeletedInstanceCleanupWorker.cs
+++ b/src/Microsoft.Health.Dicom.Api/Features/BackgroundServices/DeletedInstanceCleanupWorker.cs
@@ -8,6 +8,7 @@ using System.Diagnostics.CodeAnalysis;
 using System.Threading;
 using System.Threading.Tasks;
 using EnsureThat;
+using Microsoft.Data.SqlClient;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Microsoft.Health.Dicom.Core.Configs;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Health.Dicom.Api/Features/BackgroundServices/DeletedInstanceCleanupWorker.cs
```diff
@@ -57,8 +59,6 @@ public async Task ExecuteAsync(CancellationToken stoppingToken)
                 _deleteMeter.OldestRequestedDeletion.Add(metrics.OldestDeletion.ToUnixTimeSeconds());
                 _deleteMeter.CountDeletionsMaxRetry.Add(metrics.TotalExhaustedRetries);
 
-                await Task.Delay(_pollingInterval, stoppingToken);
```
bcarthic, line 60: Should we do before the GetMetrics so that if the service is getting started and the background service won't error immediately if the datastore is not available right away?

bcarthic (approved):

bcarthic (approved):

PR #3275, opened 2024-01-03
Fix update test to use different value for extended query tag

bcarthic:
## Description
Fix update test to use different value for extended query tag

## Related issues
Addresses [issue #].

## Testing


test/Microsoft.Health.Dicom.Web.Tests.E2E/Rest/UpdateInstanceTests.cs
```diff
@@ -143,7 +143,7 @@ public async Task WhenUpdatingDicomMetadataWithExtendedQueryTagForASingleStudy_T
     {
         DicomTag ageTag = DicomTag.PatientAge;
         DicomTag patientSexTag = DicomTag.PatientSex;
-        string tagValue = "053Y";
```
bcarthic, line 146: Extended query tag tests was using the same value so there is a probability that when tests run in parallel it could results in wrong value. So changing the value and also querying for specific study

bcarthic: Update tests are not run in parallel but the different collections are run in parallel. I hope this should solve because I use specify study to validate.

PR #3278, opened 2024-01-03
Update Pkgs for .NET 8

wsugarman:
## Description
Update some packages, including some like OTel which must only be updated for .NET 8 as it relies upon v8 versions of the Microsoft.Extensions.* libraries.

## Related issues
N/A

## Testing
Existing test suites

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -28,6 +28,12 @@
       ]
     },
     {
+      "groupName": "fo-dicom",
+      "matchPackagePrefixes": [
+        "fo-dicom"
+      ]
+    },
+    {
       "excludePackagePrefixes": [
         "Microsoft.Health.Fhir."
       ],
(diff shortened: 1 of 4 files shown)


renovate.json
```diff
@@ -27,6 +27,12 @@
         "mcr.microsoft.com/dotnet/sdk"
       ]
     },
+    {
+      "groupName": "fo-dicom",
+      "matchPackagePrefixes": [
+        "fo-dicom"
+      ]
```
bcarthic, line 34: Should we update this

```
     "allowedVersions": "<7.0.0",
      "groupName": "IdentityModel.Tokens",
      "matchPackagePatterns": [
        "IdentityModel.Tokens"
      ]
```

wsugarman: Yes!

PR #3300, opened 2024-01-11
.NET 8 Update Part 3

wsugarman:
## Description
Bringing back the changes for .NET 8. This time without any `Regex` changes for UID values.

## Related issues
[AB#112434](https://microsofthealth.visualstudio.com/Health/_workitems/edit/112434)

## Testing
Existing test suite


src/Microsoft.Health.Dicom.Core/Features/Validation/UidValidation.cs
```diff
@@ -10,44 +10,42 @@
...
 
-internal class UidValidation : StringElementValidation
+internal partial class UidValidation : StringElementValidation
 {
-    private static readonly Regex ValidIdentifierCharactersFormat =
-        new Regex("^[0-9\\.]*[0-9]$", RegexOptions.Compiled);
+    // Note: This regex allows for leading zeroes.
+    // A more strict regex may look like: "^(0|([1-9][0-9]*))(\\.(0|([1-9][0-9]*)))*$"
+#if NET7_0_OR_GREATER
+    [GeneratedRegex("^[0-9\\.]*[0-9]$", RegexOptions.CultureInvariant | RegexOptions.ExplicitCapture | RegexOptions.Singleline)]
```
wsugarman, line 18: We're preserving the `Regex` now

src/Microsoft.Health.Dicom.Core/Features/Query/QueryResponseBuilder.cs
```diff
@@ -15,7 +16,7 @@ namespace Microsoft.Health.Dicom.Core.Features.Query;
 
 internal class QueryResponseBuilder
 {
-    internal static readonly HashSet<DicomTag> DefaultStudyTags = new HashSet<DicomTag>()
+    internal static readonly ImmutableHashSet<DicomTag> DefaultStudyTags = ImmutableHashSet.CreateRange(new DicomTag[]
```
wsugarman, line 19: The only new change is that I've changed these to `ImmutableHashSet<T>` objects (not frozen ones as they're multi-targeted for .NET 6 too).

src/Microsoft.Health.Dicom.Core.UnitTests/Features/Query/QueryServiceTests.cs
```diff
@@ -54,23 +54,24 @@ public QueryServiceTests()
     }
 
     [Theory]
-    [InlineData(QueryResource.StudySeries, "123.001")]
+    [InlineData(QueryResource.StudySeries, "123.001", Skip = "Enable once UID validation rejects leading zeroes.")]
```
wsugarman, line 57: This test and the test below were written incorrectly and never asserted anything 😲 They each invoke `Assert.ThrowsAsync` but did not `await` the result. As such, when fixed, they now fail because the `001` in the UID is considered valid. I've skipped these scenarios for now

bcarthic (approved):

PR #3301, opened 2024-01-11
Enable ASPNET Core Endpoint routing

arunmk-ms:
## Description
This PR updates the files to use Aspnetcore Endpoint Routing instead of MVC. Endpoint routing is needed to add [HttpLoggingMiddleware](https://eng.ms/docs/experiences-devices/r9-sdk/docs/telemetry/logging/http-logging-middleware) to get request logs in Otel migration work.

Registration is updated following the [migration guide](https://learn.microsoft.com/en-us/aspnet/core/migration/22-to-30?view=aspnetcore-6.0&tabs=visual-studio#routing-startup-code) 

Fetching routeContext for the bundle was using mvc router which is not provided in the endpoint routing framework. So adding the method to build routeContext for the bundle requests.

## Related issues
Addresses [issue #113900](https://microsofthealth.visualstudio.com/Health/_workitems/edit/113900).

## Testing


src/Microsoft.Health.Dicom.Api/Features/Routing/UrlResolver.cs
```diff
@@ -146,11 +151,45 @@ private Uri RouteUri(string routeName, RouteValueDictionary routeValues)
...
+                routeValues,
+                scheme,
+                host);
+        }
+        else
+        {
+            // NOTE: Append "/" to the end of the path base to workaround a known bug in UrlHelper.RouteUrl for endpoint routing.
+            //       Remove the workaround when we pick up the fix. (https://github.com/dotnet/aspnetcore/issues/53177)
+            var pathBase = httpContext.Request?.PathBase.ToString();
+            if (!string.IsNullOrEmpty(pathBase) && (string.IsNullOrEmpty(httpContext.Request?.Path) || string.Equals(httpContext.Request?.Path, "/", StringComparison.Ordinal)))
```
wsugarman, line 179: In what case is `Path` empty? According to this issue, it doesn't actually appear to be a bug

src/Microsoft.Health.Dicom.Api/Registration/DicomServerApplicationBuilderExtensions.cs
```diff
@@ -56,6 +72,39 @@ public static IApplicationBuilder UseDicomServer(this IApplicationBuilder app)
...
         }
 
+        app.UseRouting();
+        useDevelopmentIdentityProvider?.Invoke(app);
+        useHttpLoggingMiddleware?.Invoke(app);
+
+        app.UseEndpoints(endpoints =>
+        {
+            endpoints.MapControllers();
+            endpoints.MapHealthChecks(KnownRoutes.HealthCheck,
```
wsugarman, line 82: Is `MapHealthChecks` necessary if `UseCachedHealthChecks` is above in the pipeline?

src/Microsoft.Health.Dicom.Api/Registration/DicomServerApplicationBuilderExtensions.cs
```diff
@@ -56,6 +72,39 @@ public static IApplicationBuilder UseDicomServer(this IApplicationBuilder app)
             app.UseStaticFiles();
         }
 
+        app.UseRouting();
+        useDevelopmentIdentityProvider?.Invoke(app);
```
wsugarman, line 76: I am wondering if we can rewrite this function to not require such verbose arguments. Perhaps we don't even need this method. Like, we could create an extension method for the OHIF viewer addition. Otherwise, everything else could be explicitly added by a web project in the `Program.cs` or `Startup.cs`, right?

wsugarman (changes requested):

src/Microsoft.Health.Dicom.Api/Registration/DicomServerApplicationBuilderExtensions.cs
arunmk-ms: Removed `MapHealthChecks` as discussed offline.

src/Microsoft.Health.Dicom.Api/Registration/DicomServerApplicationBuilderExtensions.cs
arunmk-ms: Moved these to `startup.cs` as per the discussion offline.

src/Microsoft.Health.Dicom.Api/Features/Routing/UrlResolver.cs
```diff
@@ -146,11 +151,38 @@ private Uri RouteUri(string routeName, RouteValueDictionary routeValues)
...
+                request.Host.Value);
+    }
+
+    private Uri GetRouteUri(HttpContext httpContext, string routeName, RouteValueDictionary routeValues, string scheme, string host)
+    {
+        var uriString = string.Empty;
+
+        if (httpContext == null)
+        {
+            uriString = UrlHelper.RouteUrl(
```
wsugarman, line 168: nit: I think these can all go on the same line

src/Microsoft.Health.Dicom.Web/Startup.cs
```diff
@@ -57,9 +67,42 @@ public virtual void ConfigureServices(IServiceCollection services)
...
+        app.UseHttpsRedirection();
+
+        app.UseCachedHealthChecks(new PathString(KnownRoutes.HealthCheck));
+
+        // Update Fellow Oak DICOM services to use ASP.NET Core's service container
+        DicomSetupBuilder.UseServiceProvider(app.ApplicationServices);
+
+        IOptions<FeatureConfiguration> featureConfiguration = app.ApplicationServices.GetRequiredService<IOptions<FeatureConfiguration>>();
+        if (featureConfiguration.Value.EnableOhifViewer)
+        {
```
wsugarman, line 83: I think the contents of this if-block could be in an extension method so callers could just do:

```csharp
if (featureConfiguration.Values.EnableOhifViewer)
{
    app.UseOhifViewer()
}
```

src/Microsoft.Health.Dicom.Web/Startup.cs
```diff
@@ -57,9 +67,42 @@ public virtual void ConfigureServices(IServiceCollection services)
...
+            options.DefaultFileNames.Clear();
+            options.DefaultFileNames.Add(OhifViewerIndexPagePath);
+
+            app.UseDefaultFiles(options);
+            app.UseStaticFiles();
+        }
+
+        app.UseRouting();
+
+        app.UseEndpoints(endpoints =>
```
wsugarman, line 102: nit:
```suggestion
        app.UseEndpoints(endpoints => endpoints.MapControllers());
```

wsugarman (approved):

PR #3317, opened 2024-01-20
Update health check to handle concurrency

jnlycklama:
## Description
Write a new file for each health check, so we do not run into conflicts or not found errors during the check. 

However this means we need to be very careful that these files are cleaned up. We cant use the blob lifecycle, since that requires control plane access on the storage account. Instead we can use expiry headers - these are only valid on HNS enabled accounts, and they are only available via REST calls. They are not yet supported by the SDK. To get around this, I added in a pipeline policy that will inject the headers for every file that matches the health check file format. This will automatically delete the file after an hour if the deletes from the health check fail.

Also 
- updating errors caught since Karthik's instance was crashing https://microsofthealth.visualstudio.com/Health/_workitems/edit/114214

## Related issues
Addresses [106077].

## Testing
Ran code locally using my own storage account

### DIFF
--- a/src/Microsoft.Health.Dicom.Blob.UnitTests/Features/ExternalStore/MockRequest.cs
+++ b/src/Microsoft.Health.Dicom.Blob.UnitTests/Features/ExternalStore/MockRequest.cs
@@ -1,0 +1,45 @@
+// -------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+// -------------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using Azure.Core;
+
+namespace Microsoft.Health.Dicom.Blob.UnitTests.Features.ExternalStore;
+
+internal class MockRequest : Request
+{
+    public MockRequest()
+    {
+        MockHeaders = new List<HttpHeader>();
+    }
+
+    public List<HttpHeader> MockHeaders { get; }
+
+    public override string ClientRequestId { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
+
+    public override void Dispose() => throw new NotImplementedException();
+
+    protected override void AddHeader(string name, string value)
+    {
+        MockHeaders.Add(new HttpHeader(name, value));
+    }
+
+    protected override bool ContainsHeader(string name)
+    {
+        return MockHeaders.Any(h => h.Name == name);
+    }
+
+    protected override IEnumerable<HttpHeader> EnumerateHeaders()
+    {
+        return MockHeaders;
+    }
+
+    protected override bool RemoveHeader(string name) => throw new NotImplementedException();
+    protected override bool TryGetHeader(string name, [NotNullWhen(true)] out string value) => throw new NotImplementedException();
+    protected override bool TryGetHeaderValues(string name, [NotNullWhen(true)] out IEnumerable<string> values) => throw new NotImplementedException();
+}
--- a/src/Microsoft.Health.Dicom.Blob.UnitTests/Features/Health/DicomConnectedStoreHealthCheckTests.cs
+++ b/src/Microsoft.Health.Dicom.Blob.UnitTests/Features/Health/DicomConnectedStoreHealthCheckTests.cs
@@ -58,6 +58,8 @@ public class DicomConnectedStoreHealthCheckTests
     }
 
     [Theory]
+    [InlineData(401, "InvalidAuthenticationInfo")]
+    [InlineData(403, "InvalidAuthenticationInfo")]
     [InlineData(403, "AuthorizationFailure")]
     [InlineData(403, "AuthorizationPermissionMismatch")]
     [InlineData(403, "InsufficientAccountPermissions")]
@@ -84,6 +86,7 @@ public class DicomConnectedStoreHealthCheckTests
 
     [Theory]
     [InlineData(400, "SomeErrorFromDicomBug")]
+    [InlineData(401, "AuthErrorFromDicomBug")]
     [InlineData(403, "AuthErrorFromDicomBug")]
     [InlineData(404, "NotFoundDueToDicomBug")]
     [InlineData(409, "ConflictDueToDicomBug")]
--- a/src/Microsoft.Health.Dicom.Blob.UnitTests/Features/Health/DicomConnectedStoreHealthCheckTests.cs
+++ b/src/Microsoft.Health.Dicom.Blob.UnitTests/Features/Health/DicomConnectedStoreHealthCheckTests.cs
@@ -29,6 +29,7 @@ public class DicomConnectedStoreHealthCheckTests
 {
     private readonly BlobContainerClient _blobContainerClient = Substitute.For<BlobContainerClient>();
     private readonly BlockBlobClient _blockBlobClient = Substitute.For<BlockBlobClient>();
+    private readonly BlobLeaseClient _blobLeaseClient = Substitute.For<BlobLeaseClient>();
     private readonly IBlobClient _blobClient = Substitute.For<IBlobClient>();
     private readonly IOptions<ExternalBlobDataStoreConfiguration> _externalBlobOptions = Substitute.For<IOptions<ExternalBlobDataStoreConfiguration>>();
 
@@ -38,6 +39,7 @@ public class DicomConnectedStoreHealthCheckTests
     {
         _blobClient.BlobContainerClient.Returns(_blobContainerClient);
         _blobContainerClient.GetBlockBlobClient(Arg.Any<string>()).Returns(_blockBlobClient);
+        _blockBlobClient.GetBlobLeaseClient().Returns(_blobLeaseClient);
 
         _externalBlobOptions.Value.Returns(new ExternalBlobDataStoreConfiguration()
         {
@@ -69,7 +71,8 @@ public class DicomConnectedStoreHealthCheckTests
     [InlineData(404, "FilesystemNotFound")]
     [InlineData(409, "ContainerBeingDeleted")]
     [InlineData(409, "ContainerDisabled")]
-    public async Task GivenRequestFailedExceptionForCustomerError_RunHealthCheck_ReturnsDegradedAndDeletesBlob(int statusCode, string blobErrorCode)
+    [InlineData(400, "BlobTypeNotSupported")]
+    public async Task GivenRequestFailedExceptionForCustomerError_RunHealthCheck_ReturnsDegraded(int statusCode, string blobErrorCode)
     {
         _blockBlobClient.DownloadContentAsync(Arg.Any<CancellationToken>())
             .Throws(new RequestFailedException(statusCode, "Error", blobErrorCode, new System.Exception()));
@@ -80,8 +83,6 @@ public class DicomConnectedStoreHealthCheckTests
 
         Assert.Equal(HealthStatus.Degraded, result.Status);
         Assert.Equal(HealthStatusReason.ConnectedStoreDegraded.ToString(), healthStatusReason.ToString());
-
-        await _blockBlobClient.Received(2).DeleteIfExistsAsync(Arg.Any<DeleteSnapshotsOption>(), Arg.Any<BlobRequestConditions>(), Arg.Any<CancellationToken>());
     }
 
     [Theory]
@@ -89,20 +90,18 @@ public class DicomConnectedStoreHealthCheckTests
     [InlineData(403, "AuthErrorFromDicomBug")]
     [InlineData(404, "NotFoundDueToDicomBug")]
     [InlineData(409, "ConflictDueToDicomBug")]
-    public async Task GivenRequestFailedExceptionForServiceError_RunHealthCheck_ThrowsExceptionAndCleansUpBlob(int statusCode, string blobErrorCode)
+    public async Task GivenRequestFailedExceptionForServiceError_RunHealthCheck_ThrowsException(int statusCode, string blobErrorCode)
     {
         _blockBlobClient.DownloadContentAsync(Arg.Any<CancellationToken>())
             .Throws(new RequestFailedException(statusCode, "Error", blobErrorCode, new System.Exception()));
 
         await Assert.ThrowsAsync<RequestFailedException>(async () => await _dicomConnectedStoreHealthCheck.CheckHealthAsync(null, CancellationToken.None));
-
-        await _blockBlobClient.Received(2).DeleteIfExistsAsync(Arg.Any<DeleteSnapshotsOption>(), Arg.Any<BlobRequestConditions>(), Arg.Any<CancellationToken>());
     }
 
     [Fact]
     public async Task HostNotFound_RunHealthCheck_ReturnsDegraded()
     {
-        _blockBlobClient.StageBlockAsync(Arg.Any<string>(), Arg.Any<Stream>(), Arg.Any<BlockBlobStageBlockOptions>(), Arg.Any<CancellationToken>())
+        _blockBlobClient.UploadAsync(Arg.Any<Stream>(), cancellationToken: Arg.Any<CancellationToken>())
             .Throws(new AggregateException(new List<Exception>()
             {
                 new Exception("No such host is known."),
@@ -119,13 +118,15 @@ public class DicomConnectedStoreHealthCheckTests
     }
 
     [Fact]
-    public async Task AllDeleteFails_RunHealthCheck_ReturnsDegradedAndExceptionIsNotThrown()
+    public async Task NameOrServiceNotKnown_RunHealthCheck_ReturnsDegraded()
     {
-        _blockBlobClient.DeleteIfExistsAsync(Arg.Any<DeleteSnapshotsOption>(), Arg.Any<BlobRequestConditions>(), Arg.Any<CancellationToken>())
-            .Throws(new RequestFailedException(403, "Failure", BlobErrorCode.AuthorizationFailure.ToString(), new System.Exception()));
-
-        _blockBlobClient.DeleteAsync(Arg.Any<DeleteSnapshotsOption>(), Arg.Any<BlobRequestConditions>(), Arg.Any<CancellationToken>())
-            .Throws(new RequestFailedException(403, "Failure", BlobErrorCode.AuthorizationFailure.ToString(), new System.Exception()));
+        _blockBlobClient.UploadAsync(Arg.Any<Stream>(), cancellationToken: Arg.Any<CancellationToken>())
+            .Throws(new AggregateException(new List<Exception>()
+            {
+                new Exception("Name or service not known."),
+                new Exception("Name or service not known."),
+                new Exception("Name or service not known."),
+            }));
 
         HealthCheckResult result = await _dicomConnectedStoreHealthCheck.CheckHealthAsync(null, CancellationToken.None);
 
@@ -136,14 +137,16 @@ public class DicomConnectedStoreHealthCheckTests
     }
 
     [Fact]
-    public async Task DeleteBeforeAndAfterFails_RunHealthCheck_ReturnsHealthyAndExceptionIsNotThrown()
+    public async Task DeleteFails_RunHealthCheck_ReturnsDegradedAndExceptionIsNotThrown()
     {
-        // the extra deletes before and after the checks fail, but everything else successfully finishes
-        _blockBlobClient.DeleteIfExistsAsync(Arg.Any<DeleteSnapshotsOption>(), Arg.Any<BlobRequestConditions>(), Arg.Any<CancellationToken>())
+        _blockBlobClient.DeleteAsync(Arg.Any<DeleteSnapshotsOption>(), Arg.Any<BlobRequestConditions>(), Arg.Any<CancellationToken>())
             .Throws(new RequestFailedException(403, "Failure", BlobErrorCode.AuthorizationFailure.ToString(), new System.Exception()));
 
         HealthCheckResult result = await _dicomConnectedStoreHealthCheck.CheckHealthAsync(null, CancellationToken.None);
 
-        Assert.Equal(HealthStatus.Healthy, result.Status);
+        result.Data.TryGetValue("Reason", out object healthStatusReason);
+
+        Assert.Equal(HealthStatus.Degraded, result.Status);
+        Assert.Equal(HealthStatusReason.ConnectedStoreDegraded.ToString(), healthStatusReason.ToString());
     }
 }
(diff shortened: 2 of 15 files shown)


src/Microsoft.Health.Dicom.Blob/Extensions/AzureStorageErrorExtensions.cs
```diff
@@ -46,6 +46,8 @@ public static bool IsConnectedStoreCustomerError(this RequestFailedException rfe
     public static bool IsStorageAccountUnknownHostError(this Exception exception)
     {
         return exception.Message.Contains("No such host is known", StringComparison.OrdinalIgnoreCase) ||
-            (exception is AggregateException ag && ag.InnerExceptions.All(e => e.Message.Contains("No such host is known", StringComparison.OrdinalIgnoreCase)));
+            exception.Message.Contains("Name or service not known", StringComparison.OrdinalIgnoreCase) ||
```
jnlycklama, line 49: This was added because of Karthik's instance that was failing for this error https://microsofthealth.visualstudio.com/Health/_workitems/edit/114214

src/Microsoft.Health.Dicom.Blob/Features/Health/DicomConnectedStoreHealthCheck.cs
```diff
@@ -51,26 +52,33 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
         _logger.LogInformation("Checking the health of the connected store.");
 
         BlobContainerClient containerClient = _blobClient.BlobContainerClient;
-        BlockBlobClient blockBlobClient = containerClient.GetBlockBlobClient($"{_externalStoreOptions.StorageDirectory}{_externalStoreOptions.HealthCheckFileName}");
 
-        // start trying to delete the blob, in case delete failed on the previous run
-        await TryDeleteBlob(blockBlobClient, cancellationToken);
```
jnlycklama, line 57: I removed the extra deletes because they seem unnecessary if we are using the same file name each time. If it fails to delete, it doesnt matter. The next health check will just upload new content

src/Microsoft.Health.Dicom.Blob/Features/Health/DicomConnectedStoreHealthCheck.cs
```diff
@@ -51,26 +52,33 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
...
+            await TryAcquireLease(leaseBlobLeaseClient, cancellationToken);
 
-            string blockId = Convert.ToBase64String(Guid.NewGuid().ToByteArray());
+            using Stream stream = new MemoryStream(Encoding.UTF8.GetBytes(_testContent));
 
-            // test blob/write and that block blobs are supported
-            await blockBlobClient.StageBlockAsync(blockId, stream, cancellationToken: cancellationToken);
-            await blockBlobClient.CommitBlockListAsync(new List<string> { blockId }, new CommitBlockListOptions(), cancellationToken);
+            // test blob/write
+            await healthCheckBlobClient.UploadAsync(stream, cancellationToken: cancellationToken);
```
jnlycklama, line 75: Storage will still throw "BlobTypeNotSupported" using UploadAsync if we are using a BlockBlobClient, so testing with these specific block blob functions is not neccessary

src/Microsoft.Health.Dicom.Blob/Features/Health/DicomConnectedStoreHealthCheck.cs
```diff
@@ -84,21 +92,42 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
...
     }
 
-    [System.Diagnostics.CodeAnalysis.SuppressMessage("Design", "CA1031:Do not catch general exception types", Justification = "Do not fail during final clean up")]
-    private static async Task TryDeleteBlob(BlockBlobClient blobClient, CancellationToken cancellationToken)
+    public static async Task TryReleaseLease(BlobLeaseClient leaseBlobLeaseClient, CancellationToken cancellationToken)
     {
         try
         {
-            await blobClient.DeleteIfExistsAsync(DeleteSnapshotsOption.IncludeSnapshots, new BlobRequestConditions(), cancellationToken);
+            await leaseBlobLeaseClient.ReleaseAsync(cancellationToken: cancellationToken);
```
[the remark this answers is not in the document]
jnlycklama: It seems like the InvalidBlockList error only comes up during extreme concurrency issues. The health check of each dicom pod is trying to write/delete the same file, so the error is coming up in a few instances.

I'm not sure if it's worth it to add to every blob store interaction, but I'm also not sure how likely it is that we are trying to write or delete to the same blob at the same time. Is that possible? I guess the customer could be trying to delete/write as we are trying to delete/write, but maybe an error should be expected in that case?

src/Microsoft.Health.Dicom.Blob/Extensions/AzureStorageErrorExtensions.cs
jnlycklama: Moved this file to Core so it can be shared with DataStoreRequestFailedException

src/Microsoft.Health.Dicom.Blob.UnitTests/Features/Health/DicomConnectedStoreHealthCheckTests.cs
```diff
@@ -80,29 +83,25 @@ public async Task GivenRequestFailedExceptionForCustomerError_RunHealthCheck_Ret
...
         Assert.Equal(HealthStatusReason.ConnectedStoreDegraded.ToString(), healthStatusReason.ToString());
-
-        await _blockBlobClient.Received(2).DeleteIfExistsAsync(Arg.Any<DeleteSnapshotsOption>(), Arg.Any<BlobRequestConditions>(), Arg.Any<CancellationToken>());
     }
 
     [Theory]
     [InlineData(400, "SomeErrorFromDicomBug")]
     [InlineData(403, "AuthErrorFromDicomBug")]
     [InlineData(404, "NotFoundDueToDicomBug")]
     [InlineData(409, "ConflictDueToDicomBug")]
```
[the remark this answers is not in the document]
jnlycklama: This test is meant to test that any 409 that we dont classify as a client issue will throw an exception and set the health check to unhealthy. blob can return 409 for a bunch of reasons so I think its worth checking https://learn.microsoft.com/en-us/rest/api/storageservices/blob-service-error-codes

src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalStoreHealthExpiryHttpPipelinePolicy.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        _healthCheckPathRegex = Regex.Escape(uriBuilder.Uri.ToString());
+        _txtRegex = Regex.Escape(".txt");
+    }
+
+    public override void Process(HttpMessage message, ReadOnlyMemory<HttpPipelinePolicy> pipeline)
+    {
+        if (IsHealthCheckFileUpload(message.Request))
+        {
+            message.Request.Headers.Add("x-ms-expiry-time", $"{_externalStoreOptions.HealthCheckFileExpiryInMs}");
```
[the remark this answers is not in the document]
jnlycklama: Hmm doesn't look like it - in OSS they just use a string https://github.com/Azure/azure-sdk-for-net/blob/757ee0d13d1aaf9ba5d62cdeb5d53810668b72fd/sdk/storage/Azure.Storage.Blobs/src/Generated/BlobRestClient.cs#L614 
There is a `BlobHttpHeaders` class, but it only has a couple of generic header options https://github.com/Azure/azure-sdk-for-net/blob/757ee0d13d1aaf9ba5d62cdeb5d53810668b72fd/sdk/storage/Azure.Storage.Blobs/src/Models/BlobHttpHeaders.cs#L19

src/Microsoft.Health.Dicom.Web/appsettings.json
```diff
@@ -30,7 +30,8 @@
   "ExternalBlobStore": {
     "BlobContainerUri": null,
     "StorageDirectory": "DICOM/",
-    "HealthCheckFileName": "healthCheck/health.txt"
+    "HealthCheckFilePath": "healthCheck/health",
+    "HealthCheckFileExpiryInMs": 3600000 // 1 hour
```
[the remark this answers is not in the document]
jnlycklama: Oh `// 1 hour` was added, not deleted. I think its red because it's not the correct formatting for appsettings.json. I actually will remove it since its showing up red

src/Microsoft.Health.Dicom.Web/appsettings.json
```diff
@@ -30,7 +30,8 @@
   "ExternalBlobStore": {
     "BlobContainerUri": null,
     "StorageDirectory": "DICOM/",
-    "HealthCheckFileName": "healthCheck/health.txt"
+    "HealthCheckFilePath": "healthCheck/health",
+    "HealthCheckFileExpiryInMs": 3600000
```
wsugarman, line 34: nit: just make this a `TimeSpan` and convert to MS in the code

src/Microsoft.Health.Dicom.Core/Microsoft.Health.Dicom.Core.csproj
```diff
@@ -8,6 +8,7 @@
   <ItemGroup>
     <PackageReference Include="Azure.Core" />
     <PackageReference Include="Azure.Identity" />
+    <PackageReference Include="Azure.Storage.Blobs" />
```
wsugarman, line 11: There shouldn't be a blob dependency here. Perhaps there should be an exception that implements `DataStoreRequestFailedException` with blob-specific logic

src/Microsoft.Health.Dicom.Blob/Utilities/ExternalBlobDataStoreConfiguration.cs
```diff
@@ -20,7 +20,9 @@ internal class ExternalBlobDataStoreConfiguration
     // use for local testing with Azurite
     public string ContainerName { get; set; }
 
-    public string HealthCheckFileName { get; set; }
+    public string HealthCheckFilePath { get; set; }
+
+    public int HealthCheckFileExpiryInMs { get; set; }
```
wsugarman, line 25:
```suggestion
[Range(typeof(TimeSpan), "00:00:01", "1.00:00:00", ConvertValueInInvariantCulture = true, ParseLimitsInInvariantCulture = true)]
public TimeSpan HealthCheckFileExpiry { get; set; }
```

src/Microsoft.Health.Dicom.Blob/Features/Health/DicomConnectedStoreHealthCheck.cs
```diff
@@ -84,21 +80,21 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
         }
         finally
         {
-            // Just in case the blob was successfully created but failed to be deleted, try clean up again
-            await TryDeleteBlob(blockBlobClient, cancellationToken);
+            await TryDeleteAsync(healthCheckBlobClient, cancellationToken);
```
wsugarman, line 83: I think we can ignore this now, right? Do we need this `finally` block?

src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalStoreHealthExpiryHttpPipelinePolicy.cs
```diff
@@ -0,0 +1,65 @@
...
+            message.Request.Headers.Add("x-ms-expiry-option", "RelativeToNow");
+        }
+
+        return ProcessNextAsync(message, pipeline);
+    }
+
+    private bool IsHealthCheckFileUpload(Request request)
+    {
+        return Regex.IsMatch(request.Uri.ToString(), $"^{_healthCheckPathRegex}{GuidRegex}{_txtRegex}$") &&
+            (request.Method == RequestMethod.Put || request.Method == RequestMethod.Post || request.Method == RequestMethod.Patch);
```
wsugarman, line 63: nit: Can we say the following?
```suggestion
request.Method is RequestMethod.Put or RequestMethod.Post or RequestMethod.Patch
```

src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalStoreHealthExpiryHttpPipelinePolicy.cs
```diff
@@ -0,0 +1,65 @@
...
+            message.Request.Headers.Add("x-ms-expiry-time", $"{_externalStoreOptions.HealthCheckFileExpiryInMs}");
+            message.Request.Headers.Add("x-ms-expiry-option", "RelativeToNow");
+        }
+
+        return ProcessNextAsync(message, pipeline);
+    }
+
+    private bool IsHealthCheckFileUpload(Request request)
+    {
+        return Regex.IsMatch(request.Uri.ToString(), $"^{_healthCheckPathRegex}{GuidRegex}{_txtRegex}$") &&
```
wsugarman, line 62: Please pre-compile this `Regex` in the ctor instead of parsing here. E.g. `new Regex("<my string>", RegexOptions.Compiled | <more options>);`

src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalStoreHealthExpiryHttpPipelinePolicy.cs
```diff
@@ -0,0 +1,65 @@
...
+            message.Request.Headers.Add("x-ms-expiry-time", $"{_externalStoreOptions.HealthCheckFileExpiryInMs}");
+            message.Request.Headers.Add("x-ms-expiry-option", "RelativeToNow");
+        }
+
+        ProcessNext(message, pipeline);
+    }
+
+    public override ValueTask ProcessAsync(HttpMessage message, ReadOnlyMemory<HttpPipelinePolicy> pipeline)
+    {
+        if (IsHealthCheckFileUpload(message.Request))
```
wsugarman, line 51: Can you move the header append to a helper method? Perhaps even combining with `IsHealthCheckFileUpload`?

src/Microsoft.Health.Dicom.Blob.UnitTests/Features/ExternalStore/MockRequest.cs
```diff
@@ -0,0 +1,45 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using Azure.Core;
+
+namespace Microsoft.Health.Dicom.Blob.UnitTests.Features.ExternalStore;
+
+internal class MockRequest : Request
```
wsugarman, line 14: Something about `Request` couldn't be mocked via `NSubstitute`? Or this was easier?

src/Microsoft.Health.Dicom.Blob/Features/Health/DicomConnectedStoreHealthCheck.cs
jnlycklama: I had left it in to have the health check do a best-effort clean up of the blob, and the expiry header would be our last line of defense to delete it an hour later.

But you're right, since the expiry header will always work, maybe I can get rid of this and just reduce the expiry time so there aren't too many blobs waiting to be deleted at once.

src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalStoreHealthExpiryHttpPipelinePolicy.cs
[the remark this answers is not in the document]
jnlycklama: This uri is a `RequestUriBuilder`, not just a plain `Uri`, so I had to do `ToUri().AbsoluteUri` instead

src/Microsoft.Health.Dicom.Core/Microsoft.Health.Dicom.Core.csproj
jnlycklama: Ended up splitting the exception extentions into `BlobStorageErrorExtensions` and `AzureStorageErrorExtensions`, since the method that `DataStoreRequestFailedException` needs is not blob specific

src/Microsoft.Health.Dicom.Blob.UnitTests/Features/ExternalStore/MockRequest.cs
jnlycklama: Yeah, `Headers` is not virtual or abstract so cannot be mocked https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/core/Azure.Core/src/Request.cs#L108

src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalStoreHealthExpiryHttpPipelinePolicy.cs
```diff
@@ -0,0 +1,59 @@
...
+        _externalStoreOptions = EnsureArg.IsNotNull(externalStoreOptions, nameof(externalStoreOptions));
+        EnsureArg.IsNotNull(externalStoreOptions.BlobContainerUri, nameof(externalStoreOptions.BlobContainerUri));
+        EnsureArg.IsNotNull(externalStoreOptions.StorageDirectory, nameof(externalStoreOptions.StorageDirectory));
+        EnsureArg.IsNotNull(externalStoreOptions.HealthCheckFilePath, nameof(externalStoreOptions.HealthCheckFilePath));
+
+        UriBuilder uriBuilder = new UriBuilder(_externalStoreOptions.BlobContainerUri);
+        uriBuilder.Path = Path.Combine(uriBuilder.Path, _externalStoreOptions.StorageDirectory, _externalStoreOptions.HealthCheckFilePath);
+
+        string healthCheckPathRegex = Regex.Escape(uriBuilder.Uri.AbsoluteUri);
+        string txtRegex = Regex.Escape(".txt");
```
wsugarman, line 34: lol I think you can probably escape `.txt` yourself 😛

src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalStoreHealthExpiryHttpPipelinePolicy.cs
```diff
@@ -0,0 +1,59 @@
...
+        _healthCheckRegex = new Regex($"^{healthCheckPathRegex}{GuidRegex}{txtRegex}$", RegexOptions.Compiled);
```
wsugarman, line 35: nit: include the culture invariant option

src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalStoreHealthExpiryHttpPipelinePolicy.cs
```diff
@@ -0,0 +1,59 @@
...
+        AddExpiryHeaderToHealthCheckFileUploadRequest(message);
+        return ProcessNextAsync(message, pipeline);
+    }
+
+    private void AddExpiryHeaderToHealthCheckFileUploadRequest(HttpMessage message)
+    {
+        if (_healthCheckRegex.IsMatch(message.Request.Uri.ToUri().AbsoluteUri) &&
+        (message.Request.Method == RequestMethod.Put || message.Request.Method == RequestMethod.Post || message.Request.Method == RequestMethod.Patch))
+        {
+            message.Request.Headers.Add("x-ms-expiry-time", $"{_externalStoreOptions.HealthCheckFileExpiry.TotalMilliseconds}");
```
wsugarman, line 55: nit: invoke `ToString` with an invariant culture instead of a format string that will use current culture

src/Microsoft.Health.Dicom.Blob/Features/Health/DicomConnectedStoreHealthCheck.cs
```diff
@@ -51,26 +52,22 @@ public async Task<HealthCheckResult> CheckHealthAsync(HealthCheckContext context
         _logger.LogInformation("Checking the health of the connected store.");
 
         BlobContainerClient containerClient = _blobClient.BlobContainerClient;
-        BlockBlobClient blockBlobClient = containerClient.GetBlockBlobClient($"{_externalStoreOptions.StorageDirectory}{_externalStoreOptions.HealthCheckFileName}");
+        BlockBlobClient healthCheckBlobClient = containerClient.GetBlockBlobClient(Path.Combine(_externalStoreOptions.StorageDirectory, $"{_externalStoreOptions.HealthCheckFilePath}{Guid.NewGuid()}.txt"));
 
-        // start trying to delete the blob, in case delete failed on the previous run
-        await TryDeleteBlob(blockBlobClient, cancellationToken);
```
wsugarman, line 57: What about the old file? Should there be temp logic that deletes the old blob if it exists?

src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalStoreHealthExpiryHttpPipelinePolicy.cs
jnlycklama: They don't make the constructor with `CultureInfo` public for some reason... had to change the options to `RegexOptions.InvariantCulture`

src/Microsoft.Health.Dicom.Blob/Features/Health/DicomConnectedStoreHealthCheck.cs
jnlycklama: Ah good catch - Added and put a comment with reference to the work item to clean up

src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalStoreHealthExpiryHttpPipelinePolicy.cs
wsugarman: That's what I meant

src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalStoreHealthExpiryHttpPipelinePolicy.cs
```diff
@@ -0,0 +1,59 @@
...
+        _externalStoreOptions = EnsureArg.IsNotNull(externalStoreOptions, nameof(externalStoreOptions));
+        EnsureArg.IsNotNull(externalStoreOptions.BlobContainerUri, nameof(externalStoreOptions.BlobContainerUri));
+        EnsureArg.IsNotNull(externalStoreOptions.StorageDirectory, nameof(externalStoreOptions.StorageDirectory));
+        EnsureArg.IsNotNull(externalStoreOptions.HealthCheckFilePath, nameof(externalStoreOptions.HealthCheckFilePath));
+
+        UriBuilder uriBuilder = new UriBuilder(_externalStoreOptions.BlobContainerUri);
+        uriBuilder.Path = Path.Combine(uriBuilder.Path, _externalStoreOptions.StorageDirectory, _externalStoreOptions.HealthCheckFilePath);
+
+        string healthCheckPathRegex = Regex.Escape(uriBuilder.Uri.AbsoluteUri);
+        _healthCheckRegex = new Regex($"^{healthCheckPathRegex}{GuidRegex}\\.txt$", RegexOptions.CultureInvariant);
```
wsugarman, line 35: OR flags together: `RegexOptions.Compiled | RegexOptions.CultureInvariant`

wsugarman (approved):

PR #3331, opened 2024-01-31
track egress blob bytes

esmadau:
## Description
track egress blob bytes

We will only track on dicom binary blob egress since workitem and metadata continue to be stored in internal storage.

## Related issues
Addresses [AB#112842].

## Testing
Amended existing unit test to ensure correct length is set on context

### DIFF
--- a/src/Microsoft.Health.Dicom.Core.UnitTests/Features/Retrieve/RetrieveResourceServiceTests.cs
+++ b/src/Microsoft.Health.Dicom.Core.UnitTests/Features/Retrieve/RetrieveResourceServiceTests.cs
@@ -355,7 +355,7 @@ public class RetrieveResourceServiceTests
         foreach (var file in streamsAndStoredFiles)
         {
             var length = file.Value.Length;
-            _fileStore.GetFilePropertiesAsync(Arg.Any<long>(), Partition.Default, null, DefaultCancellationToken).Returns(new FileProperties { ContentLength = length });
+            _fileStore.GetFilePropertiesAsync(count, Partition.Default, null, DefaultCancellationToken).Returns(new FileProperties { ContentLength = length });
             _fileStore.GetStreamingFileAsync(count, Partition.Default, null, DefaultCancellationToken).Returns(file.Value);
             count++;
             expectedContentLength += length;


src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
```diff
@@ -224,6 +225,7 @@ public async Task<RetrieveResourceResponse> GetInstanceResourceAsync(RetrieveRes
 
         // eagerly doing getFrames to validate frame numbers are valid before returning a response
         Stream stream = await _blobDataStore.GetFileAsync(instance.VersionedInstanceIdentifier.Version, partition, instance.InstanceProperties.FileProperties, cancellationToken);
+        _dicomRequestContextAccessor.RequestContext.BlobBytesEgress = stream?.Length ?? 0;
```
bcarthic, line 228: For frames I see you are logging for retrieving the file, should we do for the whole file or just for the frames they asked?

src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
```diff
@@ -132,6 +132,7 @@ public async Task<RetrieveResourceResponse> GetInstanceResourceAsync(RetrieveRes
                 SetTranscodingBillingProperties(fileProperties.ContentLength);
 
                 using Stream stream = await _blobDataStore.GetFileAsync(version, instance.VersionedInstanceIdentifier.Partition, instance.InstanceProperties.FileProperties, cancellationToken);
+                _dicomRequestContextAccessor.RequestContext.BlobBytesEgress = stream?.Length ?? 0;
```
bcarthic, line 135: You can add this line in SetTranscodingBillingProperties

src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
```diff
@@ -306,7 +308,10 @@ private static bool NeedsTranscoding(bool isOriginalTransferSyntaxRequested, str
             long version = instanceMetadata.GetVersion(isOriginalVersionRequested);
             FileProperties fileProperties = await _blobDataStore.GetFilePropertiesAsync(version, _dicomRequestContextAccessor.RequestContext.GetPartition(), instanceMetadata.InstanceProperties.FileProperties, cancellationToken);
             Stream stream = await _blobDataStore.GetStreamingFileAsync(version, _dicomRequestContextAccessor.RequestContext.GetPartition(), instanceMetadata.InstanceProperties.FileProperties, cancellationToken);
-            streamTotalLength += fileProperties.ContentLength;
+            streamTotalLength += stream?.Length ?? 0;
```
bcarthic, line 311: what was the reason to change from fileproperties to stream?

src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
bcarthic: Seems like we want to bill for the file we are downloading

src/Microsoft.Health.Dicom.Core/Features/Retrieve/RetrieveResourceService.cs
```diff
@@ -132,6 +132,7 @@ public async Task<RetrieveResourceResponse> GetInstanceResourceAsync(RetrieveRes
                 SetTranscodingBillingProperties(fileProperties.ContentLength);
 
                 using Stream stream = await _blobDataStore.GetFileAsync(version, instance.VersionedInstanceIdentifier.Partition, instance.InstanceProperties.FileProperties, cancellationToken);
```
bcarthic, line 134: You do stream?.Length ?? 0; in other file. DO you expect to return null?

PR #3337, opened 2024-02-02
track total egress bytes on stow

esmadau:
## Description
track total egress bytes on stow

## Related issues
Addresses [AB#112842].

## Testing
amended existing stow test and added test

### DIFF
--- a/src/Microsoft.Health.Dicom.Core/Features/Context/IDicomRequestContext.cs
+++ b/src/Microsoft.Health.Dicom.Core/Features/Context/IDicomRequestContext.cs
@@ -32,5 +32,5 @@ public interface IDicomRequestContext : IRequestContext
     /// <summary>
     /// Egress bytes from Dicom server to other resources
     /// </summary>
-    long TotalEgressBytes { get; set; }
+    long TotalDicomEgressToStorageBytes { get; set; }
 }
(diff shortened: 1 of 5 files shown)


src/Microsoft.Health.Dicom.Core/Features/Context/IDicomRequestContext.cs
```diff
@@ -28,4 +28,9 @@ public interface IDicomRequestContext : IRequestContext
 
     // Opportunity for the core to change based on the caller version
     int Version { get; set; }
+
+    /// <summary>
+    /// Egress bytes from Dicom server to other resources
+    /// </summary>
+    long TotalEgressBytes { get; set; }
```
wsugarman, line 35: Let's perhaps call this `TotalStorageEgressBytes` to disambiguate it from egressing data from the service overall

src/Microsoft.Health.Dicom.Core/Features/Store/StoreService.cs
```diff
@@ -115,6 +116,7 @@ public class StoreService : IStoreService
                         long len = length.GetValueOrDefault();
                         // Update Telemetry
                         _storeMeter.InstanceLength.Record(len);
+                        totalBytesStored += len;
```
wsugarman, line 119: We could also directly assign to the context

wsugarman (approved):

src/Microsoft.Health.Dicom.Core/Features/Store/StoreService.cs
```diff
@@ -115,6 +115,7 @@ public class StoreService : IStoreService
                         long len = length.GetValueOrDefault();
                         // Update Telemetry
                         _storeMeter.InstanceLength.Record(len);
+                        _dicomRequestContextAccessor.RequestContext.TotalDicomEgressToStorageBytes += len;
```
bcarthic, line 118: Is this Ingress or egress? isn't egress is any data being downloaded or transferred out of the storage account.

PR #3341, opened 2024-02-06
add new error as customer error; add option to disable health check

jnlycklama:
## Description
Add new error when HNS is not enabled as a customer error

## Related issues
Addresses #114879

## Testing
Unit tests

### DIFF
--- a/src/Microsoft.Health.Dicom.Web/appsettings.json
+++ b/src/Microsoft.Health.Dicom.Web/appsettings.json
@@ -85,7 +85,7 @@
       "EnableOhifViewer": true,
       "EnableExternalStore": false,
       "DisableOperations": false,
-      "EnableExternalStoreHealthCheck": true
+      "EnableExternalStoreHealthCheck": false
     },
     "Services": {
       "DeletedInstanceCleanup": {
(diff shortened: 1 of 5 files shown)


src/Microsoft.Health.Dicom.Web/appsettings.json
```diff
@@ -84,7 +84,8 @@
       "EnableFullDicomItemValidation": false,
       "EnableOhifViewer": true,
       "EnableExternalStore": false,
-      "DisableOperations": false
+      "DisableOperations": false,
+      "EnableExternalStoreHealthCheck": true
```
bcarthic, line 88: This can be false by default, since there is no concept of external store in OSS

bcarthic: Once merged in paas, you can make this true by default

jnlycklama: Good call, makes sense to leave this false here

bcarthic (approved):

bcarthic (approved):

KnicKnic (approved):

PR #3342, opened 2024-02-06
Dicom Cast Validation Bug

abdullah248:
## Description
Ignore dicom cast validation bug when adding irradiation events and dose summary reports

## Related issues
Addresses [AB#114198].

## Testing
Tested locally

### DIFF
--- a/converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/FhirTransaction/Observation/ObservationParser.cs
+++ b/converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/FhirTransaction/Observation/ObservationParser.cs
@@ -35,7 +35,7 @@ internal static class ObservationParser
             Observation observation = null;
             try
             {
-                var code = new DicomCodeItem(codes);
+                var code = createDicomCode(codes);
 
                 if (ObservationConstants.IrradiationEvents.Contains(code) && TryCreateIrradiationEvent(dataset, patientReference, identifier, out Observation irradiationEvent))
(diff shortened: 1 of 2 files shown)


bcarthic (approved):

converter/dicom-cast/src/Microsoft.Health.DicomCast.Core/Features/Worker/FhirTransaction/Observation/ObservationParser.cs
```diff
@@ -32,13 +32,24 @@ private static IEnumerable<Observation> ParseDicomDataset(DicomDataset dataset,
     {
         if (dataset.TryGetSequence(DicomTag.ConceptNameCodeSequence, out DicomSequence codes) && codes.Items.Count > 0)
         {
-            var code = new DicomCodeItem(codes);
+            Observation observation = null;
+            try
```
bcarthic, line 36: Is there a unit test for this file?

abdullah248: I left a comment already on line 48 of why we are doing this. I don't think a separate method is necessary - I am just adding a try catch. I had to refactor a little since we can't yield inside of a try catch

abdullah248: Added unit test as well

bcarthic (approved):

bcarthic (approved):

PR #3357, opened 2024-02-09
Add configuration to filter health checks

jnlycklama:
## Description
Add configuration to filter health checks. Any health check can be added to this list, and it will be excluded from running. 
Also fixed an issue with running the IDP health check locally - the http pipeline policy required a blob URI to run, which does not exist when running with Azurite

## Related issues
Addresses #115147

## Testing
Ran locally. Added single & multiple health check names to the list and saw that they were removed from execution. Also tried removing the config all together or using an empty list, in which case all of the health checks will run.

### DIFF
--- a/src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalStoreHealthExpiryHttpPipelinePolicy.cs
+++ b/src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalStoreHealthExpiryHttpPipelinePolicy.cs
@@ -10,6 +10,7 @@ using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Azure.Core;
 using Azure.Core.Pipeline;
+using Azure.Storage.Blobs;
 using EnsureThat;
 using Microsoft.Health.Dicom.Blob.Utilities;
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalStoreHealthExpiryHttpPipelinePolicy.cs
```diff
@@ -24,11 +25,23 @@ internal class ExternalStoreHealthExpiryHttpPipelinePolicy : HttpPipelinePolicy
...
+        Uri blobUri;
+
+        if (_externalStoreOptions.BlobContainerUri != null)
+        {
+            blobUri = _externalStoreOptions.BlobContainerUri;
+        }
+        else
+        {
+            // For local testing with Azurite
+            BlobContainerClient blobContainerClient = new BlobContainerClient(_externalStoreOptions.ConnectionString, _externalStoreOptions.ContainerName);
```
bcarthic, line 40: Should this be more guarded with env than the uri?

jnlycklama: This is how its done in the `ExternalBlobClient` https://github.com/microsoft/dicom-server/blob/7d750be725e34aebcb78663a670bb9d393e62606/src/Microsoft.Health.Dicom.Blob/Features/ExternalStore/ExternalBlobClient.cs#L25

We could guard by the current environment, although technically environment doesn't matter here. If we for some reason wanted to switch to connecting to the blob store via ConnectionString instead of BlobContainerUri, the url generated in the `else` statement would still be accurate.

src/Microsoft.Health.Dicom.Core/Configs/HealthCheckPublisherConfiguration.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    public const string SectionName = "HealthCheckPublisher";
+
+    /// <summary>
+    /// A comma separated list of health check names to exclude.
+    /// Example: "DcmHealthCheck,MetadataHealthCheck"
+    /// </summary>
+    public string ExcludedHealthCheckNames { get; set; }
+
+    public IEnumerable<string> GetListOfExcludedHealthCheckNames()
```
bcarthic, line 21: nit: this could be IReadonlyList<string>

bcarthic (approved):

bcarthic (approved):

